PR #109, opened 2019-06-17
ci: stage all platforms

junr03:

Description: organize azure pipelines CI into independent stages.
Risk Level: low - reorganize CI.
Testing: run CI.


azure-pipelines.yml
```diff
@@ -10,151 +10,162 @@ resources:
...
+          - script: ./ci/mac_ci_setup.sh
+            displayName: 'Install dependencies'
+          - script: bazel build --config=ios //:ios_dist
+            displayName: 'Build Envoy.framework distributable'
+          - task: PublishPipelineArtifact@0
+            displayName: 'Publish Envoy.framework distributable'
+            inputs:
+              artifactName: 'Envoy.framework'
+              targetPath: 'dist/Envoy.framework'
+      - job: objc_helloworld
```
rebello95: Do you think we should prefix these with `mac` like we do for `linux` above?

junr03: fair point, would be more homogenous

PR #131, opened 2019-06-18
kotlin: Create aar files from kotlin_library jar rather than relying on android_binary

buildbreaker:
The primary goal this achieves is that it unlocks our ability to develop the Envoy mobile shim in kotlin rather than java. 

The gist of this change is that `android_library`'s implicit `aar` doesn't flatten its transitive dependencies. When using the kotlin rules, the `kt_android_library` rule creates a few underlying libraries, because of this the `classes.jar` in the `aar` we built was empty. This change separately builds the underlying `*kt.jar` file, and replaces the `aar`'s `classes.jar` with the kotlin one.

https://github.com/lyft/envoy-mobile/issues/99

Description: Enable kotlin library development
Risk Level: low
Testing: ci, local
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/examples/java/hello_world/BUILD
+++ b/examples/java/hello_world/BUILD
@@ -2,9 +2,12 @@ licenses(["notice"])  # Apache 2
 
 android_binary(
     name = "hello_envoy",
-    srcs = glob([
-        "*.java",
-    ]),
+    srcs = [
+        "MainActivity.java",
+        "Response.java",
+        "ResponseRecyclerViewAdapter.java",
+        "ResponseViewHolder.java",
+    ],
     custom_package = "io.envoyproxy.envoymobile.helloenvoy",
     manifest = "AndroidManifest.xml",
     resource_files = glob(["res/**"]),
--- a/examples/java/hello_world/BUILD
+++ b/examples/java/hello_world/BUILD
@@ -3,10 +3,7 @@ licenses(["notice"])  # Apache 2
 android_binary(
     name = "hello_envoy",
     srcs = glob([
-        "MainActivity.java",
-        "Response.java",
-        "ResponseRecyclerViewAdapter.java",
-        "ResponseViewHolder.java",
+        "*.java",
     ]),
     custom_package = "io.envoyproxy.envoymobile.helloenvoy",
     manifest = "AndroidManifest.xml",
--- a/examples/kotlin/hello_world/BUILD
+++ b/examples/kotlin/hello_world/BUILD
@@ -6,7 +6,7 @@ load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_android_library")
 envoy_package()
 
 android_binary(
-    name = "hello_envoy_kotlin",
+    name = "hello_envoy_kt",
     custom_package = "io.envoyproxy.envoymobile.helloenvoykotlin",
     manifest = "AndroidManifest.xml",
     deps = [
--- a/examples/kotlin/hello_world/BUILD
+++ b/examples/kotlin/hello_world/BUILD
@@ -16,9 +16,12 @@ android_binary(
 
 kt_android_library(
     name = "hello_envoy_kt_lib",
-    srcs = glob([
-        "*.kt",
-    ]),
+    srcs = [
+        "MainActivity.kt",
+        "Response.kt",
+        "ResponseRecyclerViewAdapter.kt",
+        "ResponseViewHolder.kt",
+    ],
     custom_package = "io.envoyproxy.envoymobile.helloenvoykotlin",
     manifest = "AndroidManifest.xml",
     resource_files = glob([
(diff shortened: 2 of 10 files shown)


buildbreaker:
Thanks @keith for the pair program to achieve this!

:)

buildbreaker:
@goaway I made a bunch of modifications so this PR would not have conflicts with your PR: https://github.com/lyft/envoy-mobile/pull/149

I'll have to make another PR to convert your API changes to Kotlin and do some clean up

library/java/io/envoyproxy/envoymobile/Envoy.java
```diff
@@ -11,7 +11,7 @@
 
   public void load() { System.loadLibrary("envoy_jni"); }
 
-  public void run(Context context, String config) {
+  public void run(final Context context, final String config) {
```
buildbreaker: I got a complaint on the build without this change but I think we can just ignore it since this file is intended to change completely

examples/java/hello_world/BUILD
```diff
@@ -3,10 +3,7 @@ licenses(["notice"])  # Apache 2
 android_binary(
     name = "hello_envoy",
     srcs = glob([
-        "MainActivity.java",
-        "Response.java",
-        "ResponseRecyclerViewAdapter.java",
-        "ResponseViewHolder.java",
+        "*.java",
```
goaway: I somewhat prefer listing these dependencies explicitly, since it allows more than one module per directory (this is used all over the Envoy project).

examples/kotlin/hello_world/BUILD
```diff
@@ -17,10 +17,7 @@ android_binary(
 kt_android_library(
     name = "hello_envoy_kt_lib",
     srcs = glob([
-        "MainActivity.kt",
-        "Response.kt",
-        "ResponseRecyclerViewAdapter.kt",
-        "ResponseViewHolder.kt",
+        "*.kt",
```
goaway: Same as above. Also I don't think we need `glob` if we list explicitly.

examples/java/hello_world/BUILD
buildbreaker: ok

library/java/io/envoyproxy/envoymobile/Envoy.java
```diff
@@ -45,7 +45,7 @@ public static void load(Context context) {
   // initialization with the configuration provided. If the Envoy native library and its
   // dependencies haven't been loaded and initialized yet, this will happen lazily when
   // the first instance is created.
-  public Envoy(Context context, String config) {
+  public Envoy(final Context context, final String config) {
```
goaway: nice

library/java/io/envoyproxy/envoymobile/EnvoyEmptyClass.kt
```diff
@@ -0,0 +1,2 @@
+package library.java.io.envoyproxy.envoymobile
+
```
goaway: Do you want to add a comment here (similar to the java class?)

library/java/io/envoyproxy/envoymobile/Envoy.java
buildbreaker: This is probably because these files are going through to the kotlin bazel rules. Wont be a problem for long!

library/kotlin/io/envoyproxy/envoymobile/EnovyKotlinEmptyClass.java
```diff
@@ -0,0 +1,12 @@
+package io.envoyproxy.envoymobile;
```
goaway: "Enovy" typo in this filename

library/kotlin/io/envoyproxy/envoymobile/EnovyKotlinEmptyClass.java
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * This class is used as a workaround to compile cc_libraries into a java_library for
+ * kt_android_library for Envoy JNI in kotlin. There's an issue with transitive dependencies
+ * related to kt_jvm_library which errors with a message:
+ * <p>
+ * ...in deps attribute of kt_jvm_library rule //:android_lib_kt:
+ * '//library/common:envoy_jni_interface_lib' does not have mandatory providers: 'JavaInfo' <p>
+ * Could be related to: https://github.com/bazelbuild/rules_kotlin/issues/132
+ */
+public class EnovyKotlinEmptyClass {}
```
goaway: (and typo here too)

library/java/io/envoyproxy/envoymobile/Envoy.java
goaway: I mean I think it's a good thing

PR #145, opened 2019-06-18
build: remove unnecessary opts

### LINKED ISSUE #52: build: audit link options being passed
Validate that we actually need all of the link options we're currently passing through Bazel, as some extra ones may have been added during initial development.

goaway:

Description: Clean up of build opts.
Risk Level: Low
Testing: Built and ran locally.

Fixes #52

### DIFF
--- a/library/common/BUILD
+++ b/library/common/BUILD
@@ -15,10 +15,5 @@ envoy_cc_library(
 cc_library(
     name = "envoy_jni_interface_lib",
     srcs = ["jni_interface.cc"],
-    copts = ["-std=c++14"],
-    linkopts = [
-        "-lm",
-        "-llog",
-    ],
     deps = [":envoy_main_interface_lib"],
 )


library/common/BUILD
```diff
@@ -15,10 +15,5 @@ envoy_cc_library(
 cc_library(
     name = "envoy_jni_interface_lib",
     srcs = ["jni_interface.cc"],
-    copts = ["-std=c++14"],
-    linkopts = [
-        "-lm",
-        "-llog",
```
keith: I think these are still necessary for the JNI. Without these JNI builds failed to link

keith: The discussion we had at the time was likely it's upstream envoy that is missing `-lm` at least, but it doesn't matter there since it's being implicitly added somewhere else in the tree

goaway: still working for me locally - @keith can you try and repro?

goaway: I'm using a Pixel 2 XL emulator on API 28

goaway:
thanks @keith !

PR #147, opened 2019-06-18
build: reinsert possibly necessary opts

goaway:

Description: Welp. Maybe we need this after all.
Risk Level: Low
Testing: Saw more ceilf errors and this fixed them again.

### DIFF
--- a/library/common/BUILD
+++ b/library/common/BUILD
@@ -15,5 +15,10 @@ envoy_cc_library(
 cc_library(
     name = "envoy_jni_interface_lib",
     srcs = ["jni_interface.cc"],
+    copts = ["-std=c++14"],
+    linkopts = [
+        "-lm",
+        "-llog",
+    ],
     deps = [":envoy_main_interface_lib"],
 )


library/common/BUILD
```diff
@@ -15,5 +15,10 @@ envoy_cc_library(
 cc_library(
     name = "envoy_jni_interface_lib",
     srcs = ["jni_interface.cc"],
+    copts = ["-std=c++14"],
```
keith: I'm the least sure about needing this one

rebello95:
Out of curiosity, why didn't CI catch the issues when building the removal PR?

buildbreaker:
@rebello95 It errors upon runtime so we'll need some basic CI UI test job to catch this

goaway:
Honestly, everything was working after multiple `bazel clean --expunge`s, clearing out `/dist` and rebuilding the apps with a brand new emulator, so I really don't know how it was working.

goaway:
But I did some more dev on Android and once I had, bam, more `ceilf` errors. :(

goaway:
I think it's possible our build rules are somehow not clearing everything out they should even after expunge cc @junr03

mattklein123:
FWIW based on some other symptoms that people have expressed, I have a feeling that bazel might not be rebuilding everything in all cases. There are probably some bugs in the rules somewhere.

Issue #46, opened 2019-06-04
library: handle threading within library


goaway:
This pretty easy to do, standard in networking libraries, and simplifies usage.

goaway:
Completed with #149 and #164

PR #149, opened 2019-06-18
lib: make Android interface instance-based and threadsafe

goaway:

Description: Updates Android interface to statically load and initialize the library, but allow multiple Envoy instances to be created with their own configuration and tied to their own thread.
Risk Level: Moderate
Testing: Built and ran library and apps locally

### DIFF
--- a/library/java/io/envoyproxy/envoymobile/Envoy.java
+++ b/library/java/io/envoyproxy/envoymobile/Envoy.java
@@ -9,22 +9,69 @@ import java.io.*;
 // Wrapper class that allows for easy calling of Envoy's JNI interface in native Java.
 public class Envoy {
 
-  public void load() { System.loadLibrary("envoy_jni"); }
+  // Internal reference to helper object used to load and initialize the native library.
+  // Volatile to ensure double-checked locking works correctly.
+  private static volatile EnvoyLoader loader = null;
 
-  public void run(Context context, String config) {
-    Thread thread = new Thread(new Runnable() {
-      @Override
-      public void run() {
-        initialize((ConnectivityManager)context.getSystemService(Context.CONNECTIVITY_SERVICE));
-        runEnvoy(config.trim());
-      }
-    });
-    thread.start();
+  // Dedicated thread for running this instance of Envoy.
+  private final Thread runner;
+
+  // Private helper class used by the load method to ensure the native library and its
+  // dependencies are loaded and initialized at most once.
+  private static class EnvoyLoader {
+
+    EnvoyLoader(Context context) {
+      System.loadLibrary("envoy_jni");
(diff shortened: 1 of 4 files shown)


goaway:
Part of issue #46

goaway:
I'm starting the thread on object initialization, because it provides the strongest run-once guarantee without resorting to synchronization.

library/java/io/envoyproxy/envoymobile/Envoy.java
```diff
@@ -9,22 +9,67 @@
...
+  private static class EnvoyLoader {
+
+    EnvoyLoader(Context context) {
+      System.loadLibrary("envoy_jni");
+      initialize((ConnectivityManager)context.getSystemService(Context.CONNECTIVITY_SERVICE));
+    }
+  }
+
+  // Load and initialize Envoy and its dependencies, but only once.
+  public static void load(Context context) {
```
buildbreaker: I feel like I am missing something but shouldn't just making this method `synchronized` achieve the same behavior?

library/java/io/envoyproxy/envoymobile/Envoy.java
```diff
@@ -9,22 +9,67 @@
...
+  private static volatile EnvoyLoader loader = null;
 
-  public void run(Context context, String config) {
-    Thread thread = new Thread(new Runnable() {
+  // Dedicated thread for running this instance of Envoy.
+  private final Thread runner;
+
+  // Private helper class used by the load method to ensure the native library and its
+  // dependencies are loaded and initialized at most once.
+  private static class EnvoyLoader {
```
buildbreaker: If/when we translate this to Kotlin we can avoid nesting this class within `Envoy` :)

library/java/io/envoyproxy/envoymobile/Envoy.java
goaway: Because we support lazy loading, making this synchronized requires this call to acquire a lock every time an Envoy instance is created - but the behavior we execute in here only ever has to happen once. If there was something that needed to happen every time, with high contention, synchronizing the full method would be more appropriate.

library/java/io/envoyproxy/envoymobile/Envoy.java
goaway: cool

goaway:
Correct. You don't need to actually call `Envoy.load(*)` because this supports lazy loading of the native library. However, for those that want to start loading the lib before creating their request instance(s), the call is made available statically.

library/java/io/envoyproxy/envoymobile/Envoy.java
goaway: Another way to look at is that as soon as we have a fully-created instance of `EnvoyLoader` we know the native library has been loaded and no longer need to synchronize.

PR #152, opened 2019-06-19
docs: note using git hooks needs envoy dependences

adamawolf:
Description:

As an uninitiated person improving the sample projects or documentation, running `./envoy/support/bootstrap` alone doesn't cut it. The `pre-push` hook uses a bunch of dependencies from the core envoy project (realpath, clang-format, buildifier).

This change clarifies this in Contributing.md since some folks may try to contribute without necessarily having full context on envoy's dependencies. An alternative would be to install dependencies in `bootstrap`.

Risk Level: None
Testing: None
Docs Changes: Only
Release Notes: N/a

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,10 +32,11 @@ maximize the chances of your PR being merged.
   Please see [envoy/support/README.md](https://github.com/envoyproxy/envoy/tree/master/support) for
   more information on these hooks.
 
-  Note: these git hooks depend upon `realpath` which is part of `coreutils`. Install `coreutils` on macOS with:
+  Note: the git hooks depend upon many of the dependencies used to build envoy. Install these on macOS with:
 
   ```bash
-  brew install coreutils
+  brew install coreutils wget cmake libtool go bazel automake ninja clang-format autoconf aspell
+  go get github.com/bazelbuild/buildtools/buildifier
   ```
 
 * Create your PR.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -32,6 +32,13 @@ maximize the chances of your PR being merged.
   Please see [envoy/support/README.md](https://github.com/envoyproxy/envoy/tree/master/support) for
   more information on these hooks.
 
+  Note: the git hooks depend upon many of the dependencies used to build envoy. Install these on macOS with:
+
+  ```bash
+  brew install coreutils wget cmake libtool go bazel automake ninja clang-format autoconf aspell
```
rebello95: This dependency list can get out of sync with upstream Envoy relatively easily - what do you think about linking to the [upstream's readme section](https://github.com/envoyproxy/envoy/blob/master/bazel/README.md#quick-start-bazel-build-for-developers) instead? This also has the benefit of being non-specific to macOS

CONTRIBUTING.md
```diff
@@ -32,6 +32,8 @@ maximize the chances of your PR being merged.
   Please see [envoy/support/README.md](https://github.com/envoyproxy/envoy/tree/master/support) for
   more information on these hooks.
 
+  Note: the git hooks depend upon many of the dependencies used to build envoy via Bazel. Follow the [Quick start Bazel build for developers](https://github.com/envoyproxy/envoy/blob/master/bazel/README.md#quick-start-bazel-build-for-developers) guide to install these.
```
rebello95: Would you mind breaking this into multiple lines like the rest of the file so as not to exceed 110 characters if possible?

rebello95: I think you may have accidentally dropped the commit when rebasing

adamawolf: yah, I retracted the commit simply breaking it to two lines. but if it's ok for URLs to overflow, then let's go with that

PR #155, opened 2019-06-19
bazel: add release configuration

rebello95:
Adds a `--config=release` option for building with Bazel. At the moment, this ensures that iOS builds include all 4 architectures instead of only x86. In the future, more configurations will likely be added here for better optimization settings.

Risk Level: Low
Testing: Locally
Docs Changes: In PR


.bazelrc
```diff
@@ -19,3 +19,7 @@ build \
 build:ios --define=manual_stamp=manual_stamp
 
 build:android --fat_apk_cpu=x86,x86_64,armeabi-v7a,arm64-v8a
+
+build:release \
+  --compilation_mode=opt \
```
mattklein123: This is present in the OSS .bazelrc:

```
# Optimize build for binary size reduction.
build:sizeopt -c opt --copt -Os
```

I think this is what we want for now. I'm not sure if the OSS .bazelrc can be imported into this one somehow and we can just use that? @keith?

mattklein123: Also while we are here I would also add `--copt --gdb3` so we have full symbols across all deps for ease of debugging.

keith: Yopu can import the config from envoy, https://docs.bazel.build/versions/master/guide.html#imports and then you could do something like `build:release --config=whateverfromenvoy`

rebello95: Nice, done!

mattklein123: You can remove the above line now, since it's implied by the next line.

rebello95: Updated

rebello95:
Currently getting an interesting error when building with this configuration. @keith is helping investigate:

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: warning same member name (watcher_impl.o) in output file used for input files: bazel-out/ios-armv7-min10.0-applebin_ios-ios_armv7-opt/bin/external/envoy/source/common/filesystem/libwatcher_lib.lo_b07d3fadbdc4eeeaf15501e9ec6c9e21.o(watcher_impl.o) and: bazel-out/ios-armv7-min10.0-applebin_ios-ios_armv7-opt/bin/external/envoy/source/common/init/libwatcher_lib.lo_22a8cf39b04661ce3a6f40a88d82617d.o(watcher_impl.o) due to use of basename, truncation and blank padding
ERROR: /Users/mrebello/Development/envoy-mobile/BUILD:12:1: Bundling ios_framework failed (Exit 1): bundletool failed: error executing command 
  (cd /private/var/tmp/_bazel_mrebello/c8c34ff25de1ec7ce8f729468e440838/execroot/__main__ && \
  exec env - \
  bazel-out/host/bin/external/build_bazel_rules_apple/tools/bundletool/bundletool bazel-out/darwin-opt/bin/ios_framework-intermediates/bundletool_control.json)
Execution platform: @bazel_tools//platforms:host_platform
Traceback (most recent call last):
  File "/private/var/tmp/_bazel_mrebello/c8c34ff25de1ec7ce8f729468e440838/execroot/__main__/bazel-out/host/bin/external/build_bazel_rules_apple/tools/bundletool/bundletool.runfiles/build_bazel_rules_apple/tools/bundletool/bundletool.py", line 236, in <module>
    _main(sys.argv[1])
  File "/private/var/tmp/_bazel_mrebello/c8c34ff25de1ec7ce8f729468e440838/execroot/__main__/bazel-out/host/bin/external/build_bazel_rules_apple/tools/bundletool/bundletool.runfiles/build_bazel_rules_apple/tools/bundletool/bundletool.py", line 224, in _main
    bundler.run()
  File "/private/var/tmp/_bazel_mrebello/c8c34ff25de1ec7ce8f729468e440838/execroot/__main__/bazel-out/host/bin/external/build_bazel_rules_apple/tools/bundletool/bundletool.runfiles/build_bazel_rules_apple/tools/bundletool/bundletool.py", line 116, in run
    f.get('contents_only', False), out_zip)
  File "/private/var/tmp/_bazel_mrebello/c8c34ff25de1ec7ce8f729468e440838/execroot/__main__/bazel-out/host/bin/external/build_bazel_rules_apple/tools/bundletool/bundletool.runfiles/build_bazel_rules_apple/tools/bundletool/bundletool.py", line 153, in _add_files
    self._write_entry(dest, f.read(), fexec, out_zip)
  File "/private/var/tmp/_bazel_mrebello/c8c34ff25de1ec7ce8f729468e440838/execroot/__main__/bazel-out/host/bin/external/build_bazel_rules_apple/tools/bundletool/bundletool.runfiles/build_bazel_rules_apple/tools/bundletool/bundletool.py", line 215, in _write_entry
    out_zip.writestr(zipinfo, data)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/zipfile.py", line 1232, in writestr
    self._writecheck(zinfo)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/zipfile.py", line 1114, in _writecheck
    " would require ZIP64 extensions")
zipfile.LargeZipFile: Filesize would require ZIP64 extensions
Target //:ios_dist failed to build
INFO: Elapsed time: 1571.248s, Critical Path: 242.78s
INFO: 5180 processes: 5180 local.
FAILED: Build did NOT complete successfully
```

keith:
Ideally we could fix that issue by updating rules_apple with this:

```patch
diff --git i/tools/bundletool/bundletool.py w/tools/bundletool/bundletool.py
index e8a8a23..14ee471 100644
--- i/tools/bundletool/bundletool.py
+++ w/tools/bundletool/bundletool.py
@@ -106,7 +106,7 @@ class Bundler(object):
     bundle_merge_zips = self._control.get('bundle_merge_zips', [])
     root_merge_zips = self._control.get('root_merge_zips', [])
 
-    with zipfile.ZipFile(output_path, 'w') as out_zip:
+    with zipfile.ZipFile(output_path, 'w', allowZip64=True) as out_zip:
       for z in bundle_merge_zips:
         dest = os.path.normpath(os.path.join(bundle_path, z['dest']))
         self._add_zip_contents(z['src'], dest, out_zip)
```

But there appears to be a bug in python2 that prevents this from working too https://bugs.python.org/issue23306

rebello95:
I'm attempting locally without `--copt -ggdb3` to see if that helps with the size overload. We were seeing a ~4GB artifact before it failed.

rebello95:
Another update: Dropping the above flags still resulted in the zip error, as the artifacts are still ~800MB. Going to circle back to this with @keith when he's back on Monday

.bazelrc
```diff
@@ -19,3 +21,8 @@ build \
 build:ios --define=manual_stamp=manual_stamp
 
 build:android --fat_apk_cpu=x86,x86_64,armeabi-v7a,arm64-v8a
+
+build:release \
+  --config=sizeopt \
+  --copt -ggdb3 \
+  --ios_multi_cpus=i386,x86_64,armv7,arm64
```
goaway: I _think_ this can be updated to use `---config=fat` now. If not we should add Android's archs in here also (since it's no longer fat by default).

rebello95: Yep, should be able to

rebello95:
Update: Ran without specifying the `--copt -ggdb3` option, and artifacts are still ~740MB each and thus yield the same size error. @keith is going to help investigate this a bit further

keith:
Ok so overall here's what I discovered:

- The _massive_ size is definitely caused by debug info for all of envoy's dependencies. Right now it passes `-ggdb3` for everything https://github.com/envoyproxy/envoy/blob/5dd7cc99af303cb92741718836ad4e73639d630f/bazel/envoy_internal.bzl#L36-L37 which includes the maximum amount of debug info (although AFAICT in clang there is no difference between this and just `-g`)
- With no debug info the entire archive is down from ~4gb -> ~250mbs, so that's the smallest expected size, but we probably want _some_ of this debug info, but probably not _everything_ as it is now. Likely there's an opportunity in core envoy to not pass this flag to everything
- Using python3 instead of python2 (which is the default if we upgrade to bazel 0.27.0) to build fixes the python exception here because python3 sets `allowZip64` to True by default, and doesn't have the same bug as python2 when creating a zip this large

For now I would recommend using python3 with bazel 0.27.0 (which upstream envoy supports as of last week) and ignoring the size issue, and once we're further along we can try and reduce that. But the 4gb size might be too large for now too, in which case we'll need to find a way to disable debug info for envoy libraries.

rebello95:
Thanks so much for investigating, @keith! I'm on board with the temporary fix of using python3 / bazel 0.27.0 if that seems reasonable to @mattklein123 and @goaway, and we can track the upstream improvement in a separate issue. If this is the direction we want to move in, I can look into seeing if we have any issues with upgrading to 0.27.0 (I think @junr03 ran into a few with Android a few weeks back which may or may not still be relevant)

mattklein123:
Feel free to skip `-ggdb3` for now if that makes things easier. I would add a TODO about this. Optimally we would produce full symbols and just strip them and put them elsewhere, but I'm not sure what is possible.

keith:
For future reference here's the python2 bug https://stackoverflow.com/questions/30376224/python-gzip-overflowerror-size-does-not-fit-in-an-int/38177538#38177538

rebello95:
Going to put this on hold until https://github.com/lyft/envoy-mobile/issues/116 is addressed, upgrading bazel to 0.27.0

rebello95:
@keith looks like this still fails with the latest master and bazel 0.27 with a similar error:

```
ERROR: /Users/mrebello/Desktop/envoy-mobile/library/swift/src/BUILD:9:1: Creating framework zip for Envoy failed (Segmentation fault): zipper failed: error executing command 
  (cd /private/var/tmp/_bazel_mrebello/b847463c9e3427e81077e7fbd7d85e6e/execroot/__main__ && \
  exec env - \
  external/bazel_tools/tools/zip/zipper/zipper c bazel-out/darwin-opt/bin/library/swift/src/Envoy.zip 'Envoy.framework/Envoy=bazel-out/darwin-opt/bin/library/swift/src/Envoy.fat' 'Envoy.framework/Modules/Envoy.swiftmodule/arm64.swiftdoc=bazel-out/ios-arm64-min10.0-applebin_ios-ios_arm64-opt/bin/library/swift/src/Envoy.swiftdoc' 'Envoy.framework/Modules/Envoy.swiftmodule/arm64.swiftmodule=bazel-out/ios-arm64-min10.0-applebin_ios-ios_arm64-opt/bin/library/swift/src/Envoy.swiftmodule' 'Envoy.framework/Modules/Envoy.swiftmodule/arm.swiftdoc=bazel-out/ios-armv7-min10.0-applebin_ios-ios_armv7-opt/bin/library/swift/src/Envoy.swiftdoc' 'Envoy.framework/Modules/Envoy.swiftmodule/arm.swiftmodule=bazel-out/ios-armv7-min10.0-applebin_ios-ios_armv7-opt/bin/library/swift/src/Envoy.swiftmodule' 'Envoy.framework/Modules/Envoy.swiftmodule/i386.swiftdoc=bazel-out/ios-i386-min10.0-applebin_ios-ios_i386-opt/bin/library/swift/src/Envoy.swiftdoc' 'Envoy.framework/Modules/Envoy.swiftmodule/i386.swiftmodule=bazel-out/ios-i386-min10.0-applebin_ios-ios_i386-opt/bin/library/swift/src/Envoy.swiftmodule' 'Envoy.framework/Modules/Envoy.swiftmodule/x86_64.swiftdoc=bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-opt/bin/library/swift/src/Envoy.swiftdoc' 'Envoy.framework/Modules/Envoy.swiftmodule/x86_64.swiftmodule=bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-opt/bin/library/swift/src/Envoy.swiftmodule')
Execution platform: @bazel_tools//platforms:host_platform
Uncompressed input jar has size 18446744072517795054, which exceeds the maximum supported output size 4294967295.
Assuming that ijar will be smaller and hoping for the best.
Target //:ios_dist failed to build
INFO: Elapsed time: 64.879s, Critical Path: 62.23s
INFO: 12 processes: 12 local.
FAILED: Build did NOT complete successfull
```

rebello95:
Circling back to this, rebuilding after merging master yields basically the same error:

```
INFO: Deleting stale sandbox base /private/var/tmp/_bazel_mrebello/19484c32c8304d4a1f2e8d399f135a39/sandbox
ERROR: /Users/mrebello/Development/envoy-mobile-2/library/swift/src/BUILD:5:1: Creating framework zip for Envoy failed (Segmentation fault): zipper failed: error executing command 
  (cd /private/var/tmp/_bazel_mrebello/19484c32c8304d4a1f2e8d399f135a39/execroot/__main__ && \
  exec env - \
  external/bazel_tools/tools/zip/zipper/zipper c bazel-out/darwin-opt/bin/library/swift/src/Envoy.zip 'Envoy.framework/Envoy=bazel-out/darwin-opt/bin/library/swift/src/Envoy.fat' 'Envoy.framework/Headers/Envoy-Swift.h=bazel-out/ios-arm64-min10.0-applebin_ios-ios_arm64-opt/bin/library/swift/src/ios_framework_archive-Swift.h' 'Envoy.framework/Modules/Envoy.swiftmodule/arm64.swiftdoc=bazel-out/ios-arm64-min10.0-applebin_ios-ios_arm64-opt/bin/library/swift/src/Envoy.swiftdoc' 'Envoy.framework/Modules/Envoy.swiftmodule/arm64.swiftmodule=bazel-out/ios-arm64-min10.0-applebin_ios-ios_arm64-opt/bin/library/swift/src/Envoy.swiftmodule' 'Envoy.framework/Headers/Envoy-Swift.h=bazel-out/ios-armv7-min10.0-applebin_ios-ios_armv7-opt/bin/library/swift/src/ios_framework_archive-Swift.h' 'Envoy.framework/Modules/Envoy.swiftmodule/arm.swiftdoc=bazel-out/ios-armv7-min10.0-applebin_ios-ios_armv7-opt/bin/library/swift/src/Envoy.swiftdoc' 'Envoy.framework/Modules/Envoy.swiftmodule/arm.swiftmodule=bazel-out/ios-armv7-min10.0-applebin_ios-ios_armv7-opt/bin/library/swift/src/Envoy.swiftmodule' 'Envoy.framework/Headers/Envoy-Swift.h=bazel-out/ios-i386-min10.0-applebin_ios-ios_i386-opt/bin/library/swift/src/ios_framework_archive-Swift.h' 'Envoy.framework/Modules/Envoy.swiftmodule/i386.swiftdoc=bazel-out/ios-i386-min10.0-applebin_ios-ios_i386-opt/bin/library/swift/src/Envoy.swiftdoc' 'Envoy.framework/Modules/Envoy.swiftmodule/i386.swiftmodule=bazel-out/ios-i386-min10.0-applebin_ios-ios_i386-opt/bin/library/swift/src/Envoy.swiftmodule' 'Envoy.framework/Headers/Envoy-Swift.h=bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-opt/bin/library/swift/src/ios_framework_archive-Swift.h' 'Envoy.framework/Modules/Envoy.swiftmodule/x86_64.swiftdoc=bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-opt/bin/library/swift/src/Envoy.swiftdoc' 'Envoy.framework/Modules/Envoy.swiftmodule/x86_64.swiftmodule=bazel-out/ios-x86_64-min10.0-applebin_ios-ios_x86_64-opt/bin/library/swift/src/Envoy.swiftmodule')
Execution platform: @bazel_tools//platforms:host_platform
Uncompressed input jar has size 18446744072738329018, which exceeds the maximum supported output size 4294967295.
Assuming that ijar will be smaller and hoping for the best.
Target //:ios_dist failed to build
INFO: Elapsed time: 71.367s, Critical Path: 28.89s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```

@goaway pointed out yesterday that `18446744072738329018` seems like a crazy-large number. Looking into this a little further, it appears that [Bazel caps the size at 4GB then attempts to do the zipping](https://github.com/bazelbuild/bazel/blob/master/third_party/ijar/zip.cc#L1080-L1087) regardless of the originally estimated size.

That said, the error above is actually a segfault:

```
Creating framework zip for Envoy failed (Segmentation fault): zipper failed: error executing command 
  (cd /private/var/tmp/_bazel_mrebello/19484c32c8304d4a1f2e8d399f135a39/execroot/__main__
```

@keith do you have any more ideas on how we might be able to work around this?

rebello95:
Updated the PR with a new set of configs per @goaway's suggestion (thank you!).

Size diff of the zipped `.framework` on this PR right now:
- `--fat`: 182.1MB
- `--release`: 81.6MB

I think we should go with this for now.

keith:
Not having symbols will degrade crash reporting at least in Xcode. But with bitcode it should be fine. Which you actually might need to enable here

rebello95:
I'll add bitcode enablement in a follow-up PR 👍

rebello95:
Being tracked in https://github.com/lyft/envoy-mobile/issues/208

rebello95:
https://github.com/lyft/envoy-mobile/pull/446

PR #164, opened 2019-06-21
lib: make iOS interface instantiable and internalize thread

goaway:

Description: Allow multiple Envoy instances to be created with their own configuration and tied to their own thread.
Risk Level: Low
Testing: Built and ran apps locally

### DIFF
--- a/examples/swift/hello_world/AppDelegate.swift
+++ b/examples/swift/hello_world/AppDelegate.swift
@@ -7,21 +7,16 @@ private enum ConfigLoadError: Error {
 
 @UIApplicationMain
 final class AppDelegate: UIResponder, UIApplicationDelegate {
+  var envoy: Envoy?
   var window: UIWindow?
 
   func application(
     _ application: UIApplication,
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool
   {
-    do {
-      let configYaml = try self.loadEnvoyConfig() as NSString
-      NSLog("Loading config:\n\(configYaml)")
-      Thread.detachNewThread {
-        run_envoy(configYaml.utf8String)
-      }
-    } catch let error {
-      NSLog("Failed to load config: \(error)")
-    }
+    let configYaml = try! self.loadEnvoyConfig()
+    NSLog("Loading config:\n\(configYaml)")
+    self.envoy = Envoy(config: configYaml)
 
     let window = UIWindow(frame: UIScreen.main.bounds)
     window.rootViewController = ViewController()
--- a/examples/swift/hello_world/AppDelegate.swift
+++ b/examples/swift/hello_world/AppDelegate.swift
@@ -7,16 +7,17 @@ private enum ConfigLoadError: Error {
 
 @UIApplicationMain
 final class AppDelegate: UIResponder, UIApplicationDelegate {
-  var envoy: Envoy?
+  private var envoy: Envoy!
   var window: UIWindow?
 
   func application(
     _ application: UIApplication,
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool
   {
-    let configYaml = try! self.loadEnvoyConfig()
-    NSLog("Loading config:\n\(configYaml)")
-    self.envoy = Envoy(config: configYaml)
+    NSLog("Loading config")
+    let envoyConfig = try! self.loadEnvoyConfig()
+    NSLog("Loaded config:\n\(envoyConfig)")
+    self.envoy = Envoy(config: envoyConfig)
 
     let window = UIWindow(frame: UIScreen.main.bounds)
     window.rootViewController = ViewController()
--- a/library/objective-c/BUILD
+++ b/library/objective-c/BUILD
@@ -16,9 +16,5 @@ objc_library(
     ],
     copts = ["-std=c++14"],
     visibility = ["//visibility:public"],
-    #sdk_dylibs = [
-    #    "resolv.9",
-    #    "c++",
-    #],
     deps = ["//library/common:envoy_main_interface_lib"],
 )
--- a/library/objective-c/BUILD
+++ b/library/objective-c/BUILD
@@ -1,0 +1,24 @@
+licenses(["notice"])  # Apache 2
+
+filegroup(
+    name = "envoy_framework_headers",
+    srcs = [
+        "Envoy.h",
+    ],
+    visibility = ["//visibility:public"],
+)
+
+objc_library(
+    name = "envoy_objc_interface_lib",
+    srcs = [
+        "Envoy.h",
+        "Envoy.mm",
+    ],
+    copts = ["-std=c++14"],
+    visibility = ["//visibility:public"],
+    #sdk_dylibs = [
+    #    "resolv.9",
+    #    "c++",
+    #],
+    deps = ["//library/common:envoy_main_interface_lib"],
+)
--- a/library/objective-c/Envoy.h
+++ b/library/objective-c/Envoy.h
@@ -1,8 +1,14 @@
 #import <Foundation/Foundation.h>
 
 @interface Envoy : NSObject
+
+/// Indicates whether this Envoy instance is currently active and running.
 @property (nonatomic, readonly, getter=isRunning) BOOL running;
+
+/// Returns whether the Envoy instance is terminated.
 @property (nonatomic, readonly, getter=isTerminated) BOOL terminated;
 
+/// Create a new Envoy instance. The Envoy runner NSThread is started as part of instance
+/// initialization with the configuration provided.
 - (id)initWithConfig:(NSString*)config;
 @end
--- a/library/objective-c/Envoy.h
+++ b/library/objective-c/Envoy.h
@@ -1,0 +1,8 @@
+#import <Foundation/Foundation.h>
+
+@interface Envoy : NSObject
+@property (nonatomic, readonly, getter=isRunning) BOOL running;
+@property (nonatomic, readonly, getter=isTerminated) BOOL terminated;
+
+- (id)initWithConfig:(NSString*)config;
+@end
--- a/library/objective-c/Envoy.h
+++ b/library/objective-c/Envoy.h
@@ -10,5 +10,5 @@
 
 /// Create a new Envoy instance. The Envoy runner NSThread is started as part of instance
 /// initialization with the configuration provided.
-- (id)initWithConfig:(NSString*)config;
+- (instancetype)initWithConfig:(NSString*)config;
 @end
--- a/library/objective-c/Envoy.h
+++ b/library/objective-c/Envoy.h
@@ -3,10 +3,10 @@
 @interface Envoy : NSObject
 
 /// Indicates whether this Envoy instance is currently active and running.
-@property (nonatomic, readonly, getter=isRunning) BOOL running;
+@property (nonatomic, readonly) BOOL isRunning;
 
-/// Returns whether the Envoy instance is terminated.
-@property (nonatomic, readonly, getter=isTerminated) BOOL terminated;
+/// Indicates whether the Envoy instance is terminated.
+@property (nonatomic, readonly) BOOL isTerminated;
 
 /// Create a new Envoy instance. The Envoy runner NSThread is started as part of instance
 /// initialization with the configuration provided.
--- a/library/objective-c/Envoy.mm
+++ b/library/objective-c/Envoy.mm
@@ -1,0 +1,45 @@
+#import "library/objective-c/Envoy.h"
+
+#import "library/common/main_interface.h"
+
+
+@interface Envoy ()
+@property (nonatomic, strong) NSThread *runner;
+@end
+
+@implementation Envoy
+
+@synthesize runner;
+
+// Create a new Envoy instance. The Envoy runner NSThread is started as part of instance
+// initialization with the configuration provided.
+- (id)initWithConfig:(NSString *)config {
+  self.runner = [[NSThread alloc] initWithTarget:self selector:@selector(run:) object:config];
+  [self.runner start];
+}
+
+// Returns whether this Envoy instance is currently active and running.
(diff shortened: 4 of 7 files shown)


examples/swift/hello_world/AppDelegate.swift
```diff
@@ -7,21 +7,16 @@ private enum ConfigLoadError: Error {
 
 @UIApplicationMain
 final class AppDelegate: UIResponder, UIApplicationDelegate {
+  var envoy: Envoy?
```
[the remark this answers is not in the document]
rebello95: Or even simpler (the yaml loading function would have to be made static/privately global):

```swift
private let envoy = Envoy(config: loadYaml())
```

library/objective-c/BUILD
```diff
@@ -0,0 +1,24 @@
...
+
+objc_library(
+    name = "envoy_objc_interface_lib",
+    srcs = [
+        "Envoy.h",
+        "Envoy.mm",
+    ],
+    copts = ["-std=c++14"],
+    visibility = ["//visibility:public"],
+    #sdk_dylibs = [
```
rebello95: Do we need these?

library/objective-c/Envoy.mm
```diff
@@ -0,0 +1,45 @@
...
+
+@interface Envoy ()
+@property (nonatomic, strong) NSThread *runner;
+@end
+
+@implementation Envoy
+
+@synthesize runner;
+
+// Create a new Envoy instance. The Envoy runner NSThread is started as part of instance
```
rebello95: These docstrings should probably live in the header file, right?

library/objective-c/Envoy.mm
```diff
@@ -0,0 +1,45 @@
...
+// initialization with the configuration provided.
+- (id)initWithConfig:(NSString *)config {
```
rebello95:
```suggestion
- (instancetype)initWithConfig:(NSString *)config {
```

library/objective-c/Envoy.mm
```diff
@@ -0,0 +1,45 @@
...
+  self.runner = [[NSThread alloc] initWithTarget:self selector:@selector(run:) object:config];
```
rebello95: Would it make sense to move these to the common `init` from `NSObject` so we can avoid the null checks below?

library/objective-c/Envoy.mm
```diff
@@ -0,0 +1,45 @@
...
+  [self.runner start];
```
rebello95: What are your thoughts on exposing a `start` function instead of starting the instance here? The reason I ask is that this implementation requires consumers to create an instance of `Envoy` and keep it alive, but they may never need to call into it. In Swift this can lead to unused variable warnings. In practice, consumers would usually call `start()` immediately after instantiating

library/objective-c/Envoy.mm
```diff
@@ -0,0 +1,45 @@
...
+}
+
+// Returns whether this Envoy instance is currently active and running.
+- (BOOL)isRunning {
+  return self.runner != NULL && self.runner.isExecuting;
```
rebello95: I don't think these `!= NULL` checks are required, since this should return `FALSE` if it doesn't exist

library/objective-c/Envoy.mm
```diff
@@ -0,0 +1,45 @@
...
+
+#pragma mark private
+
+- (void)run:(NSString *)config {
+  try {
+    run_envoy(config.UTF8String);
+  } catch (NSException *e) {
+    NSLog(@"Envoy exception: %@", e);
+    NSDictionary *userInfo = @{ @"exception": e};
+    [NSNotificationCenter.defaultCenter postNotificationName:@"EnvoyException"
```
rebello95: Do you think we should prefix this notification name with something to better differentiate? Would also be nice to document it somewhere

examples/swift/hello_world/AppDelegate.swift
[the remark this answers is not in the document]
goaway: I just made it optional to get rid of a warning about it being unable to be synthesized without an initial value. Is it safe/reasonable to be read and parse a resource in order to load this up statically? I'm not opposed to it, but it seemed like a lot to be doing outside of the standard execution path.

examples/swift/hello_world/AppDelegate.swift
```diff
@@ -7,21 +7,16 @@ private enum ConfigLoadError: Error {
...
-      let configYaml = try self.loadEnvoyConfig() as NSString
-      NSLog("Loading config:\n\(configYaml)")
-      Thread.detachNewThread {
-        run_envoy(configYaml.utf8String)
-      }
-    } catch let error {
-      NSLog("Failed to load config: \(error)")
-    }
+    let configYaml = try! self.loadEnvoyConfig()
+    NSLog("Loading config:\n\(configYaml)")
```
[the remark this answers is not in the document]
goaway: It references it though? I might be misunderstanding

library/objective-c/BUILD
goaway: Apparently not :)

library/objective-c/Envoy.mm
goaway: Good call

library/objective-c/Envoy.mm
goaway: oo that's right, things have gotten fancier :)

goaway: I should be calling super init as well, huh

library/objective-c/Envoy.mm
goaway: Hm. I guess maybe if we had a default config we could pass through?

goaway: At this point, not sure what would make sense there though, tbh.

library/objective-c/Envoy.mm
goaway: I originally had it set up this way. I backtracked because Java's best run-once guarantees are built into object initialization. So I updated this to match.

Once the direct API is in place, consumers will be making other calls to the library. And on the other hand, someone may have a real reason to want a separate `start` call at some point.

If this ends up being an issue, I don't mind changing it, but for now, it simplified things on the Java side.

library/objective-c/Envoy.mm
goaway: That's right, I had forgotten about objective-c's safety/fanciness there. Will update.

library/objective-c/Envoy.mm
goaway: If we think this is good pattern (and I think maybe it's a useful one), then yeah we should document.

What sort of prefix did you have in mind?

goaway: (I was sort of thinking of "Envoy" as a prefix fwiw)

examples/swift/hello_world/AppDelegate.swift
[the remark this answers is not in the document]
rebello95: ah yea you're right. Maybe makes more sense to log `Loading config`, then `Loading config: %@` afterwards

examples/swift/hello_world/AppDelegate.swift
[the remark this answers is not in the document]
rebello95: I see. Another option would be to make it a `private lazy var`, then call `.run()` on it in the app launch function (making it non-optional and also preventing instantiation until it's needed)

library/objective-c/Envoy.mm
rebello95: Heh

library/objective-c/Envoy.mm
rebello95: Objective-C normally asks for 3-letter class prefixes, so maybe we could do the same for notifications 🤔

library/objective-c/Envoy.mm
rebello95: Okay that's fine with me, we can revisit later if needed

library/objective-c/Envoy.h
```diff
@@ -0,0 +1,14 @@
+#import <Foundation/Foundation.h>
+
+@interface Envoy : NSObject
+
+/// Indicates whether this Envoy instance is currently active and running.
+@property (nonatomic, readonly, getter=isRunning) BOOL running;
```
rebello95: Just realized these are already `readonly`:
```suggestion
@property (nonatomic, readonly) BOOL isRunning;
```

library/objective-c/Envoy.h
```diff
@@ -0,0 +1,14 @@
...
+
+/// Returns whether the Envoy instance is terminated.
+@property (nonatomic, readonly, getter=isTerminated) BOOL terminated;
```
rebello95:
```suggestion
@property (nonatomic, readonly) BOOL isTerminated;
```

library/objective-c/Envoy.mm
```diff
@@ -0,0 +1,45 @@
...
+@end
+
+@implementation Envoy
+
+@synthesize runner;
+
+- (instancetype)initWithConfig:(NSString *)config {
+  if (self = [super init]) {
+    self.runner = [[NSThread alloc] initWithTarget:self selector:@selector(run:) object:config];
+    [self.runner start];
```
rebello95: I think we need to `return self;` here?

rebello95: Actually, probably simplest to remove the `else` and simply `return self;` below

examples/swift/hello_world/AppDelegate.swift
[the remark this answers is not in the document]
goaway: `lazy` sounds like exactly what we want. Good call.

library/objective-c/Envoy.mm
goaway: Cool, yep can definitely revisit.

library/objective-c/Envoy.mm
goaway: How strict is that convention these days? My inclination is to eschew it, unless it's really going to bother people.

library/objective-c/Envoy.h
goaway: Yeah I forget why I did it that other way.

library/objective-c/Envoy.mm
rebello95: It can still cause collisions in Objective-C classes 🤷‍♂

rebello95: Though, notifications might also be fine

library/objective-c/Envoy.mm
goaway: sg

library/objective-c/Envoy.mm
goaway: I don't care that much, but I can't think of a good letter code that's not going to risk just making things more confusing. "Envoy" as a prefix seems (to me) pretty low-risk in terms of likelihood to cause a collision.

library/objective-c/Envoy.mm
goaway: Good here? Since obj-c magics away the nil checks for us anyways.

rebello95: Yea that's fine

examples/swift/hello_world/AppDelegate.swift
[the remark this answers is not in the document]
rebello95: Looks like this still needs to be updated

library/objective-c/Envoy.mm
rebello95: Fine with me to leave this as-is for now

examples/swift/hello_world/AppDelegate.swift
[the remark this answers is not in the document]
goaway: So it turns out `lazy` isn't quite the same as Kotlin's `lateinit`. One pattern folks seem to use for late-init members in Swift (e.g. view components) is implicitly-unwrapped optionals. So I gave that a whirl. Let me know if you think that's a good compromise

library/objective-c/Envoy.mm
goaway: 👍

examples/swift/hello_world/AppDelegate.swift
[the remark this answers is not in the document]
rebello95: The IUO does work (probably better than optional here). I'm fine leaving it as is, but for the record you could also do something like this (not necessarily suggesting you do so, though):

```swift
private lazy var envoy: Envoy = {
    let envoyConfig = try! self.loadEnvoyConfig()
    return Envoy(config: envoyConfig)
}()
```

library/objective-c/Envoy.mm
```diff
@@ -0,0 +1,45 @@
...
+  }
+  return self;
+}
+
+- (BOOL)isRunning {
+  return self.runner.isExecuting;
+}
+
+- (BOOL)isTerminated {
+  return self.runner.isFinished;
```
rebello95: This one we might want `self.runner == nil || self.runner.isFinished` I think?

examples/swift/hello_world/AppDelegate.swift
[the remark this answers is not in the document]
goaway: Yeah, I did something very similar, but I realized I'd still need to reference Envoy somehow to actually trigger initialization. And couldn't think of anything that made more sense than simply carrying out initialization explicitly. No really strong feelings though.

library/objective-c/Envoy.mm
goaway: yep, good catch

goaway: Oh actually, I think the current logic is fine - Terminated is only true if it's gone through all other lifecycle phases and then stopped. So `runner` being `nil` means it hasn't started yet, in other words, it's not terminated. (I think eventually we can formalize this all better.)

goaway: i.e. the current logic is the same as what was originally there; it just takes advantage of objective-c's nil-dispatch behavior.

goaway:
Thanks @rebello95 !

goaway:
This now also completes https://github.com/lyft/envoy-mobile/issues/46

PR #165, opened 2019-06-21
bazel: Move aar build specific rules into library directory

buildbreaker:
We're not yet able to have specific kotlin and java directory specific rules just yet because of the transitive dependency problem highlighted in the comment.

Description: move aar build specific rules into the `library/BUILD`
Risk Level: low
Testing: ci, local
Docs Changes: na
Release Notes: na
[Optional Fixes #Issue]
[Optional Deprecated:]


buildbreaker:
The JNI file needs to be updated. Will fix in a moment :(

BUILD
```diff
@@ -29,44 +26,9 @@ touch $@
     stamp = True,
 )
 
-aar_with_jni(
```
goaway: I think we should have a top-level alias to this rule, in keeping with our primary build targets being buildable from the root.

library/java/io/envoyproxy/envoymobile/EnvoyEngine.java
```diff
@@ -1,4 +1,4 @@
-package io.envoyproxy.envoymobile;
+package io.envoyproxy.envoymobile.envoyengine;
```
goaway: Why the package change? Not that big a deal, just curious.

buildbreaker: It was a bit strange to have the kotlin layer have the same package name as the envoy engine layer since they're being built separately.

BUILD
buildbreaker: It's still buildable from the root since we do this in the `//:android_dist` rule. The current docs are still valid 😄 

The current implementation detail is that it's an `aar_with_jni` macro. We can change this however we want to and the root level declaration will stay the same.

goaway: What I mean is I think we should also have a root level `android_aar` target. This parallels iOS, and is a "purer" target from a bazel perspective, since bazel philosphy is that builds should be hermetic. This implies, among other things, that they shouldn't modify the workspace.

Our dist rule adopts a different philosophy to allow us to build the apps and more easily locate distributable artifacts. So that was sort of the idea behind having both options. If someone imports our project via bazel rules or as a submodule into another bazel project, they will prefer the `android_aar` rule as their primary target still.

Does that kinda make sense? I think this is achievable with a simple alias.

library/java/io/envoyproxy/envoymobile/EnvoyEngine.java
goaway: Sounds good. Should we just call it 'engine'?

PR #166, opened 2019-06-21
android: Update examples to display errors with a red background

buildbreaker:
This is just a change to allow us to more easily visualize when Envoy fails. We're repurposing the examples here to also display this. This way others can also install the example application and be able to visualize the success/failures of Envoy in the wild.

Changes:
1. An error view item background is: #ff3651
1. Added a request count on the top of the item view (Screen below)
1. Moved all the duplicated classes for java and kotlin examples under a shared lib

Related to https://github.com/lyft/envoy-mobile/issues/129: App lifecycle & network condition handling.

Description: Update examples to display errors with a red background
Risk Level: low
Testing: manual
Docs Changes: n/a
Release Notes: na/
[Optional Fixes #Issue]
[Optional Deprecated:]

Screenshot_20190621-164638


examples/java/hello_world/MainActivity.java
```diff
@@ -44,14 +45,14 @@ protected void onCreate(Bundle savedInstanceState) {
...
-    String config = null;
+    String config;
     try {
       config = loadEnvoyConfig(getBaseContext(), R.raw.config);
     } catch (RuntimeException e) {
       Log.d("MainActivity", "exception getting config.", e);
       throw new RuntimeException("Can't get config to run envoy.");
     }
-    envoy = new Envoy(context, config);
+    Envoy envoy = new Envoy(context, config);
```
goaway: Even if it's not strictly necessary at this time, I think our apps should follow a pattern of retaining a reference to their Envoy instance, since eventually they'll need to do this anyways to utilize it. (This is a side effect of us making the decision to use an instantiated/non-singleton approach.)

examples/kotlin/hello_world/BUILD
```diff
@@ -18,18 +18,16 @@ kt_android_library(
...
     ],
     custom_package = "io.envoyproxy.envoymobile.helloenvoykotlin",
     manifest = "AndroidManifest.xml",
-    resource_files = glob([
-        "res/**/*.xml",
-        "res/**/*.yaml",
-    ]),
+    resource_files = [
+        "res/layout/activity_main.xml",
+        "res/raw/config.yaml",
```
goaway: nice, thanks for cleaning this up

examples/java/hello_world/MainActivity.java
goaway: Sorry, so what I also meant by this, is can we put the L36 back (`private Envoy envoy;`) member variable. :)

PR #171, opened 2019-06-25
build: add arch options configuration

goaway:

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -16,6 +16,6 @@ build \
 # static framework.
 # More information on stamping can be found here:
 # https://github.com/envoyproxy/envoy/tree/master/bazel#enabling-optional-features
-build:ios --define=manual_stamp=manual_stamp
+build:ios --define=manual_stamp=manual_stamp --ios_multi_cpus=i386,x86_64,armv7,arm64
 
 build:android --fat_apk_cpu=x86,x86_64,armeabi-v7a,arm64-v8a
--- a/.bazelrc
+++ b/.bazelrc
@@ -16,7 +16,7 @@ build \
 # static framework.
 # More information on stamping can be found here:
 # https://github.com/envoyproxy/envoy/tree/master/bazel#enabling-optional-features
-build:ios --define=manual_stamp=manual_stamp 
+build:ios --define=manual_stamp=manual_stamp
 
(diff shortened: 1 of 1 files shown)


keith:
I don't think you want this all the time since it will 4x builds where you pass `--config=ios` which is why @rebello95 was adding a new config for it https://github.com/lyft/envoy-mobile/pull/155

goaway:
Was just thinking that it simplified dev and paralleled Android.

goaway:
Didn't see @rebello95 's PR before - I still kinda think it's simpler to just build the archs up front by default. Bazel will still cache most artifacts right?

keith:
It will cache them yea, but if you change something build related that causes an envoy rebuild it will be 4 times longer than before

goaway:
Yeah, true. I guess my line of thinking is: this project is building something for distribution and so it's useful for the default to be the most supportive/permissive configuration possible.

One less thing for someone to miss when they go and try to use the project. No terribly strong feelings about it though.

goaway:
(I do still think a release configuration makes sense regardless of where we land on this.)

rebello95:
> Yeah, true. I guess my line of thinking is: this project is building something for distribution and so it's useful for the default to be the most supportive/permissive configuration possible.

I agree that this is generally true, but I am a bit concerned about the hit to the local development workflow for us and other contributors when changes are made that trigger rebuilds as @keith mentioned.

What are your thoughts on adding another build configuration that includes these architectures which can be used in `release` and optionally appended for local builds? Something like `--config=devices`?

.bazelrc
```diff
@@ -16,6 +16,10 @@ build \
 # static framework.
 # More information on stamping can be found here:
 # https://github.com/envoyproxy/envoy/tree/master/bazel#enabling-optional-features
-build:ios --define=manual_stamp=manual_stamp
+build:ios --define=manual_stamp=manual_stamp
```
rebello95: nit: extra whitespace

goaway:
Updated. So, e.g., you can do:
```
--config:ios --config:device
--config:android --config:fat
```

One difference is if you don't specify an arch target, Android will default to _only_ emulator now (which will also mean faster builds).

.bazelrc
goaway: good catch ty

PR #173, opened 2019-06-25
logging: add configurable log-level to startup params

goaway:
Description: Adds configurable log-level to common interface, and support to both iOS and Android interfaces. Defaults to log-level info.
Risk Level: Low
Testing: Locally compiled and tested.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -26,7 +26,7 @@ IndentWidth: 2
 ObjCBlockIndentWidth: 2
 ObjCSpaceAfterProperty: true
 ObjCSpaceBeforeProtocolList: true
-PointerAlignment: Left
+PointerAlignment: Right
 SortIncludes: false
 ...
 
--- a/library/kotlin/io/envoyproxy/envoymobile/Envoy.kt
+++ b/library/kotlin/io/envoyproxy/envoymobile/Envoy.kt
@@ -4,7 +4,7 @@ import android.content.Context
 import io.envoyproxy.envoymobile.engine.EnvoyEngine
 
 // Wrapper class that allows for easy calling of Envoy's JNI interface in native Java.
-class Envoy(
+class Envoy @JvmOverloads constructor(
     context: Context,
     config: String,
     logLevel: String = "info"
--- a/library/kotlin/io/envoyproxy/envoymobile/Envoy.kt
+++ b/library/kotlin/io/envoyproxy/envoymobile/Envoy.kt
@@ -6,7 +6,8 @@ import io.envoyproxy.envoymobile.engine.EnvoyEngine
 // Wrapper class that allows for easy calling of Envoy's JNI interface in native Java.
 class Envoy(
     context: Context,
-    config: String
+    config: String,
+    logLevel: String = "info"
 ) {
 
   // Dedicated thread for running this instance of Envoy.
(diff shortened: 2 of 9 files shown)


library/kotlin/io/envoyproxy/envoymobile/Envoy.kt
```diff
@@ -4,9 +4,10 @@ import android.content.Context
 import io.envoyproxy.envoymobile.engine.EnvoyEngine
 
 // Wrapper class that allows for easy calling of Envoy's JNI interface in native Java.
-class Envoy(
+class Envoy @JvmOverloads constructor(
     context: Context,
-    config: String
+    config: String,
+    logLevel: String = "info"
```
buildbreaker: What do you think about introducing an enum for this?

buildbreaker:
Looks good. What are your thoughts on having this be part of a configuration object in the future? Like bundling the log level and config in an object we pass into this layer.

library/kotlin/io/envoyproxy/envoymobile/Envoy.kt
goaway: I did think about it, and Envoy uses one internally.

However, currently we have to pass this as a string, and there's no portable enum type we could expose for this (so we'd have to define the enum on both platforms and still pass a string ultimately).

This seemed good enough for now to me. ¯\\\_(ツ)\_/¯

goaway:
@buildbreaker I think bundling this into the configuration interface could definitely make sense depending on where we land with it.

.clang-format
```diff
@@ -26,7 +26,7 @@ IndentWidth: 2
 ObjCBlockIndentWidth: 2
 ObjCSpaceAfterProperty: true
 ObjCSpaceBeforeProtocolList: true
-PointerAlignment: Left
+PointerAlignment: Right
```
goaway: Just gonna sneak this in here.

goaway: (It's Apple's convention and `Left` wasn't getting applied consistently by clang-format anyways.)

library/kotlin/io/envoyproxy/envoymobile/Envoy.kt
buildbreaker: Yeah, I think we'll have to create one eventually. Let's keep that in mind on the next iteration for configurations/logs

.clang-format
rebello95: No opinion here, but for context it was `Left` because that's what upstream Envoy uses for C++

library/kotlin/io/envoyproxy/envoymobile/Envoy.kt
rebello95: I definitely agree that this should be an enum at some point. What about making constants to use instead for now? Feels a little too much like a magic string

library/objective-c/Envoy.h
```diff
@@ -10,5 +10,10 @@
 
 /// Create a new Envoy instance. The Envoy runner NSThread is started as part of instance
 /// initialization with the configuration provided.
-- (instancetype)initWithConfig:(NSString*)config;
+- (instancetype)initWithConfig:(NSString *)config;
+
+/// Create a new Envoy instance. The Envoy runner NSThread is started as part of instance
+/// initialization with the configuration provided.
+- (instancetype)initWithConfig:(NSString *)config logLevel:(NSString *)logLevel;
```
rebello95: Would also be nice to document what valid log levels are available here

goaway: I agree enums would be better. Tell you all what I'll open another PR with enums after I merge and re-test iOS. :)

goaway:
thank you @rebello95 and @buildbreaker!

Issue #181, opened 2019-06-25
ci: size regression


junr03:
CI should have a job that does relative (percentage deltas) and absolute size regressions of the demos.

CI should fail if the PR introduces a regression beyond the agreed upon thresholds.

The analysis should be done on the final demo sizes, and the reduces test_binary_size target.

junr03:
closed by https://github.com/lyft/envoy-mobile/pull/212

PR #182, opened 2019-06-25
perf: binary size analysis 

### LINKED ISSUE #17: reduce binary size
In order for us to be able to ship Envoy on the client we need to optimize for binary size. 

We will need to reduce Envoy Mobile’s final application bloat (compressed) down to < 5MB to be able to ship in a production mobile application

junr03:

Description: This PR concludes binary size investigation slated for issue #17. The three deliverables of this PR are:

1.  Developer documentation that solidifies the building and analysis platform used for binary size analysis.
2.  A sorted list of issues for next steps in binary size reduction under the [`perf/size`](https://github.com/lyft/envoy-mobile/labels/perf%2Fsize) label.
3.  An final baseline size for the binary:

> As of https://github.com/lyft/envoy-mobile/tree/f17caebcfce09ec5dcda905dc8418fea4d382da7
The test_binary_size_size as built by the toolchain against the architecture described (arm64 with clang and lld)
compiles to a stripped size of 8.9mb and a compressed size of 3mb.

Additionally #181 will add CI jobs to add size regression analysis on every PR.

Risk Level: low - add new bazel target and docs.
Docs Changes: added developer documentation.
Fixes #17


junr03:
@Reflejo deliverables we talked about this morning. ptal when you have a chance.

.bazelrc
```diff
@@ -23,3 +23,5 @@ build:android --fat_apk_cpu=x86,x86_64
 build:device --ios_multi_cpus=armv7,arm64 --fat_apk_cpu=armeabi-v7a,arm64-v8a
 
 build:fat --ios_multi_cpus=i386,x86_64,armv7,arm64 --fat_apk_cpu=x86,x86_64,armeabi-v7a,arm64-v8a
+
+build:sizeopt -c opt --copt -Os --copt=-ggdb3 --linkopt=-fuse-ld=lld
```
keith: I don't think lld would be compatible with the iOS build?

mattklein123: Can this also just get rolled into the release PR that @rebello95 is working on instead?

docs/root/development/development.rst
```diff
@@ -0,0 +1,11 @@
...
+
+Development
+===========
+
+.. toctree::
+  :maxdepth: 2
+
+  performance/performance
+
+This section of the docs describe information useful to engineers actively developing Envoy Mobile.
```
mattklein123: nit: move about TOC

docs/root/development/performance/binary_size.rst
```diff
@@ -0,0 +1,135 @@
...
+
+Open issues regarding size
+--------------------------
+
+
+``perf/size`` is a label tagging all current open issues that can improve binary size.
+Check out the issues `here <https://github.com/lyft/envoy-mobile/labels/perf%2Fsize>`_.
+After performing any change that tries to address these issues you should run through the analysis pipeline described above, and make sure your changes match expectations.
+
+The following issues are listed in priority order maximizing complexity vs. size win:
```
mattklein123: I don't think this is going to reasonably stay up to date, so personally I would just kill this and just leave the label link.

docs/root/development/performance/binary_size.rst
```diff
@@ -0,0 +1,135 @@
...
+- https://github.com/lyft/envoy-mobile/issues/177
+- https://github.com/lyft/envoy-mobile/issues/179
+- https://github.com/lyft/envoy-mobile/issues/176
+
+Current status
+~~~~~~~~~~~~~~
+
+As of https://github.com/lyft/envoy-mobile/tree/f17caebcfce09ec5dcda905dc8418fea4d382da7
+The test_binary_size_size as built by the toolchain against the architecture described above
+compiles to a stripped size of 8.9mb and a compressed size of 3mb.
```
mattklein123: nice: this is actually a lot smaller than I thought, and I think we can make this smaller still.

docs/root/development/performance/performance.rst
```diff
@@ -0,0 +1,13 @@
...
+Performance Analysis
+====================
+
+.. toctree::
+  :maxdepth: 2
+
+  binary_size
+
+Performance analysis can take several shapes in mobile applications.
+These docs describe the process (tools, analysis, measurements) that the Envoy Mobile team uses in their performance investigations.
```
mattklein123: nit: reflow and wrap to 100 cols here and elsewhere.

library/common/test_binary_size.cc
```diff
@@ -0,0 +1,8 @@
+#include "main_interface.h"
```
mattklein123: I would move this into a test folder.

.bazelrc
junr03: @keith does iOS use gold?

I can work with @rebello95 to consolidate. I do think that for release we want to do `--strip=never`, while here we want to keep symbols and manually strip after bazel.

docs/root/development/performance/binary_size.rst
junr03: yep, lots of room

.bazelrc
keith: Apple has their own linker, ld64, and we have to use that one

junr03: I see, thanks for the info. My idea with the sizeopt config was to use to get binaries for size analysis, not necessarily for release. So the linker stipulation here is just for standardization of the tooling used in that type of investigation.

I'll adjust the bazel config accordingly once we settle on the release config.

keith: This would likely just mean that we wouldn't be able to use `--config=sizeopt` with iOS, right now @rebello95's open PR is using that config in the release mode

junr03: gotcha, ok. I just looked at @rebello95's PR. I think I was confused about the ordering here. Now I see that #155 depends on this. I have fixed in a way that I think makes everything work together lmk what you think @keith

.bazelrc
```diff
@@ -23,3 +23,5 @@ build:android --fat_apk_cpu=x86,x86_64
 build:device --ios_multi_cpus=armv7,arm64 --fat_apk_cpu=armeabi-v7a,arm64-v8a
 
 build:fat --ios_multi_cpus=i386,x86_64,armv7,arm64 --fat_apk_cpu=x86,x86_64,armeabi-v7a,arm64-v8a
+
+build:sizeopt -c opt --copt -Os
```
rebello95: Instead of duplicating this from upstream envoy, you should be able to import that `.bazelrc` file like I'm doing here: https://github.com/lyft/envoy-mobile/pull/155/files#diff-f55b2f358053ad76fb5ac3f776f78ef8R1

docs/root/development/development.rst
```diff
@@ -0,0 +1,12 @@
+.. _dev:
+
+Development
+===========
+
+This section of the docs describe information useful to engineers actively
```
rebello95:
```suggestion
This section of the docs describes information useful to engineers actively
```

docs/root/development/performance/binary_size.rst
```diff
@@ -0,0 +1,164 @@
...
+analysis (otool, objdump, jtool), `Bloaty <https://github.com/google/bloaty>`_
+has been the tool of choice to run object file analysis.
+
+Bloaty's layering of data sources is extremely helpful in being able to explode
+the binary in all sorts of different ways. For example, one can look at the
+composition of each compile unit in terms of sections::
+
+  $ bloaty --debug-file=bin/test_binary_size -c envoy.bloaty -d sections,bloaty_package,compileunits bin/test_binary_size.stripped
+  ...
+  7.7%   109Ki   7.7%   110Ki bazel-out/aarch64-opt/bin/external/envoy_api/envoy/api/v2/route/route.pb.cc
```
buildbreaker: Is this an intentional format choice?

```suggestion
    7.7%   109Ki   7.7%   110Ki bazel-out/aarch64-opt/bin/external/envoy_api/envoy/api/v2/route/route.pb.cc
```

docs/root/development/performance/binary_size.rst
```diff
@@ -0,0 +1,164 @@
...
+    3.1%  3.45Ki   3.1%  3.45Ki .eh_frame_hdr
+    1.3%  1.48Ki   1.3%  1.48Ki .data
+    0.0%       0   0.6%     672 .bss
+    0.1%      72   0.1%      72 .rodata
+
+Or one might want to see how sections of the binary map to compilation units::
+
+  $ bloaty --debug-file=bin/test_binary_size -c envoy.bloaty -d bloaty_package,compileunits,sections bin/test_binary_size.stripped
+  ...
+  13.2%   929Ki  13.0%   929Ki .rodata
```
buildbreaker: Same here

docs/root/development/performance/binary_size.rst
```diff
@@ -0,0 +1,164 @@
...
+binary size. Check out the issues `here
+<https://github.com/lyft/envoy-mobile/labels/perf%2Fsize>`_. After performing
+any change that tries to address these issues you should run through the
+analysis pipeline described above, and make sure your changes match
+expectations.
+
+Current status
+~~~~~~~~~~~~~~
+
+As of
```
buildbreaker: Should we move this information to a release and link to our releases here?

docs/root/development/performance/binary_size.rst
junr03: yep that is bloaty's indentation to show groups of data.

docs/root/development/performance/binary_size.rst
```diff
@@ -0,0 +1,164 @@
...
+  mkdir build
+  cd build
+  cmake ..
+  make -j6
+  cp bloaty /usr/local/bin/bloaty
+
+The binary being compiled is ``//test/performance:test_binary_size``.
+The binary is getting built with the following build command::
+
+  bazel build //test/performance:test_binary_size --config=sizeopt --copt=-ggdb3 --linkopt=-fuse-ld=lld
```
buildbreaker: It's a bit unclear to me from reading this how the flags below map to these configuration flags

docs/root/development/performance/binary_size.rst
junr03: hmmm that is an interesting suggestion. I think once we do 0.2, where we will want people to import into apps and test, we should definitely have a release section in the docs. I am going to leave a comment here.

.bazelrc
junr03: Neat, didn't notice that in your PR.

docs/root/development/performance/binary_size.rst
junr03: clarified below.

PR #190, opened 2019-06-26
build: clean up java dep structure

goaway:

### DIFF
--- a/library/java/io/envoyproxy/envoymobile/engine/BUILD
+++ b/library/java/io/envoyproxy/envoymobile/engine/BUILD
@@ -1,0 +1,17 @@
+licenses(["notice"])  # Apache 2
+
+# These source files must re-exported to the kotlin custom library rule to ensure their inclusion.
+filegroup(
+    name = "envoy_engine_srcs",
+    srcs = ["EnvoyEngine.java"],
+    visibility = ["//visibility:public",],
+)
+
+android_library(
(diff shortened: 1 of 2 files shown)


library/java/io/envoyproxy/envoymobile/engine/BUILD
```diff
@@ -0,0 +1,18 @@
+licenses(["notice"])  # Apache 2
+
+# These source files must be re-exported to the kotlin custom library rule to ensure their
+# inclusion.
+filegroup(
```
buildbreaker: Ah! This is how to achieve this

buildbreaker:
I didn't realize we could use `filegroup` to expose a file to an external `BUILD` module.

Could we also move the current `library/BUILD` into the `library/kotlin` directory to further consolidate the build rules to the platform?

goaway:
@buildbreaker yep was going to try that next

buildbreaker:
Talked offline, we'll move the kotlin rules in another PR

goaway:
thanks @buildbreaker!

library/java/io/envoyproxy/envoymobile/engine/BUILD
goaway: Yep, it also means if we ever add more java srcs, we don't have to worry about missing them in the kotlin rule. :)

PR #194, opened 2019-06-27
logging: add LogLevel enum to kotlin interface

goaway:

### DIFF
--- a/library/kotlin/io/envoyproxy/envoymobile/Envoy.kt
+++ b/library/kotlin/io/envoyproxy/envoymobile/Envoy.kt
@@ -7,17 +7,13 @@ import io.envoyproxy.envoymobile.engine.EnvoyEngine
  * Available logging levels for an Envoy instance. Note some levels may be compiled out.
  */
 enum class LogLevel {
-  Trace,
-  Debug,
-  Info,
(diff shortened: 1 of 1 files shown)


library/kotlin/io/envoyproxy/envoymobile/Envoy.kt
```diff
@@ -3,11 +3,26 @@ package io.envoyproxy.envoymobile
 import android.content.Context
 import io.envoyproxy.envoymobile.engine.EnvoyEngine
 
+// Available logging levels for an Envoy instance. Note some levels may be compiled out.
+enum class LogLevel {
```
buildbreaker: We can just do something like this:
```

enum class Foo(internal val representation: String) {
  BAR("Bar"),
  BAZ("Baz"),
}

fun method(foo: Foo) {
  val logLevel = foo.representation
}
```
Then we don't have to overload the `toString()` of our enum and keep the convention of all caps for enum definitions.

goaway: Sure, I considered that as well, and don't have a preference. Will update. 👍

goaway:
(rebased)

PR #196, opened 2019-06-27
build: clean up kotlin deps

goaway:


goaway:
Honestly, not 100% my preferred way of handling this. I'd still like to build the aar at the top level, but this led to visibility issues I couldn't work around with the artifacts produced by the current kotlin rules. Still, this seems like an improvement.

goaway:
thanks @buildbreaker :)

PR #211, opened 2019-07-01
ci: binary size regression job

junr03:

Description: this PR adds a CI job to test for relative size increases as well as absolute size of the test_binary_size target as a check on size for mobile platform constraints. 
Risk Level: low - added new CI jobs.
Testing: CI. Also tested locally for failure messages.


ci/test_size_regression.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# Checks the absolute size, and the relative increase of a file.
+MAX_SIZE=4000000
+MAX_PERC=1
```
goaway: This might be too strict right now with so much in flux. Can we run the check but not have it block PRs?

azure-pipelines.yml
```diff
@@ -55,6 +55,76 @@ stages:
...
+  # Note: the size regression jobs are not ideal.
+  # The size analysis described in the docs was ran on arm64 machines.
+  # However, it seems that azure pipelines will support arm64 machines sometime soon.
+  # Whenever that is available this jobs should be moved to that architecture.
+  # Issue: https://github.com/lyft/envoy-mobile/issues/216
+  - stage: performance
+    dependsOn: [] # this removes the implicit dependency on previous stage and causes this to run in parallel.
+    jobs:
+      - job: size_current
+        timeoutInMinutes: 60
```
goaway: Does the timeout apply to just while the job is running or does it include time in queue as well. If it's just the former, can this be shorter?

ci/test_size_regression.sh
junr03: I think I'd rather change the constants to laxer terms if you want, but I would like this to block because otherwise we won't pay attention.

azure-pipelines.yml
junr03: Yep, this is only for running time. Yeah this can probably be lower. Looks like 30 mins is good enough.

azure-pipelines.yml
```diff
@@ -55,6 +55,76 @@ stages:
...
+              artifactName: 'current'
+              targetPath: bazel-bin/test/performance/test_binary_size
+      - job: size_master
+        timeoutInMinutes: 60
+        pool:
+          vmImage: 'Ubuntu 16.04'
+        steps:
+          - checkout: self
+            submodules: true
+          - script: git checkout master && git pull origin master
```
rebello95: In the future, we may want to consider:
- Running this against the SHA that the PR was opened against (i.e., if this branch is 5 commits behind master, compare against that SHA instead of the tip of master which may have changed since the PR was opened/rebased)
- Caching these artifacts so we don't have to rebuild master against each PR

junr03: Good point on caching: https://github.com/lyft/envoy-mobile/issues/229.

I see your point. But why should we prefer the time at which the PR was forked from master than the current tip?

rebello95: I think so, using master here almost feels like a side effect of sorts

PR #221, opened 2019-07-03
kotlin: Adding bazel macro for more convenient test target declaration

buildbreaker:
Creating test targets for Kotlin is a bit tedious. This change is to help alleviate that.

Some callouts:
1. This will *only* run tests within the `io.envoyproxy.envoymobile` package space
2. This will fail if there are no tests in the provided test target

I feel these are reasonable requirements for this project and can rely on this for the near future (we can revisit when we there is a solution which is more official).

Bazel out of the box:
```
load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_jvm_test")

kt_jvm_test(
    name = "example_kotlin_test",
    test_target = "com.example.myproject.mymodule.ExampleTest"
    srcs = [
        "ExampleTest.kt",
    ],
    deps = [
        "//:another_target",
        "@maven//:org_assertj_assertj_core",
        "@maven//:junit_junit",
    ]
)
```

With this change:
```
load("//bazel:kotlin_test.bzl", "envoy_mobile_kt_test)

envoy_mobile_kt_test(
    name = "example_kotlin_test",
    srcs = [
        "ExampleTest.kt",
    ],
    deps = [
        "//:another_target",
    ]
)
```

Took this solution from: https://stackoverflow.com/questions/46365464/how-to-run-all-tests-in-bazel-from-a-single-java-test-rule

Description: Adding bazel macro for more convenient test target declaration
Risk Level: low
Testing: locally
Docs Changes: n/a
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -111,9 +111,8 @@ maven_install(
         "junit:junit:4.12"
     ],
     repositories = [
-        "https://jcenter.bintray.com/",
-        "https://maven.google.com",
         "https://repo1.maven.org/maven2",
+        "https://jcenter.bintray.com/",
     ]
 )
 
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -102,6 +102,21 @@ git_repository(
     shallow_since = "1552938175 -0400",
 )
 
+load("@rules_jvm_external//:defs.bzl", "maven_install")
+
+maven_install(
+    artifacts = [
+        # Test artifacts
+        "org.assertj:assertj-core:3.9.0",
+        "junit:junit:4.12"
+    ],
+    repositories = [
+        "https://jcenter.bintray.com/",
+        "https://maven.google.com",
+        "https://repo1.maven.org/maven2",
+    ]
+)
+
 # Bazel Kotlin 1.3 patch: https://github.com/bazelbuild/rules_kotlin/issues/159
 # bazelbuild/rules_kotlin currently doesn't work with Kotlin 1.3
 # TODO: https://github.com/lyft/envoy-mobile/issues/68
--- a/bazel/EnvoyMobileTestSuite.kt
+++ b/bazel/EnvoyMobileTestSuite.kt
@@ -1,0 +1,55 @@
+package io.envoyproxy.envoymobile.bazel
+
+import junit.framework.JUnit4TestAdapter
+import junit.framework.TestSuite
+import org.junit.runner.RunWith
+import java.io.File
+import java.net.URLClassLoader
+import java.util.*
+import java.util.zip.ZipFile
+
+/**
+ * This is class is taken from https://stackoverflow.com/questions/46365464/how-to-run-all-tests-in-bazel-from-a-single-java-test-rule
+ *
+ * Translated to Kotlin and slightly modified
+ *
+ */
+@RunWith(org.junit.runners.AllTests::class)
+object EnvoyMobileTestSuite {
+  private const val CLASS_SUFFIX = ".class"
+
+  @JvmStatic
+  fun suite(): TestSuite {
+    val suite = TestSuite()
+    val classLoader = Thread.currentThread().contextClassLoader as URLClassLoader
+
+    // The first entry on the classpath contains the srcs from java_test
+    val classesInJar = findClassesInJar(File(classLoader.urLs[0].path))
+    for (clazz in classesInJar) {
+      val name = Class.forName(clazz)
+      if (name != EnvoyMobileTestSuite::class.java) {
+        suite.addTest(JUnit4TestAdapter(name))
+      }
+    }
+
+    return suite
+  }
+
+  private fun findClassesInJar(jarFile: File): Set<String> {
+    val classNames = mutableSetOf<String>()
+
+    ZipFile(jarFile).use { zipFile ->
+      val entries = zipFile.entries()
+      for (entry in entries) {
+        val entryName = entry.name
+
+        if (entryName.endsWith(CLASS_SUFFIX)) {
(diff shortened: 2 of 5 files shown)


bazel/kotlin_test.bzl
```diff
@@ -0,0 +1,30 @@
+load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_jvm_test")
+
+# A basic macro to make it easier to declare and run kotlin tests
+#
+# Ergonomic improvements include:
+# 1. Avoiding the need to declare the test_class which requires a fully qualified class name (example below)
+# 2. Avoiding the need to redeclare common unit testing dependencies like JUnit
+# 3. Ability to run more than one test file per target
```
buildbreaker: If this is too controversial, we can have the sources parameter take a single file rather than a list of files

buildbreaker:
Going to keep the array because [envoy's test macro](https://github.com/envoyproxy/envoy/blob/master/test/common/buffer/BUILD#L49) allows for an array of sources

WORKSPACE
```diff
@@ -102,6 +102,21 @@ git_repository(
...
+
+maven_install(
+    artifacts = [
+        # Test artifacts
+        "org.assertj:assertj-core:3.9.0",
+        "junit:junit:4.12"
+    ],
+    repositories = [
+        "https://jcenter.bintray.com/",
+        "https://maven.google.com",
```
artem-zinnatullin: no need for google repo for these artifacts

WORKSPACE
```diff
@@ -102,6 +102,21 @@ git_repository(
...
+        "https://repo1.maven.org/maven2",
```
artem-zinnatullin: pls prefer maven central to jcenter, basically put it in front jcenter :)

Reasons:

- Reliability
- Latency/Throughput
- Security (jcenter has had multiple instances of malicious artifacts uploaded to valid looking groupIds)

bazel/EnvoyMobileTestSuite.kt
```diff
@@ -0,0 +1,54 @@
...
+object EnvoyMobileTestSuite {
+  private const val CLASS_SUFFIX = ".class"
+
+  @JvmStatic
+  fun suite(): TestSuite {
+    val suite = TestSuite()
+    val classLoader = Thread.currentThread().contextClassLoader as URLClassLoader
+
+    // The first entry on the classpath contains the srcs from java_test
+    val classesInJar = findClassesInJar(File(classLoader.urLs[0].path))
```
artem-zinnatullin: Java can run class files directly, in which case there won't be `.jar` file I'd assume, but I guess since we build and run with Bazel it's not going to be an issue

buildbreaker: Is there an alternative here which we should explore?

artem-zinnatullin: I mean, it's kind of bad that we rely on particular index like `0`, ideally you'd scan entire classpath (can be a bit slow), so that Bazel update doesn't break your tests (they'll prob silently finish in this case which is worst thing for tests)

buildbreaker: With the current set up we have the following items in the `classLoader.urLs`:
```
/private/var/tmp/_bazel_achiu/e95f0937739315919b86dd2afc724e3e/execroot/__main__/bazel-out/darwin-fastbuild/bin/library/kotlin/test/sample_test.jar
/private/var/tmp/_bazel_achiu/e95f0937739315919b86dd2afc724e3e/execroot/__main__/bazel-out/darwin-fastbuild/bin/bazel/envoy_mobile_test_suite.jar
/private/var/tmp/_bazel_achiu/e95f0937739315919b86dd2afc724e3e/external/maven/v1/https/repo1.maven.org/maven2/junit/junit/4.12/junit-4.12.jar
/private/var/tmp/_bazel_achiu/e95f0937739315919b86dd2afc724e3e/external/maven/v1/https/repo1.maven.org/maven2/org/hamcrest/hamcrest-core/1.3/hamcrest-core-1.3.jar
/private/var/tmp/_bazel_achiu/e95f0937739315919b86dd2afc724e3e/external/com_github_jetbrains_kotlin/lib/kotlin-stdlib.jar
/private/var/tmp/_bazel_achiu/e95f0937739315919b86dd2afc724e3e/external/maven/v1/https/repo1.maven.org/maven2/org/assertj/assertj-core/3.9.0/assertj-core-3.9.0.jar
/private/var/tmp/_bazel_achiu/install/a134bf748a9bb448b9801b24812c57ec/_embedded_binaries/embedded_tools/tools/jdk/TestRunner_deploy.jar
```

I'm not sure what a more reliable heuristic would be to find the test class files. Should we instead just unpack all the `classLoader.urLs` jars and look for classes which have the name `*Test`?

buildbreaker: Another option is for us to suffix the bazel rules with `_test` but it'd be confusing for the person defining a test target

buildbreaker: I tried running a look up of all the jars in the context. It took ~300s to run a single test target (versus 0.3s without the exhaustive look up).

I don't think this is a reasonable solution to ensure we are not breaking on bazel updates since we would incur an incredible amount of development productivity hit here.

I've updated the PR to exclude non-envoy packages (`io.envoyproxy.envoymobile`). The bazel tests will fail if there are no tests in the test target.

This _should_ cover the case where the first `urL` is not the test target since we would fail if there are no tests within the envoy package space.

PR #222, opened 2019-07-03
perf: update envoy to include libevent patch

rebello95:
Updating Envoy to include https://github.com/envoyproxy/envoy/pull/7453 (and subsequently https://github.com/libevent/libevent/pull/849), which should fix the performance issues we're seeing on iOS. Resolves https://github.com/lyft/envoy-mobile/issues/215.


rebello95:
Had to update our extensions due to an upstream change for TLS. Please take another look 😄

PR #225, opened 2019-07-03
core api: add header definitions for native interface

goaway:
Co-authored-by: Jose Ulises Nino Rivera <junr03@users.noreply.github.com>

Description: Introduces the interfaces for the core API.
Risk Level: Low
Testing: CI

### DIFF
--- a/library/common/main_interface.h
+++ b/library/common/main_interface.h
@@ -41,18 +41,18 @@ typedef struct {
 } envoy_stream_pair;
 
 typedef struct {
-  char *name;
-  char *value;
+  char* name;
+  char* value;
 } envoy_header;
 
 typedef struct {
   size_t length;
-  envoy_header *headers;
+  envoy_header* headers;
 } envoy_headers;
 
 typedef struct {
   size_t length;
-  uint8_t *bytes;
+  uint8_t* bytes;
 } envoy_data;
 
 const envoy_data envoy_nodata = {0, NULL};
@@ -60,9 +60,8 @@ const envoy_data envoy_nodata = {0, NULL};
 typedef struct {
   envoy_error_code_t error_code;
   size_t length;
-  char *string;
+  char* string;
 } envoy_error;
-
 
 #ifdef __cplusplus
 extern "C" { // function pointers
@@ -83,7 +82,6 @@ typedef struct {
   on_trailers t;
   on_error e;
 } envoy_observer;
-
 
 #ifdef __cplusplus
 extern "C" { // functions
--- a/library/common/main_interface.h
+++ b/library/common/main_interface.h
@@ -36,6 +36,14 @@ typedef struct {
 } envoy_string;
 
 /**
+ * Holds raw binary data as an array of bytes.
+ */
+typedef struct {
+  uint64_t length;
+  uint8_t* bytes;
+} envoy_data;
+
+/**
  * Holds data about an HTTP stream.
  */
 typedef struct {
@@ -64,14 +72,6 @@ typedef struct {
   // Array of headers.
   envoy_header* headers;
 } envoy_headers;
-
-/**
- * Holds raw binary data as an array of bytes.
- */
-typedef struct {
-  uint64_t length;
-  uint8_t* bytes;
-} envoy_data;
 
 // Convenience constant to pass to function calls with no data.
 // For example when sending a headers-only request.
@@ -181,14 +181,14 @@ envoy_status_t send_trailers(envoy_stream_t stream, envoy_headers trailers);
  * @param stream, the stream to close.
  * @return envoy_status_t, the resulting status of the operation.
  */
-envoy_status_t close_stream(envoy_stream_t stream);
+envoy_status_t locally_close_stream(envoy_stream_t stream);
 
 /**
  * Detach all observers from a stream and send an interrupt upstream if supported by transport.
  * @param stream, the stream to evict.
  * @return envoy_status_t, the resulting status of the operation.
  */
-envoy_status_t evict_stream(envoy_stream_t stream);
+envoy_status_t interrupt_stream(envoy_stream_t stream);
 
 /**
  * External entrypoint for library.
--- a/library/common/main_interface.h
+++ b/library/common/main_interface.h
@@ -3,10 +3,138 @@
 // NOLINT(namespace-envoy)
 
 /**
+ * Handle to an Envoy engine instance. Valid only for the lifetime of the engine and not intended
+ * for any external interpretation or use.
+ */
+typedef int envoy_engine_t;
+
+/**
+ * Handle to an outstanding Envoy HTTP request. Valid only for the duration of the request and not
+ * intended for any external interpretation or use.
+ */
+typedef int envoy_request_t;
+
+/**
+ * Handle to an outstanding Envoy HTTP stream. Valid only for the duration of the stream and not
+ * intended for any external interpretation or use.
+ */
+typedef int envoy_stream_t;
+
+/**
+ * Result codes returned by all calls made to this interface.
+ */
+typedef enum {} envoy_status_t;
+
+/**
+ * Error code associated with terminal status of a HTTP stream or request.
+ */
+typedef enum {} envoy_error_code_t;
+
+typedef struct {
+  envoy_status_t status;
+  envoy_request_t request;
+} envoy_request_pair;
+
+typedef struct {
+  envoy_status_t status;
+  envoy_stream_t stream;
+} envoy_stream_pair;
+
+typedef struct {
+  char *name;
+  char *value;
+} envoy_header;
+
+typedef struct {
+  size_t length;
+  envoy_header *headers;
+} envoy_headers;
+
+typedef struct {
+  size_t length;
+  uint8_t *bytes;
+} envoy_data;
+
+const envoy_data envoy_nodata = {0, NULL};
+
+typedef struct {
+  envoy_error_code_t error_code;
+  size_t length;
+  char *string;
+} envoy_error;
+
+
+#ifdef __cplusplus
+extern "C" { // function pointers
+#endif
+
+typedef void (*on_headers)(envoy_headers headers, bool end_stream);
+typedef void (*on_data)(envoy_data data, bool end_stream);
+typedef void (*on_trailers)(envoy_headers headers);
+typedef void (*on_error)(envoy_error error);
+
+#ifdef __cplusplus
+} // function pointers
+#endif
+
+typedef struct {
+  on_headers h;
+  on_data d;
+  on_trailers t;
+  on_error e;
+} envoy_observer;
+
+
+#ifdef __cplusplus
+extern "C" { // functions
+#endif
+
+/**
+ * Submit a request.
+ */
+envoy_request_pair submit_request(envoy_headers headers, envoy_data data, envoy_observer observer);
+
+/**
+ * Cancel a request.
+ */
+envoy_status_t cancel_request(envoy_request_t request);
+
+/**
+ * Open an underlying HTTP stream.
+ */
+envoy_stream_pair open_stream(envoy_headers headers, envoy_data data, envoy_observer observer);
+
+/**
+ * Send headers over an open HTTP stream.
+ */
+envoy_status_t send_headers(envoy_stream_t stream, envoy_headers header, boolean end_stream);
+
+/**
+ * Send metadata over an HTTTP stream.
+ */
+envoy_status_t send_metadata(envoy_stream_t stream, envoy_headers metadata, boolean end_stream);
+
+/**
+ * Send trailers over an open HTTP stream.
+ */
+envoy_status_t send_trailers(envoy_stream_t stream, envoy_headers trailers, boolean end_stream);
+
+/**
+ * Half-close an HTTP stream (stream will be observable and may return further data, but nothing
+ * further may be sent).
+ */
+envoy_status_t close_stream(envoy_stream_t stream);
+
+/**
+ * Detach all observers from a stream and send an interrupt upstream if supported by transport.
+ */
+envoy_status_t evict_stream(envoy_stream_t stream);
+
+/**
  * External entrypoint for library.
  */
+envoy_status_t run_engine(const char* config, const char* log_level);
+
 #ifdef __cplusplus
-extern "C" int run_envoy(const char* config, const char* log_level);
-#else
-int run_envoy(const char* config, const char* log_level);
+} // functions
 #endif
--- a/library/common/main_interface.h
+++ b/library/common/main_interface.h
@@ -23,7 +23,7 @@ typedef uint64_t envoy_stream_t;
 typedef enum { ENVOY_SUCCESS, ENVOY_FAILURE } envoy_status_t;
 
 /**
- * Error code associated with terminal status of a HTTP stream or request.
+ * Error code associated with terminal status of a HTTP stream.
  */
 typedef enum {} envoy_error_code_t;
 
@@ -34,17 +34,6 @@ typedef struct {
   uint64_t length;
   char* data;
 } envoy_string;
-
-/**
- * Holds data about an HTTP request.
- */
-typedef struct {
-  // Status of the Envoy HTTP request. Note that the request might have failed inline.
-  // Thus the status should be checked before pursuing other operations on the request.
-  envoy_status_t status;
-  // Handle to the Envoy HTTP request.
-  envoy_request_t request;
-} envoy_request;
 
 /**
  * Holds data about an HTTP stream.
--- a/library/common/main_interface.h
+++ b/library/common/main_interface.h
@@ -67,7 +67,7 @@ typedef struct {
  * Holds an HTTP header map as an array of envoy_header structs.
  */
 typedef struct {
-  // Size of the array.
+  // Number of header elements in the array.
   uint64_t length;
   // Array of headers.
   envoy_header* headers;
--- a/library/common/main_interface.h
+++ b/library/common/main_interface.h
@@ -153,37 +153,68 @@ envoy_status_t cancel_request(envoy_request_t request);
 
 /**
  * Open an underlying HTTP stream.
+ * @param observer, the observer that will run the stream callbacks.
+ * @return envoy_stream, with a stream handle and a success status, or a failure status.
  */
-envoy_stream start_stream(envoy_headers headers, envoy_data data, envoy_observer observer);
+envoy_stream start_stream(envoy_observer observer);
 
 /**
- * Send headers over an open HTTP stream.
+ * Send headers over an open HTTP stream. This method can be invoked once and needs to be called
+ * before send_data.
+ * @param stream, the stream to send headers over.
+ * @param headers, the headers to send.
+ * @param end_stream, supplies wether this is headers only.
+ * @return envoy_status_t, the resulting status of the operation.
  */
 envoy_status_t send_headers(envoy_stream_t stream, envoy_headers header, bool end_stream);
 
 /**
- * Send metadata over an HTTP stream.
+ * Send data over an open HTTP stream. This method can be invoked multiple times.
+ * @param stream, the stream to send data over.
+ * @param data, the data to send.
+ * @param end_stream, supplies wether this is the last data in the stream.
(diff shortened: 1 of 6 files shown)


library/common/main_interface.h
```diff
@@ -1,12 +1,141 @@
...
+ * Handle to an Envoy engine instance. Valid only for the lifetime of the engine and not intended
+ * for any external interpretation or use.
+ */
+typedef int envoy_engine_t;
+
+/**
+ * Handle to an outstanding Envoy HTTP request. Valid only for the duration of the request and not
+ * intended for any external interpretation or use.
+ */
+typedef int envoy_request_t;
```
rebello95: Assuming the intent for this is to have the buffer/lifecycle managed by the shim (or platform) library, and pass an `envoy_request_t` as a pointer to the underlying C library?

library/common/main_interface.h
```diff
@@ -1,12 +1,141 @@
...
+ */
+typedef int envoy_stream_t;
+
+/**
+ * Result codes returned by all calls made to this interface.
+ */
+typedef enum { ENVOY_SUCCESS, ENVOY_FAILURE } envoy_status_t;
+
+/**
+ * Error code associated with terminal status of a HTTP stream or request.
```
rebello95: If this is specific to HTTP, do you think that's worth reflecting in the name of the type?

library/common/main_interface.h
```diff
@@ -1,12 +1,141 @@
...
+ */
+typedef enum {} envoy_error_code_t;
+
+typedef struct {
+  envoy_status_t status;
+  envoy_request_t request;
+} envoy_request_pair;
```
rebello95: `pair` can become outdated relatively quickly if we add a third field to the type. Thoughts on alternative names (or is that not a concern)?

library/common/main_interface.h
```diff
@@ -1,12 +1,141 @@
...
+
+typedef struct {
+  envoy_status_t status;
+  envoy_stream_t stream;
+} envoy_stream_pair;
+
+typedef struct {
+  char* name;
+  char* value;
```
rebello95: Is this intended to be a pointer to a string with comma-separated header values? I know we had decided to allow multiple header values for the same name [here](https://docs.google.com/document/d/1N0ZFJktK8m01uqqgfDRVB9mpC1iEn9dqkQaa_yMn_kE/edit#heading=h.blzdzy72yuk1)

library/common/main_interface.h
```diff
@@ -1,12 +1,141 @@
...
+  size_t length;
+  uint8_t* bytes;
+} envoy_data;
+
+const envoy_data envoy_nodata = {0, NULL};
+
+typedef struct {
+  envoy_error_code_t error_code;
+  size_t length;
+  char* string;
```
rebello95: I think it's potentially confusing to name the field using a word that's a keyword in other languages. Maybe `value` or `message`?

library/common/main_interface.h
```diff
@@ -1,12 +1,141 @@
...
+typedef void (*on_data)(envoy_data data, bool end_stream);
+typedef void (*on_trailers)(envoy_headers headers);
+typedef void (*on_error)(envoy_error error);
+
+#ifdef __cplusplus
+} // function pointers
+#endif
+
+typedef struct {
+  on_headers h;
```
rebello95: Why not name these more clearly?

library/common/main_interface.h
```diff
@@ -1,12 +1,141 @@
...
+} envoy_observer;
+
+#ifdef __cplusplus
+extern "C" { // functions
+#endif
+
+/**
+ * Submit a request.
+ */
+envoy_request_pair submit_request(envoy_headers headers, envoy_data data, envoy_observer observer);
```
rebello95: Should this be `send_request` to match up with the other `send*` functions?

library/common/main_interface.h
junr03: There are several ways to do this. The way @goaway and I were discussion with this "handle" was to actually let the c/c++ implementation keep track of all the request (e.g on a static table) via these integer handles as IDs.

library/common/main_interface.h
junr03: I like this suggestion. Specially because it creates the distinction that the above type, for instance, is not related to http and this one _is_.

library/common/main_interface.h
junr03: Yeah that would be the encoding. I will add comments to these fields to clarify that.

library/common/main_interface.h
junr03: We could go further and just call this send, and the stream function start, per the convention in Envoy's AsyncClient. Further, given the return types (envoy_request_pair, envoy_stream_pair) I think the names would be clear that way.

library/common/main_interface.h
```diff
@@ -1,12 +1,141 @@
...
+
+/**
+ * Cancel a request.
+ */
+envoy_status_t cancel_request(envoy_request_t request);
```
buildbreaker: Is the intent here to have a builder pattern for creating the streaming request?

library/common/main_interface.h
```diff
@@ -1,12 +1,141 @@
...
+  size_t length;
+  char* string;
+} envoy_error;
+
+#ifdef __cplusplus
+extern "C" { // function pointers
+#endif
+
+typedef void (*on_headers)(envoy_headers headers, bool end_stream);
+typedef void (*on_data)(envoy_data data, bool end_stream);
```
rebello95: How often is this called? For streams, data can be continually coming down the wire

library/common/main_interface.h
junr03: I am open to calling these structs `envoy_<request/stream>`. @goaway any strong opinions?

library/common/main_interface.h
```diff
@@ -1,12 +1,141 @@
...
+  envoy_error_code_t error_code;
+  size_t length;
+  char* string;
+} envoy_error;
+
+#ifdef __cplusplus
+extern "C" { // function pointers
+#endif
+
+typedef void (*on_headers)(envoy_headers headers, bool end_stream);
```
rebello95: Looks like these APIs are designed for streaming. How does this translate to unary requests?

library/common/main_interface.h
rebello95: Looks like unary upstream uses `send`, and streaming uses `start`. Do we intend to differentiate on this level or some other level (platform/shim)? https://github.com/envoyproxy/envoy/blob/master/include/envoy/http/async_client.h#L210-L221

rebello95: (Seems like maybe we should expose both the streaming and unary APIs?)

library/common/main_interface.h
junr03: We left this direction open so that the implementation of the observer can handle as desired. A simple implementation on top of these callbacks for a unary http request could follow a similar interface to:

https://github.com/envoyproxy/envoy/blob/master/include/envoy/http/async_client.h#L30-L44

library/common/main_interface.h
junr03: I don't think I understand your last point. Both send(request) and start(stream) will be available to the platform layer. Internally the common(c/c++) layer will use the `send_<headers/data/trailers>` functions to implement send(request). This is very similar to the Envoy implementation: https://github.com/envoyproxy/envoy/blob/master/source/common/http/async_client_impl.cc#L185 where the AsyncRequestImpl derives from the AsyncStreamImpl.

library/common/main_interface.h
junr03: Not sure I understand your question in relation to this function declaration.

library/common/main_interface.h
junr03: When a complete data frame is received. I added comments to the functions to clarify some of these questions.

library/common/main_interface.h
buildbreaker: I think I misread

library/common/main_interface.h
mattklein123: Do we need to expose the non-streaming API from core? I know we have discussed this a bit in the past, but my general feeling is that we should propagate streaming as far up the stack as possible as it will avoid issues and assumptions later.

library/common/main_interface.h
```diff
@@ -1,12 +1,223 @@
...
+  envoy_status_t status;
+  // Handle to the Envoy HTTP Stream.
+  envoy_stream_t stream;
+} envoy_stream;
+
+/**
+ * Holds a single name/value header.
+ */
+typedef struct {
+  char* name;
```
mattklein123: Strong preference to create a small `string_view` struct which has a pointer and length and use that everywhere. We should not rely on null termination and have run into problems with this in the main Envoy code (it is now removed).

library/common/main_interface.h
```diff
@@ -1,12 +1,223 @@
...
+  // Multiple header values for the same header name are supported via a comma-delimited string.
+  char* value;
+} envoy_header;
+
+/**
+ * Holds an HTTP header map as an array of envoy_header structs.
+ */
+typedef struct {
+  // Size of the array.
+  size_t length;
```
mattklein123: general preference for strongly typed integers everywhere, e.g., uint32_t or similar.

buildbreaker:
@mattklein123 I am assuming that you're talking about _any_ unary API. Could you help me understand your preference of not exposing an unary APIs at this layer?

From my perspective, there needs to be some layer which exposes a unary/streaming API for the end consumer. If we don't do it on this layer, the platform layers would need to handle the distinction between the two types of calls.

mattklein123:
> Could you help me understand your preference of not exposing an unary APIs at this layer?

It is my experience that not accounting for streaming at every layer creates large architectural issues/debt that will have to be dealt with later. I think it's in our best interest to only expose streaming as far up the stack as possible. Unary APIs are trivial wrappers around the streaming APIs and can be done at the top layer.

buildbreaker:
Thanks for the explanation @mattklein123!

library/common/main_interface.h
rebello95: I agree with this sentiment, will definitely make this a bit more ergonomic

library/common/main_interface.h
rebello95: > The way @goaway and I were discussion with this "handle" was to actually let the c/c++ implementation keep track of all the request (e.g on a static table) via these integer handles as IDs.

Got it. That makes sense, though I think it's worth documenting here since it's not immediately obvious by the type as to what this represents.

> Do we need to expose the non-streaming API from core? I know we have discussed this a bit in the past, but my general feeling is that we should propagate streaming as far up the stack as possible as it will avoid issues and assumptions later.

I don't think we _need_ to expose the non-streaming API. However, this does mean that we'll be exposing unary calls from the Swift/Kotlin libraries, and writing (minimal?) bridging logic back to the C streaming APIs in both places. Is that the intent?

library/common/main_interface.h
junr03: good idea. changed everywhere we had string-like constructs.

rebello95:
Regarding exposing streaming APIs, I replied in a thread [here](https://github.com/lyft/envoy-mobile/pull/225#discussion_r301692892) but want to reiterate for visibility:

I primarily want to make sure we understand the trade-offs of exposing the unary APIs from the C layer versus the Swift/Kotlin (platform) layers and that we're all on the same page with what those mean:

**Exposing from C layer**
- Unary + streaming APIs are available from C layer
- Unary will call into streaming APIs under the hood within this layer
- Platform layers can call directly into unary or streaming depending on their need (for v0.2, it'll just be unary), minimizing their logic

**Exposing from platform layer**
- C layer only has streaming APIs
- Native layers have both streaming and unary APIs (just unary for v0.2)
- Both native layers have logic to bridge from unary requests to C streaming APIs

I definitely see the value in only exposing streaming APIs from the C layer, as long as we're okay with duplicating some amount of logic between iOS/Android with regard to bridging unary APIs back to streaming requests. Thoughts @mattklein123?

library/common/main_interface.h
junr03: yep, that is the intent. Will respond on main thread.

junr03:
Thanks for the summary @rebello95, that is representative of my thoughts with a couple additional things

**Exposing unary signatures from C layer**

-  Depending on implementation we have to implement buffer handling/ownership for both cases.
-  In fact for 0.2, I think, there would be no streaming callsites in the platform layer

**Exposing unary signatures from the platform layer**

-  Implementing duplicate code in both platform layers is the biggest point against pushing the unary signature up the layers. I guess as far as implementing twice goes, this is probably one of the most innocuous points. But I think that it is something that shouldn’t go unchecked.

I think we can make it work both ways and both have tradeoffs. @goaway brought up a good point with the fact that we gave both call patterns the same callback struct so both are handled the same in that direction w.r.t the C layer. Moreover, this would allow us to reevaluate more easily if need be.

mattklein123:
> I definitely see the value in only exposing streaming APIs from the C layer, as long as we're okay with duplicating some amount of logic between iOS/Android with regard to bridging unary APIs back to streaming requests. Thoughts @mattklein123?

Yeah, my preference is for duplication right now at higher layers, mainly because the logic to do this is trivial. This will force us to think through all of this from a "streaming first" perspective. Once more code is written, if we decide to push some unary interfaces back down the stack, I think that is reasonable. WDYT?

rebello95:
Thanks for clarifying @junr03.

> Yeah, my preference is for duplication right now at higher layers, mainly because the logic to do this is trivial. This will force us to think through all of this from a "streaming first" perspective. Once more code is written, if we decide to push some unary interfaces back down the stack, I think that is reasonable. WDYT?

This is fine with me. If we decide to move some unary interfaces into the C layer later on, it should be relatively trivial to toss code out from the native/platform layers and shift it. For now we can expose streaming-only APIs from C and provide ergonomics in Swift/Kotlin for unary.

library/common/main_interface.h
rebello95: This also applies to `envoy_stream_t` and its docs

library/common/main_interface.h
```diff
@@ -1,12 +1,214 @@
...
+/**
+ * Common abstraction for strings.
+ */
+typedef struct {
+  uint64_t length;
+  char* data;
+} envoy_string;
+
+/**
+ * Holds data about an HTTP request.
```
rebello95: Similarly throughout this file, if types are specific to HTTP do you think that detail should be reflected in their name? Or maybe omitted from the docstrings?

library/common/main_interface.h
```diff
@@ -1,12 +1,214 @@
...
+  envoy_header* headers;
+} envoy_headers;
+
+/**
+ * Holds raw binary data as an array of bytes.
+ */
+typedef struct {
+  uint64_t length;
+  uint8_t* bytes;
+} envoy_data;
```
rebello95: Should we group this with `envoy_string` since they represent more basic types?

library/common/main_interface.h
```diff
@@ -1,12 +1,214 @@
...
+
+/**
+ * Holds an HTTP header map as an array of envoy_header structs.
+ */
+typedef struct {
+  // Size of the array.
+  uint64_t length;
+  // Array of headers.
+  envoy_header* headers;
+} envoy_headers;
```
rebello95: Out of curiosity, why do we need this type as opposed to using a hashmap?

library/common/main_interface.h
```diff
@@ -1,12 +1,214 @@
...
+
+/**
+ * Send headers over an open HTTP stream. This method can be invoked once and needs to be called
+ * before send_data.
+ * @param stream, the stream to send headers over.
+ * @param headers, the headers to send.
+ * @param end_stream, supplies whether this is headers only.
+ * @return envoy_status_t, the resulting status of the operation.
+ */
+envoy_status_t send_headers(envoy_stream_t stream, envoy_headers header, bool end_stream);
```
rebello95:
```suggestion
envoy_status_t send_headers(envoy_stream_t stream, envoy_headers headers, bool end_stream);
```

library/common/main_interface.h
```diff
@@ -1,12 +1,214 @@
...
+ */
+envoy_status_t send_metadata(envoy_stream_t stream, envoy_headers metadata, bool end_stream);
+
+/**
+ * Send trailers over an open HTTP stream. This method can only be invoked once per stream.
+ * Note that this method implicitly ends the stream.
+ * @param stream, the stream to send trailers over.
+ * @param trailers, the trailers to send.
+ * @return envoy_status_t, the resulting status of the operation.
+
```
rebello95: nit: remove

library/common/main_interface.h
junr03: There are no maps in C. We could build that in this layer. But I'd rather use this simple struct and build better data structures in the c++ implementation of this header.

library/common/main_interface.h
junr03: would like to hear what @goaway has to say. My two cents for not grouping:

1. It gives semantic clarity that this struct is byte data versus a string.
2. Often char _is_ 8bits. But not grouping allows us to make the bytes field here uint8_t explicitly.

library/common/main_interface.h
junr03: The reason why I have not done this is because I feel including http in the type name might be misleading some places. For instance in the error code type we are could create misunderstanding between it and actual http codes (5xx, etc).

library/common/main_interface.h
```diff
@@ -1,12 +1,202 @@
...
+ * @return envoy_status_t, the resulting status of the operation.
+ */
+envoy_status_t close_stream(envoy_stream_t stream);
+
+/**
+ * Detach all observers from a stream and send an interrupt upstream if supported by transport.
+ * @param stream, the stream to evict.
+ * @return envoy_status_t, the resulting status of the operation.
+ */
+envoy_status_t evict_stream(envoy_stream_t stream);
```
junr03: @goaway nit: should we name start/close/evict_stream, just start/close/evict given that the return value for the first, and the args for the latter two make it evident what is getting started, closed, and evicted?

library/common/main_interface.h
```diff
@@ -1,12 +1,202 @@
...
+#ifdef __cplusplus
+extern "C" { // functions
+#endif
+
+/**
+ * Open an underlying HTTP stream.
+ * @param observer, the observer that will run the stream callbacks.
+ * @return envoy_stream, with a stream handle and a success status, or a failure status.
+ */
+envoy_stream start_stream(envoy_observer observer);
```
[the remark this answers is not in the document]
rebello95: (I'm assuming this is once per request and that the connections are handled under the hood)

library/common/main_interface.h
```diff
@@ -1,12 +1,202 @@
...
+  envoy_string name;
+  // Multiple header values for the same header name are supported via a comma-delimited string.
+  envoy_string value;
+} envoy_header;
+
+/**
+ * Holds an HTTP header map as an array of envoy_header structs.
+ */
+typedef struct {
+  // Size of the array.
```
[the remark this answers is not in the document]
goaway: Good call, will update.

library/common/main_interface.h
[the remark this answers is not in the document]
goaway: This is once per request - in h2+ parlance a "stream" is more or less the same as a "request" and a "session" is more or less the same as a "connection".

library/common/main_interface.h
goaway: Definitely open to other naming options. This is technically a callback function that's _part_ of an observer, and not an observer itself (rather, the struct/object that holds it and also contains error handling is the observer). There's also some historical precedent for `on_x` naming of callback functions. It could be called `header_callback_function_ptr`.

library/common/main_interface.h
goaway: Since the argument is technically still a primitive and there's no overloading by signature, I think it's maybe still better to be explicit in the function name.

That said, I dislike this name now.

What do you think about `locally_close_stream` and `interrupt_stream`?

junr03: +1 I think those are good.

library/common/main_interface.h
rebello95: No strong opinion, that's fine

junr03:
@mattklein123 any other comments?

goaway:
thanks all 🚀

PR #235, opened 2019-07-08
ci: Remove libtool from macOS

keith:
On macOS with brew installing libtool installs `glibtool` which isn't
ever referenced. Whatever this is for must work fine with the default
`libtool` vendored with Xcode.

Upstream PR: https://github.com/envoyproxy/envoy/pull/7499


keith:
It looks like autoreconf really does expect glibtool, we're not hitting the codepath that needs this right now, but I guess we should leave it

PR #240, opened 2019-07-09
kotlin: add library request interfaces

buildbreaker:
https://github.com/lyft/envoy-mobile/issues/119

Adding a primitive `Request` and `RequestBuilder` with basic features:
* URL
* method
* headers
* trailers
* retry policy

Tests are basic setting a builder and ensuring the result is expected. Additionally, the request build transform creates an equivalent object when built back.

Description: kotlin: add library request interfaces 
Risk Level: low
Testing: unit
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


library/kotlin/io/envoyproxy/envoymobile/Request.kt
```diff
@@ -0,0 +1,50 @@
+package io.envoyproxy.envoymobile
+
+import java.net.URL
+
+class Request internal constructor(
+    val method: RequestMethod,
+    val url: URL,
+    val headers: Map<String, List<String>>,
+    val trailers: Map<String, List<String>>,
+    val body: ByteArray?,
```
goaway: https://github.com/lyft/envoy-mobile/pull/242#discussion_r302657799

library/kotlin/io/envoyproxy/envoymobile/RequestBuilder.kt
```diff
@@ -0,0 +1,90 @@
...
+  fun addHeader(header: String, value: String): RequestBuilder {
+    if (headers.containsKey(header)) {
+      headers[header]!!.add(value)
+    }
+
+    headers[header] = mutableListOf(value)
+    return this
+  }
+
+  fun removeHeader(header: String): RequestBuilder {
```
goaway: https://github.com/lyft/envoy-mobile/pull/242#discussion_r302659560

library/kotlin/io/envoyproxy/envoymobile/RequestBuilder.kt
```diff
@@ -0,0 +1,90 @@
...
+
+
+/**
+ * Builder used for constructing instances of `Request` types.
+ *
+ * @param url URL for the request.
+ * @param method Method for the request.
+ */
+class RequestBuilder(
+    val url: URL,
```
goaway: Do we want to support passing `scheme`, `authority`/`host`, and `path` as separate parameters?

goaway: Maybe phrasing it a little less open-ended: I can think of a couple reasons we might consider supporting it.

The builder could support both options, of course.

Internally (in the request object), I think we should have a split representation.
- Actual protocols handle these attributes entirely separately.
- There are cases where the separate values could encode more information than the combined url.
- You can still re-compose the url if you want it back from the request.

rebello95: @buildbreaker and I had discussed this offline as a consideration. A few thoughts:

- `URL` is required when initializing the builders right now, as is the `RequestMethod`, in order to avoid having `build()` be throwable. We could potentially add another initializer that takes `scheme`/`host`/`path` separately if we want to maintain this and store those values separately under the hood
- I don't think adding methods such as `addHost`/`addScheme` really make sense here since all 3 need to be set to create a valid request - so we'd likely have to pass all of them at once (which could work)

Essentially, I'm okay with storing the 3 values separately, the question is more along the lines of how we want to expose it (my suggestion would be another initializer that takes all 3 in place of the `URL`)

goaway: Makes sense. That sounds good to me. 👍

buildbreaker: We can always revisit this in the future. I think having just this right now is a fine solution

rebello95: https://envoyproxy.slack.com/archives/CKQ2LK23G/p1562880255143900

library/kotlin/io/envoyproxy/envoymobile/RequestBuilder.kt
```diff
@@ -0,0 +1,90 @@
...
+    this.body = body
+    return this
+  }
+
+  fun addRetryPolicy(retryPolicy: RetryPolicy?): RequestBuilder {
+    this.retryPolicy = retryPolicy
+    return this
+  }
+
+  fun addHeader(header: String, value: String): RequestBuilder {
```
rebello95: I think the general semantic here is `name` instead of `header` (i.e., the key that goes with the value). Same for trailers

library/kotlin/io/envoyproxy/envoymobile/RetryPolicy.kt
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Specifies how a request may be retried, containing one or more rules.
+ *
+ * @param maxRetryCount Maximum number of retries that a request may be performed.
+ * @param retryOnCodes Whitelist of rules used for retrying.
+ * @param perRetryTimeoutMs Timeout (in milliseconds) to apply to each retry.
+ */
+data class RetryPolicy(
+    val maxRetryCount: Int,
```
rebello95: This should be an unsigned int

library/kotlin/io/envoyproxy/envoymobile/RetryPolicy.kt
```diff
@@ -0,0 +1,26 @@
...
+    val retryOnCodes: List<RetryRule>,
+    val perRetryTimeoutMs: Long
```
rebello95: Should also be an unsigned int

library/kotlin/io/envoyproxy/envoymobile/RetryPolicy.kt
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Specifies how a request may be retried, containing one or more rules.
+ *
+ * @param maxRetryCount Maximum number of retries that a request may be performed.
+ * @param retryOnCodes Whitelist of rules used for retrying.
+ * @param perRetryTimeoutMs Timeout (in milliseconds) to apply to each retry.
+ */
+data class RetryPolicy(
+    val maxRetryCount: Int,
+    val retryOnCodes: List<RetryRule>,
```
rebello95: This is optional

library/kotlin/io/envoyproxy/envoymobile/RetryPolicy.kt
buildbreaker: We don't have unsigned ints

library/kotlin/test/io/envoyproxy/envoymobile/RequestBuilderTest.kt
```diff
@@ -0,0 +1,105 @@
+package io.envoyproxy.envoymobile
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import java.net.URL
+
+class RequestBuilderTest {
+
+  @Test
+  fun `adding request data should have body present in request`() {
```
buildbreaker: Probably worth talking about how we want to write our test names.

The one I chose was `[condition] should [expectation]`. I tend to like to avoid adding the method or prefix with `test` (more explanation on the alternatives)

Alternatives:
* `should [expectation] when [condition]`
* `[method name] should [expectation] when [condition]` -- this adds overhead onto identifying what the actual method is that we want to test
* `test [method name] should [expectation] when [condition]` -- I personally like this the least because I feel having a prefix of `test` is redundant

rebello95: I generally agree with the sentiment of `[condition] should [expectation]`:
- iOS tests are identified by the `test` prefix, so we **must** have these prefixes in Swift/ObjC
- `[method name] should [expectation] when [condition]` - I think this one depends a bit on the test. In some cases I could see it being valuable to clarify

buildbreaker: The other thing which I am doing here is leveraging Kotlin's feature of declaring methods as sentence strings -- not sure if you have opinions here @junr03 @rebello95 @goaway

rebello95: Can't do the sentence strings in Swift, but not opposed to using it in Kotlin

library/kotlin/io/envoyproxy/envoymobile/Request.kt
buildbreaker: Resolving since the linked comment's decision is to discuss it further

library/kotlin/test/io/envoyproxy/envoymobile/RequestBuilderTest.kt
buildbreaker: > [method name] should [expectation] when [condition] - I think this one depends a bit on the test. In some cases I could see it being valuable to clarify

Yeah, there might be exceptions but for the most part the cases that I've run across where the method name is present are when the method actually describes the behavior (like `track event should publish to the pipeline`).

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
```diff
@@ -0,0 +1,26 @@
...
+    val retryOn: List<RetryRule>,
+    val perRetryTimeoutMs: Long?
+)
+
+/**
+ * These are retry rules specified in Envoy's router filter.
+ * @see <a href="https://www.envoyproxy.io/docs/envoy/latest/configuration/http_filters/router_filter#x-envoy-retry-on">x-envoy-retry-on</a>
+ */
+enum class RetryRule {
+  ALL_5XX,
```
rebello95: These don't quite match what's defined in Swift. Do you want to update Kotlin or should we change Swift's? https://github.com/lyft/envoy-mobile/blob/master/library/swift/src/RetryPolicy.swift#L7

Personal preference is to update Kotlin since the Swift statuses are reflective of Envoy upstream

library/kotlin/test/io/envoyproxy/envoymobile/RequestBuilderTest.kt
goaway: My take:

- I think this looks good. I think it works especially well when your unit under test has a small scope with limited behavior (also a characteristic of good design/architecture).
- I think we should still be ok using `[method name] should [expectation] (when [condition])` for more complex cases where it would add clarity. I do think we should be consistent in format for a given file/unit.
- I think the "sentence" function names in kotlin are nice and see no reason not to use them.

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
goaway: I agree with mapping to Envoy's options directly.

library/kotlin/test/io/envoyproxy/envoymobile/RequestBuilderTest.kt
buildbreaker: > I think we should still be ok using [method name] should [expectation] (when [condition]) for more complex cases where it would add clarity. I do think we should be consistent in format for a given file/unit.

I agree!

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
buildbreaker: I made a mental note to update this when I saw what you did in your PR @rebello95 😬

library/kotlin/src/io/envoyproxy/envoymobile/RequestBuilder.kt
```diff
@@ -0,0 +1,159 @@
...
+   * @param name the header key.
+   * @param value the value associated to the header key.
+   * @return this builder.
+   */
+  fun addHeader(name: String, value: String): RequestBuilder {
+    if (headers.containsKey(name)) {
+      headers[name]!!.add(value)
+    }
+
+    headers[name] = mutableListOf(value)
```
rebello95: This is a bug - I think you need to early-return above to avoid re-setting the array and clearing previously present values

library/kotlin/src/io/envoyproxy/envoymobile/RequestBuilder.kt
```diff
@@ -0,0 +1,159 @@
...
+  /**
+   * Remove the value in the specified header.
+   *
+   * @param name the header key to remove.
+   * @param value the value to be removed.
+   * @return this builder.
+   */
+  fun removeHeader(name: String, value: String): RequestBuilder {
+    if (headers.containsKey(name)) {
+      headers[name]!!.remove(value)
```
rebello95: I would additionally remove the header entirely if it becomes empty after executing this

library/kotlin/src/io/envoyproxy/envoymobile/RequestBuilder.kt
```diff
@@ -0,0 +1,159 @@
...
+   * @param name the trailer key.
+   * @param value the value associated to the trailer key.
+   * @return this builder.
+   */
+  fun addTrailer(name: String, value: String): RequestBuilder {
+    if (trailers.containsKey(value)) {
+      trailers[name]!!.add(value)
+    }
+
+    trailers[name] = mutableListOf(value)
```
rebello95: Same bug here

library/kotlin/src/io/envoyproxy/envoymobile/RequestBuilder.kt
```diff
@@ -0,0 +1,159 @@
...
+  /**
+   * Remove the value in the specified trailer.
+   *
+   * @param name the trailer key to remove.
+   * @param value the value to be removed.
+   * @return this builder.
+   */
+  fun removeTrailer(name: String, value: String): RequestBuilder {
+    if (trailers.containsKey(name)) {
+      trailers[name]!!.remove(value)
```
rebello95: Same comment as for headers regarding removing the empty array

library/kotlin/src/io/envoyproxy/envoymobile/RequestBuilder.kt
```diff
@@ -0,0 +1,159 @@
...
+        headers,
+        trailers,
+        body,
+        retryPolicy
+    )
+  }
+
+  internal fun setHeaders(headers: Map<String, List<String>>): RequestBuilder {
+    this.headers.clear()
+    headers.map { entry -> this.headers[entry.key] = entry.value.toMutableList() }
```
rebello95: Can we avoid side-effects within `map`? Same with `setTrailers`

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Specifies how a request may be retried, containing one or more rules.
+ *
+ * @param maxRetryCount Maximum number of retries that a request may be performed.
+ * @param retryOn Whitelist of rules used for retrying.
+ * @param perRetryTimeoutMs Timeout (in milliseconds) to apply to each retry.
+ */
+data class RetryPolicy(
+    val maxRetryCount: Int,
+    val retryOn: List<RetryRule>,
```
rebello95: This should be optional

buildbreaker: Why not just keep it an empty list? If there's no retry rules, shouldn't it be the same as if it were to be unset?

rebello95: Hmm, that's true. Ok yea - I'll update Swift

rebello95: 🤦‍♂ already non-optional in Swift, ignore haha

library/kotlin/src/io/envoyproxy/envoymobile/RequestBuilder.kt
goaway: I think what you maybe really want here is `forEach`.

rebello95: Yep agreed

library/kotlin/test/io/envoyproxy/envoymobile/RequestBuilderTest.kt
```diff
@@ -0,0 +1,128 @@
...
+
+  @Test
+  fun `adding retry policy should have policy present in request`() {
+
+    val retryPolicy = RetryPolicy(23, listOf(RetryRule.FIVE_XX, RetryRule.CONNECT_FAILURE), 1234)
+    val request = RequestBuilder(URL("http://0.0.0.0:9001/api.lyft.com/demo.txt"), RequestMethod.GET)
+        .addRetryPolicy(retryPolicy)
+        .build()
+
+    assertThat(request.retryPolicy).isEqualTo(RetryPolicy(23, listOf(RetryRule.FIVE_XX, RetryRule.CONNECT_FAILURE), 1234))
```
rebello95: Why not compare directly to `retryPolicy`?

library/kotlin/src/io/envoyproxy/envoymobile/RequestBuilder.kt
buildbreaker: What's the intent here?

library/kotlin/test/io/envoyproxy/envoymobile/RequestBuilderTest.kt
buildbreaker: It's easier to see the values which we are expecting on an assert line

library/kotlin/src/io/envoyproxy/envoymobile/RequestBuilder.kt
rebello95: To not send/set an empty header

buildbreaker: I'll add a test for this case too

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
```diff
@@ -30,3 +30,13 @@ kt_android_library(
     visibility = ["//visibility:public"],
     deps = ["//library/java/io/envoyproxy/envoymobile/engine:envoy_engine_lib"],
 )
+
+kt_jvm_library(
+    name = "envoy_interfaces_lib",
```
rebello95: Shouldn't this be part of `envoy_lib`?

buildbreaker: It'll pull in all the dependencies with the JNI, Android, etc. I wanted to keep this more vanilla kotlin for now because I hope to isolate most of the classes away from Android and Envoy. The rationale is because I want to keep the tests lighter weight and we can more easily make decisions independent from the other layers.

We can also revisit the organization and responsibility boundaries down the line

buildbreaker:
thanks for the review @rebello95 @goaway

PR #242, opened 2019-07-09
swift: add library request interfaces

rebello95:
Introduces request interfaces for the Swift library.

**Builder pattern**

We elected to use a builder pattern for the following reasons:
- Allows us to avoid making all properties mutable (such as those on `Request`). This is especially valuable because these types are classes instead of structs due to the fact that they need to be visible to Objective-C from Swift
- Provides a clear separation between mutable/immutable states of these types since only builders can be modified
- Allows consumers to easily convert between builders and the models, which will become particularly useful when we introduce features such as runtime filters in the future:

```swift
func filter(request: Request) -> Request {
    return request
        .newBuilder()
        .addHeader(name: "x-new-header", value: "foo")
        .build()
}
```

**Swift example**

```swift
let request = RequestBuilder(method: .post, url: url)
    .addBody(data)
    .addHeader(name: "x-some-header", value: "foo")
    .addTrailer(name: "x-some-trailer", value: "foo")
    .build()
```

**Objective-C example**

```objc
Request *request = [Request withMethod:RequestMethodGet url:url build:^(RequestBuilder *builder) {
    [builder addBody:bodyData];
    [builder addHeaderWithName:@"x-some-header" value:@"foo"];
    [builder addTrailerWithName:@"x-some-trailer" value:@"foo"];
}];
```

**Notes**

- `NSObject` and `@objc` classes had to be used in order for these types to be visible to Objective-C
- Enums had to inherit from `Int` to be visible to Objective-C

Resolves https://github.com/lyft/envoy-mobile/issues/122

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,3 +14,4 @@ repos:
   rev: v1.15.0
   hooks:
     - id: codespell
+      args: [-L uint]
--- a/library/swift/Request.swift
+++ b/library/swift/Request.swift
@@ -3,41 +3,41 @@ import Foundation
 /// Represents an Envoy HTTP request. Use `RequestBuilder` to construct new instances.
 @objcMembers
 public final class Request: NSObject {
-    /// Method for the request.
-    public let method: RequestMethod
-    /// URL for the request.
-    public let url: URL
-    /// Headers to send with the request.
-    /// Multiple values for a given name are valid, and will be sent as comma-separated values.
-    public let headers: [String: [String]]
-    /// Trailers to send with the request.
-    /// Multiple values for a given name are valid, and will be sent as comma-separated values.
-    public let trailers: [String: [String]]
-    // Serialized data to send as the body of the request.
-    public let body: Data?
-    // Retry policy to use for this request.
-    public let retryPolicy: RetryPolicy?
+  /// Method for the request.
+  public let method: RequestMethod
+  /// URL for the request.
+  public let url: URL
+  /// Headers to send with the request.
+  /// Multiple values for a given name are valid, and will be sent as comma-separated values.
+  public let headers: [String: [String]]
+  /// Trailers to send with the request.
+  /// Multiple values for a given name are valid, and will be sent as comma-separated values.
+  public let trailers: [String: [String]]
+  // Serialized data to send as the body of the request.
+  public let body: Data?
+  // Retry policy to use for this request.
+  public let retryPolicy: RetryPolicy?
 
-    /// Converts the request back to a builder so that it can be modified (i.e., by a filter).
-    ///
-    /// - returns: A new builder including all the properties of this request.
-    public func newBuilder() -> RequestBuilder {
-        return RequestBuilder(request: self)
-    }
+  /// Converts the request back to a builder so that it can be modified (i.e., by a filter).
+  ///
+  /// - returns: A new builder including all the properties of this request.
+  public func newBuilder() -> RequestBuilder {
+    return RequestBuilder(request: self)
+  }
 
-    /// Internal initializer called from the builder to create a new request.
-    init(method: RequestMethod,
-         url: URL,
-         headers: [String: [String]] = [:],
-         trailers: [String: [String]] = [:],
-         body: Data?,
-         retryPolicy: RetryPolicy?)
-    {
-        self.method = method
-        self.url = url
-        self.headers = headers
-        self.trailers = trailers
-        self.body = body
-        self.retryPolicy = retryPolicy
-    }
+  /// Internal initializer called from the builder to create a new request.
+  init(method: RequestMethod,
+       url: URL,
+       headers: [String: [String]] = [:],
+       trailers: [String: [String]] = [:],
+       body: Data?,
+       retryPolicy: RetryPolicy?)
+  {
+    self.method = method
+    self.url = url
+    self.headers = headers
+    self.trailers = trailers
+    self.body = body
+    self.retryPolicy = retryPolicy
+  }
 }
--- a/library/swift/Request.swift
+++ b/library/swift/Request.swift
@@ -1,0 +1,43 @@
+import Foundation
+
+/// Represents an Envoy HTTP request. Use `RequestBuilder` to construct new instances.
+@objcMembers
+public final class Request: NSObject {
+    /// Method for the request.
+    public let method: RequestMethod
+    /// URL for the request.
+    public let url: URL
+    /// Headers to send with the request.
+    /// Multiple values for a given name are valid, and will be sent as comma-separated values.
+    public let headers: [String: [String]]
+    /// Trailers to send with the request.
+    /// Multiple values for a given name are valid, and will be sent as comma-separated values.
+    public let trailers: [String: [String]]
+    // Serialized data to send as the body of the request.
+    public let body: Data?
+    // Retry policy to use for this request.
+    public let retryPolicy: RetryPolicy?
+
+    /// Converts the request back to a builder so that it can be modified (i.e., by a filter).
+    ///
+    /// - returns: A new builder including all the properties of this request.
+    public func newBuilder() -> RequestBuilder {
+        return RequestBuilder(request: self)
+    }
+
+    /// Internal initializer called from the builder to create a new request.
+    init(method: RequestMethod,
+         url: URL,
+         headers: [String: [String]] = [:],
+         trailers: [String: [String]] = [:],
+         body: Data?,
+         retryPolicy: RetryPolicy?)
+    {
+        self.method = method
+        self.url = url
+        self.headers = headers
+        self.trailers = trailers
+        self.body = body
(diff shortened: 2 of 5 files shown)


library/swift/Request.swift
```diff
@@ -0,0 +1,43 @@
...
+  /// URL for the request.
+  public let url: URL
+  /// Headers to send with the request.
+  /// Multiple values for a given name are valid, and will be sent as comma-separated values.
+  public let headers: [String: [String]]
+  /// Trailers to send with the request.
+  /// Multiple values for a given name are valid, and will be sent as comma-separated values.
+  public let trailers: [String: [String]]
+  // Serialized data to send as the body of the request.
+  public let body: Data?
```
goaway: I'd suggest adding `public let stream: Stream?` here also for body data. We can adopt one of two conventions - always send body data first, and then streamed data after, or enforce that only one of these is ever set during initialization. (The latter is probably safer and more common.)

We can also discuss and punt on this, but it's supported by Apple's own APIs and seems valuable to solve now (especially in line with @mattklein123's thoughts on the core API PR).

library/swift/RequestBuilder.swift
```diff
@@ -0,0 +1,106 @@
...
+
+  @discardableResult
+  public func addHeader(name: String, value: String) -> RequestBuilder {
+    self.headers[name, default: []].append(value)
+    return self
+  }
+
+  @discardableResult
+  public func removeHeader(name: String) -> RequestBuilder {
+    self.headers.removeValue(forKey: name)
```
goaway: nit: this might lead to confusion since it will remove _all_ headers with a matching name. (e.g. `Link` headers). We might want to rename this to better indicate that, and/or provide a pair-matching API for header removal.

library/swift/RequestMethod.swift
```diff
@@ -0,0 +1,36 @@
+import Foundation
+
+/// Represents an HTTP request method.
+@objc
+public enum RequestMethod: Int {
```
goaway: If we implement `RawRepresentable` we can use this more ergonomically in Swift, I think.(?)

library/swift/RequestMethod.swift
```diff
@@ -0,0 +1,36 @@
...
+      return "OPTIONS"
+    case .patch:
+      return "PATCH"
+    case .post:
+      return "POST"
+    case .put:
+      return "PUT"
+    case .trace:
+      return "TRACE"
+    }
```
goaway: super nit, but if we put these all in a const array indexed by the enum value, we can avoid all the branching generated by the switch.

library/swift/RequestMethod.swift
rebello95: I mentioned this in the PR description, but unfortunately we have to use a variant of `Int` instead of `String` as the raw value in order for this to be visible to Objective-C

library/swift/RetryPolicy.swift
```diff
@@ -0,0 +1,47 @@
+import Foundation
+
+/// Rules that may be used with `RetryPolicy`.
+/// See the `x-envoy-retry-on` Envoy header for documentation.
+@objc
+public enum RetryRule: Int {
```
goaway: Same thoughts as on the above enum implementation.

library/swift/RequestBuilder.swift
rebello95: That's a good callout. What about:

```swift
/// Remove all headers with this name
func removeHeaders(named name: String)

/// Remove an exact match header
func removeHeader(name: String, value: String)
```

library/swift/RequestMethod.swift
rebello95: We definitely could do that, my only concern is that it's a bit fragile for cases where we add a new log level (if someone were to forget to add it to the string array). Do you think it's worth trading compiler-enforcement that each type has a `String` value in an exhaustive `switch`?

The alternative would be something like this:

```swift
/// List of log levels indexed by `LogLevel.rawValue`.
private let kLogLevelStringValues = [
  "trace",
  "debug",
  "info",
  "warn",
  "error",
  "critical",
  "off",
]

/// Available logging levels for an Envoy instance.
/// Note that some levels may be compiled out.
@objc
public enum LogLevel: Int {
  case trace = 0
  case debug
  case info
  case warn
  case error
  case critical
  case off

  /// String representation of the log level.
  var stringValue: String {
    return kLogLevelStringValues[self.rawValue]
  }
}
```

library/swift/RequestBuilder.swift
```diff
@@ -0,0 +1,106 @@
...
+  // Serialized data to send as the body of the request.
+  public private(set) var body: Data?
+  // Retry policy to use for this request.
+  public private(set) var retryPolicy: RetryPolicy?
+
+  // MARK: - Initializers
+
+  /// Internal initializer used for converting a request back to a builder.
+  init(request: Request) {
+    self.url = request.url
```
goaway: https://github.com/lyft/envoy-mobile/pull/240/files#r302691380

(cross-posting comments from your two PRs since we'll ultimately want consistency)

library/swift/RetryPolicy.swift
rebello95: https://github.com/lyft/envoy-mobile/pull/242#discussion_r302692922

library/swift/RequestMethod.swift
rebello95: Discussed further offline, and something like what I've pasted below would work. However I don't really think this gives us much benefit:
- We end up with more code because we have to do 2-way mappings from `String`
- We probably don't want to expose the raw `String` values publicly (which would be required here) - they are only needed for communicating to the C library

Thoughts?

```swift
@objc
enum Foo: Int, RawRepresentable {
    case abc

    typealias RawValue = String

    init?(rawValue: String) {
        switch rawValue {
        case "abc":
            self = .abc
        default:
            return nil
        }
    }

    var rawValue: String {
        switch self {
        case .abc:
            return "abc"
        }
    }
}
```

library/swift/RequestBuilder.swift
rebello95: https://github.com/lyft/envoy-mobile/pull/240/files#r302696592 😁

library/swift/RequestBuilder.swift
goaway: Yep, I like that.

library/swift/Request.swift
rebello95: Absolutely agree that streaming is something we need to support early on from the native interfaces. From our offline discussion, let's discuss outside of this PR's context, and add separately. Tracking in this issue: https://github.com/lyft/envoy-mobile/issues/247

library/swift/RequestMethod.swift
goaway: I mean, it's definitely a silly micro-optimization. Since I think this will very rarely change, my slight preference is still to do the optimization, but I think you could make the very fair case that the compiler check on the switch statement is more valuable. I'm good with however you'd like to proceed.

library/swift/RequestMethod.swift
goaway: To clarify on my emoji, I agree with your take that this doesn't actually end up benefiting us a lot. :) I was thinking it would let us get implicit casts to Strings with the enum values, and if it doesn't, there's no real reason to go down this path.

rebello95: Sounds good!

library/swift/RequestMethod.swift
rebello95: I'd prefer to leave it like this if you don't feel strongly. Since this codepath is rarely traversed (once per startup of Envoy at the moment), I think using a `switch` instead of a potentially faster array lookup is a good tradeoff for compile-time safety

library/swift/RequestBuilder.swift
rebello95: Done

goaway:
nice work

.pre-commit-config.yaml
```diff
@@ -14,3 +14,4 @@ repos:
   rev: v1.15.0
   hooks:
     - id: codespell
+      args: [-L uint]
```
goaway: well, nice to know how to do this

library/swift/RetryPolicy.swift
```diff
@@ -0,0 +1,47 @@
...
+/// Specifies how a request may be retried, containing one or more rules.
+/// https://www.envoyproxy.io/learn/automatic-retries
+@objcMembers
+public final class RetryPolicy: NSObject {
+  /// Maximum number of retries that a request may be performed.
+  public let maxRetryCount: UInt
+  /// Whitelist of rules used for retrying.
+  public let retryOn: [RetryRule]
+  /// Timeout (in milliseconds) to apply to each retry.
+  public let perRetryTimeoutMS: UInt?
```
buildbreaker: What should the default value be here?

rebello95: `nil`

PR #249, opened 2019-07-12
ci: add unit testing for Swift

rebello95:
Add support and examples for running Swift unit tests on CI to match up with Kotlin support that was added in https://github.com/lyft/envoy-mobile/pull/231.

This also moves the Swift source code into a `src` directory to match up with Kotlin.

Resolves https://github.com/lyft/envoy-mobile/issues/248


azure-pipelines.yml
```diff
@@ -284,6 +284,18 @@ stages:
...
+        pool:
+          vmImage: 'macos-10.14'
+        steps:
+          - checkout: self
+            submodules: true
+          - script: ./ci/mac_ci_setup.sh
+            displayName: 'Install dependencies'
+          - script: |
+              ./bazelw test --test_output=all --build_tests_only //library/swift/test/...
+            displayName: 'Run Swift library tests'
```
keith: looks like this isn't interpreted correctly because of the multiline string above

rebello95: Seems to have been working for Android, but updated

keith: maybe it's just github's rendering then, but i think this is nicer

rebello95: Yep same

azure-pipelines.yml
```diff
@@ -168,9 +168,8 @@ stages:
               export CC=clang
               export CXX=clang++
             displayName: 'Set up clang toolchain'
-          - script: |
-              ./bazelw test --test_output=all --build_tests_only //library/kotlin/test/...
-            displayName: '//library/kotlin tests'
+          - script: ./bazelw test --test_output=all --build_tests_only //library/kotlin/test/...
```
buildbreaker: Thanks

PR #251, opened 2019-07-12
core: c++ scaffolding for main api

junr03:
Description: this PR adds a few foundational constructs to implement the c API introduced in #225.
The approach take here is to create an Http dispatcher that is able to manage http streams via the APIs introduced in #225. The streams created by the dispatcher can be interacted with in the mobile platform, and ultimately map down to direct injection into the envoy router via upstream envoy constructs (async client, async stream, and the router).

The code introduced here is capable of making a headers only request from platform code, and receiving the corresponding callback in platform code.

Subsequent PRs will implement the other relevant methods to deal with data and trailers.

Risk Level: med - foundational constructs introduced.
Testing: added cpp unit tests, and a CI job to run them.


junr03:
@goaway this is a green build of what I had been working on plus what we worked on this afternoon. Note I left two FIXMEs behind.

goaway:
Will update with a list of follow up tasks.

goaway:
Issues that stem from this for v0.2:
https://github.com/lyft/envoy-mobile/issues/280
https://github.com/lyft/envoy-mobile/issues/281
https://github.com/lyft/envoy-mobile/issues/282
https://github.com/lyft/envoy-mobile/issues/283
https://github.com/lyft/envoy-mobile/issues/284
https://github.com/lyft/envoy-mobile/issues/285

goaway:
Perhaps it merits clarification that while the full test harness has been stripped from this PR, this code _is_ functional: we've used it to successfully send requests and receives response via direct injection into Envoy.

library/common/http/dispatcher.cc
```diff
@@ -0,0 +1,105 @@
...
+                                                       // getAnIncrement(); confirm or update.
+
+  event_dispatcher_.post([this, observer, new_stream_id]() -> void {
+    DirectStreamCallbacksPtr callbacks =
+        std::make_unique<DirectStreamCallbacks>(new_stream_id, observer, *this);
+    AsyncClient& async_client = cluster_manager_.httpAsyncClientForCluster("egress_cluster");
+    AsyncClient::Stream* underlying_stream = async_client.start(*callbacks, {});
+
+    DirectStreamPtr direct_stream =
+        std::make_unique<DirectStream>(underlying_stream, std::move(callbacks));
```
mattklein123: Assuming we plan on handling null above, prefer passing a reference to underlying_stream here vs. pointer.

library/common/http/dispatcher.cc
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// TODO: implement.
+envoy_status_t Dispatcher::sendData(envoy_stream_t, envoy_headers, bool) { return ENVOY_FAILURE; }
+envoy_status_t Dispatcher::sendMetadata(envoy_stream_t, envoy_headers, bool) {
+  return ENVOY_FAILURE;
+}
+envoy_status_t Dispatcher::sendTrailers(envoy_stream_t, envoy_headers) { return ENVOY_FAILURE; }
+envoy_status_t Dispatcher::locallyCloseStream(envoy_stream_t) { return ENVOY_FAILURE; }
+envoy_status_t Dispatcher::resetStream() { return ENVOY_FAILURE; }
```
mattklein123: Does't this need to take an envoy_stream_t?

library/common/http/dispatcher.cc
```diff
@@ -0,0 +1,105 @@
...
+
+Dispatcher::Dispatcher(Event::Dispatcher& event_dispatcher,
+                       Upstream::ClusterManager& cluster_manager)
+    : event_dispatcher_(event_dispatcher), cluster_manager_(cluster_manager) {}
+
+envoy_stream_t Dispatcher::startStream(envoy_observer observer) {
+  envoy_stream_t new_stream_id = current_stream_id_++; // FIXME: This needs to be equivalent to
+                                                       // getAnIncrement(); confirm or update.
+
+  event_dispatcher_.post([this, observer, new_stream_id]() -> void {
```
mattklein123: Per our offline conversation, there are various cancellation races here. Can you add relevant comments/TODOs/etc. to guide further development? This is also related to the threading model of the observer callbacks which I also asked about elsewhere.

library/common/buffer/utility.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include "envoy/buffer/buffer.h"
+
+#include "library/common/include/c_types.h"
+
+namespace Envoy {
+namespace Buffer {
+namespace Utility {
+
+/**
```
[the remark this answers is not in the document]
junr03: We are still defining this. We had a good discussion this morning, and @goaway is going to flesh out in a doc. This PR does not address any buffer ownerships concern (doing that in a subsequent PR) so perhaps it is best if we get rid of the buffer util for this PR?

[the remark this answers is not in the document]
goaway: I also created an issue to track, and yeah I will be posting a write up of the background research I've done, and the approach I'm proposing for our initial pass.

library/common/http/dispatcher.cc
```diff
@@ -0,0 +1,105 @@
...
+                                                         envoy_observer observer,
+                                                         Dispatcher& http_dispatcher)
+    : stream_(stream), observer_(observer), http_dispatcher_(http_dispatcher) {}
+
+void Dispatcher::DirectStreamCallbacks::onHeaders(HeaderMapPtr&& headers, bool end_stream) {
+  ENVOY_LOG(debug, "response headers for stream (end_stream={}):\n{}", end_stream, *headers);
+  if (end_stream) {
+    http_dispatcher_.removeStream(stream_);
+  }
+  observer_.h(stream_, Utility::transformHeaders(std::move(headers)), end_stream);
```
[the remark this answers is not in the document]
goaway: I'll write something up. Note I have another branch up which reworks the observers a little bit. At this layer the callbacks are coming from the thread Envoy issues the callbacks from. However, baked into the function pointer exposed at this layer will be platform-specific dispatch (to a dispatch_queue on iOS and an Executor on Android), that will result on the callback itself being executed on a different (and optionally caller-provided) thread.

library/common/http/dispatcher.cc
```diff
@@ -0,0 +1,105 @@
...
+Dispatcher::DirectStream::DirectStream(AsyncClient::Stream* underlying_stream,
+                                       DirectStreamCallbacksPtr callbacks)
+    : underlying_stream_(underlying_stream), callbacks_(std::move(callbacks)) {}
+
+Dispatcher::Dispatcher(Event::Dispatcher& event_dispatcher,
+                       Upstream::ClusterManager& cluster_manager)
+    : event_dispatcher_(event_dispatcher), cluster_manager_(cluster_manager) {}
+
+envoy_stream_t Dispatcher::startStream(envoy_observer observer) {
+  envoy_stream_t new_stream_id = current_stream_id_++; // FIXME: This needs to be equivalent to
```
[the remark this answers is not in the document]
goaway: Actually, it was supposed to be. I think it accidentally got stripped from this PR.

library/common/http/dispatcher.cc
```diff
@@ -0,0 +1,105 @@
...
+                                                       // getAnIncrement(); confirm or update.
+
+  event_dispatcher_.post([this, observer, new_stream_id]() -> void {
+    DirectStreamCallbacksPtr callbacks =
+        std::make_unique<DirectStreamCallbacks>(new_stream_id, observer, *this);
+    AsyncClient& async_client = cluster_manager_.httpAsyncClientForCluster("egress_cluster");
+    AsyncClient::Stream* underlying_stream = async_client.start(*callbacks, {});
```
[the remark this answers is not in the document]
goaway: In brief, our planned approach to address this race is to unhook observers immediately upon cancelation. Then we can dispatch upstream on the same dispatch thread as other ongoing work, without worrying about relative timing.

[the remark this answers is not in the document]
junr03: Yep good catch, I meant to do checking in the DirectStream's constructor. But I like your suggestion below better.

library/common/http/dispatcher.h
```diff
@@ -0,0 +1,84 @@
...
+    envoy_stream_t stream_;
+    envoy_observer observer_;
+    Dispatcher& http_dispatcher_;
+  };
+
+  using DirectStreamCallbacksPtr = std::unique_ptr<DirectStreamCallbacks>;
+
+  class DirectStream {
+  public:
+    // FIXME follow up on concerns about the lifetime of the stream pointer.
```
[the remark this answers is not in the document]
goaway: Sure thing. Basically, we have lingering concerns around the timing of relinquishing our Stream ownership (thus causing it to be freed). For instance, _inside_ the execution of stream callback that sets `end_stream` to true or indicates an error, can we safely drop the stream? Do we need to queue it on our dispatcher for removal after the callback returns? Is there some other pattern we should adopt?

library/common/http/header_utility.cc
```diff
@@ -0,0 +1,47 @@
...
+  Http::HeaderMapPtr transformed_headers = std::make_unique<HeaderMapImpl>();
+  for (uint64_t i = 0; i < headers.length; i++) {
+    transformed_headers->addCopy(LowerCaseString(convertString(headers.headers[i].key)),
+                                 convertString(headers.headers[i].value));
+  }
+  return transformed_headers;
+}
+
+envoy_headers transformHeaders(HeaderMapPtr&& header_map) {
+  envoy_header* headers = new envoy_header[header_map->size()];
```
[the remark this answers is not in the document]
junr03: Yes, that is the expectation. Afaik we can `free` in C what we have `new`ed here. Did you mean something by that part, or just that we should probably provide utility functions to make this process standard?

We have a tracking issue for this: https://github.com/lyft/envoy-mobile/issues/280

library/common/http/header_utility.h
```diff
@@ -0,0 +1,31 @@
...
+ * This function copies the content.
+ * @param headers, the envoy_headers to transform.
+ * @return HeaderMapPtr, the HeaderMap 1:1 transformation of the headers param.
+ */
+HeaderMapPtr transformHeaders(envoy_headers headers);
+
+/**
+ * Transform envoy_headers to HeaderMap.
+ * This function copies the content.
+ * Caller owns the allocated bytes for the return value, and needs to free after use.
```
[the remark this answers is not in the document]
junr03: Yep, we have a tracking issue: https://github.com/lyft/envoy-mobile/issues/280

library/common/buffer/utility.h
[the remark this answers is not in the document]
junr03: Issue is #284

library/common/include/c_types.h
```diff
@@ -0,0 +1,130 @@
...
+
+#ifdef __cplusplus
+extern "C" { // function pointers
+#endif
+/**
+ * Called when all headers get received on the async HTTP stream.
+ * @param headers, the headers received.
+ * @param end_stream, whether the response is headers-only.
+ */
+typedef void (*on_headers)(envoy_headers headers, bool end_stream, void* context);
```
rebello95: Do we also need an observer for `on_metadata`? Looks like that's missing here

goaway: Envoy doesn't actually support it yet, but this is still a good callout. Let's add it so we can have everything rigged for when support does land.

junr03:
@mattklein123 this is ready for another pass.

library/common/http/dispatcher.cc
```diff
@@ -0,0 +1,108 @@
...
+                                                         envoy_observer observer,
+                                                         Dispatcher& http_dispatcher)
+    : stream_(stream), observer_(observer), http_dispatcher_(http_dispatcher) {}
+
+void Dispatcher::DirectStreamCallbacks::onHeaders(HeaderMapPtr&& headers, bool end_stream) {
+  ENVOY_LOG(debug, "response headers for stream (end_stream={}):\n{}", end_stream, *headers);
+  if (end_stream) {
+    http_dispatcher_.removeStream(stream_);
+  }
+  observer_.on_headers_f(Utility::transformHeaders(*headers), end_stream, observer_.context);
```
mattklein123: nit: I would remove the `_f` trailer from all of these functions, but TIOLI.

library/common/http/dispatcher.cc
```diff
@@ -0,0 +1,108 @@
...
+      callbacks->onReset();
+    } else {
+      DirectStreamPtr direct_stream =
+          std::make_unique<DirectStream>(*underlying_stream, std::move(callbacks));
+      streams_.emplace(new_stream_id, std::move(direct_stream));
+      ENVOY_LOG(debug, "started stream [{}]", new_stream_id);
+    }
+  });
+
+  return new_stream_id;
```
mattklein123: Do you still want to return the stream ID in the case of an immediate failure above? Should the return value indicate the failure in some way? Either way can you document this case in the function docs?

library/common/http/dispatcher.cc
```diff
@@ -0,0 +1,108 @@
...
+}
+
+envoy_status_t Dispatcher::sendHeaders(envoy_stream_t stream_id, envoy_headers headers,
+                                       bool end_stream) {
+  event_dispatcher_.post([this, stream_id, headers, end_stream]() -> void {
+    DirectStream* direct_stream = getStream(stream_id);
+    if (direct_stream != nullptr) {
```
mattklein123: When does this happen? Should this be an ASSERT and/or failure?

library/common/http/header_utility.cc
```diff
@@ -0,0 +1,50 @@
...
+  transformed_headers.headers = headers;
+
+  header_map.iterate(
+      [](const HeaderEntry& header, void* context) -> HeaderMap::Iterate {
+        envoy_headers* transformed_headers = static_cast<envoy_headers*>(context);
+
+        const absl::string_view header_key = header.key().getStringView();
+        const absl::string_view header_value = header.value().getStringView();
+
+        envoy_string key = {header_key.size(), strdup(header_key.data())};
```
mattklein123: You can't assume that the header strings are null-terminated. Use malloc and then memcpy and then null terminate if needed. Same next line.

library/common/include/c_types.h
```diff
@@ -0,0 +1,147 @@
...
+  envoy_stream_t stream;
+} envoy_stream;
+
+/**
+ * Holds a single key/value header.
+ */
+typedef struct {
+  envoy_string key;
+  // Multiple header values for the same header key are supported via a comma-delimited string.
+  envoy_string value;
```
mattklein123: Personally I would swap with envoy_data here as a general indication that these things are not necessarily null terminated.

library/common/http/header_utility.cc
goaway: yep, we were cheating, will fix.

library/common/http/dispatcher.cc
junr03: We can't synchronously know of immediate failure because it all runs on the dispatched post. The app will get the onReset() callback if failure to get a stream happens. We will document on the function docs.

library/common/include/c_types.h
```diff
@@ -0,0 +1,139 @@
...
+ * Called when a data frame gets received on the async HTTP stream.
+ * This callback can be invoked multiple times if the data gets streamed.
+ * @param data, the data received.
+ * @param end_stream, whether the data is the last data frame.
+ * @param context, contains the necessary state to carry out platform-specific dispatch and
+ * execution.
+ */
+typedef void (*on_data)(envoy_data data, bool end_stream, void* context);
+/**
+ * Called when all trailers get received on the async HTTP stream.
```
rebello95:
```suggestion
 * Called when all metadata gets received on the async HTTP stream.
```

library/common/include/c_types.h
```diff
@@ -0,0 +1,139 @@
...
+ * Note that end stream is implied when on_trailers is called.
+ * @param metadata, the trailers received.
```
rebello95:
```suggestion
 * @param metadata, the metadata received.
```

junr03:
@mattklein123 this is ready for another pass.

mattklein123:
@goaway @junr03 friendly reminder to please not force push, it makes reviews a lot more difficult. Thanks!

library/common/http/dispatcher.cc
mattklein123: FYI it's possible this will happen before this function returns, so need to consider that as one of the cases in your cancellation follow ups.

library/common/http/dispatcher.cc
mattklein123: Ping on this question.

mattklein123: Nevermind I see your comment above, but presumably this only happens during a cancellation/failure race condition? Can you expand the comment on why/when this would actually happen?

library/common/http/header_utility.cc
```diff
@@ -0,0 +1,65 @@
...
+namespace Envoy {
+namespace Http {
+namespace Utility {
+
+static inline envoy_data copyEnvoyData(size_t length, const uint8_t* source) {
+  uint8_t* destination = static_cast<uint8_t*>(malloc(sizeof(uint8_t) * length));
+  memcpy(destination, source, length);
+  return {length, destination};
+}
+/*
```
mattklein123: ?

library/common/http/dispatcher.cc
junr03: Sorry, answered in an inline comment as it seemed valuable to preserve in comment: 

```
    // If direct_stream is not found, it means the stream has already closed or been reset
    // and the appropriate callback has been issued to the caller. There's nothing to do here
    // except silently swallow this.
```

library/common/http/header_utility.cc
junr03: Defined the function but not used yet anywhere in this code. Will delete for now.

library/common/http/dispatcher.cc
junr03: Additional to that there is the even simpler case, which is that the caller might have called with a bogus stream_id.

So even when we cover the threading issues and resolve the potential cancellation/failure race condition we still have to decide what to do with bogus input. Do we want to silently swallow, or ASSERT that the pointer should never be null?

junr03:
@mattklein123 thanks for the reviews. I responded to your remaining comments.

> friendly reminder to please not force push, it makes reviews a lot more difficult. Thanks!

Yeah sorry about that the commit history got a bit unwieldy with the subsequent branches we have opened. We'll avoid rebases and force pushes going forward.

library/common/http/dispatcher.cc
mattklein123: I think we need to sort out the error handling semantics and be consistent. IMO you should return some type of error in this case. I would add a TODO to track this.

junr03: @mattklein123 agreed. I added an inline comment and opened a tracking issue.

PR #255, opened 2019-07-12
swift: include missing library files

rebello95:
These aren't being compiled in right now.


rebello95:
I believe @goaway and @buildbreaker had a discussion about not globbing - personally I'm in favor of it 🤷‍♂

PR #257, opened 2019-07-12
swift: add macro for unit testing

rebello95:
Conceptually similar to the macro added in https://github.com/lyft/envoy-mobile/pull/221 for Kotlin testing, this PR adds a `envoy_mobile_swift_test` macro for Swift unit testing.

- Prevents consumers from having to define both swift_library and ios_unit_test targets
- Provides a set of linker options that is required to properly run tests
- Sets default visibility and OS requirements


keith:
https://github.com/bazelbuild/rules_apple/issues/490

rebello95:
Thanks for filing that @keith. I think this PR is blocked until that is resolved or we introduce a workaround in this repo

PR #258, opened 2019-07-15
cleanup: remove legacy objective-c interface

goaway:


rebello95:
Don't we still need this as it's called from the Swift layer right now?

goaway:
Swift calls into `EnvoyEngine`; I don't believe anything should be relying on this anymore.

goaway:
@rebello95 if you meant the build target though, you were totally right 🤦‍♂

PR #260, opened 2019-07-15
kotlin: add library response interfaces

buildbreaker:
https://github.com/lyft/envoy-mobile/issues/119
Adding a primitive Request and RequestBuilder with basic features:
* body
* headers
* trailers

Kotlin usages:
```
val result: Result = ...

val body = result.fold(
	{ response -> 
		response.body 
	}, 
	{ failure -> 
		throw Exception(failure.cause) 
	})

// Default to a value
val value = result.fold(
	{ response -> 
		1 
	}, 
	{ failure -> 
		0 
	})

// Side effect response
result.response({ response ->
	val body = response.body
	// do something
})

// Side effect failure
result.failure({ failure ->
	val message = failure.message
	// do something
})
```

Java usages:
```
byte[] body = result.fold(response -> {
		return response.body;
	},
	failure -> {
		throw new Exception(failure.cause)
	});

// Default to a value
byte[] body = result.fold(response -> {
		return 1;
	},
	failure -> {
		return 0;
	});

// Side effect response
result.response(response -> {
		byte[] body = response.body;
		// do something
		return null;
	});

// Side effect failure
result.failure(failure -> {
		String message = failure.message
		// do something
		return null;
	})
```

Tests are basic setting a builder and ensuring the result is expected. Additionally, the response builder transform creates an equivalent object when built back. These are basically testing the exact same behavior as the [requests](https://github.com/lyft/envoy-mobile/pull/240)

Description: Add response objects
Risk Level: low
Testing: unit
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


goaway:
This echoes my question on the Request interfaces, but have we thought at all about how might expose response data for streamed processing?

This is standard in most HTTP library interfaces, and generally processing/handling of responses tends to be something that can make progress in an online fashion faster than bytes come over the network (i.e. you shouldn't wait to start processing until all bytes have arrived).

goaway:
I was sort of thinking that an Observer pattern (similar to the c-APIs, but with platform-specific facilities and sugar) would be our first-class response interface, given the flexibility it affords.

goaway:
Something very roughly like:
```
onResponse(StatusCode x, ResponseHeaders y)
onData(ByteBuffer b)
onCompletion(ResponseTrailers t)
onError(Error e)
```

library/kotlin/src/io/envoyproxy/envoymobile/Result.kt
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * A side effect function for successful Results
+   *
+   * @param response a function which is called on a response for a request
+   */
+  fun response(response: (Response) -> Unit) = fold(response, {})
+
+
+  /**
+   * A side effect function for successful Results
```
junr03: comment needs updating

library/kotlin/src/io/envoyproxy/envoymobile/Result.kt
```diff
@@ -0,0 +1,105 @@
...
+   * This method transforms a given response or failure to a specified type which
+   * the user can define. Usages of this can be:
+   *
+   * val result: Result = ...
+   *
+   * // To throw an exception when a failure occurs
+   * result.fold({ response -> response.body }, { failure -> throw Exception(failure.cause) })
+   *
+   * // To return a default value of 0 when a failure occurs
+   * result.fold({ response -> 1 }, { failure -> 0 })
```
junr03: Sorry, this is my confused eyes as a kotlin ignorant, but how does the user interact with the result. Do they write complex folds where the logic is after `->` for response, and failure? And they have access to the Response and the NetworkError.

library/kotlin/src/io/envoyproxy/envoymobile/Result.kt
```diff
@@ -0,0 +1,105 @@
...
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (javaClass != other?.javaClass) return false
+
+    other as Response
+
+    if (status != other.status) return false
+    if (body != null) {
+      if (other.body == null) return false
+      if (!body.contentEquals(other.body)) return false
```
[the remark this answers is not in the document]
buildbreaker: We can create our own byte array class which handles this and then we'll be fine going forward. Since we are planning to swap this out with some input stream which is more common, I was thinking of addressing this issue then

library/kotlin/src/io/envoyproxy/envoymobile/Result.kt
buildbreaker: I have an example usage of this up in the description but you do have the right idea

goaway:
I should say that I do think the Response object in here is a useful interface, and that the usage examples are elegant.

But I'd like to maybe take some time to think about how it can work as part of or alongside an interface that allows application processing to proceed incrementally as data comes in. (This is possible for headers as well, but generally somewhat less important than for the body payload.)

rebello95:
> This echoes my question on the Request interfaces, but have we thought at all about how might expose response data for streamed processing?

Definitely a great callout, and I think it's worth discussing prior to merging the Swift/Kotlin initial interface implementations. Some options @buildbreaker and I were discussing earlier, just to outline them here:

_The ergonomic `Result` types available in Swift/Kotlin are unavailable in Objective-C/Java, so unfortunately we need to come up with an option that supports this functionality in all 4 languages._

1. Adding a `Callback` class that can be initialized with the success/failure lambdas
2. Passing success/failure lambdas directly in to the request function, one of which will be called on completion
3. Using the custom `Result` types here, providing a single lambda which gets called on completion with the result

That said, we could also go down a path like what you suggested @goaway and essentially expose the [interfaces available from C that are checked in today](https://github.com/lyft/envoy-mobile/blob/ae3daf9f788ee863a146424042c388c6720efc52/library/common/main_interface.h#L91-L114), and provide a bit of ergonomics on top of those (i.e., using `Response`, parsing out status codes, etc.). To be honest, **I think this is probably the most practical way to support streaming.**

If this is a path we want to explore, @buildbreaker and I can definitely take a stab at it

rebello95:
@buildbreaker opened another PR with some options for response interfaces based on the discussion above: https://github.com/lyft/envoy-mobile/pull/265

I went ahead and left some comments if you want to also take a look @goaway @junr03

buildbreaker:
Going to go in a different direction: https://github.com/lyft/envoy-mobile/pull/265

PR #261, opened 2019-07-16
swift: add response interfaces

rebello95:
Adds Swift interfaces for responses within Envoy. These should be at parity with the interfaces being added for Kotlin in https://github.com/lyft/envoy-mobile/pull/260.

Resolves https://github.com/lyft/envoy-mobile/issues/118.

Doc for reference: https://docs.google.com/document/d/1N0ZFJktK8m01uqqgfDRVB9mpC1iEn9dqkQaa_yMn_kE/edit#heading=h.i6ky65xaa9va


library/swift/src/BUILD
```diff
@@ -10,10 +10,14 @@ swift_static_framework(
     name = "ios_framework",
     srcs = [
         "Envoy.swift",
+        "EnvoyResult.swift",
```
buildbreaker: Hah, should we keep `Envoy` as a prefix here as well? Feels a bit redundant since we have the module/package naming

WDYT?

library/swift/src/NetworkError.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+/// Error representing cases when no response was received from the server.
+/// I.e., the client went offline or became disconnected.
+@objcMembers
+public final class NetworkError: NSObject, Swift.Error {
```
buildbreaker: Hm, more of a platform question:
It is a swift convention to return an error and let the caller handle the error that way, correct?

library/swift/src/BUILD
rebello95: Ah yea I meant to rename this to just `Result`

library/swift/src/NetworkError.swift
rebello95: > It is a swift convention to return an error and let the caller handle the error that way, correct?

Are you referring to doing this instead of `throw`ing?

buildbreaker: Yeah

rebello95: The benefit of returning an error is that it can be typed. Swift doesn’t allow you to specify which errors can be thrown by a function, so consumers would have to catch all errors and cast. Also throwing wouldn’t work for async

goaway:
Cross-posting from here:
https://github.com/lyft/envoy-mobile/pull/260#issuecomment-512032976

This echoes my question on the Request interfaces, but have we thought at all about how might expose response data for streamed processing?

This is standard in most HTTP library interfaces, and generally processing/handling of responses tends to be something that can make progress in an online fashion faster than bytes come over the network (i.e. you shouldn't wait to start processing until all bytes have arrived).

rebello95:
Will comment on that PR so we can discuss, thanks for reviewing! 😄

rebello95:
Replaced by: https://github.com/lyft/envoy-mobile/pull/273

PR #264, opened 2019-07-16
kotlin: add bazel macros for easier kt target declaration

buildbreaker:
We currently have to redeclare our sources in the `srcs` argument for the top level library consumed by the `aar_with_jni` macro. The reason is because the `aar_with_jni` doesn't get visibility onto the transitive dependencies of the provided library. Since we are using `kt_jvm_library`, which internally creates two rules and encloses the dependencies within them, the library we provide does not expose any dependencies for the `aar_with_jni` target.

This makes it very complicated to create new libraries because for every new library, we have to create a `filegroup` which the top level `envoy_lib` needs to depend on directly in the `srcs`.

This is a total bazel macro hack to hide away this detail for us. The limitation here is that the top level `envoy_lib` needs to directly depend on the targets.

Alternative solution is to write a rule which traverses the dependencies and takes out the appropriate `srcs` from each dependency and transitive dependency using an `envoy_mobile_kt_library`.

Before:
```
aar_with_jni(
    name = "android_aar",
    android_library = "envoy_lib",
    archive_name = "envoy",
)
kt_android_library(
    name = "envoy_lib",
    srcs = [
        "Envoy.kt",
        "//library/java/io/envoyproxy/envoymobile/engine:envoy_engine_srcs", # Redundancy
    ],
    custom_package = "io.envoyproxy.envoymobile",
    manifest = "EnvoyManifest.xml",
    deps = ["//library/java/io/envoyproxy/envoymobile/engine:envoy_engine_lib"],
)
# Redundancy
filegroup(
    name = "envoy_engine_srcs",
    srcs = ["EnvoyEngine.java"],
)
android_library(
    name = "envoy_engine_lib",
    srcs = [":envoy_engine_srcs"], # Redundancy
    custom_package = "io.envoyproxy.envoymobile.engine",
    manifest = "EnvoyEngineManifest.xml",
    deps = ["//library/common:envoy_jni_interface_lib"],
)
```
Now:
```

aar_with_jni(
    name = "android_aar",
    android_library = "envoy_lib",
    archive_name = "envoy",
)
envoy_mobile_kt_aar_android_library(
    name = "envoy_lib",
    srcs = [
        "Envoy.kt",
    ],
    custom_package = "io.envoyproxy.envoymobile",
    manifest = "EnvoyManifest.xml",
    deps = ["//library/java/io/envoyproxy/envoymobile/engine:envoy_engine_lib"],
)
envoy_mobile_android_library(
    name = "envoy_engine_lib",
    srcs = ["EnvoyEngine.java"],
    custom_package = "io.envoyproxy.envoymobile.engine",
    manifest = "EnvoyEngineManifest.xml",
    deps = ["//library/common:envoy_jni_interface_lib"],
)
```

Description: add bazel macros for easier kt target declaration
Risk Level: low
Testing: local
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/bazel/kotlin_lib.bzl
+++ b/bazel/kotlin_lib.bzl
@@ -1,0 +1,56 @@
+licenses(["notice"])  # Apache 2
+
+load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_android_library", "kt_jvm_library")
+
+# Android library drops exported dependencies from dependent rules. The kt_android_library
+#  internally is just a macro which wraps two rules into one:
+#  https://github.com/bazelbuild/rules_kotlin/blob/326661e7e705d14e754abc2765837aa61bddf205/kotlin/internal/jvm/android.bzl#L28.
+#  This causes the sources to be exported and dropped due to it being a transitive dependency.
+#  To get around this, we have to redeclare the sources from envoy_engine_lib here in order to be pulled into the
+#  kotlin jar.
+def envoy_mobile_kt_aar_android_library(name, custom_package, manifest, srcs = [], deps = []):
+    filegroups = []
+    for dep in deps:
+        filegroups.append(dep + "_srcs")
+
+    kt_android_library(
+        name = name,
+        srcs = srcs + filegroups,
+        custom_package = custom_package,
+        manifest = manifest,
+        visibility = ["//visibility:public"],
+        deps = deps,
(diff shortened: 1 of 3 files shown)


buildbreaker:
thanks @keith

PR #265, opened 2019-07-18
kotlin: response interfaces

buildbreaker:
Adding the `Client` interface and the streaming request interface method for opening a stream

Description: Adding the `Client` and the streaming request method for opening a stream
Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
```diff
@@ -0,0 +1,73 @@
...
+import java.nio.ByteBuffer
+
+interface Client {
+  fun startStream(request: Request, streamCallback: StreamCallback): StreamEmitter
+}
+
+interface StreamCallback {
+  fun onHeaders(statusCode: Int, headers: Map<String, List<String>>)
+  fun onData(byteBuffer: ByteBuffer, endStream: Boolean)
+  fun onMetadata(metadata: Map<String, List<String>>, endStream: Boolean)
```
rebello95: Guessing we probably don't need this in the initial implementation 🤔

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
```diff
@@ -0,0 +1,73 @@
...
+  fun sendTrailers(trailers: Map<String, List<String>>): Boolean
+
+  fun close(): Boolean
+  fun interrupt(): Boolean
+}
+
+class Error(
+    val code: Int,
+    val message: String,
+    val cause: Throwable? = null
```
rebello95: What is the intended use case for this field?

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
```diff
@@ -0,0 +1,73 @@
+package io.envoyproxy.envoymobile
+
+import java.nio.ByteBuffer
+
+interface Client {
+  fun startStream(request: Request, streamCallback: StreamCallback): StreamEmitter
```
rebello95: Interesting that this function returns a `StreamEmitter`. I think this makes sense assuming it's more or less to be used as the platform equivalent of a `envoy_stream_t` defined [here](https://github.com/lyft/envoy-mobile/blob/ae3daf9f788ee863a146424042c388c6720efc52/library/common/main_interface.h#L18)?

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
```diff
@@ -0,0 +1,73 @@
...
+  fun onTrailers(trailers: Map<String, List<String>>)
+  fun onError(error: Error)
+
+  fun onCanceled() // TODO: onInterrupt()?
+  fun onCompletion()
+
+}
+
+interface StreamEmitter {
+  fun sendHeaders(headers: Map<String, List<String>>): Boolean
```
rebello95: These don't really align with the request interfaces checked in upstream - is the intent for public consumers not to use these interfaces, but to rather have them be used within the library for calling the C layer? For example, I'd expect an end consumer to be able to pass a `Request` model (with a URL, body, etc.) rather than having to call each of these functions to perform an HTTP request

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
rebello95: Primarily wondering how it'd be used if we already have `code` and `message`

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
buildbreaker: We probably don't but I kept it here since I was planning to close this in favor of a different PR for submitting

My original intent was to show case what we would have and what is most consistent with the common layer

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
buildbreaker: This was primarily for exception surfacing -- I copied what Cronet does as they return this [exception](https://chromium.googlesource.com/chromium/src/+/lkgr/components/cronet/android/api/src/org/chromium/net/CronetException.java?autodive=0%2F):

```
/**
 * Base exception passed to {@link UrlRequest.Callback#onFailed UrlRequest.Callback.onFailed()}.
 */
public abstract class CronetException extends IOException {
    /**
     * Constructs an exception that is caused by {@code cause}.
     *
     * @param message explanation of failure.
     * @param cause the cause (which is saved for later retrieval by the {@link
     *         java.io.IOException#getCause getCause()} method). A null value is permitted, and
     *         indicates that the cause is nonexistent or unknown.
     */
    protected CronetException(String message, Throwable cause) {
        super(message, cause);
    }
}
```

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
buildbreaker: Yeah, I was trying to follow exactly what the common layer exposes with very little modifications

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
```diff
@@ -0,0 +1,73 @@
...
+  fun sendData(byteBuffer: ByteBuffer, endStream: Boolean): Boolean
+  fun sendMetadata(metadata: Map<String, List<String>>, endStream: Boolean): Boolean
+  fun sendTrailers(trailers: Map<String, List<String>>): Boolean
+
+  fun close(): Boolean
```
rebello95: Per offline discussion: Remove since we probably don't need this yet. Semantically, this is essentially setting `endStream=true` in one of the above

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
```diff
@@ -0,0 +1,73 @@
...
+  fun interrupt(): Boolean
```
rebello95:
```suggestion
  fun cancel(): Boolean
```

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
rebello95: Per offline discussion, we should remove this, using:
- The `Request` for headers
- `sendMetadata`
- `sendTrailers`

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
```diff
@@ -0,0 +1,73 @@
...
+  fun onCompletion()
+
+}
+
+interface StreamEmitter {
+  fun sendHeaders(headers: Map<String, List<String>>): Boolean
+  fun sendData(byteBuffer: ByteBuffer, endStream: Boolean): Boolean
+  fun sendMetadata(metadata: Map<String, List<String>>, endStream: Boolean): Boolean
+  fun sendTrailers(trailers: Map<String, List<String>>): Boolean
+
```
rebello95: We should add something like `isActive` to indicate if the stream can continue to be used (set to `false` after `endStream=true`)

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
```diff
@@ -0,0 +1,73 @@
...
+  fun onCanceled() // TODO: onInterrupt()?
+  fun onCompletion()
+
+}
+
+interface StreamEmitter {
+  fun sendHeaders(headers: Map<String, List<String>>): Boolean
+  fun sendData(byteBuffer: ByteBuffer, endStream: Boolean): Boolean
+  fun sendMetadata(metadata: Map<String, List<String>>, endStream: Boolean): Boolean
```
rebello95: Suggestions from discussions:
- Return the same `StreamEmitter` to allow for chaining semantics
- `throwable` instead of a boolean to better identify where a chain fails
- `isActive` can be used per other comment to check state later

rebello95: Also, an error returned here is not recoverable for a stream (stream can't be used if it throws)

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
```diff
@@ -0,0 +1,73 @@
...
+  fun onError(error: Error)
+
+  fun onCanceled() // TODO: onInterrupt()?
+  fun onCompletion()
+
+}
+
+interface StreamEmitter {
+  fun sendHeaders(headers: Map<String, List<String>>): Boolean
+  fun sendData(byteBuffer: ByteBuffer, endStream: Boolean): Boolean
```
rebello95: We can change the signature of these functions to prevent people from chaining on something that will throw:

```
fun sendData(byteBuffer: ByteBuffer): StreamEmitter
fun sendDataAndClose(byteBuffer: ByteBuffer): Void
```

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
rebello95: Per offline convo, we should keep this

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
buildbreaker: Hm, thinking about this some more. I think it's a bit strange to have `sendDataAndClose()` on this layer because we might want to send trailers/metadata also before we close. If we want to support this type of API, we'll have to support a matrix of methods which also send and close.

Would we be opposed to having a `close()` method and we can have better ergonomics which handle the matrix of possibilities at the layer on top of this?

Another option that would be a closer compromise is that we have a `fun close(trailers: Map): Void` method which will close the connection on sending trailers (that way we only support one case rather than 6 other methods)

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
```diff
@@ -0,0 +1,32 @@
...
+  fun onCanceled()
+  fun onCompletion()
+}
+
+interface StreamEmitter {
+  fun isActive(): Boolean
+  fun sendData(byteBuffer: ByteBuffer, endStream: Boolean): StreamEmitter
+  fun sendMetadata(metadata: Map<String, List<String>>, endStream: Boolean): StreamEmitter
+  
+  fun close(trailers: Map<String, List<String>>)
```
rebello95: Should this return a `Boolean`?

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
```diff
@@ -0,0 +1,32 @@
...
+  fun onMetadata(metadata: Map<String, List<String>>, endStream: Boolean)
+  fun onTrailers(trailers: Map<String, List<String>>)
+  fun onError(error: Error)
+  fun onCanceled()
+  fun onCompletion()
+}
+
+interface StreamEmitter {
+  fun isActive(): Boolean
+  fun sendData(byteBuffer: ByteBuffer, endStream: Boolean): StreamEmitter
```
rebello95: `sendData` and `sendMetadata` should throw if they fail to succeed (i.e., `isActive() = false`), right?

rebello95:
Looks great to me after a couple comments!

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
buildbreaker: It was a different take than what we were talking about. More explanation here:  https://github.com/lyft/envoy-mobile/pull/265#discussion_r305097771

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
buildbreaker: I don't think the exception case when `isActive() == false` needed to be declared for the compiler

We can declare that exception, the case when we are unable to send data

buildbreaker:
I'm working on the comments/docs

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
```diff
@@ -0,0 +1,12 @@
...
+
+interface Client {
+  /**
+   * For starting a stream
+   *
+   * @param request the request for opening a stream
+   * @param streamCallback the callback for receiving stream events
+   * @return the emitter for streaming data outward
+   */
+  fun startStream(request: Request, streamCallback: StreamCallback): StreamEmitter
```
rebello95: What do you think about renaming this to simply `start`?

rebello95: Also think `streamCallback` can simply be `callback`

library/kotlin/src/io/envoyproxy/envoymobile/StreamCallback.kt
```diff
@@ -0,0 +1,54 @@
...
+import java.nio.ByteBuffer
+
+interface StreamCallback {
+  /**
+   * Invoked whenever the response headers have been read
+   *
+   * @param statusCode the status code of the response
+   * @param headers the headers of the response
+   */
+  fun onHeaders(statusCode: Int, headers: Map<String, List<String>>)
```
rebello95: nit: I'd put `headers` before `statusCode` so it reads more smoothly

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
buildbreaker: The unary callback will be called `UnaryCallback` then?

Totally fine with this

rebello95: Discussed offline, let's leave streaming as the first-class citizen here

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyException.kt
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * An Envoy Exception class for describing what error may have occurred.
+ *
+ * @param code internal error code associated with the exception that occurred.
+ * @param message a description of what exception that occurred.
+ * @param cause an optional cause for the exception.
+ */
+class EnvoyException(
+    val code: Int,
```
junr03: Is this an HTTP error code? If so should we make this a type that has bound checking as not all integers are valid HTTP error codes?

library/kotlin/src/io/envoyproxy/envoymobile/StreamCallback.kt
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Invoked whenever the response headers have been read.
+   *
+   * @param headers the headers of the response.
+   * @param statusCode the status code of the response.
+   */
+  fun onHeaders(headers: Map<String, List<String>>, statusCode: Int)
+
+  /**
+   * Invoked whenever the response body is read.
```
junr03: it might not be the case that all response body gets sent back at once. i.e there might be multiple onData callbacks fired for a single stream.

In general I would say `Called when a data frame is received by the stream`. Similar in the other function comments here.

library/kotlin/src/io/envoyproxy/envoymobile/StreamEmitter.kt
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * For ending an associated stream and sending trailers.
+   *
+   * @param trailers to send with ending a stream. If no trailers are needed, empty map will be the default.
+   * @throws IllegalStateException when the stream is not active.
+   * @throws EnvoyException when there is an exception ending the stream or sending trailers.
+   */
+  @Throws(EnvoyException::class)
+  fun close(trailers: Map<String, List<String>> = emptyMap())
```
junr03: Interesting, so instead of sendTrailers you combined them into a cancel call.

library/kotlin/src/io/envoyproxy/envoymobile/StreamEmitter.kt
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * For sending data to an associated stream.
+   *
+   * @param byteBuffer the byte buffer data to send to the stream.
+   * @throws IllegalStateException when the stream is not active
+   * @throws EnvoyException when there is an exception sending data.
+   * @return this stream emitter.
+   */
+  @Throws(EnvoyException::class)
+  fun sendData(byteBuffer: ByteBuffer): StreamEmitter
```
junr03: How come this and metadata don't have an `end_stream` any more?

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyException.kt
buildbreaker: This is to mirror what we have in the [`main_interface`](https://github.com/lyft/envoy-mobile/blob/master/library/common/main_interface.h#L84) and I tool out the `envoy` prefix since it's redundant.

library/kotlin/src/io/envoyproxy/envoymobile/StreamEmitter.kt
buildbreaker: Yeah, so we'll have the user explicitly close the emitter/stream when they're done. Optionally, they can send trailers.

The rationale here is that sending trailers means that we are closing a connection (which is symmetrical to how sending headers means we want to open a connection).

buildbreaker: The method signature allows for it to be called two different ways:
1. `close()`
2. `close(trailers)`

library/kotlin/src/io/envoyproxy/envoymobile/StreamEmitter.kt
buildbreaker: Answered here: https://github.com/lyft/envoy-mobile/pull/265#discussion_r305954370

library/kotlin/src/io/envoyproxy/envoymobile/StreamCallback.kt
buildbreaker: I can make this change @rebello95 WDYT?

rebello95: Sure, SGTM

library/kotlin/src/io/envoyproxy/envoymobile/StreamEmitter.kt
junr03: We still need the ability to end the stream at any point, meaning as soon as you send headers/data/metadata, not only when you send trailers

library/kotlin/src/io/envoyproxy/envoymobile/StreamEmitter.kt
junr03: @goaway I would like your opinion here.

library/kotlin/src/io/envoyproxy/envoymobile/StreamEmitter.kt
rebello95: The approach we took here is that to close the stream, the consumer can call `close()` at any point. This can include trailers, or leave them empty (in which case none will be sent). So for example:

```
stream
  .sendData(data)
  .close() // or .close(trailers: myTrailers)
```

rebello95: This makes it explicit to the consumer when they're closing a stream, and gives them the option to include a set of trailers before the stream closes (as opposed to implicitly closing using a boolean flag)

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyException.kt
goaway: I'd suggest considering an `error_` prefix just to make it _very_ clear this is not an HTTP status code.

library/kotlin/src/io/envoyproxy/envoymobile/StreamEmitter.kt
goaway: I'm good with this approach. I think it's probably more intuitive to people who aren't familiar with the underlying protocols, and it's both simple and valid to close a stream at any point with an empty `DATA` frame.

It also neatly solves the invalid chaining issue.

library/kotlin/src/io/envoyproxy/envoymobile/StreamEmitter.kt
goaway: I like this, too. I think it slightly simplifies the interface for high-level usage, without actually sacrificing functionality.

library/kotlin/src/io/envoyproxy/envoymobile/StreamCallback.kt
```diff
@@ -0,0 +1,55 @@
+package io.envoyproxy.envoymobile
+
+import java.nio.ByteBuffer
+
+interface StreamCallback {
```
goaway: `StreamCallback`(s) is accurate, but something like `ResponseHandler` might be more intuitive for a random user of the interface? Just a thought. I'm good with it as is though.

goaway:
Looks great. I'm really liking the direction we've taken here, and I think this close mechanism is a clean way of allowing chaining while preventing the invalid, already-closed case.

@rebello95 this applies to your PR as well, though I'm posting comments here.

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyException.kt
rebello95: I'm fine with changing this to `errorCode`, will update the Swift PR

library/kotlin/src/io/envoyproxy/envoymobile/StreamCallback.kt
rebello95: `ResponseHandler` SGTM, will update Swift

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
```diff
@@ -31,10 +31,14 @@ kt_android_library(
 kt_jvm_library(
     name = "envoy_interfaces_lib",
     srcs = [
+        "Client.kt",
+        "EnvoyException.kt",
         "Request.kt",
         "RequestBuilder.kt",
         "RequestMethod.kt",
         "RetryPolicy.kt",
+        "ResponseHandler.kt",
```
rebello95: nit: alphabetical order (buildifier might fail)

library/kotlin/src/io/envoyproxy/envoymobile/StreamEmitter.kt
junr03: Ok sounds good. As long as the three of you are aligned I'll get on board. I am a little concerned about the departure from the underlying API, but if this is more ergonomic at this layer, that makes sense.

buildbreaker:
thanks @rebello95 @goaway !

library/kotlin/src/io/envoyproxy/envoymobile/StreamEmitter.kt
buildbreaker: Yeah, it differ slightly but it doesn't necessarily hide any underlying functionality

PR #269, opened 2019-07-18
swift: add tests for request builders

rebello95:
Adds tests for the request builders similar to the ones available on master for Kotlin.

Also fixes the `Equatable` comparison of `Request` by overriding `isEqual`. This is required since this type subclasses `NSObject` to be visible to Objective-C, and we want property comparison instead of pointer comparison here.

Resolves https://github.com/lyft/envoy-mobile/issues/252

### DIFF
--- a/library/swift/test/RequestBuilderTests.swift
+++ b/library/swift/test/RequestBuilderTests.swift
@@ -1,0 +1,189 @@
+import Envoy
+import Foundation
+import XCTest
+
+// swiftlint:disable:next force_unwrapping
+private let kURL = URL(string: "http://0.0.0.0:9001/api.lyft.com/demo.txt")!
+private let kBodyData = Data([1, 2, 3, 4])
+
+private let kRetryPolicy1 = RetryPolicy(maxRetryCount: 123,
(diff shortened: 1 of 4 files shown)


library/swift/test/RequestBuilderTests.swift
```diff
@@ -0,0 +1,189 @@
...
+      .removeTrailer(name: "foo", value: "2")
+      .build()
+    XCTAssertNil(request.trailers["foo"])
+  }
+
+  // MARK: - Request conversion
+
+  func testRequestsAreEqualWhenPropertiesAreEqual() {
+    let request1 = kRequestBuilder1.build()
+    let request2 = kRequestBuilder2.build()
```
buildbreaker: Shouldn't you use `kRequestBuilder1` for both? Cause if we end up modifying `kRequestBuilder2` we'll break this test no?

rebello95: I originally duplicated this since re-using `kRequestBuilder1` would mean that the builder would have the same memory address. This would result in this test passing even if the properties were _not_ the same. Updated to add a helper function for creating new (identical) builders with different memory addresses 👍

PR #272, opened 2019-07-19
bazel: add convenience configs for library tests

rebello95:
Introduces some new config options to make running library tests locally a bit less verbose.

Open to thoughts/alternatives here! I'll add some docs on running tests locally after we decide on how we want this to look.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -25,3 +25,7 @@ build:android --fat_apk_cpu=x86_64
 build:device --ios_multi_cpus=armv7,arm64 --fat_apk_cpu=armeabi-v7a,arm64-v8a
 
 build:fat --ios_multi_cpus=i386,x86_64,armv7,arm64 --fat_apk_cpu=x86,x86_64,armeabi-v7a,arm64-v8a
+
+test:ios_lib --test_output=all --build_tests_only //library/swift/test/...
+
+test:android_lib --test_output=all --build_tests_only //library/kotlin/test/...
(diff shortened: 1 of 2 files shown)


.bazelrc
```diff
@@ -25,3 +25,7 @@ build:android --fat_apk_cpu=x86_64
 build:device --ios_multi_cpus=armv7,arm64 --fat_apk_cpu=armeabi-v7a,arm64-v8a
 
 build:fat --ios_multi_cpus=i386,x86_64,armv7,arm64 --fat_apk_cpu=x86,x86_64,armeabi-v7a,arm64-v8a
+
+test:ios_lib --test_output=all --build_tests_only //library/swift/test/...
```
keith: I didn't know passing targets here even worked, but I think I'd probably prefer that you have an alias that was just `--test_output=all --build_tests_only` and still required users to pass the targets, since depending on what you're working on locally you many not want all the tests, and then you'd have to reproduce those flags manually

keith: also `--test_output=all` might be nice on CI but locally you might want something more limited like `--test_output=errors`

rebello95: Those are good callouts. If that's the case, do you think it's even worth adding this config at all? Another option if we wanted to keep it would be to split it into 2 (local configs and CI configs) or to focus on local and specify the CI configs directly in the azure yml file as they are today

keith: I'm not sure if it's worth it or not for local, this does make it easier for people who aren't familiar with bazel so I guess we should add it still.

rebello95: I think it might make more sense to just add some nice documentation on how to run tests locally - I'll do that instead, and we can revisit this later if we think it's useful

PR #275, opened 2019-07-22
swift: throw when setting restricted headers

rebello95:
Since we expose headers in the `Request` type, it's technically possible for consumers to set restricted headers that we will be setting internally within the library (any header starting with `:` in this case - `:method` for example).

To prevent this, we will throw an error when any of these headers/trailers is set manually.

Context: https://envoyproxy.slack.com/archives/CKQ2LK23G/p1563826656004700

### DIFF
--- a/library/swift/src/RequestBuilder.swift
+++ b/library/swift/src/RequestBuilder.swift
@@ -1,5 +1,12 @@
 import Foundation
 
+/// Error type that may be thrown by the request builder.
+@objc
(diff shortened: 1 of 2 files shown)


rebello95:
Will likely close this PR instead of merging in order to prevent the poor ergonomics of having to `try` when using the builders. Instead, we can strip any headers that start with `:` before executing request(s). Discussing in the thread

rebello95:
Superseded by https://github.com/lyft/envoy-mobile/pull/288

PR #279, opened 2019-07-22
swift: add conversion from RetryPolicy to headers

rebello95:
Adds converters to go from `RetryPolicy` to a set of headers that upstream Envoy understands.

This will be utilized by the request logic in the `Envoy` class.

### DIFF
--- a/library/swift/src/RetryPolicy.swift
+++ b/library/swift/src/RetryPolicy.swift
@@ -4,22 +4,22 @@ import Foundation
 /// See the `x-envoy-retry-on` Envoy header for documentation.
 @objc
 public enum RetryRule: Int, CaseIterable {
-  case fiveXX
+  case status5xx
   case gatewayError
   case connectFailure
-  case retriableFourXX
+  case retriable4xx
   case refusedUpstream
 
   /// String representation of this rule.
   var stringValue: String {
     switch self {
-    case .fiveXX:
+    case .status5xx:
       return "5xx"
     case .gatewayError:
       return "gateway-error"
     case .connectFailure:
       return "connect-failure"
-    case .retriableFourXX:
+    case .retriable4xx:
       return "retriable-4xx"
     case .refusedUpstream:
       return "refused-upstream"
--- a/library/swift/src/RetryPolicy.swift
+++ b/library/swift/src/RetryPolicy.swift
@@ -3,7 +3,7 @@ import Foundation
 /// Rules that may be used with `RetryPolicy`.
 /// See the `x-envoy-retry-on` Envoy header for documentation.
 @objc
-public enum RetryRule: Int {
+public enum RetryRule: Int, CaseIterable {
   case fiveXX
   case gatewayError
   case connectFailure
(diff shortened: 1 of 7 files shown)


library/swift/src/RetryPolicy.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 /// Rules that may be used with `RetryPolicy`.
 /// See the `x-envoy-retry-on` Envoy header for documentation.
 @objc
-public enum RetryRule: Int {
+public enum RetryRule: Int, CaseIterable {
```
goaway: nit: I know it's not strictly part of this PR, but what about `case status5xx` and `case retriable4xx`?

I just wonder if the current naming might be a little less obvious to some.

library/swift/test/RetryPolicyTests.swift
```diff
@@ -1,7 +1,40 @@
...
 final class RetryPolicyTests: XCTestCase {
+  // MARK: - Conversions
+
+  func testConvertingToHeadersWithPerRetryTimeoutIncludesAllHeaders() {
+    let policy = RetryPolicy(maxRetryCount: 123,
+                             retryOn: RetryRule.allCases,
+                             perRetryTimeoutMS: 9001)
+    let expectedHeaders = [
+      "x-envoy-max-retries": "123",
+      "x-envoy-retry-on": RetryRule.allCases
```
goaway: maybe expand `allCases` here to literals

library/swift/src/RetryPolicy.swift
rebello95: They're named this way at the moment to mirror upstream envoy (see the `stringValue` mappings below). We can rename them if you think this would make them clearer - do you have context into why they were named this way upstream?

library/swift/src/RetryPolicy.swift
```diff
@@ -44,6 +44,25 @@ public final class RetryPolicy: NSObject {
     self.retryOn = retryOn
     self.perRetryTimeoutMS = perRetryTimeoutMS
   }
+
+  /// Converts the retry policy to a set of headers recognized by Envoy.
+  ///
+  /// - returns: The header representation of the retry policy.
+  func toHeaders() -> [String: String] {
```
buildbreaker: I usually prefer to keep the conversion methods in a separate entity so the source object is decoupled from the target object

rebello95: Where would you prefer we put these in this case?

library/swift/test/RetryPolicyTests.swift
rebello95: Done

library/swift/src/RetryPolicy.swift
goaway: Right, I was actually thinking this maybe mapped closer to upstream (e.g. `retriable4xx` instead of `retriableFourXX`)?

I recognize `status5xx` is a little bit of a departure, but I think `5xx` is maybe more common/recognizable as a class of status codes than `fiveXX` might be.

No very strong feelings about it though.

library/swift/src/RetryPolicy.swift
buildbreaker: Talked f2f. Let's move it to another file

library/swift/src/RetryPolicy.swift
rebello95: SGTM, updating. cc @buildbreaker for Kotlin

rebello95: Done

PR #286, opened 2019-07-22
upstream: update envoy ref

junr03:

Description: pulling in tip of envoy master.
Risk Level: low
Testing: CI


junr03:
Note that the master test in CI for size binary will not pass because it will reference an old envoy upstream sha that is not compatible with the WORKSPACE change in this PR. We will have to merge this PR in regardless.

PR #288, opened 2019-07-23
swift: add conversion from Request to headers

rebello95:
Adds converters to go from `Request` to a set of headers that upstream Envoy understands.

This will be utilized by the request logic in the `Envoy` class.

Previous consideration: https://github.com/lyft/envoy-mobile/pull/275

### DIFF
--- a/library/swift/src/RequestMapper.swift
+++ b/library/swift/src/RequestMapper.swift
@@ -1,0 +1,25 @@
+private let kRestrictedHeaderPrefix = ":"
+
+extension Request {
+  /// Returns a set of outbound headers that include HTTP
+  /// information on the URL, method, and additional headers.
+  ///
+  /// - returns: Outbound headers to send with an HTTP request.
+  func makeOutboundHeaders() -> [String: String] {
+    var headers = self.headers
+      .filter { !$0.key.hasPrefix(kRestrictedHeaderPrefix) }
+      .mapValues { $0.joined(separator: ",") }
+      .reduce(into: [
+        ":method": self.method.stringValue,
+        ":scheme": self.scheme,
+        ":authority": self.authority,
+        ":path": self.path,
+      ]) { $0[$1.key] = $1.value }
+
+    if let retryPolicy = self.retryPolicy {
+      headers = headers.merging(retryPolicy.makeOutboundHeaders()) { _, retryHeader in retryHeader }
+    }
+
+    return headers
+  }
+}
--- a/library/swift/src/RequestMapper.swift
+++ b/library/swift/src/RequestMapper.swift
@@ -5,7 +5,7 @@ extension Request {
   /// information on the URL, method, and additional headers.
   ///
   /// - returns: Outbound headers to send with an HTTP request.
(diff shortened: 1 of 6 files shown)


library/swift/src/RequestMapper.swift
```diff
@@ -0,0 +1,25 @@
+private let kRestrictedHeaderPrefix = ":"
+
+extension Request {
+  /// Returns a set of outbound headers that include HTTP
+  /// information on the URL, method, and additional headers.
+  ///
+  /// - returns: Outbound headers to send with an HTTP request.
+  func makeOutboundHeaders() -> [String: String] {
```
buildbreaker: `outboundHeaders()` to keep the method simple?

library/swift/src/RequestMapper.swift
```diff
@@ -0,0 +1,25 @@
...
+    var headers = self.headers
+      .filter { !$0.key.hasPrefix(kRestrictedHeaderPrefix) }
```
goaway: I'm guessing this will silently drop headers with the restricted prefix - did we decide to do this instead of throw?

Another option would be to simply abort the request and pass control to the `onError` handler.

goaway: Ah, I see there's another PR which throws.

goaway: Though I just read the comment on that one that we don't want to have to `try` the builder. What's our current thinking?

library/swift/src/RequestMapper.swift
rebello95: Updated

library/swift/src/RequestMapper.swift
rebello95: We [discussed further over Slack](https://envoyproxy.slack.com/archives/CKQ2LK23G/p1563831145013100?thread_ts=1563826656.004700&cid=CKQ2LK23G), and decided that it wasn't worth making the builder for adding headers throwable since this would greatly reduce the ergonomics of using it. `URLSession` allows for setting these headers silently, which is what we prefer to do here as well

rebello95: > Another option would be to simply abort the request and pass control to the onError handler.

We could do this, though personally I'd prefer to keep it simple and strip those headers for the time being

buildbreaker: Yeah, this was a super annoying realization because Swift/ObjC doesn't have a kinder runtime exception concept which this would be a good use case for since it'll be a way for us to complain to the user that they did something unexpected. While I do think that we should throw for this case, I feel it's worse to make the user handle errors for which they have already accounted for. 

The alternatives here is to compromise on the builder object:
1. Breaking the builder pattern by returning a pair (builder + error)
2. Have the build return a pair (request + error)
3. Straight crashing the application (the nuclear rocket version in Swift/ObjC of a runtime exception)

goaway: That all makes sense. I think silently dropping them is reasonable for now.

PR #291, opened 2019-07-23
kotlin: add conversion from RetryPolicy to headers

buildbreaker:
Add conversion from RetryPolicy to headers

Changes also made to make internal classes visible

Description: Add conversion from RetryPolicy to headers
Risk Level: low
Testing: unit
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


bazel/kotlin_test.bzl
```diff
@@ -18,11 +18,17 @@ load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_jvm_test")
 # )
 #
 def envoy_mobile_kt_test(name, srcs, deps = []):
+    # This is to work around the issue where we have specific implementation functionality which
+    # we want to avoid comsumers to use but we want to unit test
+    dep_srcs = []
+    for dep in deps:
+        dep_srcs.append(dep + "_srcs")
```
rebello95: Why is this necessary if the file in question is being included in the dependency?

buildbreaker: In Kotlin, there's a concept of `internal` which allows for module level visibility.

The problem here is that in Bazel, we're using the library target as a dependency (so the test itself doesn't get visibility for `internal` entities). The reason this works is because we're basically leveraging the workaround in a [previous pr](https://github.com/lyft/envoy-mobile/pull/264) and use those defined `filegroup`s as source files in the test. Since it's now sources within the test target, we are able to access `internal` entities.

The other concept that I would like to call out is that we want to keep this entity (the mapper) an `internal` visible entity because we don't necessarily want consumers to use it. We also don't want to make it strictly private since that would make testing this behavior more complex (we'll have to do some integration black box testing or something along those lines).

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
```diff
@@ -8,19 +8,19 @@ package io.envoyproxy.envoymobile
  * @param perRetryTimeoutMs Timeout (in milliseconds) to apply to each retry.
  */
 data class RetryPolicy(
-    val maxRetryCount: Int,
-    val retryOn: List<RetryRule>,
-    val perRetryTimeoutMs: Long?
+  val maxRetryCount: Int,
+  val retryOn: List<RetryRule>,
+  val perRetryTimeoutMs: Long? = null
```
buildbreaker: @rebello95 How do you feel about this default to null?

bazel/kotlin_test.bzl
```diff
@@ -6,6 +6,7 @@ load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_jvm_test")
 # 1. Avoiding the need to declare the test_class which requires a fully qualified class name (example below)
 # 2. Avoiding the need to redeclare common unit testing dependencies like JUnit
 # 3. Ability to run more than one test file per target
+# 4. Ability to test internal entities which are envoy mobile dependencies
```
rebello95: This doesn't quite make sense

bazel/kotlin_test.bzl
```diff
@@ -18,11 +19,19 @@ load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_jvm_test")
 # )
 #
 def envoy_mobile_kt_test(name, srcs, deps = []):
+    # This is to work around the issue where we have specific implementation functionality which
+    # we want to avoid comsumers to use but we want to unit test
+    dep_srcs = []
+    for dep in deps:
+        # We'll resolve only the targets in `//library/kotlin/src/io/envoyproxy/envoymobile`
```
rebello95: Would like @keith to check this file's changes

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
rebello95: Seems fine

bazel/kotlin_test.bzl
buildbreaker: `Ability to test internal envoy mobile entities`?

rebello95: yea that would work

PR #305, opened 2019-07-26
bazel: update to Xcode 10.3

rebello95:

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -7,7 +7,7 @@ build \
   --define=tcmalloc=disabled \
   --ios_minimum_os=10.0 \
   --ios_simulator_device="iPhone X" \
-  --ios_simulator_version=12.2 \
+  --ios_simulator_version=12.4 \
   --spawn_strategy=standalone \
   --verbose_failures \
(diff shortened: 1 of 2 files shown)


.bazelrc
```diff
@@ -11,7 +11,7 @@ build \
   --spawn_strategy=standalone \
   --verbose_failures \
   --workspace_status_command=envoy/bazel/get_workspace_status \
-  --xcode_version=10.2.1 \
+  --xcode_version=10.3 \
```
keith: technically if there was a 10.3.1 it would be based on the user's local environment which to use, but that's probably fine for this project

rebello95:
Looks like we need to wait on Azure to update their images to include Xcode 10.3: https://github.com/microsoft/azure-pipelines-image-generation/blob/master/images/macos/macos-10.14-Readme.md

May also want to use the 12.4 simulators instead of 12.2

.bazelrc
rebello95: Updated to lock to `10.3.0`

rebello95:
CI passed except for 1 test which is red on master - looks like a flake with the simulator.

PR #306, opened 2019-07-26
swift: add Foundation import

rebello95:
Xcode 10.3 complains that this file uses `@objc` without importing `Foundation`.


rebello95:
Merging, Kotlin flake

PR #307, opened 2019-07-26
ios: update EnvoyStatus to be more Swift-friendly

rebello95:
- Makes `EnvoyStatus` a closed enum, since we guarantee a limited set of values
- Renames the enum cases to be more consistent with Apple's [guidelines](https://developer.apple.com/library/archive/releasenotes/ObjectiveC/ModernizationObjC/AdoptingModernObjective-C/AdoptingModernObjective-C.html#//apple_ref/doc/uid/TP40014150-CH1-SW6)

This has the effect of improving their usability in Swift.

```swift
let result = ... // EnvoyStatus

if result == .Success // Before change
if result == .success // After change
```

### DIFF
--- a/library/objective-c/EnvoyEngine.mm
+++ b/library/objective-c/EnvoyEngine.mm
@@ -19,45 +19,45 @@
     [NSNotificationCenter.defaultCenter postNotificationName:@"EnvoyException"
                                                       object:self
                                                     userInfo:userInfo];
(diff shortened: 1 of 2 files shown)


keith:
How does this affect the swift caller?

rebello95:
@keith pushed some more changes and updated the description 😄

PR #309, opened 2019-07-29
objc: add nullability regions

rebello95:
Contents of these interfaces should be assumed to be nonnull.


rebello95:
Merging, library built and other tests flaked.

PR #312, opened 2019-07-30
ios: finish bridging between ObjC/Swift

rebello95:
This PR ties together the upstream Objective-C/++ implementation with the Swift interfaces/bridging implementation. After this change, end-to-end calls from the native Swift library into the Envoy core should be supported.

- Implements the streaming and unary request methods
- Adds a simple `EnvoyStreamEmitter` which sends requests from Swift down into Objective-C
- Updates the header conversion functions and their tests to convert to the proper types (`[String: [String]]` versus `[String: String]`)
- Implements `ResponseHandler` as a class which provides a pass-through for setting Objective-C blocks directly from Swift for handling responses
- Adds a parser for status codes, as well as tests

https://github.com/lyft/envoy-mobile/pull/349 will use this interface with the Swift demo apps.


rebello95:
Waiting on https://github.com/lyft/envoy-mobile/pull/316

library/swift/src/ResponseHandler.swift
```diff
@@ -1,42 +1,98 @@
...
+  public func onError(_ closure:
+    @escaping () -> Void)
+    -> ResponseHandler
+  {
+    self.underlyingObserver.onError = closure
+    return self
+  }
 
-  /// Called when an internal Envoy exception occurs with the stream.
+  /// Specify a callback for when an internal Envoy exception occurs with the stream.
```
buildbreaker:
```suggestion
  /// Specify a callback for when there is a cancellation on the stream.
```

rebello95: updated

library/swift/src/Envoy.swift
```diff
@@ -42,3 +45,22 @@ public final class Envoy: NSObject {
...
 }
+
+extension Envoy: Client {
+  public func startStream(with request: Request, handler: ResponseHandler) -> StreamEmitter {
+    let httpStream = self.engine.startStream(with: handler.underlyingObserver)
+    httpStream.sendHeaders(request.outboundHeaders(), close: false)
+    return EnvoyStreamEmitter(stream: httpStream)
+  }
+
+  public func sendUnary(_ request: Request, body: Data?,
```
goaway: Instead of `startStream` and `sendUnary`, I'd suggest:
`send(request:handler)` and `send(request:body:handler)`

goaway: or `sendRequest(_:handler:)` / `sendRequest(_:body:handler:)`

rebello95: I'd rather lean towards being clear on the intent (stream/unary) for the end consumer's sake, rather than making them assume based on function parameter signatures

goaway: And of course trailers can be in there too, but since trailers are so rare in normal cases, I'd probably still suggest a signature which excludes or defaults them.

library/swift/src/ResponseHandler.swift
```diff
@@ -1,42 +1,98 @@
...
+  public func onHeaders(_ closure:
+    @escaping (_ headers: [String: [String]], _ statusCode: Int, _ endStream: Bool) -> Void)
+    -> ResponseHandler
+  {
+    self.underlyingObserver.onHeaders = { headers, endStream in
+      closure(headers, ResponseHandler.statusCode(fromHeaders: headers), endStream)
+    }
 
-  /// Called when response metadata is received by the stream.
+    return self
```
goaway: nice

library/swift/src/Envoy.swift
goaway: I understand the argument about clarity/specificity, but I doubt most end-users are likely to be thinking in these terms, and I have misgivings because I don't really think this actually more accurate.

The `startStream` interface is a perfectly appropriate interface for sending a "unary" request, and the current naming sort of makes it sound like you should only use it when you're trying to stream something (a naive user might even think that this is to stream a response, which is completely unrelated).

In fact, all HTTP requests are "unary" in the sense that they have a single request and response. I feel like these are semi-GRPC concepts that are leaking in because GRPC layers a stream encoding into a "unary" HTTP request.

goaway: I guess in general, since this is a top level interface, I'm pushing back a little because I feel the naming should be recognizable and understandable to users who don't have a deep familiarity with the underlying concerns.

rebello95: Talked offline - this makes sense. Going to update in a separate PR after merging this

rebello95: https://github.com/lyft/envoy-mobile/pull/354

PR #313, opened 2019-07-30
jni: add java interfaces and data structures

buildbreaker:
There's quite a few classes added but this roughly lays out the foundation which we will use to create a request which the `Client` interface would call into.

The general idea using interfaces is this:
`Client.kt` -> `Engine.java` -> `JniLibrary.java` -> `main_interface.h`

* The `EnvoyEngine` will be responsible for starting up Envoy and making calls to create a stream and translate the JNI primitives to the defined java classes (added TODOs for these)
* `AndroidEngine` will use `EnvoyEngine` and have the `c-ares` connection manager functions

There'll be two flavors here:
1. Pure Java/Kotlin engine/client which will just be a Envoy library without any Android knowledge
2. Android engine/client which will use the Java/Kotlin entities and will provide any functionalities or methods which is Android specific

Description: jni: add java interfaces and data structures
Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


bazel/kotlin_lib.bzl
```diff
@@ -38,6 +38,23 @@ def envoy_mobile_android_library(name, custom_package, manifest, visibility = No
         deps = deps,
     )
 
+def envoy_mobile_java_library(name, visibility = None, srcs = [], deps = []):
```
buildbreaker: The current aar rule expects a `*_srcs` associated with every dependency

buildbreaker:
@goaway @junr03 

There will need to be changes to conform with the updates to `main_interface.h` but the intent here is to lay out the structure of responsibilities of each layer. PTAL when you have a chance!

PR #314, opened 2019-07-30
build: Switch to better kotlin rules

keith:


keith:
It started from the same base I think, so makes sense. Lmk if something goes wrong

PR #316, opened 2019-08-01
Add native-platform callback implementation

goaway:


library/common/include/c_types.h
```diff
@@ -61,14 +81,26 @@ typedef struct {
  */
 typedef struct {
   // Number of header elements in the array.
-  uint64_t length;
+  int length;
```
rebello95: Why is this no longer unsigned?

library/common/include/c_types.h
```diff
@@ -130,10 +162,11 @@ typedef void (*on_error)(envoy_error error, void* context);
...
-  on_data on_data_f;
-  on_metadata on_metadata_f;
-  on_trailers on_trailers_f;
-  on_error on_error_f;
+  envoy_on_headers_f on_headers;
+  envoy_on_data_f on_data;
+  envoy_on_metadata_f on_metadata;
+  envoy_on_trailers_f on_trailers;
+  envoy_on_error_f on_error;
+  atomic_bool *canceled;
```
rebello95: I think the linter might complain about this

library/swift/src/Client.swift
```diff
@@ -9,6 +9,6 @@ public protocol Client {
   /// - parameter handler: Handler for receiving stream events.
   ///
   /// - returns: Emitter for sending streaming data outward.
-  func startStream(request: Request, handler: ResponseHandler)
+  func startStream(request: Request, handler: ResponseHandler, handlerQueue: DispatchQueue)
```
rebello95: Can we add a `- parameter` for this?

library/objective-c/EnvoyTypes.h
```diff
@@ -5,11 +5,11 @@ NS_ASSUME_NONNULL_BEGIN
 // MARK: - Aliases
 
 /// Handle to an outstanding Envoy HTTP stream. Valid only for the duration of the stream and not
-/// intended for any external interpretation or use.
+/// intended for any exter/nal interpretation or use.
 typedef UInt64 EnvoyStreamID;
 
 /// A set of headers that may be passed to/from an Envoy stream.
-typedef NSArray<NSDictionary<NSString *, NSString *> *> EnvoyHeaders;
+typedef NSDictionary<NSString *, NSArray<NSString *> *> EnvoyHeaders;
```
rebello95: Thanks for fixing this

library/objective-c/EnvoyTypes.h
```diff
@@ -5,11 +5,11 @@ NS_ASSUME_NONNULL_BEGIN
 // MARK: - Aliases
 
 /// Handle to an outstanding Envoy HTTP stream. Valid only for the duration of the stream and not
-/// intended for any external interpretation or use.
+/// intended for any exter/nal interpretation or use.
```
rebello95:
```suggestion
/// intended for any external interpretation or use.
```

library/objective-c/EnvoyTypes.h
```diff
@@ -54,8 +54,14 @@ typedef struct {
 // MARK: - EnvoyObserver
 
 /// Interface that can handle HTTP callbacks.
+// FIXME: can we just bridge the swift object here and/or just expose this to swift as the ResponseHandler?
```
rebello95: You can assign this to me `TODO(@rebello95)`

library/objective-c/EnvoyEngine.mm
```diff
@@ -1,16 +1,157 @@
...
 #import "library/common/main_interface.h"
 
+#pragma mark - utility functions to move elsewhere
+static envoy_headers toNativeHeaders(EnvoyHeaders *headers) {
+  int length = 0;
+  for (id headerList in headers) {
+    length += headerList.count;
+  }
+  envoy_header *header_array = (envoy_header *)malloc(sizeof(envoy_header) * length);
+  int j = 0;
```
rebello95: `i`?

library/objective-c/EnvoyEngine.mm
```diff
@@ -1,16 +1,157 @@
...
+static NSData * to_ios_data(envoy_data data) {
+  // TODO: investigate buffer ownership
+  // Possibly extend/subclass NSData to call envoy_data.release on dealloc and have release drain the underlying Envoy buffer instance
+  // FIXME: do we need to autorlease here?
+  return [[NSData dataWithBytesNoCopy:data.bytes length:data.length] autorelease];
+}
+
+static EnvoyHeaders * to_ios_headers(envoy_headers headers) {
+  NSMutableDictianary *headerDict = [NSMutableDictionary new];
+  for (uint_fast32_t i = 0; i < headers.length; i++) {
```
rebello95: Why do we use `int`, `uint64_t` and `uint_fast32_t` in different loops?

library/objective-c/EnvoyEngine.mm
```diff
@@ -1,16 +1,157 @@
...
+                                                     length:header.value.length
+                                                   encoding:NSUTF8StringEncoding];
+    NSMutableArray *headerValueList = headerDict[headerName];
+    if (headerValueList == nil) {
+      headerValueList = [NSMutableArray new];
+      headerDict[headerKey] = headerValueList;
+    }
+    [headerValueList addObject:headerValue];
+  }
+  return [headerDict autorelease];
```
rebello95: Is this necessary? I'm assuming it is because we aren't using ARC here?

library/objective-c/EnvoyEngine.mm
```diff
@@ -23,45 +164,92 @@ + (EnvoyStatus)runWithConfig:(NSString *)config logLevel:(NSString *)logLevel {
...
-  NSLog(@"%@ not implemented, returning failure", NSStringFromSelector((SEL) __func__));
-  EnvoyStream stream;
-  stream.status = EnvoyStatusFailure;
-  return stream;
++ (void)setupEnvoy {
+  setup_envoy();
+}
+
+// TODO: move to Envoy
++ (EnvoyStream *)startHttpStreamForRequest:(EnvoyRequest *)request handler:(EnvoyObserver *)handler {
```
rebello95: Let's go ahead and update this in the header file

library/objective-c/EnvoyEngine.mm
```diff
@@ -23,45 +164,92 @@ + (EnvoyStatus)runWithConfig:(NSString *)config logLevel:(NSString *)logLevel {
...
-  NSLog(@"%@ not implemented, returning failure", NSStringFromSelector((SEL) __func__));
-  return EnvoyStatusFailure;
+- (void)dealloc {
+  free(self.nativeObserver);
 }
 
-+ (EnvoyStatus)sendMetadata:(EnvoyHeaders *)metadata to:(EnvoyStream *)stream close:(BOOL)close {
-  NSLog(@"%@ not implemented, returning failure", NSStringFromSelector((SEL) __func__));
-  return EnvoyStatusFailure;
+- (void)sendHeaders:(EnvoyHeaders *)headers close:(BOOL)close {
```
rebello95: Header file should be updated to reflect these changes

library/objective-c/EnvoyEngine.mm
```diff
@@ -23,45 +164,92 @@ + (EnvoyStatus)runWithConfig:(NSString *)config logLevel:(NSString *)logLevel {
...
+                                                    request:request,
+                                                   observer:handler];
 }
 
-+ (EnvoyStatus)sendHeaders:(EnvoyHeaders *)headers to:(EnvoyStream *)stream close:(BOOL)close {
-  NSLog(@"%@ not implemented, returning failure", NSStringFromSelector((SEL) __func__));
-  return EnvoyStatusFailure;
+@end
+
+@implementation EnvoyHttpStream
```
rebello95: I think you're missing the `EnvoyHttpStream` from this PR

library/objective-c/EnvoyEngine.h
```diff
@@ -82,7 +62,7 @@ NS_ASSUME_NONNULL_BEGIN
  @param config The configuration file with which to start Envoy.
  @return A status indicating if the action was successful.
  */
-+ (EnvoyStatus)runWithConfig:(NSString *)config;
++ (envoy_engine_t)runWithConfig:(NSString *)config;
```
rebello95: Let's move these functions to a new `@protocol Engine`, then have `@interface EnvoyEngine` implement them, as discussed offline so that we can mock these from the platform. This aligns with what Android has on master

library/objective-c/EnvoyEngine.mm
goaway: It doesn't even work under ARC. I couldn't recall if ARC would correctly handle Obj-C objects created in a pure C function.

library/objective-c/EnvoyEngine.mm
goaway: Working on getting that updated.

library/objective-c/EnvoyEngine.mm
goaway: Yep :)

library/objective-c/EnvoyEngine.mm
goaway: We are - we're still moving things around a little bit.

library/objective-c/EnvoyEngine.h
goaway: Are there potential namespacing issues there? We could do `EnvoyEngineProtocol`..

library/objective-c/EnvoyEngine.mm
goaway: moving fast, changing stuff.

library/common/include/c_types.h
goaway: about which?

library/common/include/c_types.h
goaway: We went back and forth.

Some convention still suggests `size_t` (though that's technically specifically for byte sizes, which this isn't, and also C++ committee apparently is mostly is on record saying they regret making it unsigned).

Google style guide recommends regular `int` in any case where an explicit bit width/sign behavior isn't actually required, and calls out that making a type unsigned offers no actual protection against people trying to set negative values.

Java uses signed lengths by convention.

Obj-c/swift use unsigned lengths by convention.

This was becoming a rabbit hole and `int` seemed like good neutral ground for now.

library/swift/src/Client.swift
goaway: `- parameter`?

goaway: ohhhh the docstring :)

sure thing

library/objective-c/EnvoyEngine.h
rebello95: Yea, as we discussed offline, let's go with `EnvoyEngine` as the protocol and `EnvoyEngineImpl` as the class. We should also update this on the Java side as part of this PR if possible.
Just to document here, the "impl" suffix naming is a convention used in Envoy, which is why we're opting to use it in Objc/Java (but not Kotlin/Swift)

library/common/include/c_types.h
rebello95: Cool, was mostly just curious as to the thoughts behind the decision. Thanks for explaining!

library/objective-c/BUILD
```diff
@@ -8,12 +8,13 @@ exports_files([
 objc_library(
     name = "envoy_engine_objc_lib",
     srcs = [
-        "EnvoyEngine.mm",
+        "EnvoyEngine.m",
+        "EnvoyHttpStream.m",
     ],
     hdrs = [
         "EnvoyEngine.h",
+        "EnvoyHttpStream.h",
```
rebello95: I think this should be moved to `exports_files` (maybe this needs to be rebased?)

library/objective-c/EnvoyHttpStream.m
```diff
@@ -0,0 +1,250 @@
...
+
+static void ios_on_complete(void *context) {
+  ios_context *c = (ios_context *)context;
+  EnvoyObserver *observer = c->observer;
+  dispatch_async(observer.dispatchQueue, ^{
+    // TODO: release stream
+    if (atomic_load(c->canceled)) {
+      return;
+    }
+  });
```
rebello95: Shouldn't this be calling back up to the platform?

library/objective-c/EnvoyTypes.h
```diff
@@ -91,6 +98,9 @@ typedef struct {
  */
 @property (nonatomic, strong) void (^onError)(EnvoyEngineError *error);
 
+// FIXME
+@property (nonatomic, strong) void (^onCancel)();
+
```
rebello95: Should we add one for `onComplete` too? Or is the assumption that `onTrailers` encompasses this?

library/swift/src/Client.swift
rebello95: Looks like still missing 🙃

goaway: moved this to the handler

library/objective-c/EnvoyHttpStream.m
goaway: Yep, some of this is more of a sketch; working headers is the main goal for this PR (with the framework in place to get the rest working).

library/objective-c/BUILD
```diff
@@ -8,12 +8,14 @@ exports_files([
 objc_library(
     name = "envoy_engine_objc_lib",
     srcs = [
-        "EnvoyEngine.m",
+        "EnvoyEngineImpl.h",
```
rebello95: Shouldn't this be in `hdrs`?

library/objective-c/EnvoyTypes.h
goaway: Current plan is not to surface onComplete, since it's actually meant more to trigger internal bookkeeping. Other callbacks have the boolean on them and trailers/error imply closure already.

That said we _can_ surface this if it ends up being necessary/useful. Currently though, it will be a little bit redundant.

library/objective-c/BUILD
goaway: Only if it's meant to be used externally - i.e., I don't think it'd be something we'd want to expose as part of the framework's public interface. Not entirely sure of the interplay with the Swift layer though. If it needs to be surfaced to `hdrs` in order for Swift to access it, then we can move it.

library/objective-c/EnvoyEngineImpl.h
```diff
@@ -0,0 +1,33 @@
...
+/// Wrapper layer for calling into Envoy's C/++ API.
+@interface EnvoyEngineImpl : NSObject <EnvoyEngine>
+
+/**
+ Run the Envoy engine with the provided config and log level.
+
+ @param config The configuration file with which to start Envoy.
+ @return A status indicating if the action was successful.
+ */
++ (EnvoyStatus)runWithConfig:(NSString *)config;
```
rebello95: Do these need to be in the header file if this is already defined to conform to `EnvoyEngine` above?

library/objective-c/EnvoyEngineImpl.h
```diff
@@ -0,0 +1,33 @@
...
+
+ @param config The configuration file with which to start Envoy.
+ @param logLevel The log level to use when starting Envoy.
+ @return A status indicating if the action was successful.
+ */
++ (EnvoyStatus)runWithConfig:(NSString *)config logLevel:(NSString *)logLevel;
+
+/// Performs necessary setup after Envoy has initialized and started running.
+/// TODO: create a post-initialization callback from Envoy to handle this automatically.
++ (void)setupEnvoy;
```
rebello95: If this is defined here, should probably also be in the `EnvoyEngine` interface

goaway: I kept this here because the plan is to get rid of it as soon as we have the post-init callback wired up.

library/objective-c/EnvoyEngineImpl.h
goaway: nope, guess not :)

library/swift/src/Envoy.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 @objcMembers
-public final class Envoy: NSObject {
+public final class Envoy<Engine: EnvoyEngine>: NSObject {
```
rebello95: I don't think this needs to be generic. @buildbreaker and I were considering just instantiating this class with an `EnvoyEngine.Type` (for class methods, or an instance of `EnvoyEngine`)

goaway: It doesn't need to be I guess. Are there downsides?

I went this direction because it felt a little cleaner to me than passing the type as an initialization parameter that needs to be stored and referenced, esp. when the interface is static.

goaway: Maybe it makes more sense in the context that our current thinking is that EnvoyEngine will always be static, and Envoy will be the platform instance.

rebello95: Generics are harder to pass around in platform code, because everything that uses them also needs to be generic. For example, we'd have to do:

```swift
func foo<T: EnvoyEngine>(_ envoy: Envoy<T>)
```

Instead of:

```swift
func foo(_ envoy: Envoy)
```

Reducing the generic and instead keeping the type hidden within the object will remove the need for this

goaway: But nothing else should actually need the generic type right? Like `foo` doesn't need to know the specialized type of `Envoy`. Pretty sure `func foo(_ envoy: Envoy)` should still work(?)

library/objective-c/BUILD
```diff
@@ -10,18 +10,18 @@ exports_files([
...
+        "EnvoyHttpStream.m",
         "EnvoyTypes.h",
-     ],
+    ],
     hdrs = [
-#        "EnvoyEngine.h",
-#        "EnvoyEngineImpl.h",
-#        "EnvoyHttpStream.h",
-#        "EnvoyTypes.h",
+        #        "EnvoyEngine.h",
```
rebello95: Should we remove these?

goaway: I can add a TODO instead

library/objective-c/EnvoyEngineImpl.h
rebello95: Sounds good

bazel/swift_static_framework.bzl
```diff
@@ -162,7 +162,8 @@ def swift_static_framework(
     module_name = module_name or name + "_framework"
     if objc_includes:
         locations = ["$(location {})".format(x) for x in objc_includes]
-        copts = copts + ["-import-objc-header"] + locations
+        for location in locations:
+            copts = copts + ["-import-objc-header", location]
```
keith: This flag only works once, and swift picks the last one https://github.com/apple/swift/blob/e7c294a076b41a9c87be99bf28d91f864977012a/lib/Frontend/CompilerInvocation.cpp#L582-L583

goaway: Thanks for spotting this @keith! That explains some of the weird ordering errors I was seeing too. I will update this rule subsequently to generate a bridge header from all objc-includes.

rebello95:
looks like there's a lint failure on CI but green otherwise!

goaway:
pushed lint fix.

swift library tests all pass locally for me, so I think this is effectively green.

rebello95:
I'm fine with going ahead and merging

PR #317, opened 2019-08-02
docs: update iOS performance findings

rebello95:
Updating based on https://github.com/lyft/envoy-mobile/issues/128#issuecomment-516260951.

### DIFF
--- a/docs/root/development/performance/device_conditions.rst
+++ b/docs/root/development/performance/device_conditions.rst
@@ -65,7 +65,7 @@ Setting the ``URLSessionConfiguration``'s ``httpMaximumConnectionsPerHost`` to `
 of them failing after the ``timeoutIntervalForRequest`` specified on the ``URLSessionConfiguration``.
 This is the same behavior seen with libraries like gRPC which use BSD sockets.
 
-Additionally, putting the phone in airplane mode results in all requests failing immediately
+Putting the phone in airplane mode resulted in all requests failing immediately
 (instead of waiting for the specified timeout) because iOS was aware that it had no connectivity.
 
 Android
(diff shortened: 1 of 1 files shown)


docs/root/development/performance/device_conditions.rst
```diff
@@ -30,7 +33,45 @@ Experimentation method
...
+a device with debugging symbols:
+
+``bazel build ios_dist --config=ios --ios_multi_cpus=arm64 --copt=-ggdb3``
+
+In the active scheme of the app's ``Environment Variables``, set ``CFNETWORK_DIAGNOSTICS=3``
+to enable more verbose ``CFNetwork`` logs. Additionally, set Envoy's logs to ``trace``.
+
+Within the app, we performed the following:
+
+- Made single requests to a simple Python server from which we could see when the client connected, disconnected, and made requests
```
buildbreaker: Are you able to link the python server code?

docs/root/development/performance/device_conditions.rst
```diff
@@ -30,7 +33,45 @@ Experimentation method
 Modified versions of the "hello world" example apps were used to run the following experiments.
 See the iOS/Android sections below for instructions on building the examples.
 
-Throughout all of the following steps, we tested to make sure that network requests succeeded
+iOS
```
buildbreaker: I feel like these are general observations with some actions. Could we instead write up very specific steps on how you were able to reproduce the experimentation.

Something like:
```
1. Start the python service
2. Switched between WiFi to cellular
3. Wait for logs to appear
...
``

This way a reader can precisely reproduce the experimentation steps and read more about the observations you noticed in the analysis

docs/root/development/performance/device_conditions.rst
rebello95: Sure

docs/root/development/performance/device_conditions.rst
rebello95: Updated

PR #320, opened 2019-08-09
configs: add support for using dynamic forward proxy

rebello95:
We'll be using the [dynamic forward proxy configuration](https://www.envoyproxy.io/docs/envoy/latest/configuration/http_filters/dynamic_forward_proxy_filter#config-http-filters-dynamic-forward-proxy) as the defacto configuration that will ship with Envoy Mobile. This will allow us to avoid making consumers register the DNS/domains they need to support when starting up the Envoy instance.

This PR:
- Pulls in the upstream fixes to support for this in https://github.com/envoyproxy/envoy/pull/7889
- Moves our example config to use the dynamic forward proxy
- Updates our main interfaces to force register the necessary extensions

Future PRs will:
- Move this into a resource compiled with the library
- Convert it to a template
- Provide typed configurations for wrapping the template
- Enable TLS: https://github.com/lyft/envoy-mobile/issues/322

Part of https://github.com/lyft/envoy-mobile/issues/169.

### DIFF
--- a/examples/common/config.yaml
+++ b/examples/common/config.yaml
@@ -13,7 +13,7 @@ static_resources:
       - name: envoy.http_connection_manager
         typed_config:
           "@type": type.googleapis.com/envoy.config.filter.network.http_connection_manager.v2.HttpConnectionManager
-          stat_prefix: ingress_http
+          stat_prefix: client
           http_protocol_options:
             allow_absolute_url: true
           route_config:
--- a/examples/common/config.yaml
+++ b/examples/common/config.yaml
@@ -45,7 +45,8 @@ static_resources:
         dns_cache_config:
           name: dynamic_forward_proxy_cache_config
           dns_lookup_family: AUTO
-    # TODO: Enable TLS as part of https://github.com/lyft/envoy-mobile/issues/322
+    # WARNING!
+    # TODO: Enable TLS in https://github.com/lyft/envoy-mobile/issues/322
     #
     # tls_context:
     #   common_tls_context:
--- a/examples/common/config.yaml
+++ b/examples/common/config.yaml
@@ -3,8 +3,7 @@
 # https://www.envoyproxy.io/docs/envoy/latest/configuration/configuration
 static_resources:
   listeners:
-  - name: listener_0
-    address:
+  - address:
       socket_address:
         protocol: TCP
         address: 0.0.0.0
@@ -46,6 +45,8 @@ static_resources:
         dns_cache_config:
           name: dynamic_forward_proxy_cache_config
           dns_lookup_family: V4_ONLY
+    # TODO: TLS support
+    #
     # tls_context:
     #   common_tls_context:
     #     validation_context:
--- a/examples/common/config.yaml
+++ b/examples/common/config.yaml
@@ -3,38 +3,53 @@
 # https://www.envoyproxy.io/docs/envoy/latest/configuration/configuration
 static_resources:
   listeners:
-    - address:
-        socket_address: {address: 0.0.0.0, port_value: 9001, protocol: TCP}
-      filter_chains:
-        - filters:
-            - name: envoy.http_connection_manager
-              config:
-                stat_prefix: client
-                route_config:
-                  virtual_hosts:
-                    - name: all
-                      domains: ["*"]
-                      routes:
-                        - match: {prefix: "/"}
-                          route: {cluster: hello_world_api}
-                http_filters:
-                  - name: envoy.router
+  - name: listener_0
+    address:
+      socket_address:
+        protocol: TCP
+        address: 0.0.0.0
+        port_value: 9001
+    filter_chains:
+    - filters:
+      - name: envoy.http_connection_manager
+        typed_config:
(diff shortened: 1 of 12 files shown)


examples/common/config.yaml
```diff
@@ -3,38 +3,53 @@
...
-                route_config:
-                  virtual_hosts:
-                    - name: all
-                      domains: ["*"]
-                      routes:
-                        - match: {prefix: "/"}
-                          route: {cluster: hello_world_api}
-                http_filters:
-                  - name: envoy.router
+  - name: listener_0
```
goaway: nit: no real need to name the listener here

examples/common/config.yaml
```diff
@@ -3,38 +3,53 @@
...
+      name: envoy.clusters.dynamic_forward_proxy
+      typed_config:
+        "@type": type.googleapis.com/envoy.config.cluster.dynamic_forward_proxy.v2alpha.ClusterConfig
+        dns_cache_config:
+          name: dynamic_forward_proxy_cache_config
+          dns_lookup_family: V4_ONLY
+    # tls_context:
+    #   common_tls_context:
+    #     validation_context:
+    #       trusted_ca: {filename: /etc/ssl/certs/ca-certificates.crt}
```
mattklein123: We definitely need/want this. Do we know what the correct paths to load the platform trusted CA files are? Or is it more difficult than that on mobile?

examples/common/config.yaml
rebello95: Removed

examples/common/config.yaml
rebello95: Totally agree. We're planning on merging with this temporarily disabled, as we'll likely need to compile our own with the library. Tracking here, and linked within the config: https://github.com/lyft/envoy-mobile/issues/322

examples/common/config.yaml
```diff
@@ -3,38 +3,54 @@
...
+      socket_address:
+        protocol: TCP
+        address: 0.0.0.0
+        port_value: 9001
+    filter_chains:
+    - filters:
+      - name: envoy.http_connection_manager
+        typed_config:
+          "@type": type.googleapis.com/envoy.config.filter.network.http_connection_manager.v2.HttpConnectionManager
+          stat_prefix: ingress_http
```
goaway: nit: change this to back to "client" imo ("ingress_http" sounds a lot like a proxy, which technically this still is, but is increasingly misleading).

examples/common/config.yaml
```diff
@@ -3,38 +3,54 @@
...
+          http_protocol_options:
+            allow_absolute_url: true
+          route_config:
+            name: local_route
+            virtual_hosts:
+            - name: local_service
```
goaway: nit: `- name: all` is a bit closer to the intent here, I think.

examples/common/config.yaml
rebello95: Good call, fixed

examples/common/config.yaml
```diff
@@ -3,38 +3,54 @@
...
+              domains: ["*"]
+              routes:
+              - match:
+                  prefix: "/"
+                route:
+                  cluster: dynamic_forward_proxy_cluster
```
goaway: how about `default_egress`

goaway: (nit)

examples/common/config.yaml
```diff
@@ -3,38 +3,54 @@
...
+          http_filters:
+          - name: envoy.filters.http.dynamic_forward_proxy
+            config:
+              dns_cache_config:
+                name: dynamic_forward_proxy_cache_config
+                dns_lookup_family: V4_ONLY
```
goaway: I think we might be fine with `AUTO`

examples/common/config.yaml
```diff
@@ -3,38 +3,54 @@
...
+    connect_timeout: 30s
+    dns_refresh_rate: 60s
+    lb_policy: CLUSTER_PROVIDED
+    cluster_type:
+      name: envoy.clusters.dynamic_forward_proxy
+      typed_config:
+        "@type": type.googleapis.com/envoy.config.cluster.dynamic_forward_proxy.v2alpha.ClusterConfig
+        dns_cache_config:
+          name: dynamic_forward_proxy_cache_config
+          dns_lookup_family: V4_ONLY
```
goaway: `AUTO`

examples/common/config.yaml
goaway: It's a little scary to merge this as is, in case anyone's actually playing around with this library. Do you want to add logic to manually reject non-https traffic for now?

goaway:
nice work

examples/common/config.yaml
rebello95: Talked offline - opted to leave a `WARNING!` here since this is in `examples/` and we are still in demo stage.

rebello95:
Thanks for the reviews!
@mattklein123 just to close the loop here - we plan to make this work properly with the direct API when/before switching over the demos to use it instead of the native networking calls they use today.

PR #321, opened 2019-08-09
build: update bazelw to use .bazelversion

goaway:
Description: This allows the project to be wrapped and built consistently using bazelisk OR the current wrapper (which has sha checks). https://github.com/bazelbuild/bazelisk is now the official wrapper for bazel.


rebello95:
Do we need to update CI/scripts/docs?

keith:
I'm kind of against this because the previous approach required no user knowledge of the bazel ecosystem. But I know envoy moved this way as well

goaway:
@keith what if I update the current `bazelw` script to reference `.bazelversion`? Then `bazelisk` will work, but the `bazelw` script will also still work.

goaway:
@rebello95 yeah I'm working on the CI change that would be necessary to support this.

keith:
Likely to do that we'd have to drop the sha validation we have here, but if everyone else is in favor of removing this I guess we can regardless

goaway:
What do you think? I agree it's nice to be able to build without needing special ecosystem domain knowledge and extra dependencies. I also actually think leaving in the sha checks is fine.

`bazelisk` still has some nice additional features for development, so this seemed like a good middle ground?

goaway:
above comment was meant to be tagged @keith :)

bazelw
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+set -euo pipefail
+
+read -r bazel_version<.bazelversion
```
keith: this doesn't really help if we still want sha validation, since that would be specific to each version

goaway: I mean it will still hardcode validation against a specific target, but that's still useful right?

keith: yea just too bad that you'd still have to change 2 files per update. I think if no one else is really attached to this we should delete it and encourage everyone to use bazelisk instead

goaway: It's too bad, but I mean, if we do sha validation with a file-based version it's also kind of a given.

Again, I think I do like having the bash script here, that supports the same configuration as bazelisk, giving the option to use either. Would you prefer to drop the sha validation? (I don't view this as a necessity; I'm fine with keeping it.)

PR #325, opened 2019-08-13
ios: update library interfaces

rebello95:
Makes some relatively minor updates to match up with the JNI interfaces being checked in with https://github.com/lyft/envoy-mobile/pull/326. This diff was larger before, but much of it got checked in with https://github.com/lyft/envoy-mobile/pull/329.


library/objective-c/EnvoyStream.h
```diff
@@ -0,0 +1,56 @@
...
+ @param close True if the stream should be closed after sending.
+ */
+- (void)sendData:(NSData *)data close:(BOOL)close;
+
+/**
+ Send metadata over the provided stream.
+
+ @param metadata Metadata to send over the stream.
+ */
+- (void)sendMetadata:(EnvoyHeaders *)metadata;
```
buildbreaker: We should probably fix these inconsistencies between the different methods in the `main_interface.h` because not all of the send methods have the option of closing/ending the streaming

goaway: Metadata can't close the stream; it's unsupported by the protocol. For now we'll be closing all streams by sending an empty data frame.

library/objective-c/EnvoyEngine.h
```diff
@@ -133,12 +139,14 @@ typedef NSDictionary<NSString *, NSArray<NSString *> *> EnvoyHeaders;
 /**
  Opens a new HTTP stream attached to this engine.
 
- @param observer Handler for stream events.
+ @param observer Handler for observing stream events.
  */
-- (EnvoyHttpStream *)openHttpStreamWithObserver:(EnvoyObserver *)observer;
+- (EnvoyStream *)startStreamWithObserver:(EnvoyObserver *)observer;
```
goaway: My preference is to keep "Http" in here, since there are other distinct stream types, and Envoy supports them (also there's, e.g. `CFStream` vs `CFHttpStream`).

rebello95: This is to match the Java interfaces: https://github.com/lyft/envoy-mobile/pull/326/files#diff-55a9e7503aabbe4019f8a57fdf14ef86R7

library/objective-c/EnvoyEngine.m
```diff
@@ -274,9 +282,9 @@ - (int)cancel {
     ios_on_cancel(context);
     // Step 3: propagate the reset into native code.
     reset_stream(_streamHandle);
-    return 0;
+    return ENVOY_SUCCESS;
```
goaway: While this is technically correct, and ultimately what we want, these enum values don't currently match the signature of the method.

rebello95: Ah you're right, for some reason I thought they were pure integers not an enum. Reverting this change

rebello95: I do think we should change these to an enum in the near future, I can open a PR for it later

library/objective-c/EnvoyEngine.h
goaway: Note that the C API lacks it because it one possible route is for it to use the same interface for general stream types (and the stream handle will distinguish handling). That's not the case at this layer; it's very tied to HTTP-specific concerns and functionality.

goaway: Also, lower level core functionality that *is* tied to HTTP is namespaced with "Http".

library/objective-c/EnvoyEngine.m
goaway: (Changing this over is currently held up by https://github.com/lyft/envoy-mobile/issues/333)

library/objective-c/EnvoyEngine.h
goaway: I do think the Java interfaces should include "Http" as well. I had planned to update them when we started working in that layer.

rebello95: Talked offline with @buildbreaker and including HTTP here seems fine since upstream Envoy also namespaces its streams with `Http::`

rebello95: Updated

PR #326, opened 2019-08-13
jni: update java library interfaces

buildbreaker:
Updating java library layer to conform to the streaming interfaces changes. Also, tweaks to make the java layer more consistent with the ObjC layer in https://github.com/lyft/envoy-mobile/pull/327

Description: Updating java library interfaces
Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngine.java
```diff
@@ -2,22 +2,21 @@
...
+  private final EnvoyEngine envoyEngine;
 
   // Private helper class used by the load method to ensure the native library and its
   // dependencies are loaded and initialized at most once.
   private AndroidEngine(Context context) {
     System.loadLibrary("envoy_jni");
-    initialize((ConnectivityManager)context.getSystemService(Context.CONNECTIVITY_SERVICE));
-    engine = new EnvoyEngine();
+    initialize((ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE));
+    envoyEngine = new EnvoyEngineImpl();
```
rebello95: Assuming an upcoming PR will allow for specifying a custom/mockable implementation of `EnvoyEngine`

library/java/src/io/envoyproxy/envoymobile/engine/BUILD
```diff
@@ -20,8 +20,9 @@ envoy_mobile_android_library(
 envoy_mobile_java_library(
     name = "envoy_base_engine_lib",
     srcs = [
-        "Engine.java",
         "EnvoyEngine.java",
+        "EnvoyEngineImpl.java",
+        "EnvoyHTTPStream.java",
```
rebello95: ah ok I'll update mine to match this name

library/java/src/io/envoyproxy/envoymobile/engine/JniLibrary.java
```diff
@@ -24,4 +24,6 @@ protected static native int sendMetadata(EnvoyStream stream, EnvoyHeaders metada
   protected static native int resetStream(EnvoyStream stream);
 
   protected static native int runEngine(String config, String logLevel);
+
+  public static native int cancel(EnvoyStream stream);
```
rebello95: The rest of these are `protected`, not `public` - should this match?

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyHTTPStream.java
```diff
@@ -0,0 +1,61 @@
...
+    JniLibrary.sendData(stream, data, endStream);
+  }
+
+  /**
+   * Send metadata over an HTTP stream. This method can be invoked multiple times.
+   *
+   * @param metadata,  the metadata to send.
+   * @param endStream, supplies whether this is the last data in the stream.
+   */
+  public void sendMetadata(EnvoyHeaders metadata, boolean endStream) {
```
rebello95: We should remove `endStream` here - it's not available in the C interfaces

library/java/src/io/envoyproxy/envoymobile/engine/BUILD
buildbreaker: No need

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngine.java
buildbreaker: Yup. This isn't as important at the moment because we're not "using" the new interfaces within our example applications. I plan to take some time to update this when we do a full end to end wire

library/java/src/io/envoyproxy/envoymobile/engine/types/EnvoyObserver.java
```diff
@@ -45,4 +45,9 @@
    *               execution.
    */
   void onError(EnvoyError error);
+
+  /**
+   * Called when the async HTTP stream has completed.
+   */
+  void onComplete();
```
goaway: I'm sorta of the opinion that we should drop this for now. We're already surfacing this information via booleans in other callbacks. It has the potential to differ in some unexpected ways from those callbacks if people/interfaces rely on it.

library/java/src/io/envoyproxy/envoymobile/engine/JniLibrary.java
```diff
@@ -3,25 +3,24 @@
...
-  protected static native int sendTrailers(EnvoyStream stream, EnvoyHeaders trailers);
+  protected static native int sendTrailers(long stream, EnvoyHeaders trailers);
 
-  protected static native int locallyCloseStream(EnvoyStream stream);
+  protected static native int locallyCloseStream(long stream);
 
-  protected static native int resetStream(EnvoyStream stream);
+  protected static native int resetStream(long stream);
 
   protected static native int runEngine(String config, String logLevel);
```
goaway: Maybe move this out from the middle of the stream interface and delineate with comments.

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngine.java
```diff
@@ -2,22 +2,21 @@
 
 import android.content.Context;
 import android.net.ConnectivityManager;
-import io.envoyproxy.envoymobile.engine.types.EnvoyStatus;
 
 public class AndroidEngine {
```
goaway: Didn't catch this name change before, but it seems fine. We should update iOS if this is the direction we're going.

goaway:
Overall I think this is great starting point for implementing the Android bridging logic. I'm generally in favor of moving forward and merging it. See comments above. :)

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngine.java
buildbreaker: It's primarily to separate out the Android layer so we can have an easier time testing our code with vanilla JUnit

library/java/src/io/envoyproxy/envoymobile/engine/types/EnvoyObserver.java
buildbreaker: This _should_ be our hook the native layer uses to signal to the platform that the native layer is "done" with the observer, right?

I was more or less erroring on the side of mirroring the current `main_interfaces.h`

library/java/src/io/envoyproxy/envoymobile/engine/JniLibrary.java
buildbreaker: mirroring the `main_interface.h`

examples/java/hello_world/MainActivity.java
```diff
@@ -26,7 +26,7 @@
 
 public class MainActivity extends Activity {
   private static final String ENDPOINT =
-      "http://0.0.0.0:9001/api.lyft.com/static/demo/hello_world.txt";
+      "http://s3.amazonaws.com/api.lyft.com/static/demo/hello_world.txt";
```
rebello95: Envoy won't be proxying these requests if they aren't pointed at the address/port it's listening to. Maybe merge master into this branch to pull in the config fixes that merged yesterday?

buildbreaker: It didn't seem to work but we can try again

library/java/src/io/envoyproxy/envoymobile/engine/types/EnvoyObserver.java
goaway: It is our hook for cleaning up, but I don't think it's needed here.

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngine.java
goaway: how about AndroidEnvoyEngineImpl?

goaway: or maybe just AndroidEngineImpl is fine

goaway: The structure here is a little odd to me. I sort of feel like AndroidEngineImpl should extend a JavaEngineImpl that has placeholder logic/methods the the AndroidEngineImpl can override with the Android-specific setup.

library/java/src/io/envoyproxy/envoymobile/engine/JniLibrary.java
```diff
@@ -3,25 +3,84 @@
...
+   * Note that this method implicitly ends the stream.
+   *
+   * @param stream,   the stream to send trailers over.
+   * @param trailers, the trailers to send.
+   * @return int, the resulting status of the operation.
+   */
+  protected static native int sendTrailers(long stream, EnvoyHeaders trailers);
 
-  protected static native int locallyCloseStream(EnvoyStream stream);
+  protected static native int locallyCloseStream(long stream);
```
junr03: This function is now gone from the main interface.

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngine.java
```diff
@@ -2,22 +2,21 @@
...
   // Volatile to ensure double-checked locking works correctly.
   private static volatile AndroidEngine loader = null;
 
-  private final Engine engine;
+  private final EnvoyEngine envoyEngine;
 
   // Private helper class used by the load method to ensure the native library and its
   // dependencies are loaded and initialized at most once.
   private AndroidEngine(Context context) {
     System.loadLibrary("envoy_jni");
```
goaway: Why is this in the initializer? And wouldn't we want to load `envoy_jni` in the pure java case?

goaway: Ok, sorry, I see the loader pattern below now. Should this be AndroidLoader instead? And there be a pure JavaLoader?

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngine.java
buildbreaker: In its current form, it's not implementing the `EnvoyEngine` interface but I do intend to do that. If you're fine with the general layering idea, we can go with this as is for now.

The general idea is that we have one of two structures:
Option 1 (Composition):
`AndroidEngineImpl` implements `EnvoyEngine` and internally delegates to `EnvoyEngineImpl` for all the interface overriden methods

Option 2 (Inheritance):
`AndroidEngineImpl` extends `EnvoyEngineImpl`

I personally prefer option 1 here but we can have a more thorough discussion about this in another PR

buildbreaker: WDYT @goaway

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngine.java
buildbreaker: Let's fix that in another PR and get the interfaces in?

library/java/src/io/envoyproxy/envoymobile/engine/JniLibrary.java
```diff
@@ -3,25 +3,84 @@
...
 
-  protected static native int resetStream(EnvoyStream stream);
+  /**
+   * Half-close an HTTP stream. The stream will be observable and may return further data
+   * via the observer callbacks. However, nothing further may be sent.
+   *
+   * @param stream, the stream to close.
+   * @return int, the resulting status of the operation.
+   */
+  protected static native int resetStream(long stream);
```
goaway: This should have the docstring from `cancel`.

library/java/src/io/envoyproxy/envoymobile/engine/JniLibrary.java
```diff
@@ -3,25 +3,84 @@
...
 
+  /**
+   * Detach all observers from a stream and send an interrupt upstream if supported by transport.
+   *
+   * @param stream, the stream to evict.
+   * @return int, the resulting status of the operation.
+   */
+  protected static native int cancel(long stream);
```
goaway: This isn't on the main interface any more.

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngine.java
goaway: Sounds good.

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngine.java
goaway: Agreed, let's discuss subsequently.

buildbreaker:
@goaway The original intent of the data structures defined in `types/` was to make the JNI methods easier to use from the `jni_inferface.cc` side. This hopefully allows us to more easily implement a pass through from the `jni_interface.cc` to the `main_interface.h`

Open to changing it and I agree it's separate from this PR

goaway:
From conversation in slack, after consideration, I agree we should keep the stuff in `types/` for now, since it may prove useful, and we can evaluate better as we work on the bridging layer.

buildbreaker:
thanks @goaway !

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyStream.java
```diff
@@ -0,0 +1,63 @@
...
+    JniLibrary.sendTrailers(streamHandle, trailers);
+  }
+
+  /**
+   * Cancel the streamHandle. This functions as an interrupt, and aborts further callbacks and
+   * handling of the streamHandle.
+   *
+   * @return Success, unless the streamHandle has already been canceled.
+   */
+  public int resetStream() { return JniLibrary.resetStream(streamHandle); }
```
rebello95: I think this might make sense to be a boolean

PR #327, opened 2019-08-14
ios: utilize engine handle at platform layer

goaway:

### DIFF
--- a/library/objective-c/EnvoyEngine.m
+++ b/library/objective-c/EnvoyEngine.m
@@ -1,21 +1,36 @@
 #import "library/objective-c/EnvoyEngine.h"
 
 #import "library/common/main_interface.h"
-#import "library/common/include/c_types.h"
 
 #import <stdatomic.h>
 
 @implementation EnvoyObserver
 @end
 
-@implementation EnvoyEngineImpl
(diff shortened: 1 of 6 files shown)


library/objective-c/EnvoyEngine.m
```diff
@@ -205,16 +205,15 @@ @implementation EnvoyHttpStream {
   EnvoyHttpStream *_strongSelf;
   EnvoyObserver *_platformObserver;
   envoy_observer *_nativeObserver;
-  envoy_stream_t _nativeStream;
-  envoy_engine_t _engineHandle;
+  envoy_stream_t _streamHandle;
 }
 
-- (instancetype)initWithEngine:(envoy_engine_t)engine observer:(EnvoyObserver *)observer {
+- (instancetype)initWithHandle:(envoy_stream_t)streamHandle observer:(EnvoyObserver *)observer {
```
buildbreaker: I actually like this approach more than the original engine approach

This is a similar concept to what I was thinking about doing for the java side

Definitely can chat more tomorrow in case we think of a better solution

goaway:
recreated here: https://github.com/lyft/envoy-mobile/pull/329

PR #328, opened 2019-08-14
core & ios: implement typed configurations & template

rebello95:
Introduces a `Configuration` type which consumers will use to start instances of `Envoy`. This change allows us to:

- Provide an ergonomic interface for using Envoy Mobile without having to learn Envoy's configuration YAML structure
- Utilize a solid set of default settings for this library
- Prevent consumers from accessing features that are unsupported
- Validate configuration at compile time rather than runtime

This PR:
- Adds a templated configuration in a `.cc` file, exposed through the common layer all the way through to the platform layers
- Adds a `Configuration` type for accessing this configuration
- Updates the iOS example apps to use the new typed configuration instead of their own yaml files
- Maintains an existing initializer for exposing a string configuration for consumers who may still want this functionality

Tests will be added in a separate PR that updates this interface further with a builder for `Envoy` instances.

Resolves https://github.com/lyft/envoy-mobile/issues/169.

Related issue (why this is in C++ instead of a `yaml` resource): https://github.com/lyft/envoy-mobile/issues/341


bazel/swift_static_framework.bzl
```diff
@@ -142,6 +142,7 @@ def swift_static_framework(
         name,
         module_name = None,
         srcs = [],
+        resources = [],
```
goaway: I think the builtin objc framework uses `data` for this and imo we should follow that example.

bazel/swift_static_framework.bzl
```diff
@@ -168,6 +170,7 @@ def swift_static_framework(
     swift_library(
         name = archive_name,
         srcs = srcs,
+        data = resources,
```
goaway: I don't think we need this, since we'll need to manually put this in the framework bundle.

library/common/include/config_template.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+// NOLINT(namespace-envoy)
+
+/**
+ * Template configuration compiled with the Envoy Mobile library.
+ * More information about Envoy's config can be found at:
+ * https://www.envoyproxy.io/docs/envoy/latest/configuration/configuration
+ */
+const char* config_template_str;
```
goaway: `extern const char*`

library/common/main_interface.cc
```diff
@@ -16,6 +16,9 @@
 #include "library/common/buffer/utility.h"
 #include "library/common/http/dispatcher.h"
 #include "library/common/http/header_utility.h"
+#include "library/common/include/config_template.h"
+
+char* config_template() { return strdup(config_template_str); }
```
goaway: Make the return type const and you can drop the `strdup`.

library/objective-c/EnvoyEngine.m
```diff
@@ -47,6 +47,14 @@ - (EnvoyHttpStream *)openHttpStreamWithObserver:(EnvoyObserver *)observer {
 
 @end
 
+@implementation EnvoyConfiguration
+
++ (NSString *)templateString {
+    return [[NSString alloc] initWithUTF8String:config_template()];
```
goaway: Isn't there an init that specifies it won't copy? (I kinda recall that there is?)

library/swift/src/Configuration.swift
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/// Builder used for creating configurations for new Envoy instances.
+@objcMembers
+public final class Configuration: NSObject {
+    /// The address at which Envoy should listen for requests.
+    public private(set) var address: String = "0.0.0.0"
+
+    /// The port at which Envoy should listen for requests.
+    public private(set) var port: String = "9001"
```
goaway: I'd suggest dropping `address` and `port` since default configuration will be targeted at the direct API, which won't have a socket-based listener.

library/common/main_interface.cc
goaway: Actually, better yet, let's drop this function altogether, and just put the `extern`'d const on the `main_interface.h`. No real benefit to indirection here.

goaway: You can still define the `const` in a separate file.

library/swift/src/Configuration.swift
rebello95: Fine with me

library/objective-c/EnvoyEngine.m
rebello95: This could maybe work, but TBH I don't really think copying the bytes here is a big deal since it's a 1-time setup. Also trying this fails to parse when passing it to an Envoy initializer:

```objc
return [[NSString alloc] initWithBytesNoCopy:&config_template
                                      length:strlen(config_template)
                                    encoding:kCFStringEncodingUTF8
                                freeWhenDone:NO];
```

library/common/main_interface.cc
rebello95: Done

rebello95:
> If we structure the configuration in swift we are opening up the gates for having to do the same in every language we want to support.

I outlined some of the concerns with a purely string-based approach in the [linked issue](https://github.com/lyft/envoy-mobile/issues/169), but to add some more color:

- Envoy Mobile supports a small subset of the configurations that are available in upstream Envoy. This will likely always be the case, and going outside those configurations can cause undefined behavior at runtime. Using types can help alleviate this
- Having consumers manually load configuration file resources and pass them to Envoy Mobile at runtime is additional overhead and can lead to bugs discovered at runtime instead of build time (invalid config files, mis-loading resources, etc.)
- Envoy's current configurations are non-trivial and require a learning curve to be able to use them at all. The status quo of networking libraries on mobile generally requires little up-front configuration cost, and we should try to match that here to encourage wider adoption
- Supporting this in each language (currently Swift/Kotlin) does amount to some overhead, but that can be said about every other aspect of the library as well (i.e., Envoy Mobile isn't all C++ 😄)

> What are your thoughts about maintenance of this typed configuration approach?

- Configs will be forwards compatible. Using a builder pattern on both platforms will allow us to add new defaults without making consumers update their code unless they want to take advantage of the new features
- On the other hand, we can always deprecate/remove options from the configuration if necessary (especially pre-1.0), and consumers will find out at build time rather than when loading a config file at runtime

In summary: I think the reduced barrier to entry, ease of use, and simplicity of typed configurations far out-weighs the minimal maintenance cost this will incur.

junr03:
Thanks for that explanation. It goes to show my bias from the far more complex land of service mesh configurations. I think all the reasons listed above really assuage any immediate concerns I have with this approach :)

> Envoy's current configurations are non-trivial and require a learning curve to be able to use them at all. The status quo of networking libraries on mobile generally requires little up-front configuration cost, and we should try to match that here to encourage wider adoption.

This specially makes me agree with the approach taken here. And reminds me that a very similar pattern is being used in cronet; with, as you say here, a very small configuration surface.

goaway:
I talked with @junr03 a bit about this offline. In general, while I think @junr03's concerns merit consideration, I tend to agree with @rebello95's points above. To summarize a couple of the things we discussed:

I think the challenges the Network Team has faced are a bit beyond the scope of what we'll encounter here. The configuration of Lyft's mesh is far more complex than our default template configuration will ever have to be.

The proto generation, serialization and deserialization that would be required to support the proto based approach is a fair amount of overhead, whereas string interpolation is cheap, and at least for a smaller set of configurable touch points, easy.

Finally, I think we can avoid the risks of implementing across languages by having the templating be well-documented. If this ends up being a pain point, we can also consider migrating templating logic down into the common layer at some point down the road.

rebello95:
> Finally, I think we can avoid the risks of implementing across languages by having the templating be well-documented. If this ends up being a pain point, we can also consider migrating templating logic down into the common layer at some point down the road.

Yep I think documenting this is very valuable, and we can also use upstream Envoy's comprehensive docs as a point of reference here too. I also think it's useful to allow consumers to instantiate Envoy Mobile instances using raw string files (which is maintained in this PR), for those who are interested in doing something ultra-customized.

library/objective-c/EnvoyEngine.m
```diff
@@ -51,6 +51,16 @@ - (EnvoyHTTPStream *)startStreamWithObserver:(EnvoyObserver *)observer {
 
 @end
 
+#pragma mark - EnvoyConfiguration
+
+@implementation EnvoyConfiguration
+
++ (NSString *)templateString {
+  return [[NSString alloc] initWithUTF8String:config_template];
```
goaway: nit: how about putting
```
static NSString * const kEnvoyTemplateString = [[NSString alloc] initWithUTF8String:config_template];
```
above the `@implementation` and then just returning that each time.

library/swift/src/Envoy.swift
```diff
@@ -15,12 +15,20 @@ public final class Envoy: NSObject {
...
-    self.runner = RunnerThread(config: config, engine: self.engine, logLevel: logLevel)
+  public convenience init(config: Configuration = Configuration(), logLevel: LogLevel = .info) throws {
+    self.init(configFile: try config.build(), logLevel: logLevel)
+  }
+
+  /// Initialize a new Envoy instance using a string configuration.
+  ///
+  /// - parameter configFile: Configuration file to use for starting Envoy.
+  /// - parameter logLevel:   Log level to use for this instance.
+  public init(configFile: String, logLevel: LogLevel = .info) {
```
goaway: nit: let's maybe call this `configYaml` instead of `configFile`, since there's more than one type of config file that can be used to run envoy. (Also this matches the option name that Envoy uses internally, for consistency.)

goaway:
Implementation looks great; just a couple minor suggestions.

library/swift/src/Envoy.swift
rebello95: sgtm

library/objective-c/EnvoyEngine.m
rebello95: Unfortunately this doesn't work since we hit the following error:

> Initializer element is not a compile-time constant

goaway: Ah, makes sense. Silly, since of course it is constant at compile-time, but I understand the limitations at this scope.

You _could_ work around this by using `#define` instead of the extern'd const... but, let's not worry about it.

PR #331, opened 2019-08-15
core: init functions

goaway:

### DIFF
--- a/library/common/main_interface.cc
+++ b/library/common/main_interface.cc
@@ -45,6 +45,7 @@ envoy_status_t send_trailers(envoy_stream_t, envoy_headers) { return ENVOY_FAILU
 envoy_status_t reset_stream(envoy_stream_t stream) { return http_dispatcher_->resetStream(stream); }
 
 envoy_engine_t init_engine() {
+  // TODO(goaway): return new handle once multiple engine support is in place.
   return 1;
 }
 
--- a/library/common/main_interface.cc
+++ b/library/common/main_interface.cc
@@ -44,6 +44,7 @@ envoy_status_t reset_stream(envoy_stream_t stream) { return http_dispatcher_->re
 
 envoy_engine_t init_engine() {
   // TODO(goaway): return new handle once multiple engine support is in place.
+  // https://github.com/lyft/envoy-mobile/issues/332
   return 1;
 }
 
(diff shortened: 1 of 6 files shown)


library/common/main_interface.cc
```diff
@@ -45,6 +45,7 @@ envoy_status_t send_trailers(envoy_stream_t, envoy_headers) { return ENVOY_FAILU
 envoy_status_t reset_stream(envoy_stream_t stream) { return http_dispatcher_->resetStream(stream); }
 
 envoy_engine_t init_engine() {
+  // TODO(goaway): return new handle once multiple engine support is in place.
```
junr03: I think it is worth it to open an issue for this work item.

junr03: I think it is worth it to open an issue for this work item.

library/common/main_interface.cc
```diff
@@ -21,9 +22,15 @@
 
 static std::unique_ptr<Envoy::MainCommon> main_common_;
 static std::unique_ptr<Envoy::Http::Dispatcher> http_dispatcher_;
+static std::atomic<envoy_stream_t> current_stream_id_{0};
```
junr03: I am assuming the vision is to make both this and the dispatcher owned by the engine instance once that work is done?

library/common/main_interface.h
```diff
@@ -11,12 +11,19 @@
 extern "C" { // functions
 #endif
 
+/**
+ * Initialize an underlying HTTP stream.
+ * @param engine, handle to the engine that will manage this stream.
+ * @return envoy_stream_t, handle to the underlying stream.
+ */
+envoy_stream_t init_stream(envoy_engine_t);
```
junr03: Should the comment also say that the stream must be started before other stream operations happen? Is that confusing?

library/common/main_interface.cc
goaway: Nope - stream handles will be necessarily globally unique. This is so that we can locate the appropriate the engine/dispatcher by just the handle.

goaway: just realized I didn't update the name here though, let me fix

library/common/main_interface.h
goaway: sure

goaway: I don't think it's particularly confusing, but I do agree it's worth calling out explicitly. Updated.

library/common/main_interface.cc
junr03: And hence why init_stream takes an envoy_engine_t gotcha.

PR #336, opened 2019-08-15
bridge: remove unneeded struct

goaway:


goaway:
Missed that this isn't used anywhere now.

PR #337, opened 2019-08-16
objc: fix envoy_observer native_init

rebello95:
This was being instantiated in the wrong order, and warning at compile time that the wrong closure types were being assigned to `envoy_observer`:

```
library/objective-c/EnvoyEngine.m:228:33: warning: incompatible pointer types initializing 'envoy_on_error_f' (aka 'void (*)(envoy_error, void *)') with an expression of type 'void (void *)' [-Wincompatible-pointer-types]
                                ios_on_complete, ios_on_error, context};
                                ^~~~~~~~~~~~~~~
library/objective-c/EnvoyEngine.m:228:50: warning: incompatible pointer types initializing 'envoy_on_complete_f' (aka 'void (*)(void *)') with an expression of type 'void (envoy_error, void *)' [-Wincompatible-pointer-types]
                                ios_on_complete, ios_on_error, context};
                                                 ^~~~~~~~~~~~
```

### DIFF
--- a/library/objective-c/EnvoyEngine.m
+++ b/library/objective-c/EnvoyEngine.m
@@ -224,8 +224,8 @@ static void ios_on_error(envoy_error error, void *context) {
 
   // Create native observer
   envoy_observer *native_obs = (envoy_observer *)malloc(sizeof(envoy_observer));
-  envoy_observer native_init = {ios_on_headers,  ios_on_data,  ios_on_trailers, ios_on_metadata,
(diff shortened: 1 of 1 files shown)


rebello95:
@keith that's a great callout. Tried running locally with warnings as errors, but as you mentioned it fails with core Envoy. What would be the best way to run this only on specific Envoy Mobile `cc_library` targets?

rebello95:
Filed https://github.com/lyft/envoy-mobile/issues/339

PR #340, opened 2019-08-16
Create an AndroidJniLibrary and move specific loader logic into the respective JniLibrary classes

buildbreaker:
Create an `AndroidJniLibrary` and move specific loader logic into the `*JniLibrary` classes. Cleaning up the other layers in Kotlin due to this change.

Description: Create an `AndroidJniLibrary` and move specific loader logic into the `*JniLibrary` classes
Risk Level: low
Testing: local
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/examples/java/hello_world/MainActivity.java
+++ b/examples/java/hello_world/MainActivity.java
@@ -10,6 +10,8 @@ import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
 import android.util.Log;
 import io.envoyproxy.envoymobile.Envoy;
+import io.envoyproxy.envoymobile.engine.AndroidEngineImpl;
+import io.envoyproxy.envoymobile.engine.EnvoyEngine;
 import io.envoyproxy.envoymobile.shared.Failure;
 import io.envoyproxy.envoymobile.shared.Response;
 import io.envoyproxy.envoymobile.shared.ResponseRecyclerViewAdapter;
@@ -43,7 +45,6 @@ public class MainActivity extends Activity {
     setContentView(R.layout.activity_main);
 
     Context context = getBaseContext();
-    Envoy.load(context);
 
     // Create envoy instance with config.
     String config;
@@ -53,7 +54,9 @@ public class MainActivity extends Activity {
       Log.d("MainActivity", "exception getting config.", e);
       throw new RuntimeException("Can't get config to run envoy.");
     }
-    envoy = new Envoy(context, config);
+
+    EnvoyEngine envoyEngine = new AndroidEngineImpl(context);
+    envoy = new Envoy(envoyEngine, config);
 
     recyclerView = (RecyclerView)findViewById(R.id.recycler_view);
     recyclerView.setLayoutManager(new LinearLayoutManager(this));
--- a/examples/kotlin/hello_world/MainActivity.kt
+++ b/examples/kotlin/hello_world/MainActivity.kt
@@ -10,6 +10,7 @@ import android.support.v7.widget.LinearLayoutManager
 import android.support.v7.widget.RecyclerView
 import android.util.Log
 import io.envoyproxy.envoymobile.Envoy
+import io.envoyproxy.envoymobile.engine.AndroidEngineImpl
 import io.envoyproxy.envoymobile.shared.Failure
 import io.envoyproxy.envoymobile.shared.Response
 import io.envoyproxy.envoymobile.shared.ResponseRecyclerViewAdapter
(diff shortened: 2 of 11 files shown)


examples/java/hello_world/MainActivity.java
```diff
@@ -53,7 +54,9 @@ protected void onCreate(Bundle savedInstanceState) {
       Log.d("MainActivity", "exception getting config.", e);
       throw new RuntimeException("Can't get config to run envoy.");
     }
-    envoy = new Envoy(context, config);
+
+    EnvoyEngine envoyEngine = new AndroidEngineImpl(context);
```
buildbreaker: At the moment, this will have to do. A more ideal ergonomic solution is to use a Builder or Factory to create an `Envoy` instance

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngineImpl.java
```diff
@@ -13,9 +14,8 @@
 
   // Private helper class used by the load method to ensure the native library and its
   // dependencies are loaded and initialized at most once.
-  private AndroidEngineImpl(Context context) {
-    System.loadLibrary("envoy_jni");
-    initialize((ConnectivityManager)context.getSystemService(Context.CONNECTIVITY_SERVICE));
+  public AndroidEngineImpl(Context context) {
+    initialize((ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE));
```
buildbreaker: We have a few options here and this is something we can figure out separate from this PR:

If `c-ares` is initialized, we can:
1. Flip a global static bit in the platform layer to signal all other `AndroidEngineImpl` instances to skip this initialization call
2. Flip a global static bit in the native layer
3. Have a method which exposes if `c-ares` is initialized (we have a native method already)

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyEngineImpl.java
```diff
@@ -6,7 +6,10 @@
 
   private final long engineHandle;
 
-  public EnvoyEngineImpl() { this.engineHandle = JniLibrary.initEngine(); }
+  public EnvoyEngineImpl() {
+    System.loadLibrary("envoy_jni");
```
buildbreaker: Moved here since we need the library here. The only odd thing is that we also have native bindings in the `AndroidEngineImpl` class -- maybe we can put this load functionality into the `JniLibrary`?

examples/java/hello_world/MainActivity.java
buildbreaker: Spoke with @rebello95 offline. We're going to go with the Builder pattern.

His [PR](https://github.com/lyft/envoy-mobile/pull/328) introduces a `Configuration` concept for Envoy. We'll rename the class and have that responsible for creating Envoy instances with the user specified configuration

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngineImpl.java
```diff
@@ -2,8 +2,9 @@
 
 import android.content.Context;
 import android.net.ConnectivityManager;
+import io.envoyproxy.envoymobile.engine.types.EnvoyObserver;
 
-public class AndroidEngineImpl {
+public class AndroidEngineImpl implements EnvoyEngine {
```
goaway: I think this will simplified by subclassing. In particular, I think you want the both the Java and Android versions to share the same `synchronized` loader state, which will only be possible via subclassing.

I also think this really isn't covering the same set of concerns as `EnvoyEngine` over in iOS. In fact, I'm having trouble following the logic here now.

This class was originally designed to be instantiated once only (IIRC), and have that instantiation guarded in a synchronized fashion, to ensure the native DLL is only loaded and initialized once. I'd suggest moving this functionality to an `AndroidEngineLoader` that subclasses an `EnvoyEngineLoader` and then simply having a single `EnvoyEngine(Impl)`.

examples/kotlin/hello_world/MainActivity.kt
```diff
@@ -33,10 +34,9 @@ class MainActivity : Activity() {
     super.onCreate(savedInstanceState)
     setContentView(R.layout.activity_main)
 
-    Envoy.load(baseContext)
-
     // Create Envoy instance with config.
-    envoy = Envoy(baseContext, loadEnvoyConfig(baseContext, R.raw.config))
+    val envoyEngine = AndroidEngineImpl(baseContext)
+    envoy = Envoy(envoyEngine, loadEnvoyConfig(baseContext, R.raw.config))
```
goaway: In some ideal future, the public interface will only require a single line here (but I think we're all on the same page on this).

buildbreaker: @rebello95 and I were thinking about having a builder class for creating an Envoy instance so we don't have to do this. It'll bundle together the configuration and log levels also.

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngineImpl.java
buildbreaker: Paired and we went with this: https://github.com/lyft/envoy-mobile/pull/340/commits/c6228343270117789894a5ec01c9de5572c99c56

examples/java/hello_world/MainActivity.java
rebello95: @buildbreaker will update this to look like iOS: https://github.com/lyft/envoy-mobile/pull/343

examples/kotlin/hello_world/MainActivity.kt
rebello95: iOS example: https://github.com/lyft/envoy-mobile/pull/343

PR #343, opened 2019-08-16
ios: implement Envoy builder & add tests

rebello95:
Implements a new `EnvoyBuilder` type which consumers will now use to create new instances of `Envoy` going forward (rather than initializing it directly).

This supports the following using a builder pattern:
- Adding log levels
- Adding configuration options that were previously part of `Configuration`
- Specifying a YAML file override for consumers who may want to use a custom configuration instead of the default template
- Creating an instance of `Envoy` with these configurations using an underlying `EnvoyEngine` implementation
- Injecting mock engines so we don't start real instances of Envoy when testing

Example usage:

```swift
let envoy = try EnvoyBuilder()
  .addStatsFlushSeconds(30)
  .addDNSRefreshSeconds(30)
  .addLogLevel(.trace)
  .build()
```


rebello95:
> Aw man... :(
> Too bad the original design didn't work

In response to fixing the segfault 😬 - I also linked the issue inline

PR #345, opened 2019-08-19
core: comment on setup_envoy/run_engine ordering

junr03:
Description: comment describing current ordering between `setup_envoy` and `run_engine` in the main_interface.
Risk Level: low - code comment

### DIFF
--- a/library/common/main_interface.cc
+++ b/library/common/main_interface.cc
@@ -50,7 +50,7 @@ envoy_engine_t init_engine() {
 
 /*
  * Setup envoy for interaction via the main interface.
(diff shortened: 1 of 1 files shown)


library/common/main_interface.cc
```diff
@@ -50,6 +50,10 @@ envoy_engine_t init_engine() {
 
 /*
  * Setup envoy for interaction via the main interface.
+ * As it stands this function __has__ to be executed after run engine.
```
rebello95:
```suggestion
 * As it stands this function __must__ be executed after calling `run_engine`.
```

PR #347, opened 2019-08-19
ios: add request unary interfaces

rebello95:
Adds unary interfaces (related to https://github.com/lyft/envoy-mobile/issues/118) which will call into the streaming interfaces under the hood.

These are to be used as convenience convenience accessors for the common case of non-streaming requests, and will be added on Android as part of https://github.com/lyft/envoy-mobile/issues/119.

This also removes `data` and `trailers` from `Request`, as the unary function will take care of this for unary requests, and streams will be expected to handle this through the `StreamEmitter`.

The underlying implementation will look something like this:

```swift
public func sendUnary(_ request: Request, body: Data?,
                      trailers: [String: [String]] = [:], handler: ResponseHandler)
{
  let emitter = self.startStream(with: request, handler: handler)
  if let body = body {
    emitter.sendData(body)
  }

  emitter.close(trailers: trailers)
}
```

### DIFF
--- a/library/swift/src/Client.swift
+++ b/library/swift/src/Client.swift
@@ -8,7 +8,25 @@ public protocol Client {
   /// - parameter request: The request for opening a stream.
   /// - parameter handler: Handler for receiving stream events.
   ///
-  /// - returns: Emitter for sending streaming data outward.
-  func startStream(request: Request, handler: ResponseHandler)
-    -> StreamEmitter
+  /// - returns: Emitter for sending streaming data outward,
+  ///            or nil if the stream couldn't be started.
+  func startStream(with request: Request, handler: ResponseHandler) -> StreamEmitter
+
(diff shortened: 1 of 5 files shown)


library/swift/src/Client.swift
```diff
@@ -8,7 +8,25 @@ public protocol Client {
...
+  /// - returns: Emitter for sending streaming data outward,
+  ///            or nil if the stream couldn't be started.
+  func startStream(with request: Request, handler: ResponseHandler) -> StreamEmitter
+
+  /// Convenience function for sending a unary request.
+  ///
+  /// - parameter request:  The request to send.
+  /// - parameter body:     Serialized data to send as the body of the request.
+  /// - parameter trailers: Trailers to send with the request.
+  func sendUnary(_ request: Request, body: Data?,
```
buildbreaker: 💯

PR #349, opened 2019-08-20
swift: update demos to use direct API

rebello95:
**This is a working end-to-end iOS implementation using Envoy Mobile as a library!**

Depends on https://github.com/lyft/envoy-mobile/pull/312 and https://github.com/lyft/envoy-mobile/pull/353.

Updates our Swift example app to use the Envoy Mobile direct interfaces rather than using `URLSession` and Envoy as a listener.

Note: For now, we are no longer showing the text response of the response body due to the fact that the underlying implementation is not yet complete for bodied requests (headers-only).

Simulator Screen Shot - TestDevice - 2019-08-20 at 13 57 44


examples/swift/hello_world/ViewController.swift
```diff
@@ -34,47 +40,38 @@ final class ViewController: UITableViewController {
...
-    let untypedHeaders = response.allHeaderFields
-    let headers = Dictionary(uniqueKeysWithValues: untypedHeaders.map
-    { header -> (String, String) in
-      return (header.key as? String ?? String(describing: header.key), "\(header.value)")
-    })
+    let requestID = self.requestCount
+    let request = RequestBuilder(method: .get, scheme: kRequestScheme,
+                                 authority: kRequestAuthority,
+                                 path: kRequestPath).build()
+    var status = -1
```
buildbreaker: This is pretty lame to do in order to be able to log/read the status code of a given error and success -- especially for the unary case.

rebello95: Yea agreed - I think we should discuss providing a way for `onError` to pipe through the status code as well (or maybe introduce a new stream handler)

rebello95: Actually, thinking about this more - `onError` here will be called if an error occurs within Envoy (not if a 4xx/5xx is returned). So in that case, seems like we should return an `EnvoyError` in that lambda and not worry about the HTTP status code being piped through there. I can update the logs in this example here too. WDYT?

PR #351, opened 2019-08-20
ci: enable github actions

junr03:
Description: this PR adds envoy mobile's CI workflow to github actions.
Risk Level: med - adding a new CI workflow. Parity is required with Azure Pipelines


.github/workflows/android.yml
```diff
@@ -0,0 +1,98 @@
...
+        run: |
+          export ANDROID_NDK_HOME=/usr/local/lib/android/sdk/ndk-bundle
+          ./bazelw build --fat_apk_cpu=x86 //examples/java/hello_world:hello_envoy
+      - name: 'Build kotlin app'
+        run: |
+          export ANDROID_NDK_HOME=/usr/local/lib/android/sdk/ndk-bundle
+          ./bazelw build --fat_apk_cpu=x86 //examples/kotlin/hello_world:hello_envoy_kt
+  macdist:
+    name: mac_dist
+    runs-on: macos-latest
```
rebello95: Are we sure that running on the latest here is always safe?

junr03: probably not. I can pin to a version in a subsequent PR.

rebello95: sounds good!

PR #353, opened 2019-08-20
ios: e2e hello world demo for swift

junr03:
Description: this PR fixes a couple bugs to allow e2e use of the Envoy Mobile library in the swift iOS demo.
Risk Level: low - fixing bugs
Testing: `./bazelw build --config=ios --xcode_version=10.3.0.10G8 //:ios_dist` and then `./bazelw run --config=ios --xcode_version=10.3.0.10G8 //examples/swift/hello_world:app`


library/objective-c/EnvoyEngine.h
```diff
@@ -137,10 +137,6 @@ typedef NSDictionary<NSString *, NSArray<NSString *> *> EnvoyHeaders;
  */
 - (int)runWithConfig:(NSString *)config logLevel:(NSString *)logLevel;
 
-/// Performs necessary setup after Envoy has initialized and started running.
-/// TODO: create a post-initialization callback from Envoy to handle this automatically.
-- (void)setup;
```
rebello95: 💯

library/objective-c/EnvoyEngine.m
```diff
@@ -237,19 +234,17 @@ - (instancetype)initWithHandle:(uint64_t)handle observer:(EnvoyObserver *)observ
   _platformObserver = observer;
 
   // Create callback context
-  ios_context *context = (ios_context *)malloc(sizeof(ios_context));
```
rebello95: Did these actually cause problems?

library/swift/src/Envoy.swift
```diff
@@ -49,6 +49,7 @@ public final class Envoy: NSObject {
 extension Envoy: Client {
   public func startStream(with request: Request, handler: ResponseHandler) -> StreamEmitter {
     let httpStream = self.engine.startStream(with: handler.underlyingObserver)
+    NSLog("Sending request with headers: \(request.outboundHeaders())")
```
rebello95: Let's remove this

library/objective-c/EnvoyEngine.m
junr03: nope, the cast is just not needed. I'll open another PR to homogenize this style.

library/swift/src/Envoy.swift
junr03: sure. I thought this was a pretty good log line to have. Is there a way to define log levels with NSLog?

junr03: For the future. I will remove this for now.

rebello95: Nope, unfortunately no log levels with `NSLog`. I think this would be good to have in the demos but maybe not the engine for now

PR #354, opened 2019-08-20
ios: update unary/streaming & add test

rebello95:
Per the discussion [here](https://github.com/lyft/envoy-mobile/pull/312#discussion_r315820916), updating the unary and streaming interfaces.

- Moved the non-streaming convenience function into an extension on the protocol type so that it's available to all consumers
- Added a test to validate the default behavior of this extension function
- Added a `CancelableStream` protocol which includes a subset of functionality from the `StreamEmitter`, allowing consumers of the unary function to cancel requests without having the ability to send additional data into the stream

### DIFF
--- a/library/swift/src/Client.swift
+++ b/library/swift/src/Client.swift
@@ -13,7 +13,7 @@ public protocol Client {
 }
 
 extension Client {
-  /// Convenience function for sending and completing a non-streaming request.
+  /// Convenience function for sending a complete (non-streamed) request.
   ///
   /// - parameter request:  The request to send.
   /// - parameter data:     Serialized data to send as the body of the request.
--- a/library/swift/src/Client.swift
+++ b/library/swift/src/Client.swift
@@ -9,23 +9,28 @@ public protocol Client {
   /// - parameter handler: Handler for receiving stream events.
   ///
   /// - returns: Emitter for sending streaming data outward.
-  func startStream(with request: Request, handler: ResponseHandler) -> StreamEmitter
-
-  /// Convenience function for sending a unary request.
-  ///
-  /// - parameter request:  The request to send.
-  /// - parameter body:     Serialized data to send as the body of the request.
-  /// - parameter trailers: Trailers to send with the request.
-  func sendUnary(_ request: Request, body: Data?,
-                 trailers: [String: [String]], handler: ResponseHandler)
+  func send(_ request: Request, handler: ResponseHandler) -> StreamEmitter
 }
 
 extension Client {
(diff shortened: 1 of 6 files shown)


library/swift/src/Client.swift
```diff
@@ -9,23 +9,28 @@ public protocol Client {
...
-  public func sendUnary(_ request: Request, body: Data?, handler: ResponseHandler) {
-    self.sendUnary(request, body: body, trailers: [:], handler: handler)
+  /// - parameter data:     Serialized data to send as the body of the request.
+  /// - parameter trailers: Trailers to send with the request.
+  ///
+  /// - returns: A cancelable request.
+  @discardableResult
+  public func send(_ request: Request, data: Data?,
+                   trailers: [String: [String]] = [:], handler: ResponseHandler)
+    -> CancelableStream
```
goaway: `CancelableStream` is very clever; I like this approach.

library/swift/src/Envoy.swift
```diff
@@ -47,20 +47,9 @@ public final class Envoy: NSObject {
 }
 
 extension Envoy: Client {
-  public func startStream(with request: Request, handler: ResponseHandler) -> StreamEmitter {
+  public func send(_ request: Request, handler: ResponseHandler) -> StreamEmitter {
```
goaway: Would `send(request:` be clearer than `send(_ request:`? Genuine question - I don't actually know which is more idiomatic Swift.

rebello95: I think this is common practice since type inference of `Request` is available from the compiler

library/swift/src/Client.swift
```diff
@@ -9,23 +9,28 @@ public protocol Client {
...
-  /// - parameter body:     Serialized data to send as the body of the request.
-  /// - parameter trailers: Trailers to send with the request.
-  func sendUnary(_ request: Request, body: Data?,
-                 trailers: [String: [String]], handler: ResponseHandler)
+  func send(_ request: Request, handler: ResponseHandler) -> StreamEmitter
 }
 
 extension Client {
-  /// Convenience function for sending a unary request without trailers.
+  /// Convenience function for sending and completing a non-streaming request.
```
goaway: nit/minor suggestion:
`/// Convenience function for sending a complete (non-streamed) request.`

rebello95: updated

rebello95:
@buildbreaker thanks for the callout - we can do something like `CancelEmitter` in the future if that makes sense 🤷‍♂

PR #359, opened 2019-08-21
ci: pin macOS and linux versions

rebello95:
This will prevent machines from switching beneath us.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -5,7 +5,7 @@ on: [push]
 jobs:
   linuxdist:
     name: linux_dist
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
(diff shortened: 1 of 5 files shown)


rebello95:
Re-running

junr03:
I am seeing the same failures across all the linux runs when installing deps. I am investigating

rebello95:
Yep, appears to be red on an unrelated PR too: https://github.com/lyft/envoy-mobile/pull/360

junr03:
:/ #351 was all green for the deps. Something must have happened with the mirror they use for package management. I am trying to see if 16.04 has the same issue. If they do I will talk to github/msft people tomorrow.

rebello95:
Sounds good 👍 looks like something's failing to download internally

PR #362, opened 2019-08-21
android: add android envoy builder

buildbreaker:
`AndroidEnvoyBuilder` just subclasses `EnvoyBuilder` but just uses the `AndroidEnvoyEngineImpl` instead of the vanilla `EnvoyEngineImpl`

I had to declare the `EnvoyBuilder` to be an open class in order for the subclassing to be available

The other change is to lift the method `EnvoyBuilder.addEngineType()` to the constructor since it' guarantee that the engine remains the same throughout the lifecycle of the builder

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


library/kotlin/src/io/envoyproxy/envoymobile/EnvoyBuilder.kt
```diff
@@ -7,18 +7,25 @@ import io.envoyproxy.envoymobile.engine.EnvoyEngineImpl
...
 
-  constructor() : this(EnvoyConfigurationImpl())
+  constructor() : this(EnvoyConfigurationImpl(), { EnvoyEngineImpl() })
+
+  /**
+   * Initialize with a specific implementation of `EnvoyEngine` to use for starting Envoy.
+   *
+   * A new instance of this engine will be created when `build()` is called.
+   */
+  internal constructor(engineType: () -> EnvoyEngine) : this(EnvoyConfigurationImpl(), engineType)
```
buildbreaker: Not that opinionated to undo

rebello95: Don't feel too strongly but one reason to revert would be for parity with iOS (since we have to do it the other way to prevent a segfault bug). Don't think adding a guarantee like this really provides much value since it's `internal`

buildbreaker:
@rebello95 Yeah I remembered that point as I stood up from my desk

PR #365, opened 2019-08-21
kotlin: implement unary and streaming interfaces

buildbreaker:
Kotlin unary/stremaing interfaces implementation

Tests verify the interaction between the `Envoy` and the `EnvoyEngine` along with the underlying `EnvoyHTTPStream` objects

Description: kotlin: implement unary and streaming interfaces
Risk Level: low
Testing: Unit
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
```diff
@@ -18,15 +20,19 @@ interface Client {
    * @param body Serialized data to send as the body of the request.
    * @param trailers Trailers to send with the request.
    * @param responseHandler the callback for receiving stream events.
+   * @return CancellableStream, a cancelable request.
```
rebello95: nit: Here and everywhere else in the diff should be `CancelableStream` for parity with iOS and the rest of the spellings

library/kotlin/src/io/envoyproxy/envoymobile/Envoy.kt
```diff
@@ -53,4 +54,22 @@ class Envoy @JvmOverloads constructor(
   fun isTerminated(): Boolean {
     return runner.state == Thread.State.TERMINATED
   }
+
+  override fun startStream(request: Request, responseHandler: ResponseHandler): StreamEmitter {
```
rebello95: We don't have `startStream` anymore, just the two `send` functions: https://github.com/lyft/envoy-mobile/blob/master/library/swift/src/Client.swift

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyStreamEmitter.kt
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * For sending data to an associated stream.
+   *
+   * @param byteBuffer the byte buffer data to send to the stream.
+   * @throws IllegalStateException when the stream is not active
+   * @throws EnvoyException when there is an exception sending data.
+   * @return StreamEmitter, this stream emitter.
+   */
+  override fun sendData(byteBuffer: ByteBuffer): StreamEmitter {
```
rebello95: `data` instead of `byteBuffer`?

library/kotlin/src/io/envoyproxy/envoymobile/StreamEmitter.kt
```diff
@@ -2,13 +2,22 @@ package io.envoyproxy.envoymobile
 
 import java.nio.ByteBuffer
 
-interface StreamEmitter {
+/**
+ * Interface for a stream that may be canceled.
+ */
+interface CancellableStream {
```
rebello95:
```suggestion
interface CancelableStream {
```

library/kotlin/src/io/envoyproxy/envoymobile/Client.kt
```diff
@@ -10,29 +10,29 @@ interface Client {
...
   /**
    * Convenience function for sending a unary request.
    *
    * @param request The request to send.
    * @param body Serialized data to send as the body of the request.
    * @param responseHandler the callback for receiving stream events.
-   * @return CancellableStream, a cancelable request.
+   * @return CancelableStream, a cancelable request.
    */
   fun send(request: Request, body: ByteBuffer?,
```
rebello95: Is the only reason for this function to allow consumers to avoid specifying trailers?

buildbreaker: This is equivalent to this, right: https://github.com/lyft/envoy-mobile/blob/537be8e3a58ad4d17baeef18e727b2171e84c30d/library/swift/src/Envoy.swift#L50

rebello95: Sorry, to clarify: You have 3 functions instead of the two here: https://github.com/lyft/envoy-mobile/blob/537be8e3a58ad4d17baeef18e727b2171e84c30d/library/swift/src/Client.swift

buildbreaker: We can do an extension function but I don't think the Java consumer would get something that is expected

The other option is to have a default value but we'll have to change the signature to have the trailers be the last parameter (which is kind of lame because I like that the `ResponseHandler` is last)

library/kotlin/src/io/envoyproxy/envoymobile/Envoy.kt
```diff
@@ -53,4 +56,22 @@ class Envoy @JvmOverloads constructor(
...
   }
+
+  override fun send(request: Request, responseHandler: ResponseHandler): StreamEmitter {
+    val stream = engine.startStream(responseHandler.underlyingObserver)
+    stream.sendHeaders(request.headers, false)
+    return EnvoyStreamEmitter(stream)
+  }
+
+  override fun send(request: Request, data: ByteBuffer?, trailers: Map<String, List<String>>, responseHandler: ResponseHandler): CancelableStream {
+    val stream = engine.startStream(responseHandler.underlyingObserver)
```
rebello95: In iOS, I actually call into the other `send` function instead of duplicating some of the logic here. Not a blocker, but some nice cleanup

PR #368, opened 2019-08-22
ios: split out objc sources with single bridging header

goaway:

Description: Separates Objective-C sources into separate files, separating logic a little bit better. A single bridging header is still required (for now).


goaway:
While we still can't compose the header via imports (and thus still need a single header), this at least improves the Objective-C landscape a bit by breaking out the monolithic sources.

PR #372, opened 2019-08-22
bazel: fix Objective-C compilation & demo

rebello95:
This updates our static framework rule to properly copy the `-Swift.h` header file from the bazel output directory into our final `Envoy.framework/Headers/Envoy-Swift.h`.

There are a few workarounds here for issues that @kastiglione has been nice enough to file upstream:

**https://github.com/bazelbuild/rules_swift/issues/291**

The code required to find and copy the `-Swift.h` header file could be simplified with this issue.

**https://github.com/bazelbuild/rules_apple/issues/557**

From the Objective-C demo app, we also depend on an empty Swift library to force dependencies like `swiftFoundation` to link.

Resolves https://github.com/lyft/envoy-mobile/issues/230

### DIFF
--- a/bazel/swift_static_framework.bzl
+++ b/bazel/swift_static_framework.bzl
@@ -43,6 +43,13 @@ def _swift_static_framework_impl(ctx):
             fail("Unhandled platform '{}'".format(platform))
 
         swift_info = archive[SwiftInfo]
+
+        # We can potentially simplify this if this change lands upstream:
+        # https://github.com/bazelbuild/rules_swift/issues/291
+        objc_headers = archive[apple_common.Objc].header.to_list()
+        print("** Swift info: {}".format(archive[apple_common.Objc].header.to_list()))
+        #        print("** Swift info: {}".format(dir(archive[CcInfo].compilation_context.)))
+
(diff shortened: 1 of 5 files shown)


bazel/swift_static_framework.bzl
```diff
@@ -43,6 +49,14 @@ def _swift_static_framework_impl(ctx):
             fail("Unhandled platform '{}'".format(platform))
 
         swift_info = archive[SwiftInfo]
+
+        # We can potentially simplify this if this change lands upstream:
+        # https://github.com/bazelbuild/rules_swift/issues/291
+        objc_headers = archive[apple_common.Objc].header.to_list()
+        objc_header = objc_headers[-1]
```
kastiglione: I don't know there's a guarantee that the generated header will be last.

kastiglione: Perhaps filter for `-Swift.h`, optionally assert that there's only one, and take that header.

rebello95: Sounds good, I'll add that in a separate PR shortly.

rebello95: https://github.com/lyft/envoy-mobile/pull/374

PR #375, opened 2019-08-22
ci: add timeouts

junr03:
Description: add tighter timeouts to CI than the default 360 mins.
Risk Level: low

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -6,6 +6,7 @@ jobs:
   linuxdist:
     name: linux_dist
     runs-on: ubuntu-18.04
+    timeout-minutes: 45
     steps:
       - uses: actions/checkout@v1
         with:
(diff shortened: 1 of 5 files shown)


junr03:
@buildbreaker all of our current PRs are running for forever on the android jobs because logcat is not grepping for liveliness in the apps, given that they are crashing because of the config problem we are discussing.

They will run for 360 minutes (the default). This PR makes sure that in the future we more tightly kill jobs like that.

rebello95:
> @buildbreaker all of our current PRs are running for forever on the android jobs because logcat is not grepping for liveliness in the apps, given that they are crashing because of the config problem we are discussing.

Should we also fix this?

junr03:
> > @buildbreaker all of our current PRs are running for forever on the android jobs because logcat is not grepping for liveliness in the apps, given that they are crashing because of the config problem we are discussing.
> 
> Should we also fix this?

😂 of course! I have a stop-gap, and @buildbreaker can fix long term.

PR #376, opened 2019-08-22
java: call Jni.load() before accessing the configuration template

buildbreaker:
This is the fast solution for this problem

A few options to really root fix here:
1. We hide all `native` methods behind another method which mirrors the current `JniLibrary` interfaces but calls the static `load()` method prior to using any methods (super safe but removes some of the control from when we load the native lib)
2. Only call the `templateString()` method *after* the initialization of the `EnvoyEngine` (feels kind of brittle still)
3. (@goaway's suggestion) Move the configuration resolution down to the Engine layer since that is a lower level detail

Description: java: call Jni.load() before accessing the configuration template
Risk Level: low
Testing: local
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


junr03:
I am going to merge this temp solution to get master back to green. We can go with the long term approach when @buildbreaker is back on monday.

PR #385, opened 2019-08-26
android: move common yaml configuration down to java layer

buildbreaker:
A small structural change to the configuration and builder relationship since we are unable to read the default configuration provided by the native layer without initializing the JNI library.

The change here is to move most of the configuration resolution down to the `Engine` layer since it is responsible for the JNI library interaction.

Description: move common yaml configuration down to java layer
Risk Level: low
Testing: unit
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -1,11 +1,38 @@
...
+
+  public EnvoyConfiguration(String configYAML, int connectTimeoutSeconds, int dnsRefreshSeconds,
+                            int statsFlushSeconds) {
+    this.configYAML = configYAML;
+    this.connectTimeoutSeconds = connectTimeoutSeconds;
+    this.dnsRefreshSeconds = dnsRefreshSeconds;
+    this.statsFlushSeconds = statsFlushSeconds;
+  }
+
+  public String resolve(String defaultConfigurationYAML) {
```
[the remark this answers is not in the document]
rebello95: +1 for making this private/internal.

Additionally, `resolveTemplate(String templateYAML)` would make sense IMO.

One other thing that I think we should do here is not check `configYAML` within this function, and instead have it simply resolve the `templateYAML` it's passed. The caller can then be responsible for determining whether to use a pre-resolved YAML string or if it needs to resolve a configuration against a template. This would reduce the amount of magic done in this function, especially if we take the approach suggested above which treats user-specified configurations as pre-resolved (not resolving them as templates)

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -1,11 +1,38 @@
...
-   */
-  String templateString();
+public class EnvoyConfiguration {
+
+  public final String configYAML;
+  public final int connectTimeoutSeconds;
+  public final int dnsRefreshSeconds;
+  public final int statsFlushSeconds;
+
+  public EnvoyConfiguration(String configYAML, int connectTimeoutSeconds, int dnsRefreshSeconds,
```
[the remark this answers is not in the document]
rebello95: Talked offline with @buildbreaker about this and I agree with the 2-initializer approach. A couple other things to note though:

- The initializer that takes the `configYAML` (pre-resolved) will not be used today, but will act as a convenience
- The motivation for this PR/change is to defer calling into the JNI to get the template string until after the JNI is loaded (i.e., when `resolve` is called). Can we inline some documentation for this here @buildbreaker?

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyBuilder.kt
```diff
@@ -80,7 +71,7 @@ open class EnvoyBuilder internal constructor(
    * @return A new instance of Envoy.
    */
   fun build(): Envoy {
-    return Envoy(engineType(), resolvedYAML(), logLevel)
+    return Envoy(engineType(), EnvoyConfiguration(configYAML, connectTimeoutSeconds, dnsRefreshSeconds, statsFlushSeconds), logLevel)
```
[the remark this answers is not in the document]
buildbreaker: I don't quite understand what you have in mind here. Do you mean to have a way for us to create an `EnvoyConfiguration` with the appropriate values from the builder: initialize `EnvoyConfiguration` with a custom configuration versus with the `connectTimeoutSeconds, dnsRefreshSeconds, statsFlushSeconds` values?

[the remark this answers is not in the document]
rebello95: Discussed further offline and we will:
- Have `EnvoyConfiguration` _only_ be used for template-based configurations and avoid conflating custom YAML configs with the template resolution
- Add a constructor to Envoy Engine so we have one that takes an `EnvoyConfiguration` (which resolves properties to the default template) _or_ a custom string YAML file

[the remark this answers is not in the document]
goaway: ^ SGTM

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -1,11 +1,64 @@
...
+  public final Integer connectTimeoutSeconds;
+  public final Integer dnsRefreshSeconds;
+  public final Integer statsFlushSeconds;
+
+  /**
+   * Create an EnvoyConfiguration with a user provided configuration YAML.
+   *
+   * @param configYAML configuration to be used for starting Envoy.
+   */
+  public EnvoyConfiguration(String configYAML) {
```
rebello95: This should be removed since the string should no longer be stored within the config

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -1,11 +1,64 @@
 package io.envoyproxy.envoymobile.engine;
 
-public interface EnvoyConfiguration {
+public class EnvoyConfiguration {
+
+  public final String configYAML;
```
rebello95: This property can be removed

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -1,11 +1,50 @@
...
-  String templateString();
+  public EnvoyConfiguration(int connectTimeoutSeconds, int dnsRefreshSeconds,
+                            int statsFlushSeconds) {
+    this.connectTimeoutSeconds = connectTimeoutSeconds;
+    this.dnsRefreshSeconds = dnsRefreshSeconds;
+    this.statsFlushSeconds = statsFlushSeconds;
+  }
+
+  /**
+   * Resolves the provided default configuration with the field values:
```
rebello95: I'd probably remove the list of values below and simply say `Resolves the provided configuration template using properties on this configuration.`, that way we don't need to update this docstring each time we add a config property

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -1,11 +1,50 @@
...
+   * - connectTimeoutSeconds
+   * - dnsRefreshSeconds
+   * - statsFlushSeconds
+   * This default configuration is provided by the native layer.
+   *
+   * @param defaultConfigurationYAML the default template configuration.
+   * @return String, the resolved template.
+   * @throws ConfigurationException, when the template provided is not fully resolved.
+   */
+  String resolveTemplate(String defaultConfigurationYAML) {
```
rebello95: I think @goaway had suggested this be named `templateYAML`

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyEngine.java
```diff
@@ -12,19 +12,19 @@
   EnvoyHTTPStream startStream(EnvoyObserver observer);
 
   /**
-   * Run the Envoy engine with the provided config and log level.
+   * Run the Envoy engine with the provided envoyConfiguration and log level.
```
rebello95: This function takes a string, not a configuration

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyEngineImpl.java
```diff
@@ -24,30 +24,30 @@ public EnvoyHTTPStream startStream(EnvoyObserver observer) {
   }
 
   /**
-   * Run the Envoy engine with the provided config and log level.
+   * Run the Envoy engine with the provided envoyConfiguration and log level.
```
rebello95: Same here

library/java/test/io/envoyproxy/envoymobile/engine/EnvoyConfigurationTest.kt
```diff
@@ -0,0 +1,33 @@
...
+class EnvoyConfigurationTest {
+
+  @Test
+  fun `resolving with default configuration resolves with values`() {
+    val envoyConfiguration = EnvoyConfiguration(123, 234, 345)
+
+    val resolvedTemplate = envoyConfiguration.resolveTemplate(TEST_CONFIG)
+    assertThat(resolvedTemplate).contains("connect_timeout: 123s")
+    assertThat(resolvedTemplate).contains("dns_refresh_rate: 234s")
+    assertThat(resolvedTemplate).contains("stats_flush_interval: 345s")
```
rebello95: Maybe also assert that the template doesn't contain any `{{`?

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyBuilder.kt
```diff
@@ -80,7 +71,12 @@ open class EnvoyBuilder internal constructor(
    * @return A new instance of Envoy.
    */
   fun build(): Envoy {
-    return Envoy(engineType(), resolvedYAML(), logLevel)
+    val configurationYAML = configYAML
```
rebello95: Not sure this really provides any benefit 🤷‍♂

buildbreaker: It's a variable and the class isn't thread safe so the field could be changed between the lines of code. This way at the start of the method call we are guaranteed to have the correct configuration yaml...

Basically, I get it's silly but the IDE was complaining to me so I changed it :(

library/java/test/io/envoyproxy/envoymobile/engine/EnvoyConfigurationTest.kt
buildbreaker: The behavior that is baked in the method is that if it contains any unresolved templates, we'll throw an exception so that test is: `resolve templates with invalid templates will throw on build`

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -1,11 +1,46 @@
...
+
+  /**
+   * Resolves the provided configuration template using properties on this configuration.
+   * This default configuration is provided by the native layer.
+   *
+   * @param templateYAML the default template configuration.
+   * @return String, the resolved template.
+   * @throws ConfigurationException, when the template provided is not fully resolved.
+   */
+  String resolveTemplate(String templateYAML) {
```
goaway: Thoughts on moving the logic in this function into `EnvoyConfigurationImpl` (could be a helper method called out of `runWithConfig`)? That way the configuration really doesn't need to know or care about any implementation details of how its attributes are used.

buildbreaker: I can create another file which holds this method and just port the unit tests there -- it'll make this class a pure data class then. Wdyt?

cc: @rebello95

goaway: Sure, I'm good with that

buildbreaker: Actually, let's revisit at a later date on this 😬

PR #388, opened 2019-08-27
upstream: update envoy submodule to tip of master

junr03:
Description: taking in the tip of master
Risk Level: low
Testing: CI and local for iOS


junr03:
The size of the perf size binary grew by 4% compared to master. I will need to investigate what happened.

junr03:
will close for now until I have time to investigate the size differential

junr03:
I am going to merge this regardless of size increase -- as we think this is acceptable for the 0.2 release. However, I have cut https://github.com/lyft/envoy-mobile/issues/447 to track.

junr03:
https://github.com/envoyproxy/envoy/commit/d0e1db6b2ea424a228ac83585f80bf7c2ad2dcc4 broke relative paths for the python format checker. However, we don't need to run the python format checker in this project.

PR #394, opened 2019-08-29
java: data passthrough

goaway:
Description: this PR adds functionality to send data to HTTP streams and receive data from callbacks for Android.
Risk Level: med - adds core functionality.
Testing: tested locally with the kotlin app and an echo endpoint. Data was seen on both the send and the received direction.

### DIFF
--- a/library/common/jni_interface.cc
+++ b/library/common/jni_interface.cc
@@ -108,13 +108,17 @@ static void pass_headers(JNIEnv* env, envoy_headers headers, jobject j_context) 
 }
 
 // Platform callback implementation
-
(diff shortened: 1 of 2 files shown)


library/common/jni_interface.cc
```diff
@@ -158,6 +180,22 @@ static void jvm_on_complete(void* context) {
 }
 
 // Utility functions
+static void jni_delete_global_ref(void* context) {
```
buildbreaker: Would it be better to move these utility functions to a different file?

junr03: Yeah, we can start cleaning this up in subsequent PRs

PR #395, opened 2019-08-30
android: make response handler arguments non-optional

junr03:
Description: these fields should not be optional. The callbacks _must_ send up that data.
Risk Level: low
Testing: tested with e2e demo


rebello95:
> Let's double check swift side too

Yep Swift is non-optional

PR #401, opened 2019-08-30
core: malloc protection for c_types

junr03:
Description: this PR introduces `safe_malloc` which asserts non null returns from malloc. Additionally it converts all uses of malloc in the library to safe_malloc.
Risk Level: med - change memory treatment of c_types.


library/common/buffer/utility.cc
```diff
@@ -17,10 +17,13 @@ Buffer::InstancePtr toInternalData(envoy_data data) {
   return buf;
 }
 
-envoy_data toBridgeData(Buffer::Instance& data) {
+absl::optional<envoy_data> toBridgeData(Buffer::Instance& data) {
   envoy_data bridge_data;
   bridge_data.length = data.length();
   bridge_data.bytes = static_cast<uint8_t*>(malloc(sizeof(uint8_t) * bridge_data.length));
```
goaway: We need to guard against data length 0, for which malloc might still return nullptr (but this would be valid).

library/common/types/c_types.h
```diff
@@ -50,7 +50,7 @@ void envoy_noop_release(void* context);
  * Holds raw binary data as an array of bytes.
  */
 typedef struct {
-  size_t length;
+  int length;
```
goaway: I'd make this `size_t`, but I'd store the actual length - when we see that `bytes` is nullptr, but `length` is non-zero, we know an error occurred.

library/common/types/c_types.h
```diff
@@ -113,6 +113,12 @@ envoy_data copy_envoy_data(size_t length, const uint8_t* src_bytes);
 // For example when sending a headers-only request.
 extern const envoy_data envoy_nodata;
 
+// Convenience constant to return when data allocation fails.
+extern const envoy_data envoy_nulldata;
+
+// Convenience constant to return when data allocation fails.
+extern const envoy_headers envoy_nullheaders;
+
```
goaway: And so I think these can be removed (since we'll use structs with a non-zero length but with nullptr bytes).

library/common/types/c_types.h
goaway: We can put a comment on the utility functions' docstrings clarifying this.

test/common/buffer/utility_test.cc
```diff
@@ -30,16 +30,19 @@ TEST(DataConstructorTest, FromCToCpp) {
 TEST(DataConstructorTest, FromCppToCEmpty) {
   OwnedImpl empty_data;
 
-  envoy_data c_data = Utility::toBridgeData(empty_data);
+  absl::optional<envoy_data> maybe_c_data = Utility::toBridgeData(empty_data);
```
goaway: I'd suggest moving from `optional` to hard checks. If we see ourselves making a lot of these, we could introduce a c utility function `is_valid` (though I don't know whether this is warranted).

library/common/buffer/utility.cc
junr03: done here and everywhere else we malloc for the bridge types in common code.

junr03:
@goaway this is updated!

goaway:
This is great, thanks @junr03 !

goaway:
I know I approved, but @mattklein123 has some pretty strong feelings about crashing when we detect this (and to be fair, any app that's failing to allocate memory may not be likely to recover).

My philosophy was that if the app is going to crash, let it be the app that crashes and not have the library be the one that does it, but then again, it's not like there's a near term plan to safeguard the _rest_ of envoy from malloc failures, so...

An option Matt suggested was to use a safe_malloc macro that just asserts we get something other than `NULL` (iff `size > 0`). This seems reasonable, too.

junr03:
@buildbreaker @rebello95 I would like to hear what you all have to say w.r.t this, in particular to convention in the Lyft app.

rebello95:
> @buildbreaker @rebello95 I would like to hear what you all have to say w.r.t this, in particular to convention in the Lyft app.

As a library, I don't think we should be crashing for any reason. If possible, we should bubble errors up to the end consumer (or swallow them if they're not actually problematic).

One reason for this is that crashes in mobile apps are more difficult to fix in a timely manner than a crash on a server due to the fact that a release cadence is followed, and shipping new releases is much more time consuming than deploying (plus users actually need to update from the broken version).

Passing errors through rather than crashing would also allow consumers to implement their own error handling that may make sense for their app. For example, if we decide to ship an experiment with Envoy versus another networking library within our app, we can fall back to the old networking stack if we encounter an error with Envoy, whereas a crash is harder to recover from.

buildbreaker:
I feel like I don't have enough of an understanding for this malloc protection change and how expansive it is. Does this only happen when the application using Envoy Mobile is unable to allocate more memory which will cause a pointer to be null? If this the only root issue, I feel like we can probably handle it and avoid exposing this issue to the user?

Generally for errors, I agree with @rebello95 that we should expose some way for the user to handle the errors (like using the response handler to pass errors)

That said, it feels to me like this is more of a responsibility question and how we want to handle **known errors** as a library: do we expose this malloc issue because we can expect our consumers to handle this error or do we handle this because we can resolve this within the library because it's inconsequential to our users?

My rationale (using the experimentation example) is that we might not want to fallback to another networking stack if the error that occurred is inconsequential  and we want to self recover at the library layer

mattklein123:
Handling OOM conditions is a special kind of error that I strongly think should be made fatal, since almost no caller handles it properly. With that said, if we decide to handle OOM errors, I strongly recommend hiding the allocator behind an interface that can be mocked so that you can have tests for all OOM conditions (extremely painful). If you don't want to do this now, I recommend using SAFE_MALLOC with a crash in the meantime and revisiting this later.

library/common/types/BUILD
```diff
@@ -1,13 +1,20 @@
...
-cc_library(
+load("@envoy//bazel:envoy_build_system.bzl", "envoy_cc_library", "envoy_package")
+
+envoy_package()
+
+envoy_cc_library(
     name = "c_types_lib",
     srcs = [
-        "c_types.c",
+        "c_types.cc",
```
goaway: Let's revert his and...

library/common/types/c_types.cc
```diff
@@ -1,6 +1,15 @@
...
 #include "library/common/types/c_types.h"
 
-#include <string.h>
+#include <string>
+
+#include "common/common/assert.h"
+
+void* safe_malloc(size_t size) {
+  void* ptr = malloc(size);
+  RELEASE_ASSERT(ptr != nullptr, "malloc failure");
```
goaway: ...use `exit(EXIT_FAILURE)`

goaway: Also you need to check that `size > 0` before crashing.

junr03: Ah yes, good call. I keep forgetting that is a possibility re: `size > 0` check.

junr03:
@goaway I think I feel ok about converting the implementation file to a .cc file given the functionality we get for free. The header is still pure C.

PR #402, opened 2019-08-30
core: partially address race with native retain on stream state

goaway:
Description: Our memory model requires a native retain on stream state that must be available for the duration of the lifecycle of the native stream. This partially addresses a race where the retain might not be in place before the stream is reset - resulting in a leak. In order to fully address this, the retain must be completed such that it prevents re-ordering and is visible to other threads.

### DIFF
--- a/library/objective-c/EnvoyHTTPStreamImpl.m
+++ b/library/objective-c/EnvoyHTTPStreamImpl.m
@@ -189,6 +189,7 @@ static void ios_on_error(envoy_error error, void *context) {
 
   // We need create the native-held strong ref on this stream before we call start_stream because
   // start_stream could result in a reset that would release the native ref.
+  // TODO: To be truly safe we probably need stronger guarantees of operation ordering on this ref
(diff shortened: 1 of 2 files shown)


goaway:
rebased

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -187,12 +187,15 @@ - (instancetype)initWithHandle:(uint64_t)handle observer:(EnvoyObserver *)observ
                                ios_on_error,   ios_on_complete, context};
   _nativeObserver = native_obs;
 
+  // We need create the native-held strong ref on this stream before we call start_stream because
+  // start_stream could result in a reset that would release the native ref.
+  _strongSelf = self;
```
junr03: Can we leave a TODO here to address: 

> In order to fully address this, the retain must be completed such that it prevents re-ordering and is visible to other threads.

PR #403, opened 2019-09-03
bazel: Add avoid_deps to swift_static_framework

keith:
This field is analogous to `ios_framework`'s `avoid_deps`. It makes sure
that specifically mentioned static libraries are not included in the
final archive. This is required for us to produce a static archive that
doesn't include BoringSSL, since the Lyft app already includes BoringSSL
as part of gRPC, and we need to link that one instead.

Fixes https://github.com/lyft/envoy-mobile/issues/200

### DIFF
--- a/bazel/swift_static_framework.bzl
+++ b/bazel/swift_static_framework.bzl
@@ -69,8 +69,12 @@ def _swift_static_framework_impl(ctx):
 
         libraries = archive[CcInfo].linking_context.libraries_to_link
         archives = []
+        avoid_labels = [target.label for target in ctx.attr.avoid_deps]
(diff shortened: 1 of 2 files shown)


keith:
This fails for the test projects because they need to link BoringSSL somehow. Maybe this should be a separate target or config we build

keith:
this is still needed once you figure out how you want to setup the builds

rebello95:
This was originally being done with the intent of supporting compiling SwiftGRPC (+ BoringSSL) and Envoy Mobile into the same app, but I don't think this is something we need to support at the moment anymore. Thoughts?

keith:
Sounds fine

PR #406, opened 2019-09-03
cleanup: rename instances of 'observer' to 'callbacks'

goaway:
Description: I've come to be of the opinion that 'observer' is an unfortunate and confusing name for the objects containing our callbacks:

- In Envoy itself these are called callbacks.
- In our filter proposal we're calling these callbacks.
- These entities have deviated from the reactive pattern from which they take their namesake.
Risk Level: Low
Testing: CI

### DIFF
--- a/test/common/http/dispatcher_test.cc
+++ b/test/common/http/dispatcher_test.cc
@@ -59,23 +59,23 @@ public:
   Http::ContextImpl http_context_;
   AsyncClientImpl client_;
   Dispatcher http_dispatcher_;
-  envoy_observer observer_;
+  envoy_http_callbacks callbacks_;
 };
 
 TEST_F(DispatcherTest, BasicStreamHeadersOnly) {
   envoy_stream_t stream = 1;
-  // Setup observer to handle the response headers.
-  envoy_observer observer;
+  // Setup callbacks to handle the response headers.
+  envoy_http_callbacks callbacks;
   callbacks_called cc = {0, 0, 0, 0};
-  observer.context = &cc;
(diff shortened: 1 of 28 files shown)


goaway:
This is really a nit for myself, since I named it "Observer" in the first place. But the discrepancy from Envoy has been bothering me, and the deviation from the pure Reactive pattern has also been bothering me. I think this is clearer. I also think it will only get harder to change as time goes on.

test/common/http/dispatcher_test.cc
```diff
@@ -127,32 +127,32 @@ TEST_F(DispatcherTest, BasicStreamHeadersOnly) {
                      .counter("internal.upstream_rq_200")
                      .value());
 
-  // Ensure that the callbacks on the observer were called.
+  // Ensure that the callbacks on the callbacks were called.
```
junr03: nit: this sentence is a bit weird now. Do you mind updating here and everywhere in the file?

test/common/http/dispatcher_test.cc
```diff
@@ -59,23 +59,23 @@ class DispatcherTest : public testing::Test {
   Http::ContextImpl http_context_;
   AsyncClientImpl client_;
   Dispatcher http_dispatcher_;
-  envoy_observer observer_;
+  envoy_http_callbacks callbacks_;
```
junr03: this name collides with the MockAsyncClientCallbacks: https://github.com/lyft/envoy-mobile/pull/406/files#diff-04eab2469b3cc12377df7e5b0015914aR49

PR #407, opened 2019-09-04
ios: bubble up error to platform

junr03:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


library/swift/src/ResponseHandler.swift
```diff
@@ -63,7 +63,7 @@ public final class ResponseHandler: NSObject {
   /// - parameter closure: Closure which will be called when an error occurs.
   @discardableResult
   public func onError(_ closure:
-    @escaping () -> Void)
+    @escaping (_ error: EnvoyError) -> Void)
```
junr03: having this type here is causing swift compilation to segfault. I think I am doing things correctly in objc with the interface, implementation, and including the new EnvoyError.m file in the BUILD file. 

@rebello95 @goaway I would appreciate your eyes on this, I am sure I am missing something silly

library/objective-c/EnvoyError.m
```diff
@@ -0,0 +1,13 @@
+#import "library/objective-c/EnvoyEngine.h"
+
+@implementation EnvoyError {
+  uint64_t _errorCode;
```
rebello95: Shouldn't these be publicly visible in the header file?

library/swift/src/ResponseHandler.swift
rebello95: This looks correct to me - we previously saw some segfaults due to issues with the Swift compiler rather than actual bugs in our code, so this could be the same thing. Do you hit the same thing if you try passing an `NSError` instead of an `EnvoyError`?

junr03:
closing this in favor of #410

PR #411, opened 2019-09-05
android: wire up cancelation

junr03:
Description: cancelation was not wired up in Android. This PR implements and wires up stream cancelation.
Risk Level: med - core functionality for android
Testing: local build


library/java/src/io/envoyproxy/envoymobile/engine/EnvoyHTTPStream.java
```diff
@@ -62,22 +62,29 @@ public void sendTrailers(Map<String, List<String>> trailers) {
...
+    if (!callbacksContext.canceled.getAndSet(true)) {
+      // Step 2: directly fire the cancel callback.
+      callbacksContext.onCancel();
+      // Step 3: propagate the reset into native code.
+      JniLibrary.resetStream(streamHandle);
+      return 0;
+    } else {
+      return 1;
+    }
+  }
```
goaway: If we delegate this logic to `JvmCallbackContext`, the `AtomicBoolean` can remain private.

junr03:
@goaway, good call on not making the atomic boolean public. I delegated the logic back to the context.

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyStreamEmitter.kt
```diff
@@ -50,6 +50,7 @@ class EnvoyStreamEmitter(
    * @throws EnvoyException when there is an exception cancelling the stream.
    */
   override fun cancel() {
+    // TODO: surface result of operation.
```
goaway: tbh I don't know that we need to, since we don't/can't surface the synchronous results of any other of our other stream operations.

goaway: (with the current interfaces, anyways)

library/swift/src/EnvoyStreamEmitter.swift
```diff
@@ -26,6 +26,7 @@ extension EnvoyStreamEmitter: StreamEmitter {
   }
 
   func cancel() {
+    // TODO: surface result of operation.
```
goaway: think we can leave off the TODO (as above)

PR #412, opened 2019-09-05
jni: add support for on-heap buffers

goaway:

### DIFF
--- a/library/java/src/io/envoyproxy/envoymobile/engine/EnvoyHTTPStream.java
+++ b/library/java/src/io/envoyproxy/envoymobile/engine/EnvoyHTTPStream.java
@@ -36,6 +36,8 @@ public class EnvoyHTTPStream {
    *
    * @param data,      the data to send.
    * @param endStream, supplies whether this is the last data in the streamHandle.
+   * @throws UnsupportedOperationException - if the provided buffer is neither a direct ByteBuffer
+   *     nor backed by an on-heap byte array.
    */
   public void sendData(ByteBuffer data, boolean endStream) {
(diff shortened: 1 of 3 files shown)


goaway:
We could do it in native code, but it would unnecessarily add the overhead of the JNI call.

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyHTTPStream.java
```diff
@@ -38,7 +38,13 @@ public void sendHeaders(Map<String, List<String>> headers, boolean endStream) {
    * @param endStream, supplies whether this is the last data in the streamHandle.
    */
```
junr03: lets put a @throws here

library/java/src/io/envoyproxy/envoymobile/engine/JniLibrary.java
```diff
@@ -69,6 +69,17 @@ public static void load() {
...
+  /**
+   * Send data over an open HTTP stream. This method can be invoked multiple
+   * times.
+   *
+   * @param stream,    the stream to send data over.
+   * @param data,      the data to send.
+   * @param endStream, supplies whether this is the last data in the stream.
+   * @return int, the resulting status of the operation.
+   */
+  protected static native int sendData(long stream, byte[] data, boolean endStream);
```
buildbreaker: Should we update the other method's documentation to call out that we need to direct buffer?

buildbreaker:
I think it's fine to do this check in the java layer

PR #413, opened 2019-09-05
interface: use 'body' as parameter name for send

goaway:


goaway:
Rationale: `data` is a keyword in kotlin, and I think this maps slightly better to the method's purpose and HTTP naming conventions.

PR #415, opened 2019-09-05
core: clean up streams in http dispatcher on reset

junr03:
Description: clean up streams in the http dispatcher when the underlying stream fires an onReset callback. Previously we were leaking this state.
Risk Level: med - stream memory management changes. However, we have a guarantee from the underlying async client implementation that terminal callbacks (onReset, onComplete) will only fire once.
Testing: unit tests.

### DIFF
--- a/library/common/http/dispatcher.cc
+++ b/library/common/http/dispatcher.cc
@@ -37,20 +37,20 @@ void Dispatcher::DirectStreamCallbacks::onTrailers(HeaderMapPtr&& trailers) {
 void Dispatcher::DirectStreamCallbacks::onComplete() {
   ENVOY_LOG(debug, "[S{}] complete stream", stream_handle_);
   bridge_callbacks_.on_complete(bridge_callbacks_.context);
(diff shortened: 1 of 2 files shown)


library/common/http/dispatcher.cc
```diff
@@ -37,12 +37,21 @@ void Dispatcher::DirectStreamCallbacks::onTrailers(HeaderMapPtr&& trailers) {
 void Dispatcher::DirectStreamCallbacks::onComplete() {
   ENVOY_LOG(debug, "[S{}] complete stream", stream_handle_);
   bridge_callbacks_.on_complete(bridge_callbacks_.context);
+  // Very important: onComplete and onReset both cleanup stream state in the http dispatcher because
```
goaway: nit: s/cleanup/clean up/

goaway:
Looks good - let's add some tests here to cover this.

junr03:
@goaway I added tests here.

PR #416, opened 2019-09-06
WIP: implement gRPC

rebello95:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/library/swift/src/GRPC/GRPCResponseHandler.swift
+++ b/library/swift/src/GRPC/GRPCResponseHandler.swift
@@ -1,0 +1,175 @@
+import Envoy
+import Foundation
+
+/// <#function description#>
+@objcMembers
+public final class GRPCResponseHandler: NSObject {
+    private var onHeaders: ((_ headers: [String: [String]], _ grpcStatus: Int?) -> Void)?
+    private var onMessage: ((_ data: Data) -> Void)?
+    private var onComplete: ((_ trailers: [String: [String]]) -> Void)?
+    private var state = State.expectingCompressionFlag
+    private var buffer: Data!
+
+    /// Underlying response handler which should be called with response data.
+    let underlyingHandler = ResponseHandler()
+
+    private enum State {
+        case expectingCompressionFlag
+        case expectingMessageLength
+        case expectingMessage(length: UInt32)
+    }
+
+    /// <#function description#>
+    ///
+    /// - parameter closure: <#closure description#>
+    ///
+    /// - returns: <#return value description#>
+    @discardableResult
+    public func onHeaders(_ closure:
+        @escaping (_ headers: [String: [String]], _ grpcStatus: Int?) -> Void)
+        -> GRPCResponseHandler
+    {
+        self.onHeaders = closure
+        return self
+    }
+
+    /// <#function description#>
+    ///
+    /// - parameter closure: <#closure description#>
+    ///
+    /// - returns: <#return value description#>
+    @discardableResult
+    public func onMessage(_ closure:
+        @escaping (_ data: Data) -> Void)
+        -> GRPCResponseHandler
+    {
+        self.onMessage = closure
+        return self
+    }
+
+    /// <#function description#>
+    ///
+    /// - parameter closure: <#closure description#>
+    ///
+    /// - returns: <#return value description#>
+    @discardableResult
+    public func onComplete(_ closure:
+        @escaping (_ trailers: [String: [String]]) -> Void)
+        -> GRPCResponseHandler
+    {
+        self.onComplete = closure
+        return self
+    }
+
+    // MARK: - Internal
+
+    override init() {
+        super.init()
+        self.underlyingHandler
+            .onHeaders { [weak self] headers, _, endStream in
+                self?.handleHeaders(headers)
+                if endStream {
+                    self?.handleCompletion(trailers: [:])
+                }
+            }
+            .onData { [weak self] data, endStream in
+                self?.handleData(data)
+                if endStream {
+                    self?.handleCompletion(trailers: [:])
+                }
+            }
+            .onTrailers { [weak self] trailers in
+                self?.handleCompletion(trailers: trailers)
+            }
+            .onError { [weak self] _ in
+                // TODO: Switch to call a different handler with the error
+                self?.handleCompletion(trailers: [:])
+            }
+    }
+
+    // MARK: - Private
+
+    private func handleHeaders(_ headers: [String: [String]]) {
+        let grpcStatus = headers["grpc-status"]?
+            .first
+            .flatMap(Int.init)
+        self.onHeaders?(headers, grpcStatus)
+    }
+
+    private func handleData(_ data: Data) {
+        if data.isEmpty {
+            return
+        }
+
+        if var existingBuffer = self.buffer {
+            existingBuffer.append(contentsOf: data)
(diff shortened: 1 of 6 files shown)


library/swift/src/GRPC/GRPCResponseHandler.swift
```diff
@@ -0,0 +1,175 @@
...
+            self.buffer.removeFirst(lengthAsInt)
+            self.state = .expectingCompressionFlag
+        }
+
+        self.processBuffer()
+    }
+}
+
+private extension Data {
+    func integer<T: FixedWidthInteger>(at index: Int) -> T? {
```
Reflejo: Not sure how "final" this is, but there are better ways of doing this unsafe cast.

rebello95: Updated, let me know if you had something else in mind 😃

library/swift/src/GRPCRequestBuilder.swift
```diff
@@ -0,0 +1,79 @@
+import Foundation
+
+/// Builder used for creating new gRPC `Request` instances.
+@objcMembers
+public final class GRPCRequestBuilder: NSObject {
+  private let underlyingBuilder: RequestBuilder
+
+  /// Initialize a new builder.
+  ///
+  /// - parameter path:      Path of the RPC (i.e., `/pb.api.v1.Foo/GetBar`).
```
buildbreaker: Should we rename this to rpcPath?

library/swift/test/GRPCRequestBuilderTests.swift
```diff
@@ -0,0 +1,46 @@
+@testable import Envoy
+import Foundation
+import XCTest
+
+final class GRPCRequestBuilderTests: XCTestCase {
+  func testUsesCorrectAuthorityWithHTTPS() {
+    let request = GRPCRequestBuilder(path: "/foo/bar", authority: "foo.bar.com",
```
buildbreaker:
```suggestion
    let request = GRPCRequestBuilder(path: "/pb.api.v1.Foo/GetBar", authority: "foo.bar.com",
```
?

library/swift/src/GRPCRequestBuilder.swift
rebello95: I think it's implied by `GRPC`, what do you think?

library/swift/test/GRPCRequestBuilderTests.swift
rebello95: Sure

library/swift/src/GRPCRequestBuilder.swift
buildbreaker: Probably fine

library/swift/src/GRPCClient.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+/// Client that supports sending and receiving gRPC traffic.
+@objcMembers
+public final class GRPCClient: NSObject {
+  private let client: Client
```
goaway: Not directly related to this PR but per offline discussion, suggest renaming `Client` to something a little more descriptive, e.g. `HTTPClient`.

rebello95: Yep, good callout. Updating here: https://github.com/lyft/envoy-mobile/pull/425

library/swift/src/Data+Extension.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+extension Data {
+  /// Returns the next integer in the data, using the size of `T`. Nil if the data is too small.
+  ///
+  /// - returns: The next integer in the data, or nil.
+  func nextInteger<T: FixedWidthInteger>() -> T? {
```
goaway: The docstring and function name sort of lead me to infer that calling this multiple times on a `Data` object should return subsequent integers (with some sort of implicit internal index).

library/swift/src/GRPCRequestBuilder.swift
```diff
@@ -0,0 +1,79 @@
+import Foundation
+
+/// Builder used for creating new gRPC `Request` instances.
+@objcMembers
+public final class GRPCRequestBuilder: NSObject {
```
goaway: This class seems to almost exclusively delegate to `RequestBuilder`. Would it be worth either extending `RequestBuilder` (either via subclassing or class extension)?

library/swift/src/GRPCResponseHandler.swift
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  /// Specify a callback for when response headers are received by the stream.
+  /// If `endStream` is `true`, the stream is complete.
+  ///
+  /// - parameter closure: Closure which will receive the headers, gRPC status,
+  ///                      and flag indicating if the stream is headers-only.
+  @discardableResult
+  public func onHeaders(_ closure:
+    @escaping (_ headers: [String: [String]], _ grpcStatus: Int, _ endStream: Bool) -> Void)
```
goaway: Since GRPC is guaranteed to send trailers with every response, you maybe don't need any `endStream` booleans on the callbacks.

library/swift/src/GRPCStreamEmitter.swift
```diff
@@ -0,0 +1,46 @@
...
+    // Compression flag (1 byte) - 0, not compressed
+    lengthPrefixedData.append(0)
+
+    // Message length (4 bytes)
+    var length = UInt32(messageData.count).bigEndian
+    lengthPrefixedData.append(UnsafeBufferPointer(start: &length, count: 1))
+
+    // Message data
+    lengthPrefixedData.append(contentsOf: messageData)
+    self.underlyingEmitter.sendData(lengthPrefixedData)
```
goaway: Rather than creating a new buffer, I'd suggest simply calling `sendData` once with the prefix, and a second time with the `messageData`.

library/swift/src/Data+Extension.swift
rebello95: Updating to `integer(atIndex:)` 👍

library/swift/src/GRPCRequestBuilder.swift
rebello95: I had considered doing this, but the `GRPCRequestBuilder` actually exposes less than the existing `RequestBuilder`. I'd prefer to keep this minimal rather than adding gRPC-specific functionality to a more generic HTTP request builder

library/swift/src/GRPCStreamEmitter.swift
rebello95: Sure

library/swift/src/GRPCResponseHandler.swift
rebello95: Nice. Based on gRPC [docs](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md):

> Response-Headers & Trailers-Only are each delivered in a single HTTP2 HEADERS frame block. Most responses are expected to have both headers and trailers but Trailers-Only is permitted for calls that produce an immediate error. Status must be sent in Trailers even if the status code is OK.

So we should be able to remove the `endStream` from `onMessage`

rebello95:
Thanks for the review and offline discussions, @goaway! Ready for another look.

PR #417, opened 2019-09-06
config: enable HTTP/2

rebello95:
Updates the standard configuration to send requests over HTTP/2 by default.

Note that there are future improvements that should be made here to support negotiating between HTTP/1.1 and HTTP/2, being tracked in https://github.com/lyft/envoy-mobile/issues/502.

More info on the config change: https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cds.proto#envoy-api-msg-cluster


rebello95:
@goaway is investigating this

junr03:
Drive by to update with the context I had that I missed to post here. 

S3 does not support h2. So when envoy sends h2 frames to S3, S3 sends back an http/1.1 frames and nghttp2 fails and errors in callbacks to the router which gets charged to the client as a 503.

rebello95:
Thanks @junr03. Additional details from what we've seen recently:
- Sending non-bodied requests (i.e., `GET`) over HTTP/2 works fine
- Sending requests with bodies (i.e., `POST`) fail, and it seems like upstream receives invalid bodies

rebello95:
PR is green now. To summarize findings from related investigations:
- S3 does not support HTTP/2, which is why our demo apps were failing. **Fix: Make a request to a sample `ping` endpoint which supports HTTP/2**
- gRPC expects client-side closes to be done using an empty data frame, not trailers. **Fix: https://github.com/lyft/envoy-mobile/pull/500**

PR #420, opened 2019-09-06
ios: move configuration to Objective-C

rebello95:
This is a mirror of the Android change made in https://github.com/lyft/envoy-mobile/pull/385.

The change was originally motivated by the fact that Android was unable to obtain the configuration template within C++ before the JNI was loaded. Moving the template resolution logic to Java fixed this issue by allowing the JNI to be initialized before loading the template.

Key changes in this PR:
- Moved template resolution to Objective-C
- Custom YAML files are no longer treated as templates, and are not resolved (but are passed through as-is). This matches Android
- Updated initializers as necessary to match Android

### DIFF
--- a/library/objective-c/EnvoyEngineImpl.m
+++ b/library/objective-c/EnvoyEngineImpl.m
@@ -16,15 +16,21 @@
   return self;
 }
 
-- (int)runWithConfig:(NSString *)config {
-  return [self runWithConfig:config logLevel:@"info"];
+- (int)runWithConfig:(EnvoyConfiguration *)config logLevel:(NSString *)logLevel {
+  NSString *templateYAML = [[NSString alloc] initWithUTF8String:config_template];
+  NSString *resolvedYAML = [config resolveTemplate:templateYAML];
+  if (resolvedYAML == nil) {
+    return 1;
+  }
+
+  return [self runWithConfigYAML:resolvedYAML logLevel:logLevel];
 }
 
-- (int)runWithConfig:(NSString *)config logLevel:(NSString *)logLevel {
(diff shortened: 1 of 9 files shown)


library/swift/src/Envoy.swift
```diff
@@ -16,32 +16,53 @@ public final class Envoy: NSObject {
...
 
   private final class RunnerThread: Thread {
     private let engine: EnvoyEngine
-    private let configYAML: String
+    private let config: ConfigurationType
     private let logLevel: LogLevel
 
-    init(configYAML: String, logLevel: LogLevel, engine: EnvoyEngine) {
-      self.configYAML = configYAML
+    enum ConfigurationType {
```
rebello95: This is essentially a `oneof` that ends up being cleaner than having two optional properties

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -16,15 +16,21 @@ - (instancetype)init {
   return self;
 }
 
-- (int)runWithConfig:(NSString *)config {
-  return [self runWithConfig:config logLevel:@"info"];
+- (int)runWithConfig:(EnvoyConfiguration *)config logLevel:(NSString *)logLevel {
+  NSString *templateYAML = [[NSString alloc] initWithUTF8String:config_template];
+  NSString *resolvedYAML = [config resolveTemplate:templateYAML];
+  if (resolvedYAML == nil) {
+    return 1;
```
buildbreaker: This is the "throwing" when the yaml resolution fails?

rebello95: More or less yeah

PR #428, opened 2019-09-11
core: queue API calls while underlying event dispatcher is not ready

### LINKED ISSUE #285: core: create post-init callback
This allows some setup to happen automatically, and would allow us to queue and then drain requests submitted during startup, rather than failing them.

The hack to work around this is to `sleep` longer than it takes Envoy to start up and then call `setup`, which is pretty unfortunate.

junr03:
Description: this PR adds a drainable queue to the http dispatcher that allows it to queue API calls while the underlying event dispatcher is not ready. It hooks into envoy's lifecycle notifier to flush the queue once envoy fires the post init lifecycle event.
Risk Level: med - changes startup behavior of the engine.
Testing: with example app and lyft alpha app. Will add unit tests in subsequent PR (#449)

Fixes #285 

Co-authored-by: Mike Schore mike.schore@gmail.com


junr03:
This PR requires https://github.com/envoyproxy/envoy/pulls to land before it.

library/common/main_interface.cc
```diff
@@ -99,9 +91,13 @@ envoy_status_t run_engine(const char* config, const char* log_level) {
   // https://github.com/lyft/envoy-mobile/issues/34
   try {
     main_common_ = std::make_unique<Envoy::MainCommon>(5, envoy_argv);
-    // TODO: this call should be done in a post init callback.
-    // related issue: https://github.com/lyft/envoy-mobile/issues/285.
-    setup_envoy();
+    // init_engine must have been called prior to calling run_engine or the creation of the envoy
+    // runner thread.
+    stageone_callback_handler_ = main_common_->server()->lifecycleNotifier().registerCallback(
```
goaway: I would add a comment here:

Technically speaking we're waiting longer than we might need to, to drain to the main thread's dispatcher. This is because we're not simply waiting for its availability and for it to have started, but also because we're waiting for clusters to have done their first attempt at DNS resolution. When we improve synchronous failure handling and/or move to dynamic forwarding, we only need to wait until the dispatcher is running (and can drain by enqueueing a drain callback on it, as we did previously).

library/common/http/dispatcher.h
```diff
@@ -131,16 +134,23 @@ class Dispatcher : public Logger::Loggable<Logger::Id::http> {
...
+   * @param callback, the functor to post.
+   */
+  void post(Event::PostCb callback);
   // Everything in the below interface must only be accessed from the event_dispatcher's thread.
   // This allows us to generally avoid synchronization.
   DirectStream* getStream(envoy_stream_t stream_handle);
   void cleanup(envoy_stream_t stream_handle);
 
+  // The initialization lock and queue, and the event_dispatcher are the only member state that may
+  // be accessed from a thread other than the event_dispatcher's own thread.
```
goaway: Maybe add `(And assertions are in place to validate this)`.

library/common/http/dispatcher.cc
```diff
@@ -51,16 +52,60 @@ Dispatcher::DirectStream::DirectStream(envoy_stream_t stream_handle,
...
+    // initialization_queue_ has been drained into the event_dispatcher_ to maintain ordering
+    // guarantees in the event_dispatcher_'s queue.
+    event_dispatcher_.store(&event_dispatcher);
+    cluster_manager_.store(&cluster_manager);
+  }
+}
+
+void Dispatcher::post(Event::PostCb callback) {
+  // If the event_dispatcher is present, dispatch the functor directly to it.
+  auto event_dispatcher = event_dispatcher_.load();
```
goaway: Reading the code we've written through again, I think we could benefit down the road from encapsulating this logic in an EventQueue class. (I think we could also simplify the synchronized state handling as well by doing so.)

goaway: EventQueue/CallbackQueue/InitQueue

library/common/http/dispatcher.cc
```diff
@@ -51,16 +52,60 @@ Dispatcher::DirectStream::DirectStream(envoy_stream_t stream_handle,
...
+void Dispatcher::ready(Event::Dispatcher& event_dispatcher,
+                       Upstream::ClusterManager& cluster_manager) {
+  {
+    Thread::LockGuard lock(initialization_lock_);
+
+    for (const Event::PostCb& cb : initialization_queue_) {
+      event_dispatcher.post(cb);
+    }
+
+    // The HTTP dispatcher's post function relies on this atomic pointer to signal transition
```
mattklein123: FWIW I don't think the complexity of the atomics here is worth it right now. Can you just hold the lock in both init and post and I think that will simplify things substantially?

goaway:
So I started to factor out the `init_queue_` logic, but it didn't really feel a whole lot simpler. I've decided to simply go with @mattklein123's suggestion for now. I have a slight concern that the increased dual-locking could lead to deadlock scenarios, but I don't think it's particularly likely call pattern.

Ultimately, I'd still much rather work towards a solution where we get an earlier handle on the event_dispatcher_ which will be safer and simpler.

library/common/http/dispatcher.cc
goaway: Went ahead with locking; see below.

library/common/http/dispatcher.cc
```diff
@@ -51,16 +52,50 @@ Dispatcher::DirectStream::DirectStream(envoy_stream_t stream_handle,
...
-                       Upstream::ClusterManager& cluster_manager)
-    : event_dispatcher_(event_dispatcher), cluster_manager_(cluster_manager) {}
+Dispatcher::Dispatcher() {
+  event_dispatcher_.store(nullptr);
+  cluster_manager_.store(nullptr);
+}
+
+void Dispatcher::ready(Event::Dispatcher& event_dispatcher,
+                       Upstream::ClusterManager& cluster_manager) {
+  {
```
mattklein123: nit: there is no point in having these extra braces as you hold the lock for the length of the function. Same below.

library/common/http/dispatcher.cc
```diff
@@ -51,16 +52,50 @@ Dispatcher::DirectStream::DirectStream(envoy_stream_t stream_handle,
     : stream_handle_(stream_handle), underlying_stream_(underlying_stream),
       callbacks_(std::move(callbacks)) {}
 
-Dispatcher::Dispatcher(Event::Dispatcher& event_dispatcher,
-                       Upstream::ClusterManager& cluster_manager)
-    : event_dispatcher_(event_dispatcher), cluster_manager_(cluster_manager) {}
+Dispatcher::Dispatcher() {
+  event_dispatcher_.store(nullptr);
```
mattklein123: I think you have a compile issue here.

library/common/http/dispatcher.h
```diff
@@ -22,7 +22,9 @@ namespace Http {
  */
 class Dispatcher : public Logger::Loggable<Logger::Id::http> {
 public:
-  Dispatcher(Event::Dispatcher& event_dispatcher, Upstream::ClusterManager& cluster_manager);
+  Dispatcher() {}
```
mattklein123: nit: del

junr03:
tested locally against the unit test suite, and with all 4 demo apps.

PR #440, opened 2019-09-13
android: move liveliness log to main

junr03:
Description: the liveliness CI tests for android are flaky because the response view holder doesn't log with high fidelity without UI interaction. This PR changes the logs to be printed in the main activity.
Risk Level: low
Testing: CI


buildbreaker:
thanks for fixing!

PR #442, opened 2019-09-16
core: introduce interface for setting preferred network

goaway:
Description: This is to allow OS/platform reachability signals to be consumed by the Envoy library engine.

### DIFF
--- a/library/common/main_interface.h
+++ b/library/common/main_interface.h
@@ -84,6 +84,14 @@ envoy_status_t reset_stream(envoy_stream_t stream);
 envoy_engine_t init_engine();
 
 /**
+ * Update the network interface to the preferred network for opening new connections.
+ * Note that this state is shared by all engines.
+ * @param network_interface, the interface to be preferred for new sockets.
+ * @return envoy_status_t, the resulting status of the operation.
+ */
+envoy_status_t set_preferred_network(envoy_network_t network_interface);
+
+/**
  * External entry point for library.
  * @param config, the configuration blob to run envoy with.
  * @param log_level, the logging level to run envoy with.
--- a/library/common/main_interface.h
+++ b/library/common/main_interface.h
@@ -84,7 +84,7 @@ envoy_status_t reset_stream(envoy_stream_t stream);
 envoy_engine_t init_engine();
 
 /**
- * Update the network interface to the preferred network for opening new connections.
+ * Update the network interface to the preferred network for opening new streams.
  * Note that this state is shared by all engines.
  * @param network_interface, the interface to be preferred for new sockets.
  * @return envoy_status_t, the resulting status of the operation.
--- a/library/common/types/c_types.h
+++ b/library/common/types/c_types.h
@@ -29,6 +29,11 @@ typedef enum { ENVOY_SUCCESS, ENVOY_FAILURE } envoy_status_t;
  */
 typedef enum { ENVOY_UNDEFINED_ERROR, ENVOY_STREAM_RESET } envoy_error_code_t;
 
+/**
+ * Networks by interface.
+ */
+typedef enum { ENVOY_WLAN, ENVOY_WWAN } envoy_network_t;
+
 #ifdef __cplusplus
 extern "C" { // release function
 #endif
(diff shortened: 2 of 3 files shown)


library/common/main_interface.h
```diff
@@ -83,6 +83,14 @@ envoy_status_t reset_stream(envoy_stream_t stream);
  */
 envoy_engine_t init_engine();
 
+/**
+ * Update the network interface to the preferred network for opening new connections.
```
rebello95: Is this correct? Shouldn't we be maintaining two _connections_ and switching between them when this function is called then piping future _requests_ over this connection?

library/common/types/c_types.h
```diff
@@ -29,6 +29,11 @@ typedef enum { ENVOY_SUCCESS, ENVOY_FAILURE } envoy_status_t;
  */
 typedef enum { ENVOY_UNDEFINED_ERROR, ENVOY_STREAM_RESET } envoy_error_code_t;
 
+/**
+ * Networks by interface.
+ */
+typedef enum { ENVOY_WLAN, ENVOY_WWAN } envoy_network_t;
```
rebello95: Can we add docstrings for these? Also, I don't have strong thoughts here, but do you think `wifi` and `cellular` would be more readable?

library/common/main_interface.h
junr03: Yes, I think new _streams_ is actually more accurate. Will update comment.

library/common/types/c_types.h
junr03: We debated about this for a while yesterday. I tend to agree that wifi and cellular are more readable, and easily understood by the reader. However, @goaway's point was that these actually represent correctness, albeit a bit pedantically. I am open to switching these to wifi and cellular.

goaway: I guess I lean a bit towards WLAN and WWAN.

Some considerations that push me in that direction:
- WLAN and WWAN is consistent and correct (technically WiFi networks are cellular local networks)
- This is a purely internal struct and users of our library will never see it
- Apple uses WWAN in their enum (but confusingly also uses WiFi; though WiFi is the only widely-adopted general WLAN technology, newer IoT standards probably can be classified as WLAN as well)
- If we tune connection configuration to the technology, WWAN and WLAN are probably the classes that we care about

But it's also not that big a deal. I do agree we should at least provide a docstring here explaining that WLAN is, generally-speaking, WiFi and that mobile phone networks are all WWAN.

library/common/main_interface.h
goaway: Yep, I like @rebello95's suggestion better than what we had previously.

library/common/types/c_types.h
rebello95: This is fine with me, I don't feel that strongly 😃

library/common/main_interface.h
```diff
@@ -83,6 +83,14 @@ envoy_status_t reset_stream(envoy_stream_t stream);
  */
 envoy_engine_t init_engine();
 
+/**
+ * Update the network interface to the preferred network for opening new streams.
+ * Note that this state is shared by all engines.
```
rebello95: Is this permanent? Seems like ideally we'd be able to set this per-engine (especially since we don't expose a singleton on the platform layer).

library/common/main_interface.h
```diff
@@ -83,6 +83,14 @@ envoy_status_t reset_stream(envoy_stream_t stream);
  */
 envoy_engine_t init_engine();
 
+/**
+ * Update the network interface to the preferred network for opening new streams.
```
rebello95: Does setting this also kill existing streams? I would assume so since on iOS they'll hang until they time out in the case of wifi <> cellular

library/common/main_interface.h
goaway: The presumption here is that the same networks/interfaces are available to any thread in the process. As long as that's true, there's no real inherent need for the state to be replicated.

I could see per-instance state being slightly more testable, maybe.

library/common/main_interface.h
goaway: We intentionally will not kill streams. Often (but not always) the OS will indicate a new preferred network not because the old one rather became completely unavailable, but rather because some heuristic indicated the other should be favored.

It's more common to have both some level of wifi connectivity and cellular during the switchover (and of course I type "wifi" and "cellular", ah well). Keeping streams open allows requests--that may still be on perfectly healthy connections--to complete.

In the future we may consider re-tuning idle timeouts to be much more aggressive for a non-favored network, but this will require some Envoy changes.

PR #443, opened 2019-09-16
core: add network-specific clusters to support OS preference

goaway:
Description: Consumes OS network preference signals to classify connections based on when they're created. New streams will be opened on connections grouped under the current network indication. Note that sockets are still opened by syscalls and so it is simply assumed they are attached to the appropriate network interface.
Risk Level: Moderate
Testing: CI and local testing

### DIFF
--- a/library/common/config_template.cc
+++ b/library/common/config_template.cc
@@ -34,7 +34,7 @@ R"(
     dns_lookup_family: V4_ONLY
     lb_policy: ROUND_ROBIN
     load_assignment:
-      cluster_name: base
+      cluster_name: base_wlan
       endpoints:
         - lb_endpoints:
             - endpoint:
@@ -54,7 +54,7 @@ R"(
     dns_lookup_family: V4_ONLY
     lb_policy: ROUND_ROBIN
     load_assignment:
-      cluster_name: base
+      cluster_name: base_wwan
       endpoints:
         - lb_endpoints:
             - endpoint:
--- a/library/common/config_template.cc
+++ b/library/common/config_template.cc
@@ -29,8 +29,8 @@ R"(
       sni: {{ domain }}
     type: LOGICAL_DNS
   - name: base_wlan # Note: the direct API depends on the existence of a cluster with this name.
-    connect_timeout: {{ connect_timeout }}
-    dns_refresh_rate: {{ dns_refresh_rate }}
+    connect_timeout: {{ connect_timeout_seconds }}s
+    dns_refresh_rate: {{ dns_refresh_rate_seconds }}s
     dns_lookup_family: V4_ONLY
     lb_policy: ROUND_ROBIN
     load_assignment:
@@ -49,8 +49,8 @@ R"(
       sni: {{ domain }}
     type: LOGICAL_DNS
   - name: base_wwan # Note: the direct API depends on the existence of a cluster with this name.
-    connect_timeout: {{ connect_timeout }}
-    dns_refresh_rate: {{ dns_refresh_rate }}
+    connect_timeout: {{ connect_timeout_seconds }}s
+    dns_refresh_rate: {{ dns_refresh_rate_seconds }}s
     dns_lookup_family: V4_ONLY
     lb_policy: ROUND_ROBIN
     load_assignment:
--- a/library/common/config_template.cc
+++ b/library/common/config_template.cc
@@ -1,5 +1,7 @@
 /**
  * Templated default configuration
+ *
+ * Note: Parts of this configuration are required for the direct API to function.
  */
 const char* config_template = R"(
 static_resources:
--- a/library/common/config_template.cc
+++ b/library/common/config_template.cc
@@ -1,12 +1,10 @@
 /**
  * Templated default configuration
- *
- * Note: Parts of this configuration are required for the direct API to function.
  */
 const char* config_template = R"(
 static_resources:
   clusters:
-  - name: base
+  - name: base # Note: the direct API depends on the existence of a cluster with this name.
     connect_timeout: {{ connect_timeout_seconds }}s
     dns_refresh_rate: {{ dns_refresh_rate_seconds }}s
     dns_lookup_family: V4_ONLY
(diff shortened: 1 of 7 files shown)


library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -34,9 +34,9 @@ String resolveTemplate(String templateYAML) {
     // TODO(@buildbreaker): Implement actual domain template resolution
     String resolvedConfiguration =
         templateYAML.replace("{{ domain }}", "s3.amazonaws.com")
-            .replace("{{ connect_timeout }}", String.format("%ss", connectTimeoutSeconds))
-            .replace("{{ dns_refresh_rate }}", String.format("%ss", dnsRefreshSeconds))
-            .replace("{{ stats_flush_interval }}", String.format("%ss", statsFlushSeconds));
+            .replace("{{ connect_timeout }}", String.format("%s", connectTimeoutSeconds))
```
junr03: In this PR I am moving all the added `s`s to the template so that the platform only has to inject the number. The absence of `s` in the seconds value for iOS was causing crashes.

rebello95: The change seems reasonable. What was the crash you were seeing?

junr03: Because these fields are duration fields they needed the trailing `s` to be interpreted as seconds. The iOS templating removed that so envoy was failing to start because `30` (only the number) is an invalid value for the duration field. To fix I added the trailing `s` to the config itself, and deleted here to create consistency. Also this way we have the same behavior in both platforms (i.e both refresh rates are seconds, rather than letting the platform change the duration unit independently).

I am going to update here to make it explicit in the variable name that this value is in seconds.

rebello95: Ah, I see. Yes that makes sense

rebello95: https://github.com/lyft/envoy-mobile/pull/448 for reference

goaway:
updated and rebased

rebello95:
Base branch still needs to be rebased I think

goaway:
@rebello95 good catch, actually just needed to point the PR at master :)

library/common/types/c_types.h
```diff
@@ -31,11 +31,11 @@ typedef enum { ENVOY_UNDEFINED_ERROR, ENVOY_STREAM_RESET } envoy_error_code_t;
 
 /**
  * Networks classified by last physical link.
- * ENVOY_GENERIC is default and includes cases where network characteristics are unknown.
- * ENVOY_WLAN includes WiFi and other local area wireless networks.
- * ENVOY_WWAN includes all mobile phone networks.
+ * ENVOY_NET_GENERIC is default and includes cases where network characteristics are unknown.
```
rebello95: Do you think `ENVOY_NET_DEFAULT` makes more sense here?

library/common/http/dispatcher.h
```diff
@@ -22,13 +22,16 @@ namespace Http {
...
+  Dispatcher(std::atomic<envoy_network_t>& preferred_network);
+
   void ready(Event::Dispatcher& event_dispatcher, Upstream::ClusterManager& cluster_manager);
 
   /**
    * Attempts to open a new stream to the remote. Note that this function is asynchronous and
    * opening a stream may fail. The returned handle is immediately valid for use with this API, but
    * there is no guarantee it will ever functionally represent an open stream.
-   * @param bridge_callbacks wrapper for callbacks for events on this stream.
+   * @param stream, the stream to start.
```
rebello95: Good catch

library/common/config_template.cc
```diff
@@ -1,5 +1,7 @@
 /**
  * Templated default configuration
+ *
+ * Note: Parts of this configuration are required for the direct API to function.
```
rebello95: This seems like a given since the library only provides a direct API now. I'd either remove this or make it clearer as to which parts of the config are required for a specific purpose

library/common/config_template.cc
```diff
@@ -28,6 +30,46 @@ R"(
             - {{ domain }}
       sni: {{ domain }}
     type: LOGICAL_DNS
+  - name: base_wlan
```
rebello95: It's not ideal that most config changes to these clusters will now need to be made in 3 places (wwan/wlan/base) - is there any way we can share these within the file? Not a blocker, but it'd be a nice improvement

library/common/types/c_types.h
goaway: @junr03 and I discussed and `DEFAULT` was one option we considered. We landed on `GENERIC` because maybe it implies it covers more than just the default case where no information is available (i.e. it also covers the general case - any network for which specific handling isn't required). I don't have strong feelings though, and would be fine with either.

library/common/config_template.cc
goaway: That's fair; I just wanted to call it out somewhere for folks who might plan to pass their own configuration.

I could comment on the clusters themselves down in the YAML.

library/common/config_template.cc
goaway: Great callout. I can factor out some of the common configuration in a follow up PR, if that sounds alright.

library/common/config_template.cc
goaway: updated

library/common/config_template.cc
```diff
@@ -28,6 +28,46 @@ R"(
             - {{ domain }}
       sni: {{ domain }}
     type: LOGICAL_DNS
+  - name: base_wlan # Note: the direct API depends on the existence of a cluster with this name.
+    connect_timeout: {{ connect_timeout_seconds }}s
+    dns_refresh_rate: {{ dns_refresh_rate_seconds }}s
+    dns_lookup_family: V4_ONLY
+    lb_policy: ROUND_ROBIN
+    load_assignment:
+      cluster_name: base
```
rebello95: Should this be `base_wlan`? Same with `base_wwan` below

goaway: good catch, yes it should

PR #448, opened 2019-09-17
config: clean up and fix config

junr03:
Description: there were inconsistencies in the config that led to envoy not being able to start. This PR fixes them.
Risk Level: low
Testing: CI and locally

### DIFF
--- a/library/common/config_template.cc
+++ b/library/common/config_template.cc
@@ -1,7 +1,5 @@
 /**
  * Templated default configuration
- *
- * Note: Parts of this configuration are required for the direct API to function.
  */
 const char* config_template = R"(
 static_resources:
--- a/library/common/config_template.cc
+++ b/library/common/config_template.cc
@@ -1,61 +1,14 @@
 /**
-  * Templated default configuration
-  *
-  * Note: Parts of this configuration are required for the direct API to function.
-  */
+ * Templated default configuration
+ *
+ * Note: Parts of this configuration are required for the direct API to function.
+ */
 const char* config_template = R"(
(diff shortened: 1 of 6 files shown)


rebello95:
Does the template cc file also need to be updated to include the `s` suffixes?

library/common/config_template.cc
```diff
@@ -1,61 +1,14 @@
 /**
-  * Templated default configuration
-  *
-  * Note: Parts of this configuration are required for the direct API to function.
-  */
+ * Templated default configuration
+ *
+ * Note: Parts of this configuration are required for the direct API to function.
```
rebello95: I think this is implied since the direct API is all we expose?

library/common/config_template.cc
```diff
@@ -1,61 +1,14 @@
...
-  - address:
-      socket_address:
-        protocol: TCP
-        address: 0.0.0.0
-        port_value: 9001
-    filter_chains:
-    - filters:
-      - name: envoy.http_connection_manager
-        typed_config:
-          "@type": type.googleapis.com/envoy.config.filter.network.http_connection_manager.v2.HttpConnectionManager
```
rebello95: Personally I think we should leave this here since we plan on supporting it. Maybe add an inline note instead?

junr03: I think we should delete for now until we support it to minimize the amount of code rot we carry around. We can always come back in history and grab it.

junr03:
@rebello95 updated from comments. I also tested locally with all example apps.

PR #462, opened 2019-09-25
ios: update preferred networks via reachability

rebello95:
Adds an implementation of observing reachability on iOS which calls in to the function added in https://github.com/lyft/envoy-mobile/pull/443 to notify Envoy of a change in preferred network selection.

Note that since the core maintains a singleton ref to the preferred network, these observations are also static and are not discarded. This should be addressed as part of implementing shutdown functionality in https://github.com/lyft/envoy-mobile/issues/445.

### DIFF
--- a/library/objective-c/EnvoyEngineImpl.m
+++ b/library/objective-c/EnvoyEngineImpl.m
@@ -3,16 +3,22 @@
 #import "library/common/main_interface.h"
 #import "library/common/types/c_types.h"
 
+#import <SystemConfiguration/SystemConfiguration.h>
+
 @implementation EnvoyEngineImpl {
   envoy_engine_t _engineHandle;
 }
+
+#pragma mark - Public interface
 
 - (instancetype)init {
   self = [super init];
   if (!self) {
     return nil;
   }
+
   _engineHandle = init_engine();
+  _ensure_reachability_running();
   return self;
 }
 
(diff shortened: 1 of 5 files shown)


library/objective-c/EnvoyEngineImpl.m
```diff
@@ -43,4 +49,43 @@ - (int)runWithConfigYAML:(NSString *)configYAML logLevel:(NSString *)logLevel {
...
+  }
+
+  _reachability_ref = reachability;
+
+  SCNetworkReachabilityContext context = {0, NULL, NULL, NULL, NULL};
+  if (!SCNetworkReachabilitySetCallback(_reachability_ref, _reachability_callback, &context)) {
+    return;
+  }
+
+  dispatch_queue_t queue = dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0);
```
goaway: Thinking we should maybe use a dedicated queue here since the main thread may get blocked on slow render operations and also be initiating network requests.

goaway: Oh sorry, you are using separate queue from the main one. What do you think about making this a higher priority?

rebello95: We can create our own queue if you think that makes sense. It is worth noting that this is not the main thread though, this is a background queue with default priority

rebello95: We could use `DISPATCH_QUEUE_PRIORITY_HIGH`, sure

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -3,16 +3,22 @@
...
+#pragma mark - Public interface
+
 - (instancetype)init {
   self = [super init];
   if (!self) {
     return nil;
   }
+
   _engineHandle = init_engine();
+  _ensure_reachability_running();
```
goaway: let's put a `dispatch_one` here

goaway: `once`*

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -43,4 +49,43 @@ - (int)runWithConfigYAML:(NSString *)configYAML logLevel:(NSString *)logLevel {
...
+  if (flags == 0) {
+    return;
+  }
+
+  BOOL isUsingWWAN = flags & kSCNetworkReachabilityFlagsIsWWAN;
+  set_preferred_network(isUsingWWAN ? ENVOY_NET_WWAN : ENVOY_NET_WLAN);
+}
+
+static void _ensure_reachability_running() {
+  if (_reachability_ref) {
```
goaway: And you can skip this check once you're using `dispatch_once`

library/objective-c/EnvoyEngineImpl.m
rebello95: Good call, updated

library/objective-c/EnvoyNetworkMonitor.m
```diff
@@ -0,0 +1,51 @@
+#import "library/objective-c/EnvoyEngine.h"
+
+#import "library/common/main_interface.h"
+
+#import <SystemConfiguration/SystemConfiguration.h>
+
+@implementation EnvoyNetworkMonitor
+
++ (void)startReachabilityIfNeeded {
+  static dispatch_once_t onceToken;
```
goaway: nit: this can actually be named something meaningful like `reachabilityStarted`

rebello95: Done

rebello95:
Thanks for the review @goaway!

PR #468, opened 2019-09-27
android: create network condition monitor

goaway:
Description: Adds a platform network monitor to Android to leverage the set_preferred_network interface. This implementation only works on Lollipop and above, and does rely on some methods deprecated in the most recent versions of Android - in order to be compatible with multiple versions and not unnecessarily complicated.
Risk Level: low
Testing: local, CI, and physical device

### DIFF
--- a/library/java/src/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
+++ b/library/java/src/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
@@ -14,7 +14,9 @@ import android.net.NetworkRequest;
 import android.os.Build;
 
 /**
- * This class makes use of some deprecated APIs, but it's only current purpose is to attempt to distill some notion of a preferred network from the OS, upon which we can assume new sockets will be opened.
+ * This class makes use of some deprecated APIs, but it's only current purpose is to attempt to
+ * distill some notion of a preferred network from the OS, upon which we can assume new sockets will
+ * be opened.
  */
 @TargetApi(Build.VERSION_CODES.LOLLIPOP)
 public class AndroidNetworkMonitor extends BroadcastReceiver {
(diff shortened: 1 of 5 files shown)


library/java/src/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
```diff
@@ -0,0 +1,99 @@
...
+
+  @Override
+  public void onReceive(Context context, Intent intent) {
+    handleNetworkChange();
+  }
+
+  private void handleNetworkChange() {
+    NetworkInfo networkInfo = connectivityManager.getActiveNetworkInfo();
+    if (networkInfo == null) {
+      AndroidJniLibrary.setPreferredNetwork(ENVOY_NET_GENERIC);
```
buildbreaker: It feels a bit too loose for the `setPreferredNetwork()` be a static method since it depends on the `AndroidJniLibrary` to be loaded first.

What do you think about making it an instance method of `AndroidJniLibrary` and we can have this `AndroidNetworkMonitor` depend on `AndroidJniLibrary` to call this?

goaway: I think that's a valid concern, but it sort of applies equally to _any_ of the current JNI library methods, doesn't it?

I can get behind making both JniLibrary classes instance-based (this could make testing easier too), but would you be alright with having that be a separate issue/PR?

buildbreaker: Yep! Let's do it in a separate PR

goaway:
rebase

goaway:
thanks for the review and testing @buildbreaker!

PR #472, opened 2019-09-29
kotlin: grpc codec

buildbreaker:
gRPC codec for kotlin

We are able to hit an Envoy backed gRPC server from `lyft.com`

We ran into some issues with testing (debug pull: https://github.com/lyft/envoy-mobile/pull/495). The first is outgoing ALPN is required for gRPC connections: https://github.com/lyft/envoy-mobile/issues/502. gRPC doesn't support disabling this option when starting up a service. We'll have to revisit this effort in the future.

For now, we are primarily missing https://github.com/lyft/envoy-mobile/pull/494 for local Envoy library e2e testing

Description: gRPC codec for kotlin
Risk Level: low
Testing: unit/end-to-end
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


library/kotlin/test/io/envoyproxy/envoymobile/GRPCStreamEmitterTest.kt
```diff
@@ -0,0 +1,109 @@
...
+    val payload = "data".toByteArray(Charsets.UTF_8)
+    val message = ByteBuffer.wrap(payload)
+    grpcStreamEmitter.sendMessage(message)
+
+    assertThat(dataOutputStream.toByteArray().sliceArray(MESSAGE_HEADING_OFFSET until dataOutputStream.size()).toString(Charsets.UTF_8)).isEqualTo("data")
+
+  }
+
+  @Test
+  fun `close is called with empty data frame`() {
```
rebello95: @junr03 we should add a test like this on iOS as part of https://github.com/lyft/envoy-mobile/pull/500

library/kotlin/src/io/envoyproxy/envoymobile/GRPCStreamEmitter.kt
```diff
@@ -0,0 +1,31 @@
...
+    // Compressed-Flag = 0 / 1, encoded as 1 byte unsigned integer
+    // Message-Length = length of Message, encoded as 4 byte unsigned integer (big endian)
+    // Message = binary representation of protobuf message
+    val byteBuffer = ByteBuffer.allocate(5)
+
+    // Compression flag (1 byte) - 0, not compressed
+    byteBuffer.put(0)
+
+    // Message length
+    val messageLength = message.remaining()
```
rebello95: I think we need to do some conversion here to enforce that this number is big endian like we do here: https://github.com/lyft/Lyft-iOS/blob/f098c5aedfa27b0c0825aa779222c9596d0df511/Modules/LyftGRPC/Sources/GRPCStreamEmitter.swift#L41 (required by the gRPC spec)

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  /**
+   * Specify a callback for when trailers are received by the stream.
+   * If the closure is called, the stream is complete.
+   *
+   * @param closure: Closure which will receive the trailers.
+   * @return ResponseHandler, this ResponseHandler.
+   */
+  fun onTrailers(closure: (trailers: Map<String, List<String>>) -> Unit): GRPCResponseHandler {
```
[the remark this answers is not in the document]
buildbreaker: Is this for keeping `onTrailers` and adding it for Swift?

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  /**
+   * Specify a callback for when an internal Envoy exception occurs with the stream.
+   * If the closure is called, the stream is complete.
+   *
+   * @param closure: Closure which will be called when an error occurs.
+   * @return ResponseHandler, this ResponseHandler.
+   */
+  fun onError(closure: (error: EnvoyError) -> Unit): GRPCResponseHandler {
```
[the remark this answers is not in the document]
buildbreaker: I think having an `onError` would be good since I use it to propagate errors from Envoy + when we receive a compressed message

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,196 @@
...
+        val byteArray = bufferedStream.toByteArray()
+        if (byteArray.size < MESSAGE_HEADING_OFFSET) {
+          // We don't have enough information so we'll just return
+          return ProcessState.CompressionFlag
+        }
+
+        val compressionFlag = byteArray[0]
+        // TODO: Support gRPC compression https://github.com/lyft/envoy-mobile/issues/501
+        if (compressionFlag.compareTo(0) != 0) {
+          errorClosure(
```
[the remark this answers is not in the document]
buildbreaker: The important detail here is that we still keep the stream alive 😬 

If we can safely close the stream, that would be much much better. 

I just left it like this for now since I do consider this a volatile feature and the behavior is a bit up in the air with respect to messages which are gRPC compressed. If we appropriately close the stream, it feels more of a intentional design to exclude compression

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,196 @@
...
+          // no op the current onData and clean up
+          underlyingHandler.onData { _, _ -> }
+          bufferedStream.reset()
+        }
+
+        nextState = ProcessState.MessageLength
+      }
+      is ProcessState.MessageLength -> {
+        val byteArray = bufferedStream.toByteArray()
+        val messageLength = ByteBuffer.wrap(byteArray.sliceArray(1..4)).int
```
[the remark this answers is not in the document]
buildbreaker: It's actually checked in the `ProcessState.CompressionFlag` but I do understand that there is carry over context which could get lost. I can separate out the checks for each state and add another if here

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        val byteArray = bufferedStream.toByteArray()
+        onMessage(ByteBuffer.wrap(
+            byteArray.sliceArray(MESSAGE_HEADING_OFFSET until MESSAGE_HEADING_OFFSET + processState.messageLength)))
+        bufferedStream.reset()
+        bufferedStream.write(
+            byteArray.sliceArray(MESSAGE_HEADING_OFFSET + processState.messageLength until byteArray.size))
+
+        if (byteArray.sliceArray(MESSAGE_HEADING_OFFSET + processState.messageLength until byteArray.size).isEmpty()) {
```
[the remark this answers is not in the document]
buildbreaker: I'm actually more inclined to have early exits and returns but the Swift implementation has the `processData` at the very end so I tried to keep this style closer to that. 

IMO, I do think it'll be clearer to return on each switch case since we wont be hitting multiple cases ever. If you agree with this, we can make changes to the Swift code to do that and I'll make the changes here too

library/kotlin/test/io/envoyproxy/envoymobile/GRPCRequestBuilderTest.kt
```diff
@@ -0,0 +1,54 @@
...
+  @Test
+  fun `not using https will pass http as scheme`() {
+    val request = GRPCRequestBuilder("/pb.api.v1.Foo/GetBar", "foo.bar.com", false)
+        .build()
+
+    assertThat(request.scheme).isEqualTo("http")
+  }
+
+  @Test
+  fun `application gprc is set as content-type header`() {
```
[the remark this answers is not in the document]
buildbreaker: Not valid, I won't be able to do this

library/kotlin/test/io/envoyproxy/envoymobile/GRPCStreamEmitterTest.kt
```diff
@@ -0,0 +1,109 @@
...
+        TODO("not implemented")
+      }
+
+      override fun sendData(byteBuffer: ByteBuffer): StreamEmitter {
+        dataOutputStream.write(byteBuffer.array())
+        return this
+      }
+
+      override fun sendMetadata(metadata: Map<String, List<String>>): StreamEmitter {
+        TODO("not implemented")
```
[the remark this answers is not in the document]
buildbreaker: So a `TODO` actually causes an error if the code ever gets executed -- This wasn't a fun test to write but since we have https://github.com/lyft/envoy-mobile/issues/504, I decided to just keep it this way -- we can alternatively just throw an exception if you feel that aesthetic is better

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,196 @@
...
+    // Awaiting the length specification of the next message.
+    object MessageLength : ProcessState()
+
+    // Awaiting a message with the specified length.
+    class Message(val messageLength: Int) : ProcessState()
+  }
+
+  internal val underlyingHandler: ResponseHandler = ResponseHandler(executor)
+
+  private var errorClosure: (error: EnvoyError) -> Unit = { }
```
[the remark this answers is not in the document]
buildbreaker: We need to keep a reference to the closure because the `ResponseHandler` doesn't expose the underlying closer

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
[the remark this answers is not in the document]
rebello95: Yea

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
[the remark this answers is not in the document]
rebello95: Sg

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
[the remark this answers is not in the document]
rebello95: What if we just call `close()` here and call the consumer with `onError`? I think that'd be a good way to shut down safely

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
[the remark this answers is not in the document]
buildbreaker: Sure. I wonder if gRPC even leverages trailers at all but we can revisit and remove this if not

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,196 @@
...
+      is ProcessState.Message -> {
+        if (bufferedStream.size() < processState.messageLength + MESSAGE_HEADING_OFFSET) {
+          // We don't have enough bytes to construct the message so we'll return
+          return ProcessState.Message(processState.messageLength)
+        }
+
+        val byteArray = bufferedStream.toByteArray()
+        onMessage(ByteBuffer.wrap(
+            byteArray.sliceArray(MESSAGE_HEADING_OFFSET until MESSAGE_HEADING_OFFSET + processState.messageLength)))
+        bufferedStream.reset()
```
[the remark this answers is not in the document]
buildbreaker: Yeah, it does an additional copy. If it helps any, we do use `ByteArrayOutputStream` as our unoptimized `ByteArray` I/O sink. We can revisit this strategy pretty easily further down the road

library/kotlin/test/io/envoyproxy/envoymobile/GRPCStreamEmitterTest.kt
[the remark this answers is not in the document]
rebello95: Hm yea I think throwing an exception might be better here. My thought was to reduce false-positives when grepping the codebase for "todo" items

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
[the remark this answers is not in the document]
buildbreaker: We also don't have a `close()` handler in the response case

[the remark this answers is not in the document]
rebello95: aha right

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,177 @@
...
+    // Awaiting a message with the specified length.
+    class Message(val messageLength: Int) : ProcessState()
+  }
+
+  internal val underlyingHandler: ResponseHandler = ResponseHandler(executor)
+
+  private var errorClosure: (error: EnvoyError) -> Unit = { }
+  /**
+   * Specify a callback for when response headers are received by the stream.
+   * If `endStream` is `true`, the stream is complete.
```
rebello95: This should be removed

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,177 @@
...
+   *
+   * @param closure: Closure which will receive the headers, status code,
+   *                 and flag indicating if the stream is complete.
+   * @return ResponseHandler, this ResponseHandler.
```
rebello95: s/ResponseHandler/GRPCResponseHandler or just "response handler" (here and throughout the file)

codespell
```diff
@@ -0,0 +1 @@
+feild
```
rebello95: Are we able to inline comments here as to why these are necessary?

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,177 @@
...
+      is ProcessState.CompressionFlag -> {
+        val byteArray = bufferedStream.toByteArray()
+        if (byteArray.isEmpty()) {
+          // We don't have enough information to extract the compression flag, so we'll just return
+          return ProcessState.CompressionFlag
+        }
+
+        val compressionFlag = byteArray[0]
+        // TODO: Support gRPC compression https://github.com/lyft/envoy-mobile/issues/501
+        if (compressionFlag.compareTo(0) != 0) {
```
rebello95: Can we not do `compressionFlag != 0`?

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,177 @@
...
+          errorClosure(
+              EnvoyError(
+                  EnvoyErrorCode.ENVOY_UNDEFINED_ERROR,
+                  "Unable to accept compression enabled messages for gRPC"))
```
rebello95:
```suggestion
                  "Unable to read compressed gRPC response message"))
```

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,177 @@
...
+
+          // no op the current onData and clean up
+          underlyingHandler.onData { _, _ -> }
```
rebello95: Should we reset other handler closures here as well? I'd assume `onTrailers` might be able to be called in this case, whereas `onError` should be considered terminal

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,177 @@
...
+        nextState = ProcessState.MessageLength
+      }
+      is ProcessState.MessageLength -> {
+        if (bufferedStream.size() < GRPC_PREFIX_LENGTH) {
+          // We don't have enough information to extract the message length, so we'll just return
+          return ProcessState.MessageLength
+        }
+
+        val byteArray = bufferedStream.toByteArray()
+        val messageLength = ByteBuffer.wrap(byteArray.sliceArray(1..4)).int
```
rebello95: Do we need to convert big endian numbers to native here? we have to apply `CFSwapInt32BigToHost` on iOS

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,177 @@
...
+
+        if (messageLength == 0) {
+          onMessage(ByteBuffer.wrap(ByteArray(0)))
+          return ProcessState.CompressionFlag
```
rebello95: Because this is early returning instead of continuing, don't we need to drop the first 5 bytes of the `bufferedStream` here to unblock future responses?

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,177 @@
...
+      }
+      is ProcessState.Message -> {
+        if (bufferedStream.size() < processState.messageLength + GRPC_PREFIX_LENGTH) {
+          // We don't have enough bytes to construct the message, so we'll just return
+          return ProcessState.Message(processState.messageLength)
+        }
+
+        val byteArray = bufferedStream.toByteArray()
+        onMessage(ByteBuffer.wrap(
+            byteArray.sliceArray(GRPC_PREFIX_LENGTH until GRPC_PREFIX_LENGTH + processState.messageLength)))
```
rebello95: Curious, why use `byteArray.sliceArray(1..4)` above but `sliceArray(... until ...)` here?

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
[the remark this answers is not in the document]
rebello95: Might be worth adding a todo with optimization notes 🤷‍♂

library/kotlin/src/io/envoyproxy/envoymobile/GRPCStreamEmitter.kt
```diff
@@ -0,0 +1,41 @@
...
+    // Message-Length = length of Message, encoded as 4 byte unsigned integer (big endian)
+    // Message = binary representation of protobuf messageData
+    val byteBuffer = ByteBuffer.allocate(GRPC_PREFIX_LENGTH)
+
+    // Compression flag (1 byte) - 0, not compressed
+    byteBuffer.put(0)
+
+    // Message length
+    val messageLength = messageData.remaining()
+    byteBuffer.order(ByteOrder.BIG_ENDIAN)
```
rebello95: Nice that this is available 😃

library/kotlin/src/io/envoyproxy/envoymobile/StreamEmitter.kt
```diff
@@ -49,5 +48,5 @@ interface StreamEmitter : CancelableStream {
    * @throws EnvoyException when there is an exception ending the stream or sending trailers.
    */
   @Throws(EnvoyException::class)
-  fun close(trailers: Map<String, List<String>>?)
+  fun close(trailers: Map<String, List<String>>? = null)
```
rebello95: We discussed on @junr03's PR not to provide default values here so that it's clear to the caller how they're closing the stream. I think we should explicitly pass `trailers: null` from the emitter

library/kotlin/test/io/envoyproxy/envoymobile/GRPCStreamEmitterTest.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    assertThat(dataOutputStream.toByteArray()[0]).isEqualTo(0)
+  }
+
+  @Test
+  fun `message length is set on the 1-4 bytes of the prefix`() {
+    val grpcStreamEmitter = GRPCStreamEmitter(emitter)
+
+    val payload = "data".toByteArray(Charsets.UTF_8)
+    val message = ByteBuffer.wrap(payload)
```
rebello95: I think we'll need to validate endian-ness here

library/kotlin/test/io/envoyproxy/envoymobile/GRPCStreamEmitterTest.kt
```diff
@@ -0,0 +1,101 @@
...
+
+  }
+
+  @Test
+  fun `close is called with empty data frame`() {
+    val grpcStreamEmitter = GRPCStreamEmitter(emitter)
+
+    grpcStreamEmitter.close()
+
+    assertThat(isCloseCalled).isTrue()
```
rebello95: This actually doesn't validate whether close is called with trailers or a data frame. We should check that trailers are `nil` in this case

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
buildbreaker: byte/int incompatibility comparison

library/kotlin/test/io/envoyproxy/envoymobile/GRPCStreamEmitterTest.kt
rebello95: https://github.com/lyft/envoy-mobile/pull/509

library/kotlin/src/io/envoyproxy/envoymobile/GRPCStreamEmitter.kt
buildbreaker: This was what `codespell` was complaining about

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
buildbreaker: Just checked, we aren't allowed to use variables with the `..` operator, instead we are supposed to use `until`

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
buildbreaker: Hm, let me add a test

buildbreaker: It's also leaking message parsing responsibilities onto the message length parsing so I can move this on the `ProcessState.Message` statement

library/kotlin/src/io/envoyproxy/envoymobile/GRPCStreamEmitter.kt
buildbreaker: Updating codespell fixes it though

buildbreaker: @junr03 found the PR which removes it: https://github.com/codespell-project/codespell/pull/1189/files

codespell
buildbreaker: Removing

library/kotlin/test/io/envoyproxy/envoymobile/GRPCStreamEmitterTest.kt
```diff
@@ -34,7 +34,7 @@ class GRPCStreamEmitterTest {
       }
 
       override fun close(trailers: Map<String, List<String>>?) {
-        isCloseCalled = true
+        isCloseCalled = trailers == null
```
rebello95: Might want to rename `isCloseCalled` now (see https://github.com/lyft/envoy-mobile/pull/509 for iOS)

library/kotlin/test/io/envoyproxy/envoymobile/GRPCStreamEmitterTest.kt
rebello95: This still needs to be resolved I believe?

buildbreaker: I guess we can ensure that we're using big endian for reading the message length in case we are running the test on a machine which defaults to little endian

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -0,0 +1,175 @@
...
+      }
+      is ProcessState.MessageLength -> {
+        if (bufferedStream.size() < GRPC_PREFIX_LENGTH) {
+          // We don't have enough information to extract the message length, so we'll just return
+          return ProcessState.MessageLength
+        }
+
+        val byteArray = bufferedStream.toByteArray()
+        val buffer = ByteBuffer.wrap(byteArray.sliceArray(1..4))
+        buffer.order(ByteOrder.BIG_ENDIAN)
```
rebello95: Assuming the language does the right thing here and does the conversion to native format when you call `.int` below this should be fine

buildbreaker: Yeah, it will write it in big endian here

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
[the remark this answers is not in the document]
rebello95: I just realized why I didn't do this originally on iOS. Storing the `errorClosure` on the `GRPCResponseHandler` is problematic because nothing necessarily keeps the `GRPCResponseHandler` in memory since only the closures need to stick around. When I ended up using the `GRPCResponseHandler` in practice, this was totally fine because the closures passed to the underlying callbacks have their own lifecycle. Maybe we should just assert here instead?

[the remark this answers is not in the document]
rebello95: https://github.com/lyft/envoy-mobile/pull/515

PR #477, opened 2019-10-01
core: change handles to a more suitable type

goaway:
Description: Better fit for storing opaque memory addresses and/or sentinel values.


goaway:
Also: JVM types are signed.

This will generally give us safety when we take the final steps to supporting multiple instances.

PR #479, opened 2019-10-02
bazel: update release config

keith:
This changes the release config to use bazel's compilation_mode=opt
which implies the necessary optimization options, and some other things.
This also removes the linkopt that appears to do nothing, and requires
you pass some configuration about architectures separately.


.bazelrc
```diff
@@ -32,8 +32,6 @@ build:fat --ios_multi_cpus=i386,x86_64,armv7,arm64 --fat_apk_cpu=x86,x86_64,arme
 # TODO: Enable `--copt -ggdb3`. Issues were encountered previously with this:
```
rebello95: Is this still relevant?

keith: I guess it could be, but I'll remove it for now and we can revisit if we feel like we're missing this info

PR #481, opened 2019-10-02
android: add proguard rules and build

buildbreaker:
1. Adding proguard rules to protect native called code and native methods. 
2. Enable proguarding in our example applications to end to end test proguard obfuscation/shrink
3. Added the proguard rules in our aar:
https://github.com/lyft/envoy-mobile/issues/478
```
➜  envoy-mobile git:(master) ✗ zipinfo -1 dist/envoy.aar
AndroidManifest.xml
classes.jar
res/
R.txt
jni/arm64-v8a/libenvoy_jni.so
jni/armeabi-v7a/libenvoy_jni.so
jni/x86/libenvoy_jni.so
jni/x86_64/libenvoy_jni.so
proguard.txt
```

Description: android: add proguard rules and build
Risk Level: low
Testing: local/e2e
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


dist/BUILD
```diff
@@ -25,3 +25,9 @@ apple_static_framework_import(
     ],
     visibility = ["//visibility:public"],
 )
+
+filegroup(
+    name = "proguard_rule",
```
artem-zinnatullin:
```suggestion
    name = "proguard_rules",
```

dist/proguard.txt
```diff
@@ -0,0 +1,4 @@
+-dontwarn **
+-dontnote **
+
+-keep @io.envoyproxy.envoymobile.Native class * { *; }
```
artem-zinnatullin: This rule says to keep the class if it's annotated with `@Native`, but I think you'd want to keep only individual methods if they're annotated

Here is how Android's `@Keep` works:


```bash
# It's an example from Android SDK, actual Androidx annotation rules are in androidx.annotations aar file
cat "$ANDROID_HOME/tools/proguard/examples/annotations/lib/annotations.pro"
```

```console
# The following annotations can be specified with classes and with class
# members.

# @Keep specifies not to shrink, optimize, or obfuscate the annotated class
# or class member as an entry point.

-keep @proguard.annotation.Keep class *

-keepclassmembers class * {
    @proguard.annotation.Keep *;
}
```

buildbreaker: So `keep` doesn't work for some reason

bazel/aar_with_jni.bzl
```diff
@@ -55,17 +55,19 @@ EOF
 
     native.genrule(
         name = name,
-        srcs = [android_library + "_kt.jar", android_library + ".aar", archive_name + "_jni_unsigned.apk"],
+        srcs = [android_library + "_kt.jar", android_library + ".aar", archive_name + "_jni_unsigned.apk", "//library:proguard_rules"],
```
goaway: rather than hardcoding this file location into the bazel rule, what do you think about passing this in as an optional parameter?

examples/kotlin/hello_world/BUILD
```diff
@@ -6,6 +6,7 @@ android_binary(
     name = "hello_envoy_kt",
     custom_package = "io.envoyproxy.envoymobile.helloenvoykotlin",
     manifest = "AndroidManifest.xml",
+    proguard_specs = ["//dist:proguard_rules"],
```
goaway: Does this need to be updated? Can/should this be referencing something in the aar?

bazel/aar_with_jni.bzl
buildbreaker: I think we can make this non-optional since we're going to need to have this in our aar

examples/kotlin/hello_world/BUILD
buildbreaker: Good point. Let me double check if this actually works if we don't reference it

I'll separate out the envoy mobile specifics

bazel/aar_with_jni.bzl
goaway: I guess my point is that the aar rule is actually general not project or source-specific, but this embeds a project specific location in it.

goaway: Or rather, it just seems preferable in general to provide source locations as arguments to bazel rules, rather than have them be part of the rule itself.

examples/kotlin/hello_world/BUILD
buildbreaker: It doesn't seem like bazel does the aar inspection + merging for us. We'll have to do this sub-optimal thing for now

PR #486, opened 2019-10-04
Update Envoy upstream

rebello95:
Pulls in the fix from https://github.com/envoyproxy/envoy/pull/8467.


rebello95:
Screen Shot 2019-10-03 at 9 01 48 PM
Looks like this resulted in a bit of a size increase. Since we need these fixes, I'm going to merge and defer investigating to https://github.com/lyft/envoy-mobile/issues/447

PR #494, opened 2019-10-09
java: create java lib for local testing

buildbreaker:
Create a target for generating a Java library used for unit tests

Description: java: create java lib for local testing
Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


library/common/android_jni_utils.cc
```diff
@@ -0,0 +1,21 @@
+#include <android/log.h>
+#include <ares.h>
+#include <jni.h>
+
+// NOLINT(namespace-envoy)
+
+jint init_jvm(JavaVM* static_jvm, jobject connectivity_manager) {
```
goaway:
```suggestion
jint init_jvm(JavaVM* jvm, jobject platform_context) {
```

The `jvm` isn't static here, it's just called that in the interface layer because it's a static ref. You can add a comment that says the `context` is really a `ConnectivityManager`, but this allows it to be named something generic in the top-level interface.

goaway: Actually I'd also call this method something else since we're not really initializing the jvm here. Maybe `platform_setup`?

library/common/android_jni_utils.cc
```diff
@@ -0,0 +1,21 @@
...
+  // See note above about c-ares.
+  // c-ares jvm init is necessary in order to let c-ares perform DNS resolution in Envoy.
+  // More information can be found at:
+  // https://c-ares.haxx.se/ares_library_init_android.html
+  ares_library_init_jvm(static_jvm);
+
+  return ares_library_init_android(connectivity_manager);
+}
+
+jboolean ares_initialized() { return ares_library_android_initialized() == ARES_SUCCESS; }
```
goaway: Like we discussed, this can be deleted altogether for now.

library/common/android_jni_utils.cc
```diff
@@ -0,0 +1,21 @@
+#include <android/log.h>
+#include <ares.h>
+#include <jni.h>
+
+// NOLINT(namespace-envoy)
+
+jint init_jvm(JavaVM* static_jvm, jobject connectivity_manager) {
+  // See note above about c-ares.
```
goaway: This comment is no longer applicable (it referred to the remainder of the comments below).

library/common/jni_interface.cc
```diff
@@ -52,20 +51,14 @@ extern "C" JNIEXPORT jint JNICALL
...
+  return init_jvm(static_jvm, connectivity_manager);
 }
 
 extern "C" JNIEXPORT jboolean JNICALL
 Java_io_envoyproxy_envoymobile_engine_AndroidJniLibrary_isAresInitialized(JNIEnv* env,
                                                                           jclass // class
 ) {
-  return ares_library_android_initialized() == ARES_SUCCESS;
+  return ares_initialized();
 }
```
goaway: We can delete all of this.

library/common/jni_interface.cc
```diff
@@ -52,20 +51,14 @@ extern "C" JNIEXPORT jint JNICALL
...
                                                                    jclass, // class
                                                                    jobject connectivity_manager) {
-  // See note above about c-ares.
-  // c-ares jvm init is necessary in order to let c-ares perform DNS resolution in Envoy.
-  // More information can be found at:
-  // https://c-ares.haxx.se/ares_library_init_android.html
-  ares_library_init_jvm(static_jvm);
-
-  return ares_library_init_android(connectivity_manager);
+  return init_jvm(static_jvm, connectivity_manager);
```
goaway: With the split implementations, should we move this back to a common JNI layer and have Android call it with a`ConnectivityManager` and Java call it with `null`?

buildbreaker: The java side doesn't need to call this right now and the Android JNI layer does all this on load (and we'll have to avoid calling the java version of this on the Java JNI layer on load). It feels a bit more brittle since we'll need to remember that we don't want to call this method in the java side because it'll impact the Android side

bazel/kotlin_test.bzl
```diff
@@ -40,4 +22,47 @@ def envoy_mobile_kt_test(name, srcs, deps = []):
             "@maven//:org_mockito_mockito_inline",
             "@maven//:org_mockito_mockito_core",
         ] + deps,
+        data = data,
+        jvm_flags = jvm_flags,
+    )
+
+def envoy_mobile_jni_kt_test(name, srcs, lib = "", deps = []):
```
buildbreaker: TODO: Add comments for usages

library/common/java_support.cc
```diff
@@ -0,0 +1,7 @@
+#include <jni.h>
+
+// NOLINT(namespace-envoy)
+
+jint platform_setup(JavaVM* jvm, jobject connectivity_manager) { return 0; }
```
goaway:
```suggestion
jint platform_setup(JavaVM*, jobject) { return 0; }
```

library/common/android_support.cc
```diff
@@ -0,0 +1,18 @@
...
+  // c-ares jvm init is necessary in order to let c-ares perform DNS resolution in Envoy.
+  // More information can be found at:
+  // https://c-ares.haxx.se/ares_library_init_android.html
+  ares_library_init_jvm(jvm);
+
+  return ares_library_init_android(connectivity_manager);
+}
+
+void jni_log(const char* tag, const char* text) {
+  __android_log_write(ANDROID_LOG_ERROR, tag, text);
```
goaway: Maybe comment this out and put a `FIXME` here linking to the general pluggable logging issue.

library/common/java_support.cc
```diff
@@ -0,0 +1,7 @@
...
+
+void jni_log(const char* tag, const char* text) {}
```
goaway: also suggest putting FIXME/TODO here as well

library/common/jni_support.h
```diff
@@ -0,0 +1,7 @@
+#include <jni.h>
+
+// NOLINT(namespace-envoy)
+
+jint platform_setup(JavaVM* jvm, jobject connectivity_manager);
```
goaway: I would put a comment here for now saying that this initializes ARES on Android and is not called from pure Java (but would be a no-op if it was). Basically, explain this whole thing.

library/common/BUILD
```diff
@@ -24,14 +25,40 @@ envoy_cc_library(
 )
 
 cc_library(
-    name = "envoy_jni_interface_lib",
+    name = "android_envoy_jni_interface_lib",
```
goaway: nit: suggest `envoy_android_jni_lib` to be a little more concise

library/common/BUILD
```diff
@@ -24,14 +25,40 @@ envoy_cc_library(
...
     copts = ["-std=c++14"],
     linkopts = [
         "-lm",
         "-llog",
     ],
     deps = [":envoy_main_interface_lib"],
 )
+
+cc_library(
+    name = "java_envoy_jni_interface_lib",
```
goaway: `envoy_java_jni_lib`

library/common/jni_interface.cc
```diff
@@ -1,11 +1,12 @@
-#include <android/log.h>
-#include <ares.h>
 #include <jni.h>
 
 #include <string>
 
+#include "jni_support.h"
 #include "main_interface.h"
 
+using namespace std;
```
buildbreaker: For the following error in tests:
```
library/common/jni_interface.cc:81:5: error: 'memcpy' was not declared in this scope
     memcpy(critical_key, headers.headers[i].key.bytes, headers.headers[i].key.length);
     ^~~~~~
```

buildbreaker:
@goaway thanks :)

rebello95:
Is this something we want to merge?

junr03:
@buildbreaker going to close this for now as it has grown stale. Feel free to reopen when we take this back up!

PR #500, opened 2019-10-14
grpc: close stream with data frame

junr03:
Description: the gRPC protocol requires closing the client stream in a DATA frame. This PR fixes the GRPCStreamEmitter for iOS.
Risk Level: med - protocol level changes.
Testing: local.

### DIFF
--- a/library/swift/src/GRPCStreamEmitter.swift
+++ b/library/swift/src/GRPCStreamEmitter.swift
@@ -46,6 +46,6 @@ public final class GRPCStreamEmitter: NSObject {
   public func close(_ data: Data) {
     // The gRPC protocol requires the client stream to close with a DATA frame.
     // More information here: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
-    self.underlyingEmitter.closeWithData(data)
+    self.underlyingEmitter.close()
   }
 }
--- a/library/swift/src/GRPCStreamEmitter.swift
+++ b/library/swift/src/GRPCStreamEmitter.swift
@@ -45,7 +45,8 @@ public final class GRPCStreamEmitter: NSObject {
   /// Close this connection.
   public func close() {
     // The gRPC protocol requires the client stream to close with a DATA frame.
-    // More information here: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
-    self.underlyingEmitter.close()
+    // More information here:
+    // https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
+    self.underlyingEmitter.close(trailers: nil)
   }
 }
--- a/library/swift/src/GRPCStreamEmitter.swift
+++ b/library/swift/src/GRPCStreamEmitter.swift
@@ -43,7 +43,9 @@ public final class GRPCStreamEmitter: NSObject {
   }
 
   /// Close this connection.
-  public func close() {
-    self.underlyingEmitter.close()
+  public func close(_ data: Data) {
+    // The gRPC protocol requires the client stream to close with a DATA frame.
+    // More information here: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
+    self.underlyingEmitter.closeWithData(data)
   }
 }
--- a/library/swift/src/GRPCStreamEmitter.swift
+++ b/library/swift/src/GRPCStreamEmitter.swift
@@ -43,7 +43,7 @@ public final class GRPCStreamEmitter: NSObject {
   }
 
   /// Close this connection.
-  public func close(_ data: Data) {
+  public func close() {
     // The gRPC protocol requires the client stream to close with a DATA frame.
     // More information here: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
     self.underlyingEmitter.close()
--- a/library/swift/src/StreamEmitter.swift
+++ b/library/swift/src/StreamEmitter.swift
@@ -32,10 +32,3 @@ public protocol StreamEmitter: CancelableStream {
   ///                       Otherwise, the stream is closed with an empty data frame.
   func close(trailers: [String: [String]]?)
 }
-
-extension StreamEmitter {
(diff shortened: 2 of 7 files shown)


library/swift/src/StreamEmitter.swift
```diff
@@ -30,6 +30,11 @@ public protocol StreamEmitter: CancelableStream {
   ///
   /// - parameter trailers: Trailers to send over the stream.
   func close(trailers: [String: [String]])
+
+  /// End the stream with one last data frame.
+  ///
+  /// - parameter data: Data to send over the stream.
+  func closeWithData(_ data: Data)
```
junr03: do we want a utility function like below to close with empty data?

library/swift/src/GRPCStreamEmitter.swift
```diff
@@ -43,7 +43,9 @@ public final class GRPCStreamEmitter: NSObject {
   }
 
   /// Close this connection.
-  public func close() {
-    self.underlyingEmitter.close()
+  public func close(_ data: Data) {
```
junr03: we dont need to expose the data here, open to just having self.underlyingEmitter.closeWithData(Data()) below

rebello95: Why not leave the public interface unchanged and simply close with `Data()` under the hood? I think this is a bit confusing since it's not immediately clear what should be sent here.

rebello95: Just commented with that suggestion - I think we should avoid exposing the option to close with custom data

library/swift/src/StreamEmitter.swift
buildbreaker: I feel we could expose the endStream flag on the `sendData` handler instead (for ergonomics, we can have that default to `false`). The gRPC implementation detail would just be that we send an empty data with end stream flag being true on the `close()`

WDYT @junr03 @rebello95 ?

rebello95: @buildbreaker and I discussed this a bit offline and thing changing to `close(data: Data)` here probably makes the most sense:
- Matches `close(trailers: [String: [String]])`
- Returns `void` to make it clear that this is a terminal state (rather than returning a `StreamEmitter` for chaining like `sendData`

junr03:
@buildbreaker @rebello95 I updated with @goaway's suggestion. lmk what you think

library/swift/src/GRPCStreamEmitter.swift
```diff
@@ -43,7 +43,9 @@ public final class GRPCStreamEmitter: NSObject {
   }
 
   /// Close this connection.
-  public func close() {
+  public func close(_ data: Data) {
```
rebello95: This change can be reverted now

library/swift/src/GRPCStreamEmitter.swift
```diff
@@ -43,7 +43,9 @@ public final class GRPCStreamEmitter: NSObject {
...
+    // The gRPC protocol requires the client stream to close with a DATA frame.
+    // More information here: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
     self.underlyingEmitter.close()
```
rebello95: This is actually wrong, it'll still send trailers here rather than empty data.

library/swift/src/StreamEmitter.swift
```diff
@@ -26,10 +26,11 @@ public protocol StreamEmitter: CancelableStream {
...
+  /// End the stream.
   ///
-  /// - parameter trailers: Trailers to send over the stream.
-  func close(trailers: [String: [String]])
+  /// - parameter trailers: optional trailers to send over the stream.
+  ///                       Otherwise, the stream is closed with an empty data frame.
+  func close(trailers: [String: [String]]?)
 }
 
 extension StreamEmitter {
```
rebello95: Maybe we should remove this now since it's ambiguous as to whether the stream will close with trailers or data?

rebello95:
Let's make the appropriate changes on Android as well as part of this PR

library/swift/src/GRPCStreamEmitter.swift
junr03: How come? I thought by deleting the extension below, I was getting rid of a default value when no argument was given

library/swift/src/StreamEmitter.swift
junr03: If we do this then we will be deleting any ability to send trailers from the platform layer.

library/swift/src/GRPCStreamEmitter.swift
rebello95: This comment was before the extension was deleted. I don't think this compiles as-is, you'll need to specify `trailers: nil` I believe

library/swift/src/StreamEmitter.swift
rebello95: I was referring to the convenience method for closing without specifying trailers, which it looks like you already removed 😃

library/swift/src/StreamEmitter.swift
```diff
@@ -26,15 +26,9 @@ public protocol StreamEmitter: CancelableStream {
...
-  /// - parameter trailers: Trailers to send over the stream.
-  func close(trailers: [String: [String]])
-}
-
-extension StreamEmitter {
-  /// Convenience function for ending the stream without sending any trailers.
-  public func close() {
-    self.close(trailers: [:])
-  }
+  /// - parameter trailers: optional trailers to send over the stream.
```
rebello95: Suggestion: `Trailers with which to close the stream. If nil, stream will be closed with an empty data frame.`

library/swift/src/GRPCStreamEmitter.swift
junr03: 😂 I think my commit and your comment cross paths. Will update

library/swift/test/GRPCStreamEmitterTests.swift
```diff
@@ -20,7 +20,7 @@ private final class MockEmitter: StreamEmitter {
     return self
   }
 
-  func close(trailers: [String: [String]]) {}
+  func close(trailers: [String: [String]]?) {}
```
rebello95: I’ll add a test for this in another PR

PR #511, opened 2019-10-16
core: buffer body for retries

junr03:
Description: set stream options to buffer stream body for retries.
Risk Level: med - the client now buffers stream body for retries.
Testing: local - lyft and example apps.


junr03:
@buildbreaker going to merge this in spite of the kotlin tests, I think your fix is now on master.

PR #512, opened 2019-10-17
kotlin: make ResponseHandler closer to Swift

buildbreaker:

Description: kotlin: make ResponseHandler closer to Swift
Risk Level: low
Testing: unit/ci
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/library/kotlin/src/io/envoyproxy/envoymobile/ResponseHandler.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/ResponseHandler.kt
@@ -12,45 +12,49 @@ import java.util.concurrent.Executor;
 class ResponseHandler(val executor: Executor) {
 
   class EnvoyHTTPCallbacksAdapter(
-      private val responseHandler: ResponseHandler
+      private val executor: Executor
   ) : EnvoyHTTPCallbacks {
 
-    override fun getExecutor(): Executor = responseHandler.executor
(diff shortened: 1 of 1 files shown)


library/kotlin/src/io/envoyproxy/envoymobile/ResponseHandler.kt
```diff
@@ -12,45 +12,49 @@ import java.util.concurrent.Executor;
...
 
   class EnvoyHTTPCallbacksAdapter(
-      private val responseHandler: ResponseHandler
+      private val executor: Executor
   ) : EnvoyHTTPCallbacks {
 
-    override fun getExecutor(): Executor = responseHandler.executor
+    private var onHeadersClosure: (headers: Map<String, List<String>>, statusCode: Int, endStream: Boolean) -> Unit = { _, _, _ -> Unit }
+
+    internal var onDataClosure: (byteBuffer: ByteBuffer, endStream: Boolean) -> Unit = { _, _ -> Unit }
```
rebello95: Would be nice if these could be lazily allocated

buildbreaker: I can just do a null check if that is better, this just ensures we have something to call into regardless of the setting state

PR #517, opened 2019-10-17
ci: update artifact to only run for common/java/kotlin

buildbreaker:

Description: ci: update artifact to only run for common/java/kotlin
Risk Level: low
Testing: on master
Docs Changes: n/a
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -5,7 +5,9 @@ on:
     branches:
       - master
     paths:
-      - 'library/**'
+      - 'library/common/**'
+      - 'library/java/**'
+      - 'library/kotlin/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/artifacts.yml
```diff
@@ -5,7 +5,9 @@ on:
     branches:
       - master
     paths:
-      - 'library/**'
+      - 'library/common/**'
+      - 'library/java/**'
+      - 'library/kotlin/**'
       - 'envoy/**'
 
 jobs:
```
junr03: we also need to change the job to not compile x86_64. It seems that we have problems with that in CI

PR #519, opened 2019-10-17
kotlin: add support for total upstream timeouts

rebello95:
When users set a timeout via `perRetryTimeoutMS`, they silently get limited to 15 seconds due to the fact that Envoy assigns a [default timeout of 15 seconds](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route.proto#envoy-api-field-route-routeaction-timeout). We should expose this through our `RetryPolicy` to allow consumers to override it using the already-supported `x-envoy-upstream-rq-timeout-ms` header.

Related Swift PR: https://github.com/lyft/envoy-mobile/pull/518

Details on retry semantics: [Envoy docs](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_routing.html?highlight=exponential#retry-semantics)

### DIFF
--- a/library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
@@ -4,9 +4,9 @@ package io.envoyproxy.envoymobile
  * Specifies how a request may be retried, containing one or more rules.
  *
  * @param maxRetryCount Maximum number of retries that a request may be performed.
- * @param retryOn Whitelist of rules used for retrying. Must be <= `totalUpstreamTimeoutMS`
- * or it will be ignored.
+ * @param retryOn Whitelist of rules used for retrying.
  * @param perRetryTimeoutMs Timeout (in milliseconds) to apply to each retry.
+ * Must be <= `totalUpstreamTimeoutMS`.
  * @param totalUpstreamTimeoutMS Total timeout (in milliseconds) that includes all retries.
  * Spans the point at which the entire downstream request has been processed and when the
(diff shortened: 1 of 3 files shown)


library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
```diff
@@ -4,13 +4,18 @@ package io.envoyproxy.envoymobile
  * Specifies how a request may be retried, containing one or more rules.
  *
  * @param maxRetryCount Maximum number of retries that a request may be performed.
- * @param retryOn Whitelist of rules used for retrying.
+ * @param retryOn Whitelist of rules used for retrying. Must be <= `totalUpstreamTimeoutMS`
```
junr03: is it worth enforcing this somehow? Or at least warning? I don't know what is convention in client land

rebello95: We could, though since this validation is done within Envoy itself, it seems like maybe that should be done there?

junr03: Envoy will just silently ignore. I guess the case that is the least intuitive and could cause people headache is if they are unaware that the default is 15s and they set a per try of >15s and they see timeouts happening at 15s, but had no warning (or appropriate mobile convention).

rebello95: Talked to @buildbreaker and we decided to throw here. Updated

PR #521, opened 2019-10-17
ios: only buffer when requests support retries

rebello95:
Updates to utilize the options exposed in https://github.com/lyft/envoy-mobile/pull/520 to prevent unnecessarily buffering requests that will never be retried. This can be especially useful for preventing buffering of long-lived [gRPC] streams, for example.

We discussed having a full fledged platform level struct to parallel `envoy_stream_options`, but decided that while we only expose one option we can leave the API as a boolean argument.

Matching Android functionality is being tracked here: https://github.com/lyft/envoy-mobile/issues/524.

### DIFF
--- a/library/swift/src/EnvoyClient.swift
+++ b/library/swift/src/EnvoyClient.swift
@@ -69,7 +69,8 @@ public final class EnvoyClient: NSObject {
 
 extension EnvoyClient: HTTPClient {
   public func send(_ request: Request, handler: ResponseHandler) -> StreamEmitter {
-    let httpStream = self.engine.startStream(with: handler.underlyingCallbacks)
+    let httpStream = self.engine.startStream(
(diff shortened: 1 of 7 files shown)


library/swift/src/EnvoyClient.swift
```diff
@@ -69,7 +69,8 @@ public final class EnvoyClient: NSObject {
 
 extension EnvoyClient: HTTPClient {
   public func send(_ request: Request, handler: ResponseHandler) -> StreamEmitter {
-    let httpStream = self.engine.startStream(with: handler.underlyingCallbacks)
+    let httpStream = self.engine.startStream(
+      with: handler.underlyingCallbacks, bufferForRetry: request.retryPolicy != nil)
```
junr03: nice, the end-user does not have to know about the tie between the stream option and retrying.

rebello95: Yep!

PR #525, opened 2019-10-18
ci: update to 90m for aar master generation

buildbreaker:
These builds are taking longer than the original 45minutes since we are building multiple architectures on a release configuration

We're going to make it 90minutes and see if it works

Description: ci: update to 90m for aar master generation
Risk Level: low
Testing: on master
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


rebello95:
Could you update the description to indicate why this is necessary?

PR #528, opened 2019-10-21
docs: minor cleanup

rebello95:
- Remove `about_docs.rst`
- Fix file name
- Fix table of contents title of `What is Envoy Mobile?`
- Fix capitalization

### DIFF
--- a/docs/root/about_docs.rst
+++ b/docs/root/about_docs.rst
@@ -1,8 +1,0 @@
-About the documentation
-=======================
-
-The Envoy Mobile documentation is composed of a few major sections:
-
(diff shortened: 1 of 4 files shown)


docs/root/about_docs.rst
```diff
@@ -1,8 +0,0 @@
-About the documentation
```
junr03: I see how this section could become useful as the docs get bigger. I feel fine about leaving it in, or resurrecting when needed. Up to you.

rebello95: I'd prefer to remove it for now and revive it later if needed (doesn't provide much value as-is IMO)

PR #534, opened 2019-10-23
grpc: support chaining emitters

rebello95:
We should support chaining with gRPC stream emitters the same way we do for HTTP stream emitters.

I.e.,

```
let emitter = ...
  .sendMessage(...)
  .sendMessage(...)
```

This PR also updates tests to use chaining syntax for gRPC.


rebello95:
> Actually, do you mind updating tests for the grpc emitters for this functionality?

Done

PR #539, opened 2019-10-24
android: refactor EnvoyError to Kotlin

rebello95:
`EnvoyError` is currently written in Swift and Java. We discussed options for unifying these types into either Objective-C/Java or Swift/Kotlin, and eventually landed on using Swift/Kotlin in order to keep all public-facing APIs/types written in these high level languages.

- Combines the Java `EnvoyError` and Kotlin `EnvoyException` into one `EnvoyError` Kotlin type that aligns with Swift's `EnvoyError`
- Removes `EnvoyErrorCode` in favor of using an integer like iOS
- Removes unnecessary `@throws` annotations

Resolves https://github.com/lyft/envoy-mobile/issues/409.


library/java/src/io/envoyproxy/envoymobile/engine/types/EnvoyErrorCode.java
```diff
@@ -1,17 +0,0 @@
-package io.envoyproxy.envoymobile.engine.types;
-
-public enum EnvoyErrorCode {
```
goaway: What if we keep this enum, but just move the `EnvoyError` to Kotlin?

That way the code is still enumerated, but the error type that people interact with is still Kotlin.

rebello95: We currently don't have a concept of an `EnvoyErrorCode` on iOS (simply an integer is exposed), so was going for parity here.

I'm totally open to introducing an error code type that we expose on both platforms in Swift/Kotlin if you think that'd be valuable (can either be this PR or a different follow-up).

PR #549, opened 2019-10-24
bazel: enable cacheable binaries

kastiglione:
Description:

When building for debug with Bazel, both local and remote should have local paths excluded from the binary, or at minimum, normalized paths before being embedded in the binary.

This change enables bazel's `debug_prefix_map_pwd_is_dot` feature, which uses [clang's `-fdebug-prefix-map`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fdebug-prefix-map). For details, see https://github.com/bazelbuild/bazel/commit/3f46dd0f6d5a042fc28d265411a6014f666a40c1 and https://github.com/bazelbuild/bazel/issues/5545.

This same feature is already enabled by default for swift.

Additionally for swift, enable the `cacheable_swiftmodules` feature. This ensures that local paths are not embedded in `.swiftmodule` files. For details, see https://github.com/bazelbuild/rules_swift/pull/296.

This change was motivated and identified after seeing developer specific file paths in binaries.

Both of these features should be irrelevant for release builds.

Risk Level: Low
Testing: None

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -5,6 +5,8 @@ build \
   --define=hot_restart=disabled \
   --define=signal_trace=disabled \
   --define=tcmalloc=disabled \
+  --features=debug_prefix_map_pwd_is_dot \
+  --features=swift.cacheable_swiftmodules \
   --ios_minimum_os=10.0 \
   --ios_simulator_device="iPhone X" \
   --ios_simulator_version=13.1 \


rebello95:
Thanks @kastiglione! I'll let @keith take a look oat this one. Please be sure to fix the DCO as well.

kastiglione:
Updated to also cover `.swiftmodule` files.

rebello95:
Re-running CI

kastiglione:
Does CI normally take a while?

rebello95:
We noticed today that GitHub Actions isn't starting CI jobs on PRs from external forks, which is apparently [by design](https://github.community/t5/GitHub-Actions/Run-a-GitHub-action-on-pull-request-for-PR-opened-from-a-forked/td-p/15426). Looking into options

kastiglione:
I have built locally twice:

* `./bazelw build ios_dist --config=ios --config=release`
* `./bazelw build ios_dist --config=ios -c dbg`

For both builds, I confirmed the following:

1. The `Envoy` binary has no debug symbol paths pointing to local paths
2. `Envoy.swiftmodule` has no embedded debugging options (where the paths would end up)

Of interest, even for the debug build, there were no debug symbols pointing to any paths. That might be worth a follow up investigation.

rebello95:
Thanks!

PR #552, opened 2019-10-25
shutdown: more deterministic exit behavior

goaway:
Description:  previously there were still races if the exit call was dispatched. This avoids the dispatch and runs the exit callback on the main thread, while ensuring that nothing else can fire after.


goaway:
Thinking about this yesterday, I realized that there are still races if we dispatch the exit call. This avoids the dispatch and runs exit callback occurs on the main thread, while ensuring that nothing else can fire after.

PR #555, opened 2019-10-28
core: expose engine via weak_ptr

goaway:
Description: Access Engine via weak_ptr so that threads can extend its lifetime for the durations of their calls to it. Despite the fact that all incoming calls are ultimately dispatched to the underlying Envoy main thread, there exists a window where the thread must leverage Engine state (e.g. to access the dispatcher) that can otherwise race with the Engine lifecycle. While it's probably possible to cover these edge cases at the platform layer, this provides a common solution.
Risk Level: Moderate
Testing: local and CI

### DIFF
--- a/library/common/main_interface.cc
+++ b/library/common/main_interface.cc
@@ -71,8 +71,14 @@ envoy_status_t run_engine(envoy_engine_t, envoy_engine_callbacks callbacks, cons
                           const char* log_level) {
   // This will change once multiple engine support is in place.
   // https://github.com/lyft/envoy-mobile/issues/332
-  static auto internal =
+
+  // The shared pointer created here will keep the engine alive until static destruction occurs.
+  static auto strong_ref =
       std::make_shared<Envoy::Engine>(callbacks, config, log_level, preferred_network_);
-  engine_ = internal;
+
+  // The weak pointer we actually expose allows calling threads to atomically check if the engine
+  // still exists and acquire a shared pointer to it - ensuring the engine persists at least for
+  // the duration of the call.
+  engine_ = strong_ref;
   return ENVOY_SUCCESS;
 }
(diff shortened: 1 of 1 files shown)


goaway:
@junr03 So I’m kinda going back on what we talked about Friday - I do think leaking forever would fix our problems for now, but I still think a weak pointer is probably just necessary long-term. Because the engine’s lifecycle will(?) ultimately be tied to the platform wrapper’s lifecycle, we need to be very careful about the timing of incoming calls coming from other (arbitrary) threads. I think we will always want to manage the engine’s state via an exposed weak_ptr that we use to capture shared_ptrs for the duration of calls - to prevent destruction starting while _any_ thread might be touching Engine in its stack.

To make this non-singleton, we convert the internal shard_ptr to one the platform wrapper holds (and eventually resets).

library/common/main_interface.cc
```diff
@@ -56,6 +71,8 @@ envoy_status_t run_engine(envoy_engine_t, envoy_engine_callbacks callbacks, cons
                           const char* log_level) {
   // This will change once multiple engine support is in place.
   // https://github.com/lyft/envoy-mobile/issues/332
-  engine_ = std::make_unique<Envoy::Engine>(callbacks, config, log_level, preferred_network_);
+  static auto internal =
```
junr03: Not a big deal, but for consistency of our statics, should we declare this above as internal_engine_ or strong_engine_, or better_name_?

goaway: Definitely we can change the name, but I did think it made more sense, at present, to encapsulate its declaration in this function call:
- nothing else should ever touch it, and this limits its accessible scope, and
- its lifecycle is more deterministic here.

junr03: Thanks for the explanation. I think the first point is sufficient on its own to leave it here.

goaway:
Thanks for the review :)

PR #557, opened 2019-10-29
cleanup: remove ares init check

goaway:
Description: We don't actually call this check anywhere anymore.
Risk Level: None
Testing: CI


goaway:
@buildbreaker I can't remember if we discussed keeping this around or not, but I just noticed it again, and figured we can always pull it from the commit history if we decide we want it for something.

PR #558, opened 2019-10-30
bazel: Fix bitcode for android binaries

keith:
Because we shared the `release` bazel config between both iOS and
Android, Android builds were incorrectly having llvm bitcode enabled for
them.


.github/workflows/artifacts.yml
```diff
@@ -1,14 +1,6 @@
 name: artifacts
 
-on:
```
rebello95: Assuming this change should be reverted since the intention is for this to only run on Android changes to master?

keith: Yes, just testing here

PR #565, opened 2019-10-30
ci: add job to publish iOS release artifact on merge to master

rebello95:
Adds a job that runs on master and builds/uploads a zip artifact of the `Envoy.framework` for iOS using the release configurations. This should make it easier for consumers to utilize the most recent iOS versions of Envoy Mobile.

Resolves https://github.com/lyft/envoy-mobile/issues/547.

Tested against this PR, which resulted in the expected artifacts being uploaded successfully:

Screen Shot 2019-10-30 at 2 15 05 PM

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -1,24 +1,43 @@
 name: artifacts
 
-on:
-  push:
-    branches:
-      - master
+# on:
+#   push:
+#     branches:
+#       - master
 
 jobs:
-  master_aar_dist:
-    name: master_aar_dist
+  master_android_dist:
+    name: master_android_dist
     runs-on: macOS-10.14
     timeout-minutes: 90
     steps:
       - uses: actions/checkout@v1
         with:
           submodules: true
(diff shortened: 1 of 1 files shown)


junr03:
I thought at some point we had filters on the aar run to only execute when the common or android code was changed in order to prevent wasted CI time on iOS or documentation changes. Do we still want to do something like that, or do we want to create fresh artifacts on _every_ commit to master?

rebello95:
@junr03 please see https://github.com/lyft/envoy-mobile/pull/561 for context 😃

junr03:
> @junr03 please see #561 for context 😃

I see. I guess that is fair. Unfiltered artifacts makes a better experience for maintainers and for users, even if it means burning a bit more CI time.

rebello95:
> I see. I guess that is fair. Unfiltered artifacts makes a better experience for maintainers and for users, even if it means burning a bit more CI time.

Right. I'd rather we optimize for the CI experience we want and obtain the machines as necessary rather than limit our functionality up front

PR #567, opened 2019-10-31
metrics: add configuration for flushing stats via a metrics service

junr03:
Description: this PR adds configuration to flush metrics from Envoy Mobile to a backend metrics service via Envoy's metrics_service API.
Risk Level: low - with respect to the library, higher risk for orgs using the library and ingesting metrics from their mobile clients.
Testing: locally, and with e2e with Lyft's infrastructure. Added unit tests for the new config templating.

### DIFF
--- a/library/common/config_template.cc
+++ b/library/common/config_template.cc
@@ -24,7 +24,7 @@ static_resources:
             inline_string: |
 )"
 #include "certificates.inc"
-R"(
+                              R"(
           verify_subject_alt_name:
             - {{ domain }}
       sni: {{ domain }}
@@ -51,8 +51,45 @@ R"(
       endpoints: *base_endpoints
     tls_context: *base_tls_context
     type: LOGICAL_DNS
+  - name: stats
+    connect_timeout: {{ connect_timeout_seconds }}s
+    dns_refresh_rate: {{ dns_refresh_rate_seconds }}s
+    dns_lookup_family: V4_ONLY
+    http2_protocol_options: {}
+    lb_policy: ROUND_ROBIN
+    load_assignment:
+      cluster_name: base_wwan
+      endpoints:
+        - lb_endpoints:
+            - endpoint:
+                address:
+                  socket_address: {address: {{ stats_domain }}, port_value: 443}
+    tls_context: *base_tls_context
+    type: LOGICAL_DNS
 stats_flush_interval: {{ stats_flush_interval_seconds }}s
+stats_sinks:
+  - name: envoy.metrics_service
+    config:
+      grpc_service:
+        envoy_grpc:
+          cluster_name: stats
+stats_config:
+  stats_matcher:
+    inclusion_list:
+      patterns:
+        - safe_regex:
+            google_re2: {}
+            regex: 'cluster\.[\w]+?\.upstream_rq_total'
+        - safe_regex:
+            google_re2: {}
+            regex: 'cluster\.[\w]+?\.upstream_cx_active'
+        - safe_regex:
+            google_re2: {}
+            regex: 'cluster\.[\w]+?\.upstream_rq_time'
 watchdog:
   megamiss_timeout: 60s
   miss_timeout: 60s
+node:
+  metadata:
+    os: {{ device_os }}
 )";
--- a/library/common/config_template.cc
+++ b/library/common/config_template.cc
@@ -24,7 +24,7 @@ static_resources:
             inline_string: |
 )"
 #include "certificates.inc"
-                              R"(
+R"(
           verify_subject_alt_name:
             - {{ domain }}
       sni: {{ domain }}
@@ -58,7 +58,7 @@ static_resources:
     http2_protocol_options: {}
     lb_policy: ROUND_ROBIN
     load_assignment:
-      cluster_name: base_wwan
+      cluster_name: stats
       endpoints:
         - lb_endpoints:
             - endpoint:
--- a/library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
+++ b/library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
@@ -14,7 +14,7 @@ public class EnvoyConfiguration {
    * @param domain                The domain to use with Envoy (i.e.,
    *                              `api.foo.com`). TODO:
    *                              https://github.com/lyft/envoy-mobile/issues/433
-   * @param statsDomain           The domain to flush stats too.
+   * @param statsDomain           The domain to flush stats to.
    * @param connectTimeoutSeconds timeout for new network connections to hosts in
    *                              the cluster.
    * @param dnsRefreshSeconds     rate in seconds to refresh DNS.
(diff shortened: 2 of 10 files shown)


library/common/config_template.cc
```diff
@@ -51,8 +51,45 @@ R"(
       endpoints: *base_endpoints
     tls_context: *base_tls_context
     type: LOGICAL_DNS
+  - name: stats
```
junr03: I would like to avoid more involved templating for all of this so I feel fine about the library having this, and the config below, even if people are not interested in flushing metrics. However, open to discussion on if we want to opt in this body of config.

library/common/config_template.cc
```diff
@@ -51,8 +51,45 @@ R"(
...
 stats_flush_interval: {{ stats_flush_interval_seconds }}s
+stats_sinks:
+  - name: envoy.metrics_service
+    config:
+      grpc_service:
+        envoy_grpc:
+          cluster_name: stats
+stats_config:
+  stats_matcher:
+    inclusion_list:
```
junr03: initial testing metrics. I wanted to include a counter, a gauge, and a histogram. We can discuss what we want to include, and also if we want to make this an  injectable list.

library/kotlin/test/io/envoyproxy/envoymobile/EnvoyClientBuilderTest.kt
```diff
@@ -29,6 +31,16 @@ class EnvoyBuilderTest {
...
   }
 
+  @Test
+  fun `specifying stats domain overrides default`() {
+    clientBuilder = EnvoyClientBuilder(Domain("api.foo.com"))
+    clientBuilder.addEngineType { engine }
+
+    clientBuilder.addStatsDomain("stats.foo.com")
+    val envoy = clientBuilder.build()
+    assertThat(envoy.envoyConfiguration!!.statsDomain).isEqualTo("stats.foo.com")
```
junr03: @rebello95 I noticed that we don't have a mirror of these tests (ones that test clientBuilder.add<Config>()` in swift, or did I miss them?

library/common/config_template.cc
```diff
@@ -51,8 +51,45 @@ R"(
...
     tls_context: *base_tls_context
     type: LOGICAL_DNS
+  - name: stats
+    connect_timeout: {{ connect_timeout_seconds }}s
+    dns_refresh_rate: {{ dns_refresh_rate_seconds }}s
+    dns_lookup_family: V4_ONLY
+    http2_protocol_options: {}
+    lb_policy: ROUND_ROBIN
+    load_assignment:
+      cluster_name: base_wwan
```
rebello95: This is no longer wwan specific, correct?

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -13,14 +14,16 @@
    * @param domain                The domain to use with Envoy (i.e.,
    *                              `api.foo.com`). TODO:
    *                              https://github.com/lyft/envoy-mobile/issues/433
+   * @param statsDomain           The domain to flush stats too.
```
rebello95:
```suggestion
   * @param statsDomain           The domain to flush stats to.
```

library/objective-c/EnvoyConfiguration.m
```diff
@@ -23,12 +25,15 @@ - (instancetype)initWithDomain:(NSString *)domain
...
     @"domain" : self.domain,
+    @"stats_domain" : self.statsDomain,
     @"connect_timeout_seconds" :
         [NSString stringWithFormat:@"%lu", (unsigned long)self.connectTimeoutSeconds],
     @"dns_refresh_rate_seconds" :
         [NSString stringWithFormat:@"%lu", (unsigned long)self.dnsRefreshSeconds],
     @"stats_flush_interval_seconds" :
-        [NSString stringWithFormat:@"%lu", (unsigned long)self.statsFlushSeconds]
+        [NSString stringWithFormat:@"%lu", (unsigned long)self.statsFlushSeconds],
+    @"device_os" : @"iOS"
```
rebello95: Android uses `android`. Seems like we should be consistent and either do `Android`/`iOS` or `android`/`ios`

library/kotlin/test/io/envoyproxy/envoymobile/EnvoyClientBuilderTest.kt
rebello95: We have similar tests in `EnvoyClientBuilderTests` but it looks like we don't have specific cases for adding individual properties to the builder. Would you mind adding some to that file?

library/common/config_template.cc
```diff
@@ -24,7 +24,7 @@ const char* config_template = R"(
             inline_string: |
 )"
 #include "certificates.inc"
-R"(
+                              R"(
```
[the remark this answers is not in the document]
junr03: my local formatter keeps fighting with the unindented character. AFAICT the CI formatter, and envoy don't care about the indent, so I was going to put it in so I don't need to fight my local formatter.

library/kotlin/test/io/envoyproxy/envoymobile/EnvoyClientBuilderTest.kt
junr03: done.

library/common/config_template.cc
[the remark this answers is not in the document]
junr03: reverted.

library/objective-c/EnvoyConfiguration.m
junr03: went for the former.

junr03:
@rebello95 updated with your comments.

library/swift/test/EnvoyClientBuilderTests.swift
```diff
@@ -69,6 +69,61 @@ final class EnvoyClientBuilderTests: XCTestCase {
...
+    }
+
+    _ = try EnvoyClientBuilder(domain: "api.foo.com")
+      .addEngineType(MockEnvoyEngine.self)
+      .addConnectTimeoutSeconds(12345)
+      .build()
+    self.waitForExpectations(timeout: 0.01)
+  }
+
+    func testAddingDNSRefreshSecondsAddsToConfigurationWhenRunningEnvoy() throws {
```
rebello95: nit: indentation is off

library/swift/test/EnvoyClientBuilderTests.swift
```diff
@@ -69,6 +69,61 @@ final class EnvoyClientBuilderTests: XCTestCase {
...
+    }
+
+    _ = try EnvoyClientBuilder(domain: "api.foo.com")
+      .addEngineType(MockEnvoyEngine.self)
+      .addDNSRefreshSeconds(23)
+      .build()
+    self.waitForExpectations(timeout: 0.01)
+  }
+
+    func testAddingStatsFlushSecondsAddsToConfigurationWhenRunningEnvoy() throws {
```
rebello95: nit: indentation is off

PR #571, opened 2019-10-31
ios: fix leaking of http streams

rebello95:
At the moment, we leak all `EnvoyHTTPStreamImpl` and `EnvoyHTTPCallbacks` instances with every stream due to the deliberate retain cycle used to keep the stream in memory and pass callbacks through to the platform layer while the stream is active. This ends up also leaking the callback closures specified by the end consumer, along with anything captured by those closures.

To resolve this problem, we will release the strong reference cycle that the stream holds to itself when the stream completes.

The state/callback for stream completion has to be stored somewhere in the `ios_context` type and retained since the `ios_context` is a struct and doesn't retain its members.

We considered having the `EnvoyHTTPCallbacks` hold a reference to the `EnvoyHTTPStreamImpl` rather than having `EnvoyHTTPStreamImpl` hold a strong reference to itself. However, this posed a few problems:
- `EnvoyHTTPCallbacks` is designed to support being shared by multiple streams, and thus cannot have only 1 reference to a stream
- `EnvoyHTTPCallbacks` is instantiated by the Swift layer, which means we'd have to leak implementation details of how the stream is kept in memory and shift logic out of the stream implementation

With these in mind, we decided to have `EnvoyHTTPStreamImpl` manage its own lifecycle based on the state of the network stream.

### DIFF
--- a/library/objective-c/EnvoyHTTPStreamImpl.m
+++ b/library/objective-c/EnvoyHTTPStreamImpl.m
@@ -9,6 +9,7 @@
 
 typedef struct {
   EnvoyHTTPCallbacks *callbacks;
+  void (^streamClosed)();
   atomic_bool *canceled;
 } ios_context;
 
@@ -124,13 +125,13 @@ static void ios_on_trailers(envoy_headers trailers, void *context) {
 
 static void ios_on_complete(void *context) {
   ios_context *c = (ios_context *)context;
-  EnvoyHTTPCallbacks *callbacks = c->callbacks;
-  dispatch_async(callbacks.dispatchQueue, ^{
-    if (atomic_load(c->canceled) || !callbacks.onComplete) {
+  void (^streamClosed)() = c->streamClosed;
+  // dispatch_async(callbacks.dispatchQueue, ^{
+    if (atomic_load(c->canceled) || !streamClosed) {
       return;
     }
-    callbacks.onComplete();
-  });
+    streamClosed();
+  // });
 }
 
 static void ios_on_cancel(void *context) {
@@ -188,6 +189,11 @@ static void ios_on_error(envoy_error error, void *context) {
   context->canceled = safe_malloc(sizeof(atomic_bool));
   atomic_store(context->canceled, NO);
 
+  __weak typeof(self) weakSelf = self;
+  context->streamClosed = ^void() {
+    [weakSelf handleStreamClosed];
+  };
+
   // Create native callbacks
   envoy_http_callbacks native_callbacks = {ios_on_headers,  ios_on_data,  ios_on_trailers,
                                            ios_on_metadata, ios_on_error, ios_on_complete,
@@ -203,11 +209,6 @@ static void ios_on_error(envoy_error error, void *context) {
     _strongSelf = nil;
     return nil;
   }
-
-  __weak typeof(self) weakSelf = self;
-  _platformCallbacks.onComplete = ^void() {
-    [weakSelf handleStreamClosed];
-  };
 
   return self;
 }
--- a/library/objective-c/EnvoyHTTPStreamImpl.m
+++ b/library/objective-c/EnvoyHTTPStreamImpl.m
@@ -8,8 +8,8 @@
 #pragma mark - Utility types and functions
 
 typedef struct {
-  EnvoyHTTPCallbacks *callbacks; // Weak
-  id<EnvoyHTTPStream> stream; // Weak
+  EnvoyHTTPCallbacks *callbacks;
+  id<EnvoyHTTPStream> stream;
   atomic_bool *canceled;
 } ios_context;
 
@@ -225,6 +225,7 @@ static void ios_on_error(envoy_error error, void *context) {
 - (void)dealloc {
   ios_context *context = _nativeCallbacks.context;
   context->callbacks = nil;
+  context->stream = nil;
   free(context->canceled);
   free(context);
 }
--- a/library/objective-c/EnvoyHTTPStreamImpl.m
+++ b/library/objective-c/EnvoyHTTPStreamImpl.m
@@ -232,7 +232,8 @@ static void ios_on_error(envoy_error error, void *context) {
 }
 
 - (void)dealloc {
-  ios_context *context = self._nativeCallbacks.context;
+  envoy_http_callbacks native_callbacks = self._nativeCallbacks;
+  ios_context *context = native_callbacks.context;
   free(context->canceled);
   free(context);
 }
--- a/library/objective-c/EnvoyHTTPStreamImpl.m
+++ b/library/objective-c/EnvoyHTTPStreamImpl.m
@@ -85,6 +85,7 @@ static void ios_on_headers(envoy_headers headers, bool end_stream, void *context
     if (atomic_load(c->canceled) || !callbacks.onHeaders) {
       return;
     }
+
     callbacks.onHeaders(to_ios_headers(headers), end_stream);
   });
 }
@@ -96,6 +97,7 @@ static void ios_on_data(envoy_data data, bool end_stream, void *context) {
     if (atomic_load(c->canceled) || !callbacks.onData) {
       return;
     }
+
     callbacks.onData(to_ios_data(data), end_stream);
   });
 }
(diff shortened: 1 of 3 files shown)


library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -126,50 +137,59 @@ static void ios_on_complete(void *context) {
   ios_context *c = (ios_context *)context;
   EnvoyHTTPCallbacks *callbacks = c->callbacks;
   dispatch_async(callbacks.dispatchQueue, ^{
-    // TODO: release stream
-    if (atomic_load(c->canceled)) {
+    if (atomic_load(c->canceled) || !callbacks._onStreamClose) {
```
junr03: should we allow the callbacks to not have this set? i.e if I were writing this in a server I would `ASSERT(callbacks._onStreamClose)` because it would be a programing error not to have a way to clean up this state.

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -126,50 +137,59 @@ static void ios_on_complete(void *context) {
...
   EnvoyHTTPCallbacks *callbacks = c->callbacks;
-  // TODO: release stream
   dispatch_async(callbacks.dispatchQueue, ^{
     // This call is atomically gated at the call-site and will only happen once.
-    if (!callbacks.onCancel) {
-      return;
+    if (callbacks.onCancel) {
+      callbacks.onCancel();
+    }
+    if (callbacks._onStreamClose) {
```
junr03: same as above

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -179,9 +199,16 @@ - (instancetype)initWithHandle:(envoy_stream_t)handle
     return nil;
   }
 
-  _streamHandle = handle;
+  // Release the stream when the connection closes.
```
junr03: nit: Release the strong reference to the stream when the ~connection~ stream is closed

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -126,50 +126,59 @@ static void ios_on_complete(void *context) {
...
-    error.message.release(error.message.context);
-    callbacks.onError(error.error_code, errorMessage);
+    if (callbacks.onError) {
+      NSString *errorMessage = [[NSString alloc] initWithBytes:error.message.bytes
+                                                        length:error.message.length
+                                                      encoding:NSUTF8StringEncoding];
+      error.message.release(error.message.context);
+      callbacks.onError(error.error_code, errorMessage);
+    }
+    if (callbacks._onStreamClose) {
```
junr03: same as above

rebello95:
@goaway @junr03 curious to see your thoughts on this - totally open to suggestions.

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -179,9 +188,16 @@ - (instancetype)initWithHandle:(envoy_stream_t)handle
...
+  // Release the stream when the connection closes.
+  // It must be kept in memory while the stream is active in order to receive callbacks.
+  __weak typeof(self) weakSelf = self;
+  callbacks._onStreamClose = ^void() {
+    weakSelf._strongSelf = nil;
+  };
+
   // Retain platform callbacks
-  _platformCallbacks = callbacks;
+  self._platformCallbacks = callbacks;
```
junr03: should we always use `self.`?

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -126,50 +126,59 @@ static void ios_on_complete(void *context) {
...
-@implementation EnvoyHTTPStreamImpl {
-  EnvoyHTTPStreamImpl *_strongSelf;
-  EnvoyHTTPCallbacks *_platformCallbacks;
-  envoy_http_callbacks _nativeCallbacks;
-  envoy_stream_t _streamHandle;
-}
+@interface EnvoyHTTPStreamImpl ()
+
+@property (nonatomic, strong) EnvoyHTTPStreamImpl *_strongSelf;
+@property (nonatomic, strong) EnvoyHTTPCallbacks *_platformCallbacks;
```
junr03: Will dealloc get rid of this strong ref automatically? Or do we need to assign nil explicitly?

library/objective-c/EnvoyHTTPStreamImpl.m
rebello95: Yep that SGTM

library/objective-c/EnvoyHTTPStreamImpl.m
rebello95: Since these changed from ivars to properties, `self` is now required to use the generated getters from the compiler. I made `_strongSelf` a property so that it could be accessed within the closure above, then updated the rest for consistency. We can change the rest back if we want to

library/objective-c/EnvoyHTTPStreamImpl.m
rebello95: ivars use `__strong` references for objects by default even though it's not specified explicitly, so the previous references were also strong. To answer your question, when the type holding the reference (`EnvoyHTTPStreamImpl`) goes out of scope (`dealloc` is called), it releases its references to those objects

goaway:
@rebello95 thanks for tracking this down!

I wanted to suggest a similar, but alternative approach that would let us keep everything as ivars, and offer a somewhat more direct mechanism for releasing the stream.

What do you think about having EnvoyHTTPCallbacks hold a strong reference to EnvoyHTTPStream and create the ref-cycle that way (dropping strongSelf on EnvoyHTTPStream). Then the only thing that needs to happen in error/cancel/final callbacks is EnvoyHTTPCallbacks unsets its EnvoyHTTPStream ref.

I think this will result in less code and less indirection. What I also like about this pattern is that EnvoyHTTPCallbacks is the true state that the core layer needs to track, and in that sense the dependency relationship seems more correct.

rebello95:
Thanks for the reviews @junr03 and @goaway!

Per discussions offline, I've updated the PR and description to reflect the thought process and state of the current diff.

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -118,47 +122,56 @@ static void ios_on_trailers(envoy_headers trailers, void *context) {
...
+    if (atomic_load(c->canceled)) {
       return;
     }
-    NSString *errorMessage = [[NSString alloc] initWithBytes:error.message.bytes
-                                                      length:error.message.length
-                                                    encoding:NSUTF8StringEncoding];
-    error.message.release(error.message.context);
-    callbacks.onError(error.error_code, errorMessage);
+
+    if (callbacks.onError) {
```
junr03: Still have the comment about: https://github.com/lyft/envoy-mobile/pull/571#discussion_r341588248. But, it is not introduced by this PR so we can deal with it in a subsequent PR.

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -8,7 +8,8 @@
 #pragma mark - Utility types and functions
 
 typedef struct {
-  EnvoyHTTPCallbacks *callbacks;
+  __unsafe_unretained EnvoyHTTPCallbacks *callbacks;
+  __unsafe_unretained EnvoyHTTPStreamImpl *stream;
```
junr03: Do you mind putting a comment here that explains why we store this ref back to the stream, and how the lifetime of the stream is tied to the context rather than the callback?

library/objective-c/EnvoyHTTPStreamImpl.m
rebello95: I'll add the assertions.

library/objective-c/EnvoyHTTPStreamImpl.m
rebello95: I had it in the PR description but will add here too

rebello95:
@junr03 addressed your comments, PTAL

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -8,8 +8,13 @@
 #pragma mark - Utility types and functions
 
 typedef struct {
-  __unsafe_unretained EnvoyHTTPCallbacks *callbacks;
+  // The stream is kept in memory through a strong reference to itself. In order to free the
```
junr03: Super clear, thanks!

PR #577, opened 2019-11-05
ios: fix podspec for v0.2.0 support

rebello95:
Makes several updates to our podspec in order to fix CocoaPods integration with v0.2.0:

- Link `resolv.9` and `c++` which is being done with Bazel today
- Link `SystemConfiguration` which is being done with Bazel today
- Specify Swift version and platform version
- Specify `source_files` **with an empty `.swift` file** in order to ensure that Swift libraries are linked when building (i.e., `swiftFoundation`)

**Important notes:**
Releasing new versions of the pod requires the following manual changes:
- Unzipping `envoy_ios_framework.zip`
- Renaming the directory to `envoy_ios_cocoapods`
- Creating an empty file at `envoy_ios_cocoapods/Envoy.framework/Swift/Empty.swift`. This forces CocoaPods to link Swift libraries as necessary. We worked around this in Bazel similarly in the past [here](https://github.com/lyft/envoy-mobile/commit/94ada57a23b5e3769adbd1623b3bf863e22d9fdd#diff-6dc94efb18b54c46a32898ba3a5a0756R15)
- Copying the repo's `LICENSE` file and placing it at `envoy_ios_cocoapods/LICENSE`
- Re-zipping `envoy_ios_cocoapods` and uploading `envoy_ios_cocoapods.zip`

In the future, we should see if this can be simplified or at the very least add a script that uploads this artifact with each commit to master just like we do with other artifacts. This is being tracked in https://github.com/lyft/envoy-mobile/issues/578.

Note: There is a similar problem reported in CocoaPods that requires this empty Swift file workaround: https://github.com/CocoaPods/CocoaPods/issues/8649.

I tested this PR by using this podspec for building an app against Envoy Mobile with CocoaPods. This spec is currently published as v0.2.0.

### DIFF
--- a/EnvoyMobile.podspec
+++ b/EnvoyMobile.podspec
@@ -6,12 +6,12 @@ Pod::Spec.new do |s|
     s.homepage = 'https://envoy-mobile.github.io'
     s.documentation_url = 'https://envoy-mobile.github.io/docs/envoy-mobile/latest/index.html'
     s.social_media_url = 'https://twitter.com/EnvoyProxy'
-    s.license = { :type => 'Apache-2.0', :file => 'envoy_ios_cocoapods/LICENSE' }
+    s.license = { type: 'Apache-2.0', file: 'envoy_ios_cocoapods/LICENSE' }
     s.platform = :ios, '10.0'
     s.swift_version = '5.1'
     s.libraries = 'resolv.9', 'c++'
     s.frameworks = 'SystemConfiguration'
-    s.source = { :http => "https://github.com/lyft/envoy-mobile/releases/download/v#{s.version}/envoy_ios_cocoapods.zip" }
+    s.source = { http: "https://github.com/lyft/envoy-mobile/releases/download/v#{s.version}/envoy_ios_cocoapods.zip" }
     s.vendored_frameworks = 'envoy_ios_cocoapods/Envoy.framework'
     s.source_files = 'envoy_ios_cocoapods/Envoy.framework/Headers/*.h', 'envoy_ios_cocoapods/Envoy.framework/Swift/*.swift'
 end
--- a/EnvoyMobile.podspec
+++ b/EnvoyMobile.podspec
@@ -1,17 +1,17 @@
 Pod::Spec.new do |s|
-    s.name                = 'EnvoyMobile'
-    s.version             = '0.2.0'
-    s.author              = 'Lyft, Inc.'
-    s.summary             = 'Client networking libraries based on the Envoy project'
-    s.homepage            = 'https://envoy-mobile.github.io'
-    s.documentation_url   = 'https://envoy-mobile.github.io/docs/envoy-mobile/latest/index.html'
-    s.social_media_url    = 'https://twitter.com/EnvoyProxy'
-    s.license             = { :type => 'Apache-2.0', :file => 'envoy_ios_cocoapods/LICENSE' }
-    s.platform            = :ios, '10.0'
-    s.swift_version       = '5.1'
-    s.libraries           = 'resolv.9', 'c++'
-    s.frameworks          = 'SystemConfiguration'
-    s.source              = { :http => "https://github.com/lyft/envoy-mobile/releases/download/v#{s.version}/envoy_ios_cocoapods.zip" }
+    s.name = 'EnvoyMobile'
+    s.version = '0.2.0'
+    s.author = 'Lyft, Inc.'
+    s.summary = 'Client networking libraries based on the Envoy project'
+    s.homepage = 'https://envoy-mobile.github.io'
+    s.documentation_url = 'https://envoy-mobile.github.io/docs/envoy-mobile/latest/index.html'
+    s.social_media_url = 'https://twitter.com/EnvoyProxy'
+    s.license = { :type => 'Apache-2.0', :file => 'envoy_ios_cocoapods/LICENSE' }
+    s.platform = :ios, '10.0'
+    s.swift_version = '5.1'
+    s.libraries = 'resolv.9', 'c++'
+    s.frameworks = 'SystemConfiguration'
+    s.source = { :http => "https://github.com/lyft/envoy-mobile/releases/download/v#{s.version}/envoy_ios_cocoapods.zip" }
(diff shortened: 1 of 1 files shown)


EnvoyMobile.podspec
```diff
@@ -1,13 +1,17 @@
...
-    s.author            = 'Lyft, Inc.'
-    s.summary           = 'Client networking libraries based on the Envoy project'
-    s.homepage          = 'https://envoy-mobile.github.io'
-    s.documentation_url = 'https://envoy-mobile.github.io/docs/envoy-mobile/latest/index.html'
-    s.social_media_url  = 'https://twitter.com/EnvoyProxy'
-    s.license           = { :type => 'Apache-2.0', :file => 'LICENSE' }
-    s.platform          = :ios
-    s.source            = { :http => "https://github.com/lyft/envoy-mobile/releases/download/v#{s.version}/envoy_ios_framework.zip" }
-    s.ios.vendored_frameworks = 'Envoy.framework'
+    s.name                = 'EnvoyMobile'
```
keith: IMO you might as well not align these like we don't in our code just because of this noise it creates

EnvoyMobile.podspec
```diff
@@ -1,13 +1,17 @@
...
+    s.version             = '0.2.0'
+    s.author              = 'Lyft, Inc.'
+    s.summary             = 'Client networking libraries based on the Envoy project'
+    s.homepage            = 'https://envoy-mobile.github.io'
+    s.documentation_url   = 'https://envoy-mobile.github.io/docs/envoy-mobile/latest/index.html'
+    s.social_media_url    = 'https://twitter.com/EnvoyProxy'
+    s.license             = { :type => 'Apache-2.0', :file => 'envoy_ios_cocoapods/LICENSE' }
```
keith:
```suggestion
    s.license             = { type: 'Apache-2.0', file: 'envoy_ios_cocoapods/LICENSE' }
```

EnvoyMobile.podspec
```diff
@@ -1,13 +1,17 @@
...
+    s.platform            = :ios, '10.0'
+    s.swift_version       = '5.1'
+    s.libraries           = 'resolv.9', 'c++'
+    s.frameworks          = 'SystemConfiguration'
+    s.source              = { :http => "https://github.com/lyft/envoy-mobile/releases/download/v#{s.version}/envoy_ios_cocoapods.zip" }
```
keith:
```suggestion
    s.source              = { http: "https://github.com/lyft/envoy-mobile/releases/download/v#{s.version}/envoy_ios_cocoapods.zip" }
```

EnvoyMobile.podspec
```diff
@@ -1,13 +1,17 @@
...
+    s.vendored_frameworks = 'envoy_ios_cocoapods/Envoy.framework'
+    s.source_files        = 'envoy_ios_cocoapods/Envoy.framework/Headers/*.h', 'envoy_ios_cocoapods/Envoy.framework/Swift/*.swift'
```
keith: You shouldn't need source files since you're vendoring a framework?

EnvoyMobile.podspec
rebello95: Kinda thought it was more readable but yea we can stop aligning to reduce churn

EnvoyMobile.podspec
rebello95: Without this the framework fails `pod spec lint` because it's missing `libSwiftFoundation`, among other Swift libraries. I added some more explanations to the PR description - without this and manually adding a `.swift` file, the pod doesn't compile. Totally open to better ways to do this

keith: Is there a corresponding CocoaPods issue about this? Seems weird that no one else would have this case

rebello95: Not from what I could tell, but I'm happy to file one. Assuming it's only an issue for mixed Objective-C/Swift libraries?

keith: CP likely doesn't check that either way, but maybe you adding the `swift_version` was what broke it?

rebello95: Commenting out `swift_version` and `source_files` yields the following errors:

```
   Testing with `xcodebuild`. 
 -> EnvoyMobile (0.2.0)
    - ERROR | [iOS] xcodebuild: Returned an unsuccessful exit code.
    - NOTE  | xcodebuild:  note: Using new build system
    - NOTE  | [iOS] xcodebuild:  note: Planning build
    - NOTE  | [iOS] xcodebuild:  note: Constructing build description
    - NOTE  | [iOS] xcodebuild:  warning: Skipping code signing because the target does not have an Info.plist file and one is not being generated automatically. (in target 'App' from project 'App')
    - NOTE  | xcodebuild:  ld: warning: Could not find or use auto-linked library 'swiftFoundation'
    - NOTE  | xcodebuild:  ld: warning: Could not find or use auto-linked library 'swiftCompatibilityDynamicReplacements'
    - NOTE  | xcodebuild:  ld: warning: Could not find or use auto-linked library 'swiftDarwin'
    - NOTE  | xcodebuild:  ld: warning: Could not find or use auto-linked library 'swiftCoreFoundation'
    - NOTE  | xcodebuild:  ld: warning: Could not find or use auto-linked library 'swiftCore'
    - NOTE  | xcodebuild:  ld: warning: Could not find or use auto-linked library 'swiftCoreGraphics'
    - NOTE  | xcodebuild:  ld: warning: Could not find or use auto-linked library 'swiftObjectiveC'
    - NOTE  | xcodebuild:  ld: warning: Could not find or use auto-linked library 'swiftDispatch'
    - NOTE  | xcodebuild:  ld: warning: Could not find or use auto-linked library 'swiftSwiftOnoneSupport'
    - NOTE  | xcodebuild:  ld: warning: Could not find or use auto-linked library 'swiftCompatibility50'
    - NOTE  | xcodebuild:  clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

rebello95: Mind taking another look whenever you get a chance @keith?

rebello95:
Yep, agree. Looks like https://github.com/CocoaPods/CocoaPods/issues/8649 is actually this problem and their workaround was also to include a `.swift` file. I'll link that issue in this commit

PR #583, opened 2019-11-08
ios: make stream cleanup (more) thread-safe

goaway:
Description: Fixes a race in cleanup timing by ensuring cleanup is ALWAYS dispatched from Envoy's main thread to the callback thread for a given stream, AFTER reset or completion has been handled at the common layer.
Risk Level: Low
Testing: CI

### DIFF
--- a/library/objective-c/EnvoyHTTPStreamImpl.m
+++ b/library/objective-c/EnvoyHTTPStreamImpl.m
@@ -137,14 +137,16 @@ static void ios_on_complete(void *context) {
   EnvoyHTTPCallbacks *callbacks = c->callbacks;
   EnvoyHTTPStreamImpl *stream = c->stream;
   dispatch_async(callbacks.dispatchQueue, ^{
-    // FIXME: If the callback queue is not serial, clean up is not currently thread-safe.
+    // TODO: If the callback queue is not serial, clean up is not currently thread-safe.
     assert(stream);
     [stream cleanUp];
   });
 }
 
 static void ios_on_cancel(void *context) {
-  // This call is atomically gated at the call-site and will only happen once.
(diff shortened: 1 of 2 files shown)


library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -166,18 +164,15 @@ static void ios_on_error(envoy_error error, void *context) {
...
-    if (callbacks.onError) {
+    if (!atomic_load(c->canceled) && callbacks.onError) {
       NSString *errorMessage = [[NSString alloc] initWithBytes:error.message.bytes
                                                         length:error.message.length
                                                       encoding:NSUTF8StringEncoding];
       error.message.release(error.message.context);
       callbacks.onError(error.error_code, errorMessage);
     }
 
+    // FIXME: If the callback queue is not serial, clean up is not currently thread-safe.
```
rebello95: Same here

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -137,27 +137,25 @@ static void ios_on_complete(void *context) {
...
-    }
-
+    // FIXME: If the callback queue is not serial, clean up is not currently thread-safe.
     assert(stream);
     [stream cleanUp];
   });
 }
 
 static void ios_on_cancel(void *context) {
+  // This call is atomically gated at the call-site and will only happen once.
```
rebello95: I filed an issue to add a flowchart for this so behavior is clearer: https://github.com/lyft/envoy-mobile/issues/584

rebello95: It'd also be good if we could outline here that `onCancel()` can be propogated through callbacks even after a stream is technically completed

PR #600, opened 2019-11-26
update: envoy ref and bazel

junr03:
Description: updating bazel to 1.2.0 and envoy ref. This change also removes the ./bazelw script and makes bazelisk the official bazel manager.
Risk Level: low
Testing: CI and local build


.github/workflows/android.yml
```diff
@@ -21,16 +21,16 @@ jobs:
           export CC=clang
           export CXX=clang++
           export ANDROID_NDK_HOME=/usr/local/lib/android/sdk/ndk-bundle
-          ./bazelw build --fat_apk_cpu=x86 //:android_dist
+          bazel build --fat_apk_cpu=x86 //:android_dist
```
buildbreaker: _Feels_ like we should be using `bazelisk` here

buildbreaker:
Not entirely sure if this is because we are using bazel's recent release. There is some cc + java rule interactions which have changed

buildbreaker:
I do see the symbol for `initEngine` when I inspect the `.so` file in the `.aar`: `Java_io_envoyproxy_envoymobile_engine_JniLibrary_initEngine`

I'll have to investigate further to understand what is actually happening here

buildbreaker:
Fun news: it breaks in the example applications but if I depend on the aar with our Lyft application, it doesn't break. Makes me believe that this has something to do with our android_binary bazel rule configurations

bazel/aar_with_jni.bzl
```diff
@@ -52,6 +52,7 @@ EOF
         name = archive_name + "_jni",
         manifest = archive_name + "_generated_AndroidManifest.xml",
         custom_package = "does.not.matter",
+        # TODO: this is a hack to fix envoy builds: https://github.com/lyft/envoy-mobile/issues/604
```
buildbreaker: Remove?

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
```diff
@@ -7,6 +7,9 @@ aar_with_jni(
     name = "android_aar",
     android_library = "envoy_lib",
     archive_name = "envoy",
+    native_deps = [
```
buildbreaker: 👍

PR #606, opened 2019-12-02
build: temporary distdir

junr03:
Description: using distdir temporarily to fetch tclap while we firgure out #600 and update envoy.
Risk Level: low
Testing: CI


junr03:
the `perf / size_master` job will fail as master is broken

PR #609, opened 2019-12-04
bazel: clean up android bazel rules

buildbreaker:
The gist here is that we had a lot of macros which transformed any jvm target's sources into a `*_src` target. This was helpful to allow us to aggregate sources in order to be sure that we have all the sources at the end result `aar`. Additionally, we use this to get around the `internal` keyword usages in some of our tests (they're mostly kotlin tests).

This change is simplifying the former cases (removing any unneeded repeating of target sources). In addition, I've cleaned up our `aar_with_jni` macro (or made it more complex) by having it own any odd/unexpected logic:
1. Resolves the `.so` `cc_binary` rules
2. Create our base files (an apk) using the `.so` file 
3. Create our backing jar via `android_binary` (it ends up being our `classes.jar`)
4. Zips everything up

Description: bazel: clean up android bazel rules
Risk Level: low
Testing: local/ci


bazel/kotlin_lib.bzl
```diff
@@ -1,59 +1,5 @@
...
-    )
-
-    native.java_library(
-        name = name,
-        srcs = srcs,
-        visibility = visibility,
-        deps = deps,
-    )
-
 def envoy_mobile_kt_library(name, visibility = None, srcs = [], deps = []):
```
buildbreaker: We still use this for tests since our tests sometimes leverage `internal` functionalities to exercise some code paths

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyClientBuilder.kt
```diff
@@ -11,7 +11,7 @@ sealed class BaseConfiguration(
 class Domain(domain: String) : BaseConfiguration(domain)
 class Yaml(yaml: String) : BaseConfiguration(yaml)
 
-open class EnvoyClientBuilder internal constructor(
+open class EnvoyClientBuilder(
```
buildbreaker: Changing the visibility since the Android flavor uses these outside it's module

bazel/aar_with_jni.bzl
```diff
@@ -30,10 +30,22 @@ load("@google_bazel_common//tools/maven:pom_file.bzl", "pom_file")
...
 # kt.jar file, and replaces the aar's classes.jar with the kotlin jar
-def aar_with_jni(name, android_library, proguard_rules = "", archive_name = "", visibility = None, native_deps = []):
-    if not archive_name:
-        archive_name = name
+def aar_with_jni(name, android_library, manifest, archive_name, native_deps = [], proguard_rules = "", visibility = []):
+    manifest_name = name + "_android_manifest"
+    android_binary_name = name + "_bin"
+    jni_archive_name = archive_name + "_jni"
+    pom_name = name + "_pom"
+    cc_lib_name = name + "_jni_interface_lib"
```
goaway: I think this implicitly creating a dependency on the name of an external rule.(?) I'd propose we simply pass the name of the target as a parameter.

goaway:
thank you, great work

Issue #584, opened 2019-11-08
docs: add flowchart diagram for stream lifecycle


rebello95:
To make changes like https://github.com/lyft/envoy-mobile/pull/583 clearer, we can add a flowchart of the lifecycle of streams:

- Requests
- Filters (once implemented)
- Stream callbacks
- Terminal state
- Cancelations

PR #610, opened 2019-12-04
ios: only fire a single 'terminal' callback per stream

goaway:
Description: Updates callback logic to ensure only one of any terminal callback is fired per stream/requests: one of final-onHeaders, final-onData, onTrailers, onError, and onCancel. Additionally ensures that it will be the last callback (iff provided callback dispatch is serial).
Risk: Low
Testing: Local and CI

### DIFF
--- a/library/objective-c/EnvoyHTTPStreamImpl.m
+++ b/library/objective-c/EnvoyHTTPStreamImpl.m
@@ -263,7 +263,7 @@ static void ios_on_error(envoy_error error, void *context) {
   ios_context *context = _nativeCallbacks.context;
   // Step 1: atomically and synchronously prevent the execution of further callbacks other than
   // on_cancel.
-  if (!atomic_exchange(context->closed, YES)) {
+  if (!dispatchable(context->closed, YES)) {
     // Step 2: directly fire the cancel callback.
     ios_on_cancel(context);
     // Step 3: propagate the reset into native code.
--- a/library/objective-c/EnvoyHTTPStreamImpl.m
+++ b/library/objective-c/EnvoyHTTPStreamImpl.m
@@ -84,7 +84,8 @@ static EnvoyHeaders *to_ios_headers(envoy_headers headers) {
 
 #pragma mark - C callbacks
 
-// Return whether a callback should be allowed to continue with execution.
+// Return whether a callback should be allowed to continue with execution. This ensures at most one
+// 'terminal' callback is issued for any given stream.
 static bool dispatchable(atomic_bool *closed, bool close) {
   if (close) {
     // Set closed to true and return true if not previously closed.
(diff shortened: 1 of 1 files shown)


library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -165,7 +175,7 @@ static void ios_on_error(envoy_error error, void *context) {
   EnvoyHTTPCallbacks *callbacks = c->callbacks;
   EnvoyHTTPStreamImpl *stream = c->stream;
   dispatch_async(callbacks.dispatchQueue, ^{
-    if (!atomic_load(c->canceled) && callbacks.onError) {
+    if (!atomic_load(c->closed) && callbacks.onError) {
```
junr03: shouldn't error also close?

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -84,11 +84,21 @@ static envoy_headers toNativeHeaders(EnvoyHeaders *headers) {
 
 #pragma mark - C callbacks
 
+// Return whether a callback should be allowed to continue with execution.
```
junr03: Do you mind adding a description here about the new logic. i.e this is done in order to ensure that only one 'terminal' callback is fired.

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -252,13 +262,15 @@ - (int)cancel {
   ios_context *context = _nativeCallbacks.context;
   // Step 1: atomically and synchronously prevent the execution of further callbacks other than
   // on_cancel.
-  if (!atomic_exchange(context->canceled, YES)) {
+  if (!atomic_exchange(context->closed, YES)) {
```
junr03: why not use dispatchable here?

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -252,13 +262,15 @@ - (int)cancel {
...
     // Step 2: directly fire the cancel callback.
     ios_on_cancel(context);
     // Step 3: propagate the reset into native code.
     reset_stream(_streamHandle);
     return 0;
   } else {
+    // Propagate reset, but don't worry about callbacks.
```
junr03: I would expand the "don't worry about callbacks" to mention that another terminal callback has fired to the platform so all we need to do here is propagate the reset to the engine.

rebello95:
Would you mind also adding a PR description? Since the stream lifecycle has come up a few times, maybe it makes sense to take care of https://github.com/lyft/envoy-mobile/issues/584 now too?

goaway:
@rebello95 yep was planning to add description before merging; let me know if what I added above looks good.

I agree #584 would be super useful, but due to the number of things in flight, prefer to still keep it as a separate issue.

library/objective-c/EnvoyHTTPStreamImpl.m
goaway: good point

PR #611, opened 2019-12-04
android: only fire a single 'terminal' callback per stream

goaway:
Description: Updates callback logic to ensure only one of any terminal callback is fired per stream/requests: one of final-onHeaders, final-onData, onTrailers, onError, and onCancel. Additionally ensures that it will be the last callback (iff provided callback dispatch is serial).
Risk: Low
Testing: Local and CI

### DIFF
--- a/library/java/src/io/envoyproxy/envoymobile/engine/JvmCallbackContext.java
+++ b/library/java/src/io/envoyproxy/envoymobile/engine/JvmCallbackContext.java
@@ -17,7 +17,7 @@ class JvmCallbackContext {
     TRAILERS,
   }
 
-  private final AtomicBoolean canceled = new AtomicBoolean(false);
+  private final AtomicBoolean closed = new AtomicBoolean(false);
   private final EnvoyHTTPCallbacks callbacks;
 
   // State-tracking for header accumulation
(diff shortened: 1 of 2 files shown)


junr03:
related iOS change #610

library/java/src/io/envoyproxy/envoymobile/engine/JvmCallbackContext.java
```diff
@@ -29,6 +29,20 @@
 
   public JvmCallbackContext(EnvoyHTTPCallbacks callbacks) { this.callbacks = callbacks; }
 
+  /**
+   * Return whether a callback should be allowed to continue with execution. This ensures at most
+   * one 'terminal' callback is issued for any given stream.
+   *
+   * @param close, whether the stream should be closed as part of this determination.
+   */
+  private dispatchable(boolean close) {
```
junr03: missing return type

library/java/src/io/envoyproxy/envoymobile/engine/JvmCallbackContext.java
```diff
@@ -167,14 +181,16 @@ public void run() {
...
-   * @return boolean, whether the callback context was canceled or not.
+   * @return boolean, whether the callback context was closed or not.
    */
   public boolean cancel() {
-    boolean canceled = !this.canceled.getAndSet(true);
-    if (canceled) {
+    // Atomically close the stream if not already closed.
+    boolean closed = dispatchable(true);
+    if (closed) {
+      // Directly fire callback if closure occurred.
```
junr03: can you remind me why we don't reset the stream here?

library/java/src/io/envoyproxy/envoymobile/engine/JvmCallbackContext.java
goaway: (already fixed)

library/java/src/io/envoyproxy/envoymobile/engine/JvmCallbackContext.java
goaway: https://github.com/lyft/envoy-mobile/blob/ms/java-termcbs/library/java/src/io/envoyproxy/envoymobile/engine/EnvoyHTTPStream.java#L80

goaway: (but thank you for helping me catch a bug)

PR #613, opened 2019-12-06
android: remove kotlin bundling from aar

buildbreaker:
We're packaging kotlin with our aar classes.jar and in the aar itself.

This change is to remove that by manually excluding specific directories from the `deploy.jar`. Additionally, we removed some unnecessary artifacts from the aar itself. Here's what it looks like now:
```
Archive:  envoy.aar
  AndroidManifest.xml     
  classes.jar             
  jni/
  jni/armeabi-v7a/
  jni/armeabi-v7a/libenvoy_jni.so  
  jni/x86/
  jni/x86/libenvoy_jni.so  
  jni/arm64-v8a/
  jni/arm64-v8a/libenvoy_jni.so  
  proguard.txt     
```

https://developer.android.com/studio/projects/android-library

Description: android: remove kotlin bundling from aar
Risk Level: low
Testing: local
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -131,9 +131,9 @@ http_archive(
     urls = ["https://github.com/google/bazel-common/archive/413b433b91f26dbe39cdbc20f742ad6555dd1e27.zip"],
 )
 
-rules_kotlin_version = "legacy-1.3.0-rc1"
+rules_kotlin_version = "legacy-1.3.0-rc2"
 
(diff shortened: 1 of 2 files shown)


rebello95:
I think the submodule update is accidental?

buildbreaker:
fixed

PR #614, opened 2019-12-12
config: experiment with some moderately aggressive keepalive settings

goaway:
Description: Because we associate connections to a network interface heuristically and because we don't currently have a good request-independent signal for connection health (like h2 pings), TCP keepalives may be the next best thing. This change aims to address reports of hung requests by setting fairly aggressive keepalive settings to detect dead connections.

After a connection is idle for 5 seconds, we'll begin sending keepalives on a 10 second interval. If any go unacked in that interval, we'll terminate the connection.

Eventually we'll want to tune this and/or make it configurable and/or rely on something at L7 (like h2 pings).

Risk Level: Low
Testing: In the wild

### DIFF
--- a/library/common/config_template.cc
+++ b/library/common/config_template.cc
@@ -28,6 +28,11 @@ R"(
             - {{ domain }}
       sni: {{ domain }}
     type: LOGICAL_DNS
+    upstream_connection_options: &upstream_opts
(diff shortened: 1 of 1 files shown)


goaway:
Certainly :)

PR #616, opened 2019-12-19
core: integration with Envoy's API listener

### LINKED ISSUE #543: core: support L7 C++ filter chain
Envoy Mobile's current use of Envoy's async client doesn't currently allow for the full L7 C++ filter chain to be executed.

We plan to add support for these filters in order to build new L7 filters and allow consumers to create/register their own.

junr03:
Description: this PR integrates [Envoy's ApiListener](https://github.com/envoyproxy/envoy/pull/9516) to replace Envoy Mobile's use of the AsyncClient. Doing so gives Envoy Mobile access to all L7 facilities provided by Envoy's Http Connection Manager; including but no limited to the L7 filters.
Risk Level: HIGH! This PR changes the underlying code for Envoy Mobile's core functionality
Testing: unit tests updated for the dispatcher, integration tests, and e2e tests with the example apps. Additionally, the ApiListener code has been unit and integration tested in Envoy.

Fixes #543


junr03:
FYI @rebello95 @buildbreaker

library/common/config_template.cc
```diff
@@ -17,13 +48,13 @@ const char* config_template = R"(
...
     tls_context: &base_tls_context
-      common_tls_context:
+      common_tls_context: &base_common_tls_context
         validation_context:
           trusted_ca:
             inline_string: |
 )"
 #include "certificates.inc"
-R"(
+                              R"(
```
goaway: whitespace?

library/common/config_template.cc
```diff
@@ -54,7 +86,19 @@ const char* config_template = R"(
       cluster_name: base_wwan
       endpoints: *base_endpoints
     transport_socket: *base_transport_socket
+    upstream_connection_options: *upstream_opts
     type: LOGICAL_DNS
+  - name: dynamic_forward_proxy_cluster
```
goaway: IMO we should use this to replace the base and base_* clusters. It's really more the standard use case.

library/common/http/dispatcher.cc
```diff
@@ -8,39 +8,55 @@
...
+                "[S{}] dispatching to platform response headers for stream (end_stream={}):\n{}",
+                direct_stream_.stream_handle_, end_stream, headers);
+      bridge_callbacks_.on_headers(bridge_headers, end_stream, bridge_callbacks_.context);
+    }
+    closeRemote(end_stream);
     return;
   }
 
-  // We assume that local replies represent error conditions, having audited occurrences in
-  // Envoy today. This is not a good long-term solution.
```
goaway: Why drop this comment?

library/common/http/dispatcher.cc
```diff
@@ -49,75 +65,152 @@ void Dispatcher::DirectStreamCallbacks::onHeaders(HeaderMapPtr&& headers, bool e
...
+      ENVOY_LOG(debug,
+                "[S{}] dispatching to platform response data for stream (length={} end_stream={})",
+                direct_stream_.stream_handle_, data.length(), end_stream);
+      bridge_callbacks_.on_data(Buffer::Utility::toBridgeData(data), end_stream,
+                                bridge_callbacks_.context);
+    }
+    closeRemote(end_stream);
   } else {
+    // FIXME: before merge. should this be an ASSERT? Yes, if we are in the local response path,
+    // then Envoy finishes the sequence with an encodeData. What if it later does trailers?
```
goaway: Seems a little bit unlikely to me. But we could change this to a comment saying "This assert helps guard our reliance on current convention: an Envoy-generated error response always ends with a single data frame."

library/common/http/dispatcher.cc
```diff
@@ -49,75 +65,152 @@ void Dispatcher::DirectStreamCallbacks::onHeaders(HeaderMapPtr&& headers, bool e
...
   envoy_error_code_t code = error_code_.value_or(ENVOY_STREAM_RESET);
   envoy_data message = error_message_.value_or(envoy_nodata);
-  bridge_callbacks_.on_error({code, message}, bridge_callbacks_.context);
-  // Very important: onComplete and onReset both clean up stream state in the http dispatcher
-  // because the underlying async client implementation **guarantees** that only onComplete **or**
-  // onReset will be fired for a stream. This means it is safe to clean up the stream when either of
-  // the terminal callbacks fire without keeping additional state in this layer.
-  http_dispatcher_.cleanup(stream_handle_);
+  // Note: in the case that we received a complete remote response but envoy is resetting the stream
+  // due to an incomplete local request this on_error call will happen after an on_complete.
```
goaway:
```suggestion
  // due to an incomplete local request, this on_error call will happen after an on_complete.
```

library/common/http/dispatcher.cc
```diff
@@ -49,75 +65,152 @@ void Dispatcher::DirectStreamCallbacks::onHeaders(HeaderMapPtr&& headers, bool e
...
+}
+
+void Dispatcher::DirectStreamCallbacks::onCancel() {
+  // This call is guarded at the call-site @see Dispatcher::DirectStream::resetStream().
+  // Therefore, it is dispatched here without protection.
+  ENVOY_LOG(debug, "[S{}] dispatching to platform cancel stream", direct_stream_.stream_handle_);
+  bridge_callbacks_.on_cancel(bridge_callbacks_.context);
+
+  // FIXME: before merge. what about clean up? It all depends on if the reset we send down causes a
+  // reset coming up, which will not fire a callback, but will result in stream clean up.
```
goaway: I'd defer changing logic here, since we know we plan to move it down into the core layer.

library/common/http/dispatcher.cc
```diff
@@ -49,75 +65,152 @@ void Dispatcher::DirectStreamCallbacks::onHeaders(HeaderMapPtr&& headers, bool e
...
-                                       DirectStreamCallbacksPtr&& callbacks)
-    : stream_handle_(stream_handle), underlying_stream_(underlying_stream),
-      callbacks_(std::move(callbacks)) {}
-
-AsyncClient::StreamOptions
-Dispatcher::DirectStream::toNativeStreamOptions(envoy_stream_options stream_options) {
-  AsyncClient::StreamOptions native_stream_options;
-  native_stream_options.setBufferBodyForRetry(stream_options.buffer_body_for_retry);
-  return native_stream_options;
+bool Dispatcher::DirectStream::complete() { return local_closed_ && remote_closed_; }
```
goaway: Is this true? Do we always set `local_closed_`?

library/common/http/dispatcher.cc
```diff
@@ -49,75 +65,152 @@ void Dispatcher::DirectStreamCallbacks::onHeaders(HeaderMapPtr&& headers, bool e
...
+
+bool Dispatcher::DirectStream::dispatchable(bool close) {
+  // FIXME: before merge. discuss my thought that this state does not need to be atomic because it
+  // will all be accessed from within the context of Envoy's main thread. So all we need is an
```
goaway: Cancellation is the case that won't be necessarily called from Envoy's main thread.

library/common/http/dispatcher.cc
```diff
@@ -251,31 +343,18 @@ envoy_status_t Dispatcher::resetStream(envoy_stream_t stream) {
   post([this, stream]() -> void {
```
goaway: We need to update stream state (`closed`) and fire callbacks synchronously here, prior to dispatching the reset.

library/common/types/c_types.h
```diff
@@ -195,6 +195,14 @@ typedef void (*envoy_on_error_f)(envoy_error error, void* context);
  */
 typedef void (*envoy_on_complete_f)(void* context);
 
+/**
+ * Called when the async HTTP stream has been cancelled by the application (FIXME: before merge.
+ * word choice).
+ * @param context, contains the necessary state to carry out platform-specific dispatch and
+ * execution.
+ */
+typedef void (*envoy_on_cancel_f)(void* context);
```
goaway: Hmm. Do we want this?

junr03: If we push down the cancellation state down to the core layer we need to have a bridging callback for cancel, right?

library/common/http/dispatcher.cc
junr03: This is related to the comment below where if cancellation state is handled in dispatched code then we are doing all the gating in Envoy's main thread.

library/common/http/dispatcher.cc
junr03: This is one of the biggest things I want to discuss. Because (related to above) if we keep all the cancellation state in code executed in Envoy's main thread, couldn't we relax the atomic boolean to just a simple boolean?

library/common/http/dispatcher.cc
junr03: no, but that is handled by the reset that envoy would cause. In addition to the threading of cancellation, this (terminal states) is the other thing I'd like to talk through.

library/common/http/dispatcher.cc
junr03: I'll bring it back. Accident

library/common/http/dispatcher.cc
junr03: from offline: leave the assert. But add explanation of why this assert is here.

library/common/types/c_types.h
junr03: from offline: yes we do want this as we surface onCancel to the platform.

library/common/http/dispatcher.cc
junr03: from offline: the contract with cancellation is that it is an interrupt. In other words, by the time this function returns no further callbacks associated with this stream will be surfaced. Hence we can't dispatch exchanging the closed state. It has async logic in it but the effect is sync.

library/common/http/dispatcher.cc
junr03: from offline: the contract with cancellation is that it is an interrupt. In other words, by the time this function returns no further callbacks associated with this stream will be surfaced. Hence we can't dispatch exchanging the closed state. It has async logic in it but the effect is sync.

library/common/http/dispatcher.h
```diff
@@ -118,25 +128,59 @@ class Dispatcher : public Logger::Loggable<Logger::Id::http> {
...
+    bool complete();
+    /**
+     * Return whether a callback should be allowed to continue with execution.
+     * This ensures at most one 'terminal' callback is issued for any given stream.
+     * FIXME: before merge. should this be in the DirectStreamCallbacks?
+     *
+     * @param close, whether the DirectStream should close if it has not closed before.
+     * @return bool, whether callbacks on this stream are dispatchable or not.
+     */
+    bool dispatchable(bool close);
```
junr03: true if it is not the last one, or the first last one

library/common/http/dispatcher.cc
junr03: @goaway This is the only point that remains from our conversation yesterday. The main wrinkle in moving the dispatchable call outside of the posted body is that up until now we asserted that all DirectStream interaction happened in the threading context of the event dispatcher. 

By moving the dispatchable call outside of the post body we are now interacting with the DirectStream in the platform threading context.

library/common/main_interface.h
```diff
@@ -30,6 +30,7 @@ envoy_stream_t init_stream(envoy_engine_t);
  * can occur.
  * @param stream, handle to the stream to be started.
  * @param callbacks, the callbacks that will run the stream callbacks.
+ * @param options, DEPRECATED.
```
junr03: https://github.com/lyft/envoy-mobile/issues/643

cc @buildbreaker @rebello95

rebello95: sounds good, not opposed to removing as part of this PR either

junr03:
Alright, @goaway this has a pretty full unit test suite and a basic integration test that uses the real API listener. I think this is at a good place to merge and get exposure with alpha users pending your review. I have a couple todos and we can definitely add more integration tests.

cc @rebello95 @buildbreaker

junr03:
cc @mattklein123 in case you want to take a look as well.

library/common/engine.cc
```diff
@@ -64,7 +64,9 @@ envoy_status_t Engine::run(std::string config, std::string log_level) {
     postinit_callback_handler_ = main_common_->server()->lifecycleNotifier().registerCallback(
         Envoy::Server::ServerLifecycleNotifier::Stage::PostInit, [this]() -> void {
           Server::Instance* server = TS_UNCHECKED_READ(main_common_)->server();
-          http_dispatcher_->ready(server->dispatcher(), server->clusterManager());
+          auto api_listener = server->listenerManager().apiListener()->get().http();
+          ASSERT(api_listener.has_value());
+          http_dispatcher_->ready(server->dispatcher(), api_listener.value());
```
goaway: Is this additional delay (beyond construction) still necessary? We're no longer waiting on an initial round of DNS resolution and the filter dynamic forwarding filter contains built in buffering logic while we do wait. If we're unsure, maybe insert a TODO to follow up.

library/common/http/dispatcher.cc
```diff
@@ -49,75 +71,137 @@ void Dispatcher::DirectStreamCallbacks::onHeaders(HeaderMapPtr&& headers, bool e
...
+  // @see Dispatcher::DirectStreamCallbacks::closeRemote, and @see Dispatcher::resetStream,
+  // respectively.
+}
+
+bool Dispatcher::DirectStream::dispatchable(bool close) {
+  if (close) {
+    // Set closed to true and return true if not previously closed.
+    return !closed_.exchange(close);
+  }
+  return !closed_.load();
```
goaway: There's a bug here, and it relates to the fact that the previous platform-level implementation checked this state (in the cases it mattered) from a serialized post-dispatch context. Here, it's necessarily the pre-dispatch context, and as it stands, cancellation could race another callback (resulting in, e.g., an `on_headers` callback being dispatched after an `on_cancel`). I expect this to be a fairly rare race, and so I'd be okay rolling forward with a FIXME, but as for the fix, I can't think of anything we can do besides having a lock that a thread holds while checking this state and performing platform dispatch.

junr03: @goaway hmmmm actually don't we have the same problem today? Because the thread that cancel checks the atomic is _not_ the same as the thread the dispatch queue context runs in, right?

Could that be the race that @rebello95 was seeing?

library/common/config_template.cc
```diff
@@ -3,19 +3,51 @@
...
+            - name: api
+              domains:
+                - "*"
+              routes:
+                - match:
+                    prefix: "/"
+                  route:
+                    cluster_header: x-envoy-mobile-cluster
+        http_filters:
+          - name: envoy.filters.http.dynamic_forward_proxy
```
rebello95: awesome

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -84,62 +84,43 @@ static envoy_headers toNativeHeaders(EnvoyHeaders *headers) {
...
-  return !atomic_load(closed);
-}
-
 static void ios_on_headers(envoy_headers headers, bool end_stream, void *context) {
   ios_context *c = (ios_context *)context;
   EnvoyHTTPCallbacks *callbacks = c->callbacks;
   dispatch_async(callbacks.dispatchQueue, ^{
-    if (!dispatchable(c->closed, end_stream) || !callbacks.onHeaders) {
-      return;
+    if (callbacks.onHeaders) {
```
rebello95: 💯 that we moved this down a layer

test/common/thread/lock_guard_test.cc
```diff
@@ -0,0 +1,27 @@
...
+#include "library/common/thread/lock_guard.h"
+
+namespace Envoy {
+namespace Thread {
+
+class ThreadTest : public testing::Test {
+protected:
+  ThreadTest() = default;
+  int a_ ABSL_GUARDED_BY(a_mutex_){0};
+  MutexBasicLockable a_mutex_;
```
rebello95: Would it make more sense to create these properties inside the test case instead of within the thread class to remove state between the existing test case and potential future cases?

test/common/thread/lock_guard_test.cc
```diff
@@ -0,0 +1,27 @@
...
+};
+
+TEST_F(ThreadTest, TestOptionalReleasableLockGuard) {
+  OptionalReleasableLockGuard lock(nullptr);
+  EXPECT_EQ(1, ++b_);
+
+  OptionalReleasableLockGuard lock2(&a_mutex_);
+  EXPECT_EQ(1, ++a_);
+  lock2.release();
+}
```
rebello95: Should we add a test to validate that the [mutex unlocks on dealloc](https://github.com/lyft/envoy-mobile/pull/616/commits/f3d8ff27efe6035b122e65b35ae4bdd23c44b8d6#diff-c9f5b5ee47e72ae0286699c9caffd68dR34)?

test/common/thread/lock_guard_test.cc
junr03: each test starts with a fresh instance of the test class.

test/common/thread/lock_guard_test.cc
junr03: the first lock guard is testing that. If it did not. The test would not compile thanks to the thread annotations.

test/common/thread/lock_guard_test.cc
rebello95: Ah cool, TIL

library/common/http/dispatcher.cc
junr03: https://github.com/lyft/envoy-mobile/pull/616/commits/f3d8ff27efe6035b122e65b35ae4bdd23c44b8d6 captures how the race is solved for now. #647 opened to explore alternative solutions in the future.

library/common/config_template.cc
junr03: my local formatter changes this. But it is a noop.

.github/workflows/core.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: 'Install dependencies'
         run: ./ci/mac_ci_setup.sh
       - name: 'Run tests'
-        run: bazel test //test/common/...
+        run: bazel test --test_output=all //test/common/...
```
buildbreaker: nice

library/common/engine.cc
goaway: Still suggest adding a TODO and/or issue for this so we still have visibility to follow up.

library/common/http/dispatcher.cc
```diff
@@ -229,71 +321,122 @@ envoy_status_t Dispatcher::sendMetadata(envoy_stream_t, envoy_headers) { return
...
+  // Hence why it is scheduled for deferred deletion. If this was all that was needed then it
+  // would be sufficient to return a shared_ptr in getStream. However, deferred deletion is still
+  // required because in Dispatcher::resetStream the DirectStream needs to live for as long and
+  // the HCM's ActiveStream lives. Hence its deletion needs to live beyond the synchronous code in
+  // Dispatcher::resetStream.
+  // TODO: currently upstream Envoy does not have a deferred delete version for shared_ptr. This
+  // means that instead of efficiently queuing the deletes for one event in the event loop, all
+  // deletes here get queued up as individual posts.
+  TS_UNCHECKED_READ(event_dispatcher_)->post([direct_stream]() -> void {
+    ENVOY_LOG(debug, "[S{}] deferred deletion of stream", direct_stream->stream_handle_);
```
goaway: Just want to be sure that the optimizer won't optimize away this capture of `direct_stream` - even if `ENVOY_LOG` is compiled away. (Even if this would definitively/arguably be an optimizer bug, I've seen this sort of thing happen before.)

library/common/thread/lock_guard.h
```diff
@@ -0,0 +1,52 @@
...
+ */
+// TODO(junr03): this could be moved to Envoy's codebase.
+class ABSL_SCOPED_LOCKABLE OptionalReleasableLockGuard {
+public:
+  /**
+   * Establishes a scoped mutex-lock. If non-null, the mutex is locked upon construction.
+   *
+   * @param lock the mutex.
+   */
+  OptionalReleasableLockGuard(BasicLockable* lock) ABSL_EXCLUSIVE_LOCK_FUNCTION(lock)
```
goaway: Cool.

library/common/http/dispatcher.cc
junr03: I really enjoy your eye for things that might end up being subtle bugs in the future. https://stackoverflow.com/a/12718425 and the comments below state that an explicit capture like above will not be optimized out even if the variable is not used in the lambda body.

library/common/thread/lock_guard.h
junr03: Yeah I think LockGuards are super neat one of my favorite RAII-enabled patterns.

junr03:
Alright team @goaway @buildbreaker @rebello95 going to merge this. However we can't bump iOS until we fix https://github.com/envoyproxy/envoy/pull/9875

PR #618, opened 2019-12-31
core: apply protobuf patch to prevent duplicate symbols

rebello95:
Fixes https://github.com/lyft/envoy-mobile/issues/617, which resulted in duplicate symbols from protobuf being present in the final Envoy Mobile binary.

The [root cause](https://github.com/bazelbuild/bazel/issues/10484) is being [patched in Bazel](https://github.com/bazelbuild/bazel/pull/10493). In the meantime, we can revert these two commits in protobuf ([first](https://github.com/protocolbuffers/protobuf/commit/7b28278c7d4f4175e70aef2f89d304696eb85ae3), [second](https://github.com/protocolbuffers/protobuf/commit/a03d332aca5d33c5d4b2cd25037c9e37d57eff02)) and apply the diff as a patch until we're able to update Bazel with the fix. More context on the protobuf issue is available [here](https://github.com/protocolbuffers/protobuf/issues/7046).

Notes:
- This should be reverted after updating Bazel with the fix
- The version of protobuf specified in the `http_archive` should be in lockstep with upstream Envoy's version
- Upstream Envoy's `@envoy//bazel:protobuf.patch` patch is also included here


WORKSPACE
```diff
@@ -170,6 +170,20 @@ grpc_java_repositories(
     omit_net_zlib = True,
 )
 
+# Apply patches to protobuf to prevent duplicate symbols: https://github.com/lyft/envoy-mobile/issues/617.
+# TODO: Remove after https://github.com/bazelbuild/bazel/pull/10493 is merged to Bazel and we upgrade.
+http_archive(
+    name = "com_google_protobuf",
+    patches = ["//bazel:protobuf1.patch", "//bazel:protobuf2.patch"],
```
keith:
```suggestion
    patches = ["//bazel:protobuf_1.patch", "//bazel:protobuf_2.patch"],
```

WORKSPACE
```diff
@@ -170,6 +170,20 @@ grpc_java_repositories(
...
+    sha256 = "3df786d47a813a2912f0fcdd79c2bc49be123a8c37949e45dbb16bcec4883b11",
+    strip_prefix = "boringssl-6263268b8c1b78a8a9b65acd6f5dd5c04dd9b0e1",
+    urls = ["https://github.com/protocolbuffers/protobuf/archive/6263268b8c1b78a8a9b65acd6f5dd5c04dd9b0e1.tar.gz"],
```
keith: This is the same version as upstream right?

WORKSPACE
```diff
@@ -170,6 +170,20 @@ grpc_java_repositories(
     omit_net_zlib = True,
 )
 
+# Apply patches to protobuf to prevent duplicate symbols: https://github.com/lyft/envoy-mobile/issues/617.
+# TODO: Remove after https://github.com/bazelbuild/bazel/pull/10493 is merged to Bazel and we upgrade.
+http_archive(
```
keith: This has to be above envoy's setup otherwise the one that it loads will take precedence

WORKSPACE
rebello95: Combined these into one patch

WORKSPACE
rebello95: It is now 😄

WORKSPACE
rebello95: Updated

Issue #634, opened 2020-01-17
master artifacts: reduce android build time


junr03:
The android master artifact takes a long time. It would be valuable to see if we can reduce it.

We extended the timeout from 90s to 120s

PR #635, opened 2020-01-17
android: extend timeout for master artifact

junr03:
Description: the android master artifact is running out of time to build. Extending the timeout to allow it to build. Ultimately we need to track this time down #634 
Risk Level: low
Testing: CI run


buildbreaker:
thanks for filing a follow up issue @junr03

PR #638, opened 2020-01-22
build: move rules_swift repo rule earlier

kastiglione:
Description:

We had reports of debugging issues due to paths that don't exist on developer machines. Previously we enabled the `swift.cacheable_modules` feature to prevent this (see #549). Apparently `rules_foreign_cc_dependencies()` is loading `rules_swift`, and must use a sha that is prior to the introduction of the `cacheable_modules` feature.

This change moves the `rules_swift` repository rule earlier in the `WORKSPACE` – above `rules_foreign_cc_dependencies()` and just below the `rules_appe` repo rule.

Risk Level:

Low

The rules_swift declaration was previously inert, it wasn't being used, so this change is effectively a bump to the rules_swift version being used/dependend on.

Testing:

Used `bazel aquery //library/swift/src:ios_framework_archive` and verified that `-no-serialize-debugging-options` was present, and that `-serialize-debugging-options` was **not included**.

Docs Changes: N/A

Release Notes: N/A

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -61,6 +61,13 @@ http_archive(
     urls = ["https://github.com/bazelbuild/rules_apple/archive/a595f71b94f75d531ebdf8ae31cc8eb1ead6a480.tar.gz"],
 )
 
+http_archive(
+    name = "build_bazel_rules_swift",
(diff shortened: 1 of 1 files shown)


kastiglione:
signed off and force pushed

kastiglione:
I opened https://github.com/lyft/envoy-mobile/pull/639 to address the Github actions CI failure.

PR #639, opened 2020-01-22
ci: change runs-on: from macOS-10.14 to -latest

kastiglione:
Description:

Some CI jobs for #638 failed with the cause given as:

> The macOS virtual environment has been updated to Catalina (v10.15). Starting January 15th, jobs that include the line 'runs-on: macOS-10.14' will fail to run and return a failed check suite. Please update your workflow and change the line 'runs-on: macOS-10.14' to 'runs-on: macOS-latest'.

This makes the specified changes, moving from `runs-on: macOS-10.14` to `runs-on: macOS-latest`.

Risk Level:

Maybe Low? I'd have to hear from others. It also seems like an upstream requirement.

Testing:

This PR will test mac CI jobs.

Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -35,7 +35,7 @@ jobs:
           bazel build --fat_apk_cpu=x86 //examples/kotlin/hello_world:hello_envoy_kt
   macdist:
     name: mac_dist
-    runs-on: macOS-10.14
+    runs-on: macOS-latest
     timeout-minutes: 45
     steps:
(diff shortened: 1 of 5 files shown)


keith:
Weird breaking change. Seems risky at OS update time but we can figure it out then IMO

kastiglione:
I wonder what their plans are for offering current and prev OS simultaneously.

kastiglione:
signed off and force pushed

Issue #630, opened 2020-01-13
tls: SAN verification changed in Envoy


junr03:
SAN verification changed in Envoy in https://github.com/envoyproxy/envoy/pull/9264. Therefore, for v3 config it is not as easy as to put the configured `domain` as a SAN to verify, as there is no wildcard matching as there was before. i.e foo.company.com in the config will no longer match *.company.com in the presented certificate. 

#626 deleted SAN verification until a better option is exposed in Envoy Mobile.

junr03:
I believe this is relevant, and would allow us to do SAN based on authority. Need to read more carefully, but want to put here before I forget https://github.com/envoyproxy/envoy/pull/9863

junr03:
wildcard matching is getting fixed in Envoy in https://github.com/envoyproxy/envoy/pull/10005

junr03:
https://github.com/envoyproxy/envoy/pull/9863 made it so that the dynamic forward proxy automatically does SAN validation. via the verifySubjectAltName path based on the host/authority header. Therefore, after https://github.com/lyft/envoy-mobile/pull/663 Envoy Mobile has been verifying SAN.

PR #641, opened 2020-01-22
ios: remove domain from engine configuration

goaway:
Description: Removes domain from Envoy engine configuration on iOS. Dynamic forwarding is now default, and so a proper `:authority` header is all that's needed.
Risk Level: Low
Testing: Pending

### DIFF
--- a/library/objective-c/EnvoyEngine.h
+++ b/library/objective-c/EnvoyEngine.h
@@ -134,7 +134,6 @@ typedef NSDictionary<NSString *, NSArray<NSString *> *> EnvoyHeaders;
 /// Typed configuration that may be used for starting Envoy.
 @interface EnvoyConfiguration : NSObject
 
-@property (nonatomic, strong) NSString *domain;
 @property (nonatomic, strong) NSString *statsDomain;
 @property (nonatomic, assign) UInt32 connectTimeoutSeconds;
 @property (nonatomic, assign) UInt32 dnsRefreshSeconds;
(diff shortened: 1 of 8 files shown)


goaway:
Depends on https://github.com/lyft/envoy-mobile/pull/616

goaway:
Note example apps will still fail temporarily because this depends on config updates in the above PR.

goaway:
rebased

library/swift/test/EnvoyClientTests.swift
```diff
@@ -67,7 +67,7 @@ final class EnvoyClientTests: XCTestCase {
       method: .get, scheme: "https", authority: "www.envoyproxy.io", path: "/docs")
```
junr03: do we want to protect at the core layer against `sendHeaders` calls that do not set an authority header?

library/objective-c/EnvoyEngine.h
```diff
@@ -143,11 +142,10 @@ typedef NSDictionary<NSString *, NSArray<NSString *> *> EnvoyHeaders;
 /**
  Create a new instance of the configuration.
  */
-- (instancetype)initWithDomain:(NSString *)domain
-                   statsDomain:(NSString *)statsDomain
-         connectTimeoutSeconds:(UInt32)connectTimeoutSeconds
-             dnsRefreshSeconds:(UInt32)dnsRefreshSeconds
-             statsFlushSeconds:(UInt32)statsFlushSeconds;
+- (instancetype)initWithStatsDomain:(NSString *)statsDomain
```
junr03: I think we have a templating bug. Which is that now we are not setting SNI in the template if we don't populate domain. Which actually brings the larger question of how we want to do that (and relatedly #630) for the dynamic proxy? I need to read the docs to see if that can be dynamic as well.

goaway: According to the docs here it sounds like SNI will happen automatically as long as there's a TLS context on the cluster:
https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_proxy

goaway: Will update.

library/swift/test/EnvoyClientTests.swift
goaway: That's a good callout and something we've discussed before. Right now we don't check to see if this is a valid HTTP/2 request at this layer, but we certainly could. I think it's worth discussing and/or tracking in an issue, but somewhat orthogonal to this change (since we haven't done it before either).

goaway:
Latest change based on these docs:
https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_proxy

"When the upstream cluster has been configured with a TLS context, Envoy will automatically perform SAN verification for the resolved host name as well as specify the host name via SNI."

library/swift/test/EnvoyClientTests.swift
rebello95: FWIW `authority` is a required property when initializing the `RequestBuilder`, so consumers would have to explicitly set an invalid/empty authority value here. Could be worth asserting or logging in the core layer as you mentioned, though

PR #642, opened 2020-01-22
android: remove domain from engine configuration

goaway:
Description: Removes domain from Envoy engine configuration on Android. Dynamic forwarding is now default, and so a proper `:authority` header is all that's needed.
Risk Level: Low
Testing: Pending

### DIFF
--- a/examples/java/hello_world/MainActivity.java
+++ b/examples/java/hello_world/MainActivity.java
@@ -9,7 +9,6 @@ import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
 import android.util.Log;
 import io.envoyproxy.envoymobile.AndroidEnvoyClientBuilder;
-import io.envoyproxy.envoymobile.Domain;
 import io.envoyproxy.envoymobile.Envoy;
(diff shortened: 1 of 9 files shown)


goaway:
Depends on https://github.com/lyft/envoy-mobile/pull/616

goaway:
Note example apps will still fail temporarily because this depends on config updates in the above PR.

buildbreaker:
no stale

goaway:
rebased

buildbreaker:
> Note example apps will still fail temporarily because this depends on config updates in the above PR.

Can't we update the example applications in this PR as well?

goaway:
@buildbreaker Example apps are already updated in this PR. The comment was referring to the fact that CI would show them as failing until #616 was merged. :)

buildbreaker:
gotcha

goaway:
Latest change, based on these docs: 
https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_proxy

"When the upstream cluster has been configured with a TLS context, Envoy will automatically perform SAN verification for the resolved host name as well as specify the host name via SNI."

PR #645, opened 2020-01-28
envoy: ref update 20200128

junr03:
Description: update the envoy ref. This bump includes https://github.com/envoyproxy/envoy/pull/9767 which will enable us to turn on process logs for Android devices in a subsequent PR.
Risk Level: med, as always with a ref update.
Testing: CI


junr03:
cc @buildbreaker

PR #656, opened 2020-02-04
Bump version to 0.2.2

rebello95:


docs/root/intro/version_history.rst
```diff
@@ -1,6 +1,16 @@
...
 ---------------
 
+0.2.2 (Feb 3, 2020)
+===================
+
+Envoy Mobile v0.2.2 changes how network requests are performed to no longer use Envoy's `AsyncClient` and to instead consume the `ApiListener` directly (#616).
+
+Additional changes:
+
+- Domain is no longer required to be specified when starting Envoy Mobile (#641, #642)
```
junr03: I would have stronger language here. “No longer supported. Envoy mobile now uses the authority specified when starting a stream”. Also the API docs still show domain in the client. Do you mind doing a code search and updating? That way the v0.2.2 tagged docs will all be consistent with what is supported.

rebello95: yep good call

PR #659, opened 2020-02-04
ios/android: allow zero for total upstream timeouts

rebello95:
We should allow zero to be specified for `totalUpstreamTimeoutMS`, which effectively [disables the upstream timeout per Envoy's docs](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route_components.proto#envoy-api-field-route-routeaction-timeout).

This can be useful for consumers who only want to timeout on a per-request basis (rather than timing out all retries in aggregate).

### DIFF
--- a/library/swift/src/RetryPolicy.swift
+++ b/library/swift/src/RetryPolicy.swift
@@ -47,15 +47,15 @@ public final class RetryPolicy: NSObject {
   ///                                     retries. Spans the point at which the entire downstream
   ///                                     request has been processed and when the upstream
   ///                                     response has been completely processed.
-  ///                                     Nil may be specified to disable the upstream timeout.
+  ///                                     Nil or 0 may be specified to disable the upstream timeout.
   public init(maxRetryCount: UInt, retryOn: [RetryRule], perRetryTimeoutMS: UInt? = nil,
               totalUpstreamTimeoutMS: UInt? = 15_000)
   {
     if let perRetryTimeoutMS = perRetryTimeoutMS,
       let totalUpstreamTimeoutMS = totalUpstreamTimeoutMS
     {
-      assert(perRetryTimeoutMS <= totalUpstreamTimeoutMS,
-             "Per-retry timeout must be <= total timeout")
+      assert(perRetryTimeoutMS <= totalUpstreamTimeoutMS || totalUpstreamTimeoutMS == 0,
+             "Per-retry timeout must be <= total timeout when != 0")
     }
 
     self.maxRetryCount = maxRetryCount
--- a/library/swift/src/RetryPolicy.swift
+++ b/library/swift/src/RetryPolicy.swift
@@ -42,12 +42,12 @@ public final class RetryPolicy: NSObject {
   ///                                     performed.
   /// - parameter retryOn:                Whitelist of rules used for retrying.
   /// - parameter perRetryTimeoutMS:      Timeout (in milliseconds) to apply to each retry. Must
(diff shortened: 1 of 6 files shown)


rebello95:
> aren't we worried that API for perRetryTimeoutMS and totalUpstreamTimeoutMS are different? One of them accepts nil while the other relies on 0 being passed. Could we come up with something that would work for both of them - accepting nils for both of them or 0 for both of them?

This is a good call out. I'll update both to be nullable.

rebello95:
For context, there's a difference in how those are serialized on the wire. For upstream/route timeouts, we need to manually specify `0` in the header, whereas for request timeouts, we omit the header. But I don't think we should expose these implementation details in the public platform interface.

library/swift/src/RetryPolicy.swift
```diff
@@ -47,13 +47,15 @@ public final class RetryPolicy: NSObject {
...
-              totalUpstreamTimeoutMS: UInt = 15_000)
+              totalUpstreamTimeoutMS: UInt? = 15_000)
   {
-    if let perRetryTimeoutMS = perRetryTimeoutMS {
-      assert(perRetryTimeoutMS <= totalUpstreamTimeoutMS || totalUpstreamTimeoutMS == 0,
-             "Per-retry timeout must be <= total timeout if total timeout != 0")
+    if let perRetryTimeoutMS = perRetryTimeoutMS,
+      let totalUpstreamTimeoutMS = totalUpstreamTimeoutMS
+    {
+      assert(perRetryTimeoutMS <= totalUpstreamTimeoutMS,
```
Augustyniak: you still need to check for `totalUpstreamTimeoutMS ` here, right?

Augustyniak: assert(perRetryTimeoutMS <= totalUpstreamTimeoutMS && totalUpstreamTimeoutMS != 0)

library/swift/test/RetryPolicyMapperTests.swift
```diff
@@ -23,4 +23,9 @@ final class RetryPolicyMapperTests: XCTestCase {
     let policy = RetryPolicy(maxRetryCount: 123, retryOn: RetryRule.allCases)
     XCTAssertNil(policy.outboundHeaders()["x-envoy-upstream-rq-per-try-timeout-ms"])
   }
+
+  func testConvertingToHeadersWithoutUpstreamTimeoutIncludesZeroForTimeoutHeader() {
```
Augustyniak: let's add a test case for when per request timeout is equal to 0 (not nil) and per retry timeout is greater than 0.

rebello95: Considered that but iOS doesn't really provide a way to test that `assert`s fail, and making the builder `throw` doesn't really seem worth it here

library/swift/src/RetryPolicy.swift
rebello95: Personally I think that omitting the `totalUpstreamTimeoutMS != 0` exclusion from the assert is probably preferred here since the consumer will then change the `0` to `nil` to achieve the same behavior (in line with the documentation of this function). Thoughts?

Augustyniak: One question - What happens if I provide `perRetryTimeoutMS = 0` and `totalUpstreamTimeoutMS = 0`? I would expect this too break since the docs says that a user should use `nil` if they want `totalUpstreamTimeoutMS` to be ignored.

rebello95: That's fair, I can update to include it in the assert.

PR #683, opened 2020-02-11
ci: only build artifacts on commits that affect the VERSION file

junr03:
Description: only version bumps are used on clients. Therefore, this PR makes it so that the jobs on the artifacts.yml workflow only run when the VERSION file changes. This should save some CI time.
Risk Level: low

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - master
+    paths:
+      - VERSION
 
 jobs:
   master_android_dist:


rebello95:
I'm not sure I fully agree with this since in theory users should be able to use the most recent commit from Envoy Mobile to build with their app (similarly to how upstream Envoy rarely tags releases). Maybe this doesn't really happen in practice since users consume CocoaPods or Maven though. WDYT?

junr03:
I was against this when we were tagging only semantic versions. But given we are committed to weekly releases tagged by date I feel fine. Happy to reconsider when we hear of a use case that is not covered by this, but for now it seems to me that saving the CI compute outweighs being able to download artifacts for every single commit.

PR #684, opened 2020-02-11
android: update bazel flag config

### LINKED ISSUE #34: logging sink for Android
Currently Envoy has a file logging sink, and a stderr logging sink. Both of these do not work with android. We need to implement an Android logging sink in order to get Envoy logs in Android. We can lift up spdlog's [android_sink](https://github.com/gabime/spdlog/blob/master/include/spdlog/sinks/android_sink.h).

junr03:
Description: https://github.com/envoyproxy/envoy/pull/9767 introduced the bazel option and https://github.com/lyft/envoy-mobile/pull/663 brought it into envoy mobile. This PR changes the bazel config for android to turn on the logger by default when building on android. It also updates a couple other stale flags and docs.
Risk Level: low
Testing: tested locally and saw logs

Fixes #34 

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -28,6 +28,7 @@ build:ios --define=manual_stamp=manual_stamp
 
 # Default flags for builds targeting Android
 build:android --fat_apk_cpu=x86_64
+build:android --define logger=android
 
 # Common flags for release builds
 # TODO: Enable `--copt -ggdb3`. Issues were encountered previously with this:
--- a/.bazelrc
+++ b/.bazelrc
@@ -44,4 +44,3 @@ build:release-ios --copt=-fembed-bitcode
 # Flags for release builds targeting Android or the JVM
 build:release-android --config=android
 build:release-android --config=release-common
-build:release-android --fat_apk_cpu=x86,armeabi-v7a,arm64-v8a
--- a/.bazelrc
+++ b/.bazelrc
@@ -27,7 +27,6 @@ build --javabase=@bazel_tools//tools/jdk:jdk
 build:ios --define=manual_stamp=manual_stamp
 
 # Default flags for builds targeting Android
-build:android --fat_apk_cpu=x86_64
 build:android --define logger=android
 
 # Common flags for release builds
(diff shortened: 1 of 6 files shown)


junr03:
@buildbreaker I actually noticed a couple more places to clean up. lmk what you think

.bazelrc
```diff
@@ -42,4 +42,6 @@ build:release-ios --config=release-common
 build:release-ios --copt=-fembed-bitcode
 
 # Flags for release builds targeting Android or the JVM
+build:release-android --config=android
 build:release-android --config=release-common
+build:release-android --fat_apk_cpu=x86,armeabi-v7a,arm64-v8a
```
rebello95: I think we should omit architectures here and specify them where we do individual builds, for parity with iOS. Thoughts?

buildbreaker: +1 to this

junr03: done

docs/root/development/performance/cpu_battery_impact.rst
```diff
@@ -94,7 +94,7 @@ a process listening to traffic sent over this connection.
 
 Getting the build:
 
-1. Build the library using ``bazel build android_dist --config=android``
```
rebello95: Why don't we need this anymore? Still seems required?

buildbreaker: It's just for turning on logging, right?

rebello95: For now, yes, though that still seems like something we'd want by default

junr03: @rebello95 what do we want by default? Logging?

rebello95: Yes, basically the configuration we get with `--config=android`

junr03: Yeah we can add it back.

junr03: added

Issue #679, opened 2020-02-11
configuration: option to use h2/http1.1 based on header information


junr03:
Right now all configured clusters go over http1.1. Until https://github.com/lyft/envoy-mobile/issues/502 is supported we should create the option to have streams decide what protocol version to support, and have backing configuration for clusters configured for http1.1 and h2.

PR #689, opened 2020-02-14
core: remove envoy_stream_options

rebello95:
As of https://github.com/lyft/envoy-mobile/pull/616, this is no longer used or necessary since it's handled by the router.

Resolves https://github.com/lyft/envoy-mobile/issues/643.

### DIFF
--- a/library/common/http/dispatcher.cc
+++ b/library/common/http/dispatcher.cc
@@ -286,8 +286,7 @@ Dispatcher::Dispatcher(std::atomic<envoy_network_t>& preferred_network)
       address_(std::make_shared<Network::Address::SyntheticAddressImpl>()) {}
 
 envoy_status_t Dispatcher::startStream(envoy_stream_t new_stream_handle,
-                                       envoy_http_callbacks bridge_callbacks,
-                                       envoy_stream_options) {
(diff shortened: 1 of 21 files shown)


junr03:
I had left this behind after I realized about #679. I was going to set the option up in this struct. Given @goaway took #679, up to him if he wants to use as is or do something else.

rebello95:
Leaving the struct makes sense to me if that's how we're going to set up HTTP/2 preferences per-stream (though a lot of the deletions in this PR will still be relevant). WDYT @goaway?

goaway:
Honestly, I'm ambivalent. Headers can work for almost every potential use-case I can think of, and there's already precedent (retry policies, timeouts). The only downside of headers is that they aren't typed, but this is all internal anyways. This is a potentially useful feature for cases where there wouldn't be headers (redis connections?), but it's not like we can't revisit it when/if that's relavent.

I guess I'm (ever so) slightly in favor removing it for now, simply because it adds overhead to all the function signatures.

rebello95:
Sounds good. I agree that this could be used internally in the future, but we can always revert/revisit when that time comes (rather than keeping dead code on master).

PR #694, opened 2020-02-19
upstream: update header maps on response path

goaway:


goaway:
Distributables compile; tests need to be updated.

goaway:
Should be good to go now.

PR #697, opened 2020-02-20
upsteam protocol: add the ability to decide what protocol to use for upstream requests

### LINKED ISSUE #679: configuration: option to use h2/http1.1 based on header information

junr03:
Description: this PR adds the ability to decide what protocol to use for upstream requests. Note that this will be unnecessary once ALPN works for upstream requests (https://github.com/lyft/envoy-mobile/issues/502)
Risk Level: low
Testing: updated unit tests
Docs Changes: updated docs

Fixes #679 

### DIFF
--- a/docs/root/api/http.rst
+++ b/docs/root/api/http.rst
@@ -17,7 +17,7 @@ a previously created :ref:`Envoy instance <api_starting_envoy>`.
 
 **Kotlin**::
 
-  val request = RequestBuilder(RequestMethod.POST, "https", "api.envoyproxy.io", "/foo")
+  val request = RequestBuilder(RequestMethod.POST, "https", "api.envoyproxy.io", "/foo", upstreamHttpProtocol = UpstreamRequestProtocol.HTTP2)
     .addRetryPolicy(RetryPolicy(...))
     .addHeader("x-custom-header", "foobar")
     ...
@@ -25,7 +25,7 @@ a previously created :ref:`Envoy instance <api_starting_envoy>`.
 
 **Swift**::
 
-  let request = RequestBuilder(method: .post, scheme: "https", authority: "api.envoyproxy.io", path: "/foo", upstreamHttpProtocol: true)
+  let request = RequestBuilder(method: .post, scheme: "https", authority: "api.envoyproxy.io", path: "/foo", upstreamHttpProtocol: .http2)
     .addRetryPolicy(RetryPolicy(...))
     .addHeader(name: "x-custom-header", value: "foobar")
     ...
--- a/docs/root/api/http.rst
+++ b/docs/root/api/http.rst
@@ -25,7 +25,7 @@ a previously created :ref:`Envoy instance <api_starting_envoy>`.
 
 **Swift**::
 
-  let request = RequestBuilder(method: .post, scheme: "https", authority: "api.envoyproxy.io", path: "/foo")
+  let request = RequestBuilder(method: .post, scheme: "https", authority: "api.envoyproxy.io", path: "/foo", upstreamHttpProtocol: true)
     .addRetryPolicy(RetryPolicy(...))
     .addHeader(name: "x-custom-header", value: "foobar")
     ...
--- a/docs/root/api/http.rst
+++ b/docs/root/api/http.rst
@@ -17,16 +17,18 @@ a previously created :ref:`Envoy instance <api_starting_envoy>`.
 
 **Kotlin**::
 
-  val request = RequestBuilder(RequestMethod.POST, "https", "api.envoyproxy.io", "/foo", upstreamHttpProtocol = UpstreamRequestProtocol.HTTP2)
+  val request = RequestBuilder(RequestMethod.POST, "https", "api.envoyproxy.io", "/foo")
     .addRetryPolicy(RetryPolicy(...))
+    .addUpstreamHttpProtocol(UpstreamRequestProtocol.HTTP2)
     .addHeader("x-custom-header", "foobar")
     ...
     .build()
 
 **Swift**::
 
-  let request = RequestBuilder(method: .post, scheme: "https", authority: "api.envoyproxy.io", path: "/foo", upstreamHttpProtocol: .http2)
+  let request = RequestBuilder(method: .post, scheme: "https", authority: "api.envoyproxy.io", path: "/foo")
     .addRetryPolicy(RetryPolicy(...))
+    .addUpstreamHttpProtocol(.http2)
     .addHeader(name: "x-custom-header", value: "foobar")
     ...
     .build()
--- a/examples/swift/hello_world/ViewController.swift
+++ b/examples/swift/hello_world/ViewController.swift
@@ -49,7 +49,7 @@ final class ViewController: UITableViewController {
     let requestID = self.requestCount
     let request = RequestBuilder(method: .get, scheme: kRequestScheme,
                                  authority: kRequestAuthority,
-                                 path: kRequestPath, upstreamHttpProtocol: true).build()
+                                 path: kRequestPath, upstreamHttpProtocol: .http2).build()
     let handler = ResponseHandler()
       .onHeaders { [weak self] headers, statusCode, _ in
         NSLog("Response status (\(requestID)): \(statusCode)\n\(headers)")
--- a/examples/swift/hello_world/ViewController.swift
+++ b/examples/swift/hello_world/ViewController.swift
@@ -52,7 +52,7 @@ final class ViewController: UITableViewController {
     // end-to-end tests in CI.
     let request = RequestBuilder(method: .get, scheme: kRequestScheme,
                                  authority: kRequestAuthority,
-                                 path: kRequestPath, upstreamHttpProtocol: .http2)
+                                 path: kRequestPath)
         .addUpstreamHttpProtocol(.http2)
         .build()
     let handler = ResponseHandler()
--- a/examples/swift/hello_world/ViewController.swift
+++ b/examples/swift/hello_world/ViewController.swift
@@ -49,7 +49,7 @@ final class ViewController: UITableViewController {
     let requestID = self.requestCount
     let request = RequestBuilder(method: .get, scheme: kRequestScheme,
                                  authority: kRequestAuthority,
-                                 path: kRequestPath).build()
+                                 path: kRequestPath, upstreamHttpProtocol: true).build()
     let handler = ResponseHandler()
       .onHeaders { [weak self] headers, statusCode, _ in
         NSLog("Response status (\(requestID)): \(statusCode)\n\(headers)")
--- a/examples/swift/hello_world/ViewController.swift
+++ b/examples/swift/hello_world/ViewController.swift
@@ -49,7 +49,9 @@ final class ViewController: UITableViewController {
     let requestID = self.requestCount
     let request = RequestBuilder(method: .get, scheme: kRequestScheme,
                                  authority: kRequestAuthority,
-                                 path: kRequestPath, upstreamHttpProtocol: .http2).build()
+                                 path: kRequestPath, upstreamHttpProtocol: .http2)
+        .addUpstreamHttpProtocol(.http2)
+        .build()
     let handler = ResponseHandler()
       .onHeaders { [weak self] headers, statusCode, _ in
         NSLog("Response status (\(requestID)): \(statusCode)\n\(headers)")
--- a/examples/swift/hello_world/ViewController.swift
+++ b/examples/swift/hello_world/ViewController.swift
@@ -47,6 +47,9 @@ final class ViewController: UITableViewController {
     NSLog("Starting request to '\(kRequestPath)'")
 
     let requestID = self.requestCount
+    // Note: this request will use an h2 stream for the upstream request.
+    // The Objective-C example uses http/1.1. This is done on purpose to test both paths in
+    // end-to-end tests in CI.
     let request = RequestBuilder(method: .get, scheme: kRequestScheme,
                                  authority: kRequestAuthority,
                                  path: kRequestPath, upstreamHttpProtocol: .http2)
--- a/library/common/http/dispatcher.cc
+++ b/library/common/http/dispatcher.cc
@@ -320,7 +320,7 @@ envoy_status_t Dispatcher::sendHeaders(envoy_stream_t stream, envoy_headers head
       RequestHeaderMapPtr internal_headers = Utility::toRequestHeaders(headers);
       ENVOY_LOG(debug, "[S{}] request headers for stream (end_stream={}):\n{}", stream, end_stream,
                 *internal_headers);
-      attachPreferredNetwork(*internal_headers);
+      setDestinationCluster(*internal_headers);
       direct_stream->request_decoder_->decodeHeaders(std::move(internal_headers), end_stream);
       direct_stream->closeLocal(end_stream);
     }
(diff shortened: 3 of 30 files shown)


docs/root/api/http.rst
```diff
@@ -17,15 +17,15 @@ a previously created :ref:`Envoy instance <api_starting_envoy>`.
...
+  val request = RequestBuilder(RequestMethod.POST, "https", "api.envoyproxy.io", "/foo", upstreamHttpProtocol = UpstreamRequestProtocol.HTTP2)
     .addRetryPolicy(RetryPolicy(...))
     .addHeader("x-custom-header", "foobar")
     ...
     .build()
 
 **Swift**::
 
-  let request = RequestBuilder(method: .post, scheme: "https", authority: "api.envoyproxy.io", path: "/foo")
+  let request = RequestBuilder(method: .post, scheme: "https", authority: "api.envoyproxy.io", path: "/foo", upstreamHttpProtocol: .http2)
```
goaway: I'd suggest moving this out of the constructor, and adding it via a builder method.

library/common/http/dispatcher.cc
```diff
@@ -477,19 +477,38 @@ const LowerCaseString ClusterHeader{"x-envoy-mobile-cluster"};
 const std::string BaseCluster = "base";
 const std::string BaseWlanCluster = "base_wlan";
 const std::string BaseWwanCluster = "base_wwan";
+const LowerCaseString H2UpstreamHeader{"x-envoy-mobile-upstream-protocol"};
```
goaway: I propose simplifying this with a `x-envoy-mobile-force-h2` header. If a header with this name is present assume inferred h2 support, if not, default to h1.

library/common/http/dispatcher.cc
```diff
@@ -477,19 +477,38 @@ const LowerCaseString ClusterHeader{"x-envoy-mobile-cluster"};
...
+const std::string H2Suffix = "_h2";
+const std::string BaseClusterH2 = BaseCluster + H2Suffix;
+const std::string BaseWlanClusterH2 = BaseWlanCluster + H2Suffix;
+const std::string BaseWwanClusterH2 = BaseWwanCluster + H2Suffix;
```
goaway: nit: this would actually be one less line of code if you just used literals ;)

library/kotlin/src/io/envoyproxy/envoymobile/UpstreamHttpProtocol.kt
```diff
@@ -0,0 +1,9 @@
+package io.envoyproxy.envoymobile
+
+/**
+ * Available upstream http protocols.
+ */
+enum class UpstreamHttpProtocol(internal val stringValue: String) {
+  HTTP1("http1"),
+  HTTP2("http2"),
+}
```
goaway: Can drop this enum if we just use a single header to force h2 (which is sort of non-standard anyways and only works with known destinations).

goaway:
Another approach would be to have an engine-level configuration option that allows you to specify authorities for which you know you can force h2. In some ways I like that better than a per-request approach, since support for this functionality is non-standard and relatively uncommon, and it avoids inadvertently dispatching requests to destinations that don't support it. It's more a property of the destination (`:authority`) than it is of the request.

That said, I'm okay with the per-request approach as well (it just means the application will have to implement its own whitelist).

library/common/http/dispatcher.cc
junr03: I had originally written this PR with that approach, and actually with something similar to what you suggest below re: central configuration. 

However, I moved to this approach for a few reasons:
1.  The enum allows us to express potential expansion of future options: h3, auto (when alpn support lands on upstream connections).
2. The enum allows us to more easily migrate existing code to those future options without breaking the API if we find ourselves in a future where we care more strictly about API breakage as, hopefully, we see more adopters.
3. IMO the protocol support of your destination is an application level concern (if the library can't do protocol negotiation for you), and the application developer should tell us that prior knowledge. I am also not a big fan of centralized configuration given at boot time, as you know :)

Right now absence of the header defaults us to http1 per the implementation in the Http::Dispatcher::setDestinationCluster. I can do:

> I'd suggest moving this out of the constructor, and adding it via a builder method.

per your suggestion make this setting a less strong requirement. I think this brings us closer to the sentiment of making this engine-level configuration without having to wire more configuration through.

lmk what you think.

junr03:
@goaway thanks for all the comments. I am answering to all of them in aggregate because I think they are all related.

I had originally written this PR with that approach, and actually with something similar to what you suggest below re: central configuration. 

However, I moved to this approach for a few reasons:
1.  The enum allows us to express potential expansion of future options: h3, auto (when alpn support lands on upstream connections).
2. The enum allows us to more easily migrate existing code to those future options without breaking the API if we find ourselves in a future where we care more strictly about API breakage as, hopefully, we see more adopters.
3. IMO the protocol support of your destination is an application level concern (if the library can't do protocol negotiation for you), and the application developer should tell us that prior knowledge. I am also not a big fan of centralized configuration given at boot time, as you know :)

Right now absence of the header defaults us to http1 per the implementation in the Http::Dispatcher::setDestinationCluster. I can do:

> I'd suggest moving this out of the constructor, and adding it via a builder method.

per your suggestion make this setting a less strong requirement. I think this brings us closer to the sentiment of making this engine-level configuration without having to wire more configuration through.

lmk what you think.

docs/root/api/http.rst
rebello95: +1 to this suggestion. `addUpstreamProtocol`?

library/common/http/dispatcher.cc
rebello95: looks like this is actually defined twice too?

library/common/http/dispatcher.cc
```diff
@@ -478,19 +478,38 @@ const LowerCaseString ClusterHeader{"x-envoy-mobile-cluster"};
...
+  // Determine upstream protocol. Use http2 if selected for explicitly, otherwise (any other value,
+  // absence of value) select http1.
+  // TODO(junr03): once http3 is available this would probably benefit from some sort of struct that
+  // maps to appropriate cluster names. However, with only two options (http1/2) this suffices.
+  bool use_h2_cluster{};
+  const HeaderEntry* entry = headers.get(H2UpstreamHeader);
+  if (entry) {
+    if (entry->value() == "http2") {
+      use_h2_cluster = true;
+    }
```
rebello95: do you think it's worth asserting in the `else` case if the caller specifies a value that's unsupported?

library/kotlin/src/io/envoyproxy/envoymobile/Request.kt
```diff
@@ -7,12 +7,14 @@ package io.envoyproxy.envoymobile
...
  * @param authority The URL authority for the request (i.e., "api.foo.com").
  * @param path The URL path for the request (i.e., "/foo").
+ * @param upstreamHttpProtocol The protcol version to use for upstream requests.
  */
 data class Request internal constructor(
     val method: RequestMethod,
     val scheme: String,
     val authority: String,
     val path: String,
+    val upstreamHttpProtocol: UpstreamHttpProtocol,
```
rebello95: nit: I think we should prefer `HTTP` over `Http` on all of these, in line with how we use `MS` for milliseconds over `Ms`. wdyt?

library/kotlin/test/io/envoyproxy/envoymobile/RequestMapperTest.kt
```diff
@@ -123,7 +145,7 @@ class RequestMapperTest {
         perRetryTimeoutMS = 9001)
     val retryPolicyHeaders = retryPolicy.outboundHeaders()
 
-    val requestHeaders = RequestBuilder(method = RequestMethod.POST, scheme = "https", authority = "api.foo.com", path = "/foo")
+    val requestHeaders = RequestBuilder(method = RequestMethod.POST, scheme = "https", authority = "api.foo.com", path = "/foo", upstreamHttpProtocol = UpstreamHttpProtocol.HTTP2)
```
rebello95: Changing this to a builder as @goaway suggested would allow for reverting a lot of these test changes

junr03: Yep, I like the suggestion to move out to the builder. It would also make it easier to deprecate if we want to move into an only ALPN-based future when that is supported.

rebello95:
A couple comments based on the discussions above:

1 (enum vs bool): If http2 were to be the only protocol we need to differentiate going forward, I think a boolean would be fine. However, since http3 is coming in the near future and we probably don't want to make protocol negotiation a hard blocker or to make breaking changes to the interfaces being introduced here, an enum seems like it'd probably make sense.

2 (engine versus request): Adding this information on a per-request basis is a bit more in line with how other configurations are set up now in Envoy Mobile. We moved to the dynamic forward proxy so that we no longer have to specify a fixed set of domains when starting the library, and requiring an http2 whitelist up front when starting up seems like a step backwards as it doesn't allow for new http2 domains to be discovered without restarting the library (not currently supported). Allowing this information to be specified per-request as a builder seems reasonable, especially considering @junr03 is stripping all `x-envoy-mobile*` headers (keeping the header implementation internal to the library)

library/kotlin/src/io/envoyproxy/envoymobile/RequestBuilder.kt
```diff
@@ -22,16 +22,8 @@ class RequestBuilder(
   // Retry policy to use for this request.
   private var retryPolicy: RetryPolicy? = null
 
-  /**
```
junr03: moving below for consistent ordering with swift

examples/swift/hello_world/ViewController.swift
```diff
@@ -49,7 +49,9 @@ final class ViewController: UITableViewController {
     let requestID = self.requestCount
     let request = RequestBuilder(method: .get, scheme: kRequestScheme,
                                  authority: kRequestAuthority,
-                                 path: kRequestPath).build()
+                                 path: kRequestPath, upstreamHttpProtocol: .http2)
+        .addUpstreamHttpProtocol(.http2)
```
rebello95: Can we update objc/java too?

library/common/http/dispatcher.cc
```diff
@@ -478,19 +478,38 @@ const LowerCaseString ClusterHeader{"x-envoy-mobile-cluster"};
...
+
+  // Determine upstream protocol. Use http2 if selected for explicitly, otherwise (any other value,
+  // absence of value) select http1.
+  // TODO(junr03): once http3 is available this would probably benefit from some sort of struct that
+  // maps to appropriate cluster names. However, with only two options (http1/2) this suffices.
+  bool use_h2_cluster{};
+  const HeaderEntry* entry = headers.get(H2UpstreamHeader);
+  if (entry) {
+    if (entry->value() == "http2") {
+      use_h2_cluster = true;
```
rebello95: Also seems like this should be `use_h2_cluster = entry->value() == "http2"`

library/kotlin/test/io/envoyproxy/envoymobile/RequestBuilderTest.kt
```diff
@@ -24,6 +24,25 @@ class RequestBuilderTest {
...
+    val retryPolicy = RetryPolicy(maxRetryCount = 23, retryOn = listOf(RetryRule.STATUS_5XX, RetryRule.CONNECT_FAILURE), perRetryTimeoutMS = 1234)
+    val request = RequestBuilder(method = RequestMethod.POST, scheme = "https", authority = "api.foo.com", path = "foo")
+        .addUpstreamHttpProtocol(UpstreamHttpProtocol.HTTP2)
+        .build()
+
+    assertThat(request.upstreamHttpProtocol).isEqualTo(UpstreamHttpProtocol.HTTP2)
+  }
+
+  @Test
+  fun `not adding upstream http protocol should have null body in request`() {
```
rebello95:
```suggestion
  fun `not adding upstream http protocol should have null upstream protocol in request`() {
```

library/kotlin/test/io/envoyproxy/envoymobile/RequestMapperTest.kt
```diff
@@ -57,24 +77,29 @@ class RequestMapperTest {
   fun `invalid semicolon prefixed header keys are filtered out of outbound request headers`() {
     val requestHeaders = RequestBuilder(method = RequestMethod.POST, scheme = "https", authority = "api.foo.com", path = "/foo")
         .addHeader(":restricted", "value")
+        .addHeader("x-envoy-mobile-test", "value")
```
rebello95: I'd probably make this a different test or update the name of the test

rebello95: could use "restricted headers" like the other test below

library/swift/src/RequestMapper.swift
```diff
@@ -1,4 +1,4 @@
-private let kRestrictedHeaderPrefix = ":"
+private let kRestrictedHeaderPrefix = [":", "x-envoy-mobile"]
```
rebello95:
```suggestion
private let kRestrictedHeaderPrefixes = [":", "x-envoy-mobile"]
```

library/swift/src/UpstreamHttpProtocol.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+/// Available upstream http protocols.
+@objc
+public enum UpstreamHttpProtocol: Int {
+  case http1
+  case http2
+
+  /// String representation of the log level.
```
rebello95:
```suggestion
  /// String representation of the protocol.
```

library/swift/src/UpstreamHttpProtocol.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+/// Available upstream http protocols.
```
rebello95:
```suggestion
/// Available upstream HTTP protocols.
```

junr03:
@rebello95 sorry wasn't ready for a pass. I think I hit all your comments. Let me take a look

examples/swift/hello_world/ViewController.swift
junr03: I intentionally left this with h1 so we have coverage for both paths. I can drop a comment in the examples to make this explicit.

library/common/http/dispatcher.cc
junr03: I want the else for the assertion. I guess I could do 

```
use_h2_cluster = entry->value() == "http2"
if (!use_h2_cluster) { ASSERT(entry->value()->getStringView() == "http1"); }
```

six of one half a dozen of the other?

library/kotlin/src/io/envoyproxy/envoymobile/Request.kt
junr03: I feel conflicted because in the envoy codebase we dont use all caps...

examples/swift/hello_world/ViewController.swift
rebello95: Comment would be nice 👍

library/common/http/dispatcher.cc
rebello95: I think what you have now is fine

library/kotlin/src/io/envoyproxy/envoymobile/Request.kt
rebello95: Eh just leave it I guess

library/kotlin/src/io/envoyproxy/envoymobile/RequestMapper.kt
```diff
@@ -4,13 +4,17 @@ internal fun Request.outboundHeaders(): Map<String, List<String>> {
   val retryPolicyHeaders = retryPolicy?.outboundHeaders() ?: emptyMap()
 
   val result = mutableMapOf<String, List<String>>()
-  result.putAll(headers.filter { entry -> !entry.key.startsWith(":") })
+  result.putAll(headers.filter { entry -> !(entry.key.startsWith(":") || entry.key.startsWith("x-envoy-mobile"))})
```
rebello95: Up to you but using this might be more readable: `!entry.key.startsWith(":") && !entry.key.startsWith("x-envoy-mobile")`

junr03: I am always ambivalent about which one is more readable. I'll change to kick CI... 💩

rebello95: 😭

PR #699, opened 2020-02-20
build: disable known issue assertions

goaway:
Description: With this change https://github.com/envoyproxy/envoy/pull/10015 Envoy now supports conditionally disabling assertions that cover known issues via build options. This allows us to resolve the crash on shutdown on iOS while the upstream issue is investigated.
Risk Level: Low
Testing: Local, upstream


rebello95:
Ah I just bumped this morning which includes your change: https://github.com/lyft/envoy-mobile/pull/698
Can you drop the `envoy` submodule change and just apply the build file modification?

rebello95:
Finishes resolution of https://github.com/lyft/envoy-mobile/issues/572

PR #704, opened 2020-02-27
upstream: update ref to include stats flush hook

rebello95:
Bumping to include https://github.com/envoyproxy/envoy/pull/10097, which we'll be utilizing shortly.


rebello95:
Size comparison failed due to a minimal size change that pushed us over the 5MB edge (relative size increase is still ~0%). Bumping here: https://github.com/lyft/envoy-mobile/pull/705

PR #707, opened 2020-02-27
core: expose functionality for flushing stats manually

rebello95:
Adds functionality built on top of the upstream Envoy changes in https://github.com/envoyproxy/envoy/pull/10097 which will allow the bridging layers to manually flush stats based on lifecycle changes in the mobile application.

Note: Stats must be flushed from the main Envoy thread, as validated by the following assertion failure that occurs when attempting to flush from another thread:

```
[2020-02-27 14:38:43.181][4716580][critical][assert] [external/envoy/source/common/thread_local/thread_local_impl.cc:190] assert failure: std::this_thread::get_id() == main_thread_id_.
```

Follow-up PRs will call these functions from Java/Objective-C.

Part of https://github.com/lyft/envoy-mobile/issues/573.

### DIFF
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -108,6 +108,11 @@ Engine::~Engine() {
 }
 
 void Engine::flushStats() {
+  // Stats must be flushed from the main thread. If we aren't on the main thread, join it.
+  if (main_thread_.joinable()) {
+    main_thread_.join();
+  }
+
   // Server instance will be null if the post-init callback has not been completed within run().
   // In this case, we can simply ignore the flush.
   if (server_instance_ != nullptr) {
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -66,6 +66,7 @@ envoy_status_t Engine::run(std::string config, std::string log_level) {
           Server::Instance* server = TS_UNCHECKED_READ(main_common_)->server();
           auto api_listener = server->listenerManager().apiListener()->get().http();
           ASSERT(api_listener.has_value());
+          server_instance_ = server;
           http_dispatcher_->ready(server->dispatcher(), api_listener.value());
         });
   } // mutex_
@@ -104,6 +105,14 @@ Engine::~Engine() {
 
   // Now we wait for the main thread to wrap things up.
   main_thread_.join();
+}
+
+void Engine::flushStats() {
+  // Server instance will be null if the post-init callback has not been completed within run().
+  // In this case, we can simply ignore the flush.
+  if (server_instance_ != nullptr) {
+    server_instance_->flushStats();
+  }
 }
 
 Http::Dispatcher& Engine::httpDispatcher() { return *http_dispatcher_; }
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -66,7 +66,7 @@ envoy_status_t Engine::run(std::string config, std::string log_level) {
           Server::Instance* server = TS_UNCHECKED_READ(main_common_)->server();
           auto api_listener = server->listenerManager().apiListener()->get().http();
           ASSERT(api_listener.has_value());
-          server_instance_ = server;
+          server_ = server;
           http_dispatcher_->ready(server->dispatcher(), api_listener.value());
         });
   } // mutex_
@@ -108,15 +108,11 @@ Engine::~Engine() {
 }
 
 void Engine::flushStats() {
-  // Stats must be flushed from the main thread. If we aren't on the main thread, join it.
-  if (main_thread_.joinable()) {
-    main_thread_.join();
-  }
-
-  // Server instance will be null if the post-init callback has not been completed within run().
+  // The server will be null if the post-init callback has not been completed within run().
   // In this case, we can simply ignore the flush.
-  if (server_instance_ != nullptr) {
-    server_instance_->flushStats();
+  if (server_) {
+    // Stats must be flushed from the main thread.
+    server_->dispatcher().post([this]() -> void { server_->flushStats(); });
   }
 }
 
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -112,6 +112,7 @@ void Engine::flushStats() {
   // In this case, we can simply ignore the flush.
   if (server_) {
     // Stats must be flushed from the main thread.
+    // Dispatching should be moved after https://github.com/lyft/envoy-mobile/issues/720
     server_->dispatcher().post([this]() -> void { server_->flushStats(); });
   }
 }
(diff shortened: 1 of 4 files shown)


library/common/main_interface.h
```diff
@@ -91,6 +91,13 @@ envoy_engine_t init_engine();
  */
 envoy_status_t set_preferred_network(envoy_network_t network);
 
+/**
+ * Flush the stats sinks outside of a flushing interval.
+ * Note: stats flushing may not be synchronous.
+ * Therefore, this function may return prior to flushing taking place.
+ */
```
junr03: Might be worth to also document that flushing before the engine has initialized results in no flush.

library/common/engine.cc
```diff
@@ -106,6 +107,19 @@ Engine::~Engine() {
   main_thread_.join();
 }
 
+void Engine::flushStats() {
+  // Stats must be flushed from the main thread. If we aren't on the main thread, join it.
```
junr03: True that we need to flush from Envoy's main thread. However, I believe this won't accomplish it, as you are blocking the platform thread until Envoy's main thread finishes execution. What we need to do is:

```c++
if (server_instance_) {
  server_instance_->dispatcher().post([this]() -> void {
    server_instance_->flushStats();
  });
}
```

(sorry if there is a syntax error, you get the idea)

This way we are posting the function call into the main thread's event loop.

library/common/engine.h
```diff
@@ -28,16 +28,24 @@ class Engine {
...
   envoy_engine_callbacks callbacks_;
   Thread::MutexBasicLockable mutex_;
   Thread::CondVar cv_;
   std::thread main_thread_;
-  std::unique_ptr<Envoy::Http::Dispatcher> http_dispatcher_;
-  std::unique_ptr<Envoy::MainCommon> main_common_ GUARDED_BY(mutex_);
-  Envoy::Server::ServerLifecycleNotifier::HandlePtr postinit_callback_handler_;
+  std::unique_ptr<Http::Dispatcher> http_dispatcher_;
+  std::unique_ptr<MainCommon> main_common_ GUARDED_BY(mutex_);
+  Server::Instance* server_instance_;
```
junr03: nit: just `server_`?

library/common/engine.cc
rebello95: Ah good call. Updating

rebello95:
Thanks for the review @junr03! Updated.

library/common/engine.h
```diff
@@ -28,16 +28,24 @@ class Engine {
...
   envoy_engine_callbacks callbacks_;
   Thread::MutexBasicLockable mutex_;
   Thread::CondVar cv_;
   std::thread main_thread_;
-  std::unique_ptr<Envoy::Http::Dispatcher> http_dispatcher_;
-  std::unique_ptr<Envoy::MainCommon> main_common_ GUARDED_BY(mutex_);
-  Envoy::Server::ServerLifecycleNotifier::HandlePtr postinit_callback_handler_;
+  std::unique_ptr<Http::Dispatcher> http_dispatcher_;
+  std::unique_ptr<MainCommon> main_common_ GUARDED_BY(mutex_);
+  Server::Instance* server_;
```
goaway: So I do think this will be racy. The raw pointer is set and read from different threads, so there's at least a window where threads calling flush_stats may not see the server even though it's been set. The only time I think it would crash though is potentially on teardown. However, we've spent some time chasing down shutdown crashes, so I don't know if we really want to add to that risk.

I would suggest leveraging the http_dispatcher and its mutex for accessing Envoy's `event_dispatcher` instead for now, since I think this is a risky pattern to promulgate without further protections. Additionally, I think we should limit the surface area of cross-thread interactions just to make it easier to reason about logic and improve safety.

Long-term I'd like to split the business logic of encoding/decoding HTTP requests and handling dispatch into two separate entities, instead of having a single `HTTPDispatcher`.

junr03: I agree with the long term vision.

And with the fact that we _might_ not get a flush due to visibility ordering of the server pointer (although, not sure if we should be concerned about a stronger guarantee there?)

But I don't think I agree with tacking more on to the http dispatcher. We still need a reference to the Server::Instance somewhere because it is through that interface that we flush stats. I am not sure it makes sense to make the http::dispatcher do that. Because at that point we are really making the http::dispatcher be a kitchen sync.

I wholeheartedly agree with ` Additionally, I think we should limit the surface area of cross-thread interactions just to make it easier to reason about logic and improve safety.` And I would like to think how we can do that but without just accumulating everything in the same class.

goaway: I get the kitchen sink argument and I covered a little bit more in a slack thread, but I think that routing things through the dispatcher right now will actually make it easier to port things over to a dedicated cross-thread dispatch mechanism later, since everything that needs to be ported in the same place.

I think my main two concerns remain:
1) I think there are probably circumstances this can crash.
2) I think there's risk in needing to track down unguarded multi-threaded access later when we do clean things up.

With respect to 2), right now every multi-threaded access _does_ occur right now through a safe mechanism. Exceptional cases like preferred_network are atomic, and primary cases use the core mutexes inside the Dispatcher. This will be the first case of unguarded access to a raw pointer, which I think is something to be avoided in and of itself.

I'm a bit less worried about (temporary) cruft in the dispatcher, because everything we can clean up once dispatch is split out will be located in one place.

goaway: Okay, I think the risks both perspectives cite are probably not that big. I still think there's a chance we could crash due to unsafe access here, but hey, also we might not. Since you're not here to discuss in person @junr03 you get to win this one by default. ;)

I'm going to file an issue about splitting out a generalized interface for cross-thread access, and this server_ ref can be flagged with that issue.

goaway: https://github.com/lyft/envoy-mobile/issues/720

rebello95: This was a good discussion to have 😃 I went ahead and linked the issue inline in the code so we can update after

junr03: Sorry I disappeared, it was late here!

re: 1 and 2 above.

1. I don't think there is a risk (famous last words) because of the a) weak_ptr semantics that you introduced @goaway,  and b) the engine holding ownership of main_common, which in turn holds ownership of the server object. This means that as long as we are able to lock the engine, we will have a valid server pointer. Right?

2. Yep, I think we have probably reached a point with out accesses that we need to straighten things out in a more digestible way. Thanks for filing #720, looking forward to chatting about that when I am back.

junr03: Thanks for linking the issue inline @rebello95

PR #708, opened 2020-02-27
core: add missing header docstrings

rebello95:

### DIFF
--- a/library/common/engine.h
+++ b/library/common/engine.h
@@ -21,11 +21,25 @@ namespace Envoy {
 
 class Engine {
 public:
+  /**
+   * Constructor for a new engine instance.
+   * @param callbacks, the callbacks to be used for streams started through the engine.
+   * @param config, the Envoy configuration to use when starting the instance.
+   * @param log_level, the log level with which to configure the engine.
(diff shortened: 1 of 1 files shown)


library/common/engine.h
```diff
@@ -21,11 +21,25 @@ namespace Envoy {
 
 class Engine {
 public:
+  /**
+   * Constructor for a new engine instance.
+   * @param callbacks, the callbacks to be used for streams started through the engine.
```
goaway: these are actually callbacks specific to the engine, unrelated to streams IIRC

library/common/engine.h
```diff
@@ -21,11 +21,25 @@ namespace Envoy {
...
+   * @param config, the Envoy configuration to use when starting the instance.
+   * @param log_level, the log level with which to configure the engine.
+   * @param preferred_network, the preferred network that should be used for new streams.
```
goaway: This is a dynamic value that is updated based on OS hooks. This is simple a handle to retrieve the current value.

rebello95: ah yeah good catch

Issue #573, opened 2019-11-02
stats: expose API to flush stats based on lifecycle events


junr03:
Similar to the preferred network API introduced in https://github.com/lyft/envoy-mobile/pull/442. The Engine should expose a stat flushing API that hooks onto lifecycle events so the engine can flush stats when it is being backgrounded or killed.

This will require Envoy OSS server impl to expose the flushStats function. https://github.com/envoyproxy/envoy/blob/e7c7ca4c09acf8d389de6d5bf8f9cb005415b86a/source/server/server.h#L243

PR #709, opened 2020-02-27
ios: flush stats on background / termination

rebello95:
Builds on https://github.com/lyft/envoy-mobile/pull/707 by consuming this interface to flush stats when the app is backgrounded or terminated. This will help ensure that stats are sent to the server when possible before the app potentially loses its in-memory stats cache.

Part of #573.

### DIFF
--- a/library/objective-c/EnvoyDeviceLifecycleMonitor.m
+++ b/library/objective-c/EnvoyDeviceLifecycleMonitor.m
@@ -1,0 +1,35 @@
+#import "library/objective-c/EnvoyEngine.h"
+
+#import "library/common/main_interface.h"
+
+#import <UIKit/UIKit.h>
+
+@implementation EnvoyDeviceLifecycleMonitor
+
++ (void)startFlushingStatsOnLifecycleChanges {
+  static dispatch_once_t lifecycleMonitoringStarted;
+  dispatch_once(&lifecycleMonitoringStarted, ^{
+    [self startObservingLifecycleNotifications];
+  });
+}
+
+#pragma mark - Private
+
++ (void)startObservingLifecycleNotifications {
+  NSNotificationCenter *notificationCenter = [NSNotificationCenter defaultCenter];
+  [notificationCenter addObserver:self
+                         selector:@selector(lifecycleDidChangeWithNotification:)
+                             name:UIApplicationWillResignActiveNotification
+                           object:nil];
+  [notificationCenter addObserver:self
+                         selector:@selector(lifecycleDidChangeWithNotification:)
+                             name:UIApplicationWillTerminateNotification
+                           object:nil];
+}
+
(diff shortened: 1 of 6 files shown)


rebello95:
Going with a bit of a different approach, so ignore this PR for now 😄

library/objective-c/EnvoyDeviceLifecycleMonitor.m
```diff
@@ -0,0 +1,35 @@
...
+#import "library/common/main_interface.h"
+
+#import <UIKit/UIKit.h>
+
+@implementation EnvoyDeviceLifecycleMonitor
+
++ (void)startFlushingStatsOnLifecycleChanges {
+  static dispatch_once_t lifecycleMonitoringStarted;
+  dispatch_once(&lifecycleMonitoringStarted, ^{
+    [self startObservingLifecycleNotifications];
```
goaway: I feel sort of like `startObservingLifecycleNotification` should be the public method name, since the fact that we're flushing stats on them seems more like an implementation detail.

library/objective-c/EnvoyDeviceLifecycleMonitor.m
```diff
@@ -0,0 +1,35 @@
...
+                             name:UIApplicationWillResignActiveNotification
+                           object:nil];
+  [notificationCenter addObserver:self
+                         selector:@selector(lifecycleDidChangeWithNotification:)
+                             name:UIApplicationWillTerminateNotification
+                           object:nil];
+}
+
++ (void)lifecycleDidChangeWithNotification:(NSNotification *)notification {
+  NSLog(@"[Envoy] triggering stats flush (%@)", notification.name);
```
goaway: Is this meant to be here? (Do we generally utilize NSLog in our platform layer? Is it normal for third-part libraries to do so?)

library/objective-c/EnvoyDeviceLifecycleMonitor.m
rebello95: I can rename

library/objective-c/EnvoyDeviceLifecycleMonitor.m
rebello95: We have several of these at the moment. @junr03 and I briefly discussed the possibility of using `ENVOY_LOG` rather than `NSLog` (which would be much nicer), but it'd require the native layer to depend on upstream Envoy code directly which would be a bit of a shift. Opened an issue for this here: https://github.com/lyft/envoy-mobile/issues/719

goaway: Sounds good, and thanks for opening the issue.

rebello95:
Replacing with https://github.com/lyft/envoy-mobile/pull/717

PR #717, opened 2020-03-02
stats: flush stats based on application lifecycle

rebello95:
## Description

This change adds the ability to automatically flush stats based on application lifecycle changes (in addition to the scheduled flushing interval).

Enabling this flushing system is done **automatically** when the iOS/Android clients are instantiated.

On **iOS**, this is done by observing `NSNotificationCenter`. 

**Android** requires an `Application` context to use for registering lifecycle callbacks. To accommodate this, the `Application` is now passed in to the Android builder instead of the `Context`.

Right now this new stats flushing behavior is enabled by default. In the future, we would like to investigate making it optional using the builder pattern. For example:

```swift
EnvoyClientBuilder()
  .addAppLifecycleHandling()
  .build()
```

```kotlin
AndroidEnvoyClientBuilder(baseContext)
  .addAppLifecycleHandling(getApplication())
  .build()
```

https://github.com/lyft/envoy-mobile/issues/721 will track this work.

As part of this PR, I also updated some documentation to better reflect the differences between Android/Java interfaces and those present on iOS.

## Testing

- Tested locally and verified that stats are flushed when backgrounding the iOS and Android sample apps
- Sample apps use this new mechanism

## Risk Level

Low, this is an additive change.

## Issues

Resolves https://github.com/lyft/envoy-mobile/issues/573


library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngineImpl.java
```diff
@@ -1,12 +1,35 @@
...
+  private final EnvoyEngine envoyEngine;
 
   public AndroidEngineImpl(Context context) {
-    super();
+    envoyEngine = new EnvoyEngineImpl();
     AndroidJniLibrary.load(context);
     AndroidNetworkMonitor.load(context);
   }
+
+  public EnvoyHTTPStream startStream(EnvoyHTTPCallbacks callbacks) {
```
buildbreaker:
```suggestion
  @Overrides
  public EnvoyHTTPStream startStream(EnvoyHTTPCallbacks callbacks) {
```

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngineImpl.java
```diff
@@ -1,12 +1,35 @@
...
+    return envoyEngine.startStream(callbacks);
+  }
+
+  public int runWithConfig(String configurationYAML, String logLevel) {
```
buildbreaker:
```suggestion
  @Overrides
  public int runWithConfig(String configurationYAML, String logLevel) {
```

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngineImpl.java
```diff
@@ -1,12 +1,35 @@
...
+    return envoyEngine.runWithConfig(configurationYAML, logLevel);
+  }
+
+  public int runWithConfig(EnvoyConfiguration envoyConfiguration, String logLevel) {
```
buildbreaker:
```suggestion
  @Overrides
  public int runWithConfig(EnvoyConfiguration envoyConfiguration, String logLevel) {
```

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngineImpl.java
```diff
@@ -1,12 +1,35 @@
...
+    return envoyEngine.runWithConfig(envoyConfiguration, logLevel);
+  }
+
+  public int runWithConfig(AndroidEnvoyConfiguration envoyConfiguration, String logLevel) {
```
buildbreaker: This doesn't need the `@Overrides`

PR #724, opened 2020-03-03
android: fix app lifecycle monitoring with typed config

rebello95:
We should be registering the application for lifecycle callbacks in both the typed config and YAML config cases.

Related to https://github.com/lyft/envoy-mobile/pull/717/files.


rebello95:
Merging to fix master.

PR #735, opened 2020-03-07
core: make streams with`is_internally_created`

### LINKED ISSUE #733: core: retries do not work
## Problem

We identified an issue where specifying a `RetryPolicy` in Envoy Mobile has no effect (requests are not retried).

## Root cause

The root cause of this bug is that Envoy Mobile converts the `RetryPolicy` structures to `x-envoy-` headers and passes them through with the outbound request. This is done via headers rather than through engine configuration (which does work from my testing) in order to allow for different retry policies on a per-request basis.

When Envoy's connection manager receives these requests, it [flags Envoy Mobile's request as an external request](https://github.com/envoyproxy/envoy/blob/master/source/common/http/conn_manager_utility.cc#L138-L149) and [strips all retry headers](https://github.com/envoyproxy/envoy/blob/cb7f13c67cb8b9f02afc053f6eef743a3ed3c1af/source/common/http/conn_manager_utility.cc#L172-L181).

As a result, Envoy's [retry policy parser](https://github.com/envoyproxy/envoy/blob/cb7f13c67cb8b9f02afc053f6eef743a3ed3c1af/source/common/router/retry_state_impl.cc#L45) is never called with the original retry headers, and the request is never retried.

### 1) Pass context directly to Envoy to indicate Envoy Mobile is an internal requestor

We could potentially pass additional information directly to Envoy to indicate that the request is being created within Envoy Mobile and to avoid stripping some of these headers. This isn't necessarily a new paradigm, as something similar is done [here](https://github.com/envoyproxy/envoy/blob/c5f4302325223765b660f0f366ce25bf8570e7a5/source/common/http/conn_manager_impl.cc#L259).

This is likely the most preferable solution, though complexity will require more investigation.

### 2) Update Envoy's heuristic for "internal" traffic

The [heuristic](https://github.com/envoyproxy/envoy/blob/master/source/common/http/conn_manager_utility.cc#L138-L149) flagging Envoy Mobile's traffic as external could potentially be updated to properly flag Envoy Mobile traffic as internal.

This could potentially be risky since it's something heavily relied upon with legacy behavior. Marking this traffic as internal may also produce unwanted side effects.

### 3) Create retry policies in engine configs and update router

We could potentially add a way for users to create retry policies when starting the engine, and include those in the configuration we pass to start Envoy. At a later time, we could use a custom header (i.e., `x-envoy-mobile-retry-policy`) to route requests differently inside the [router](https://github.com/lyft/envoy-mobile/blob/bc56e30c88fd0f8961533cb027a8f2130371b458/library/common/config_template.cc#L23-L31) based on their assigned retry policy.

This is a bit clunky, and is restrictive on end users since it will prevent them from assigning new/custom retry policies at runtime and requires them to have knowledge of all possible policies when starting the library.

junr03:
Description: in order to prevent the Http::ConnectionManager from sanitizing retry headers attached by the platform, streams are now marked by `is_internally_created` true.
Risk Level: medium, but output shown below shows that the requests still have expected headers.
Testing: log and behavior confirmation. Integration tests are needed and will be focus of imminent work.

Fixes #733

### DIFF
--- a/library/common/http/dispatcher.cc
+++ b/library/common/http/dispatcher.cc
@@ -295,9 +295,11 @@ envoy_status_t Dispatcher::startStream(envoy_stream_t new_stream_handle,
 
     // Only the initial setting of the api_listener_ is guarded.
     //
-    // Note: streams created by Envoy Mobile are tagged as is_internally_created. This means that the Http::ConnectionManager _will not_ sanitize headers when creating a stream.
+    // Note: streams created by Envoy Mobile are tagged as is_internally_created. This means that
+    // the Http::ConnectionManager _will not_ sanitize headers when creating a stream.
     direct_stream->request_decoder_ =
-        &TS_UNCHECKED_READ(api_listener_)->newStream(*direct_stream->callbacks_, true /* is_internally_created */);
+        &TS_UNCHECKED_READ(api_listener_)
+             ->newStream(*direct_stream->callbacks_, true /* is_internally_created */);
 
     Thread::ReleasableLockGuard lock(streams_lock_);
     streams_.emplace(new_stream_handle, std::move(direct_stream));
@@ -322,14 +324,18 @@ envoy_status_t Dispatcher::sendHeaders(envoy_stream_t stream, envoy_headers head
     if (direct_stream) {
       RequestHeaderMapPtr internal_headers = Utility::toRequestHeaders(headers);
       setDestinationCluster(*internal_headers);
-      // Set the x-forwarded-proto header to https because Envoy Mobile only has clusters with TLS enabled.
-      // This is done here because the ApiListener's synthetic connection would make the Http::ConnectionManager set the scheme to http otherwise.
-      // In the future we might want to configure the connection instead of setting the header here.
+      // Set the x-forwarded-proto header to https because Envoy Mobile only has clusters with TLS
+      // enabled. This is done here because the ApiListener's synthetic connection would make the
+      // Http::ConnectionManager set the scheme to http otherwise. In the future we might want to
+      // configure the connection instead of setting the header here.
       // https://github.com/envoyproxy/envoy/issues/10291
       //
-      // Setting this header is also currently important because Envoy Mobile starts stream with the ApiListener setting the is_internally_created bool to true.
-      // This means the Http::ConnectionManager *will not* mutate Envoy Mobile's request headers. One of the mutations done is adding the x-forwarded-proto header if not present.
-      // Unfortunately, the router relies on the present of this header to determine if it should provided a route for a request here:
+      // Setting this header is also currently important because Envoy Mobile starts stream with the
+      // ApiListener setting the is_internally_created bool to true. This means the
+      // Http::ConnectionManager *will not* mutate Envoy Mobile's request headers. One of the
+      // mutations done is adding the x-forwarded-proto header if not present. Unfortunately, the
+      // router relies on the present of this header to determine if it should provided a route for
+      // a request here:
       // https://github.com/envoyproxy/envoy/blob/c9e3b9d2c453c7fe56a0e3615f0c742ac0d5e768/source/common/router/config_impl.cc#L1091-L1096
       internal_headers->setReferenceForwardedProto(Headers::get().SchemeValues.Https);
       ENVOY_LOG(debug, "[S{}] request headers for stream (end_stream={}):\n{}", stream, end_stream,
--- a/library/common/http/dispatcher.cc
+++ b/library/common/http/dispatcher.cc
@@ -294,8 +294,11 @@ envoy_status_t Dispatcher::startStream(envoy_stream_t new_stream_handle,
         std::make_unique<DirectStreamCallbacks>(*direct_stream, bridge_callbacks, *this);
 
     // Only the initial setting of the api_listener_ is guarded.
+    //
+    // Note: streams created by Envoy Mobile are tagged as is_internally_created. This means that the Http::ConnectionManager _will not_ sanitize headers when creating a stream.
+    // This seems like the correct thing to do, as we want the headers created by the client to be preserved.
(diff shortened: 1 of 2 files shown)


junr03:
Note that I ultimately decided to use `is_internally_created` instead of modifying the config to `use_remote_address: true` to make these requests be treated as "internal" requests. I ended up doing this because I hypothesize that it is riskier to have client requests include headers that would open up a misconfigured front proxy from incorrectly classifying them. Overall neither option is that optimal, and we might want to think about this a bit more after we start with prod experiments.

Additionally, there is no current good end-to-end way to test that functionality that is enabled by Envoy Mobile is actually enacted by Envoy for this type of problems. We are investing into this imminently. For instance, the closest thing we could do here is an integration test that validates that the headers are not stripped in the HCM but there is no good testing bed to do this right now.

junr03:
Logs to support change of behavior. The example app request builder was modified with a RetryPolicy.

Before this the changes in this PR headers were getting stripped by the Http::ConnectionManager and thus the router was not building the desired RetryStateImpl:

```
[2020-03-06 15:32:44.813][8740313][debug][http] [library/common/http/dispatcher.cc:306] [S151] start stream
[2020-03-06 15:32:44.814][8740313][debug][http] [library/common/http/dispatcher.cc:326] [S151] request headers for stream (end_stream=false):
':scheme', 'https'
':method', 'GET'
':path', '/ping'
':authority', 'api.lyft.com'
'x-envoy-retry-on', 'connect-failure,5xx'
'x-envoy-mobile-upstream-protocol', 'http2'
'x-envoy-upstream-rq-per-try-timeout-ms', '9000'
'x-envoy-upstream-rq-timeout-ms', '15000'
'x-envoy-max-retries', '123'

[2020-03-06 15:32:44.814][8740313][debug][http] [external/envoy/source/common/http/conn_manager_impl.cc:742] [C12345][S16439340388483568484] request headers complete (end_stream=false):
':scheme', 'https'
':method', 'GET'
':path', '/ping'
':authority', 'api.lyft.com'
'x-envoy-retry-on', 'connect-failure,5xx'
'x-envoy-upstream-rq-per-try-timeout-ms', '9000'
'x-envoy-upstream-rq-timeout-ms', '15000'
'x-envoy-max-retries', '123'
'x-envoy-mobile-cluster', 'base_h2'

[2020-03-06 15:32:44.814][8740313][debug][misc] [external/envoy/source/common/router/router.cc:1794] creating retry state with headers:
':scheme', 'https'
':method', 'GET'
':path', '/ping'
':authority', 'api.lyft.com'
'x-envoy-mobile-cluster', 'base_h2'
'x-forwarded-proto', 'http'
'x-request-id', 'b937b00b-2311-4ac5-8623-6ca9fccb3374'
'x-envoy-expected-rq-timeout-ms', '15000'

[2020-03-06 15:32:44.814][8740313][debug][router] [external/envoy/source/common/router/router.cc:616] [C12345][S16439340388483568484] router decoding headers:
':scheme', 'https'
':method', 'GET'
':path', '/ping'
':authority', 'api.lyft.com'
'x-envoy-mobile-cluster', 'base_h2'
'x-forwarded-proto', 'http'
'x-request-id', 'b937b00b-2311-4ac5-8623-6ca9fccb3374'
'x-envoy-expected-rq-timeout-ms', '15000'
```
Now the headers are preserved correctly through building the retry state:

```
[2020-03-06 16:06:21.797][8797407][debug][http] [library/common/http/dispatcher.cc:306] [S252] start stream
[2020-03-06 16:06:21.797][8797407][debug][http] [library/common/http/dispatcher.cc:337] [S252] request headers for stream (end_stream=false):
':authority', 'api.lyft.com'
':method', 'GET'
':scheme', 'https'
':path', '/ping'
'x-envoy-retry-on', 'connect-failure,5xx'
'x-envoy-upstream-rq-per-try-timeout-ms', '9000'
'x-envoy-max-retries', '123'
'x-envoy-upstream-rq-timeout-ms', '15000'
'x-envoy-mobile-cluster', 'base_h2'
'x-forwarded-proto', 'https'

[2020-03-06 16:06:21.797][8797407][debug][http] [external/envoy/source/common/http/conn_manager_impl.cc:742] [C12345][S7481013991204916798] request headers complete (end_stream=false):
':authority', 'api.lyft.com'
':method', 'GET'
':scheme', 'https'
':path', '/ping'
'x-envoy-retry-on', 'connect-failure,5xx'
'x-envoy-upstream-rq-per-try-timeout-ms', '9000'
'x-envoy-max-retries', '123'
'x-envoy-upstream-rq-timeout-ms', '15000'
'x-envoy-mobile-cluster', 'base_h2'
'x-forwarded-proto', 'https'

[2020-03-06 16:06:21.797][8797407][debug][misc] [external/envoy/source/common/router/router.cc:1794] creating retry state with headers:
':authority', 'api.lyft.com'
':method', 'GET'
':scheme', 'https'
':path', '/ping'
'x-envoy-retry-on', 'connect-failure,5xx'
'x-envoy-max-retries', '123'
'x-envoy-mobile-cluster', 'base_h2'
'x-forwarded-proto', 'https'
'x-envoy-expected-rq-timeout-ms', '9000'

[2020-03-06 16:06:21.797][8797407][debug][router] [external/envoy/source/common/router/router.cc:616] [C12345][S7481013991204916798] router decoding headers:
':authority', 'api.lyft.com'
':method', 'GET'
':scheme', 'https'
':path', '/ping'
'x-envoy-mobile-cluster', 'base_h2'
'x-forwarded-proto', 'https'
'x-envoy-expected-rq-timeout-ms', '9000'
```

Also note that the x-request-id header is now not present as the headers are not getting passed to `mutateRequestHeaders`. And now the 'x-envoy-expected-rq-timeout-ms' is '9000', due to respecting the `x-envoy-upstream-rq-per-try-timeout-ms` header. Otherwise, the final headers that are sent upstream are the same.

PR #747, opened 2020-03-13
binary size: Android binary size investigation -- Apply Keith's patch

buildbreaker:
Draft PR for investigation purposes.

We've noticed ~5MB per architecture increase for iOS with the inclusion of Envoy Mobile. On Android, we are observing ~10MB increase per architecture. @keith had the suggestion that since we are doing a static framework and using stripping, we are able to get a smaller artifact size for Envoy Mobile. Android uses a dynamic library here so we may not be able to get the stripping benefits.

The diff here is to attempt to capture an iOS application increase if we were to not use stripping to see if we are able to observe a similar 10MB increase. If we do, then we have a high confident signal that stripping is the reasoning for the discrepancies between iOS and Android artifacts.

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,7 +1,7 @@
 licenses(["notice"])  # Apache 2
 
 load("@io_bazel_rules_kotlin//kotlin/internal:toolchains.bzl", "define_kt_toolchain")
-
+exports_files(["Info.plist"])
 alias(
     name = "ios_framework",
     actual = "//library/swift/src:ios_framework",
--- a/Info.plist
+++ b/Info.plist
@@ -5,14 +5,14 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
 	<key>CFBundleIdentifier</key>
-	<string>com.lyft.$(PRODUCT_NAME)</string>
+	<string>io.envoyproxy.envoymobile.helloworld</string>
 	<key>CFBundleInfoDictionaryVersion</key>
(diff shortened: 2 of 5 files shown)


buildbreaker:
Worked with @keith on this a bit today

With the current state in this PR/branch, these are the steps which I've used to build an iOS app with a dynamic library:
1. `bazel run //examples/swift/hello_world:app --config=ios`
2. `cp bazel-bin/examples/swift/hello_world/app.ipa ~/tmp/dynamic.ipa`
3. `unzip dynamic.ipa dynamic`

To get a static library/app comparison:
1. `git checkout 917186e6cd080880106b9b2e0bf06d741bbb800e` (This is the static library hash prior to the changes here)
2. `bazel run //examples/swift/hello_world:app --config=ios`
3. `cp bazel-bin/examples/swift/hello_world/app.ipa ~/tmp/static.ipa`

Observations:
The dynamic library app is `152.1MB` large
The static library app is `39.3MB` large

By inspecting the unzipped internals, the `dynamic/app.app/Frameworks/dynamic.framework/dynamic` is size `135MB`

Using `strip` naively (`strip -no_code_signature_warning -xXNT dynamic/app.app/Frameworks/dynamic.framework/dynamic`), we could get this file down to `65MB`

PR #750, opened 2020-03-13
ci: update mac ci setup

junr03:
Description: move mac ci setup to use envoy's script. I don't see much of a point in keeping a separate script here. For instance, https://github.com/envoyproxy/envoy/pull/10357 updated to fix bazel after CI images broke. Instead of keeping up-to-date we can piggy back on efforts in upstream Envoy. Obviously, if diverging needs evolve we might need a second script in Envoy Mobile that calls the Envoy script, and does additional work.
Risk Level: low, CI will let us know if I screwed up.
Testing: CI


ci/mac_ci_setup.sh
```diff
@@ -1,30 +0,0 @@
...
-fi
-
-DEPS="automake cmake coreutils go libtool wget ninja gpg bazelbuild/tap/bazelisk"
-for DEP in ${DEPS}
-do
-    is_installed "${DEP}" || install "${DEP}"
-done
-
-# https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/macos/macos-10.14-Readme.md#xcode
-sudo xcode-select --switch /Applications/Xcode_11.3.app
```
rebello95: Actually this will probably break CI

rebello95:
Is the submodule update deliberate?

ci/mac_ci_setup.sh
junr03: the 10.15 machines have 11.3.1 and I updated `.bazelrc`

junr03:
Yep, because we need https://github.com/envoyproxy/envoy/pull/10357

PR #766, opened 2020-03-19
ci: add missing key

junr03:
Description: messed up gpg keys for heroku on github actions hosted machines. I opened https://github.com/heroku/cli/issues/1464 with heroku to start a discussion. For now I found in https://cli-assets.heroku.com/install-ubuntu.sh where they host their key.


junr03:
Merging this one, but leaving https://github.com/lyft/envoy-mobile/pull/767 here as an example of what to do if we want to get rid of heroku.

Issue #769, opened 2020-03-24
stats: enable CI testing for stats emissions


junr03:
The library configures a whitelist of stats to emit. This whitelist contains regexes that could be wrong. CI should have testing to ensure that regexes are valid and match the desired stats.

PR #771, opened 2020-03-24
stats: add segmentation of stats via virtual clusters

junr03:
Description: Virtual Clusters allow Envoy Mobile to emit stats based on matching criteria. This allows users to single out important traffic for their application. This PR adds an API to the client builder to add virtual cluster configuration and emission of virtual cluster stats.

Note: the client builder API exposed here is not ideal #770 tracks enhancing this.
Risk Level: low
Testing: unit tests, and local stats test. #769 tracks adding CI tests for stats
Docs Changes: added docs for the new API.

### DIFF
--- a/docs/root/api/starting_envoy.rst
+++ b/docs/root/api/starting_envoy.rst
@@ -174,15 +174,15 @@ This functionality is used for stat segmentation.
 .. attention::
 
     This API is non-ideal as it exposes lower-level internals of Envoy than desired by this project.
-    https://github.com/lyft/envoy-mobile/issues/770 tracks enhancing this API.
+    :issue:`#770 <770>` tracks enhancing this API.
 
 **Example**::
 
   // Kotlin
-  builder.addVirtualClusters("[{json_config}]")
+  builder.addVirtualClusters("[{\"name\":\"vcluster\",\"headers\":[{\"name\":\":path\",\"exact_match\":\"/v1/vcluster\"}]}]")
 
   // Swift
-  builder.addVirtualClusters("[{json_config}]")
+  builder.addVirtualClusters("[{\"name\":\"vcluster\",\"headers\":[{\"name\":\":path\",\"exact_match\":\"/v1/vcluster\"}]}]")
 
 ----------------------
 Advanced configuration
--- a/docs/root/api/starting_envoy.rst
+++ b/docs/root/api/starting_envoy.rst
@@ -163,6 +163,27 @@ This information is sent as metadata when flushing stats.
   // Swift
   builder.addAppId("com.mydomain.myapp)
 
+~~~~~~~~~~~~
+``addVirtualClusters``
+~~~~~~~~~~~~
+
+Specify the virtual clusters config for Envoy Mobile's configuration.
+The configuration is expected as a JSON list.
+This functionality is used for stat segmentation.
+
+.. attention::
+
+    This API is non-ideal as it exposes lower-level internals of Envoy than desired by this project.
+    https://github.com/lyft/envoy-mobile/issues/770 tracks enhancing this API.
+
+**Example**::
+
+  // Kotlin
+  builder.addVirtualClusters("[{json_config}]")
+
+  // Swift
+  builder.addVirtualClusters("[{json_config}]")
+
 ----------------------
 Advanced configuration
 ----------------------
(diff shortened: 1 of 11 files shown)


junr03:
@rebello95 @buildbreaker I am testing locally with the Lyft app and a local statsd server to see the new stat emissions. Waiting on https://github.com/envoyproxy/envoy/pull/10499 and https://github.com/envoyproxy/envoy/pull/10501 to land in order to bump the envoy ref. But this is ready for review

docs/root/api/starting_envoy.rst
```diff
@@ -163,6 +163,27 @@ This information is sent as metadata when flushing stats.
...
+
+.. attention::
+
+    This API is non-ideal as it exposes lower-level internals of Envoy than desired by this project.
+    https://github.com/lyft/envoy-mobile/issues/770 tracks enhancing this API.
+
+**Example**::
+
+  // Kotlin
+  builder.addVirtualClusters("[{json_config}]")
```
buildbreaker: Is there an example if a `json_config` or is it just a list of strings?

rebello95: Definitely agree that we should add an actual example here

docs/root/api/starting_envoy.rst
```diff
@@ -163,6 +163,27 @@ This information is sent as metadata when flushing stats.
...
+~~~~~~~~~~~~~~~~~~~~~~
+
+Specify the virtual clusters config for Envoy Mobile's configuration.
+The configuration is expected as a JSON list.
+This functionality is used for stat segmentation.
+
+.. attention::
+
+    This API is non-ideal as it exposes lower-level internals of Envoy than desired by this project.
+    https://github.com/lyft/envoy-mobile/issues/770 tracks enhancing this API.
```
rebello95: You should be able to use `:issue:` here

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -23,10 +24,11 @@
    * @param statsFlushSeconds            interval at which to flush Envoy stats.
    * @param appVersion                   the App Version of the App using this Envoy Client.
    * @param appId                        the App ID of the App using this Envoy Client.
+   * @param virtualClusters              the virtual cluster config.
```
rebello95: May be worth specifying more details here (`JSON list of virtual clusters`?)

library/swift/src/EnvoyClientBuilder.swift
```diff
@@ -20,6 +20,7 @@ public final class EnvoyClientBuilder: NSObject {
   private var statsFlushSeconds: UInt32 = 60
   private var appVersion: String = "unspecified"
   private var appId: String = "unspecified"
+  private var virtualClusters: String = "[]"
```
rebello95: I'm a little concerned that even though this is temporary, consumers can easily incorrectly add clusters. What do you think about doing something like this:

```
private var virtualClusters = [String]()

  func addVirtualCluster(_ virtualCluster: String) -> EnvoyClientBuilder {
    self.virtualClusters.append(virtualCluster)
    return self
  }
```

Then later you can pass it down as `"[\(self.virtualClusters.joined(separator: ",")]"`

WDYT?

junr03: I'm not sure your proposal gives much additional protection. In general I think the approach I would like to take here is to actually work with the structured config in the bridge/native layer, and expose ergonomic APIs up to the platform. So for instance the API at the platform layer would not be `addVirtualClusters` which is an implementation detail IMO, but rather `addPathSegmentation` (or something like that). Because what we are really trying to achieve here is higher dimension observability.

buildbreaker: @junr03  @rebello95 I want to make sure we're in alignment for our objective here. This is all temporary, right? I think we should be more willing to go with what we have now to see how the data looks in the next few days and make adjustments after our beta cut.

What do you think?

rebello95: Yes, 100% agree that this is temporary. Just seems like the suggestion would provide a _little more resiliency_ in the mean time, but this is fine

junr03: yeah, this approach is temporary.

PR #774, opened 2020-03-30
retries: enable retry budget for concurrent retries

junr03:
Description: previously the cluster configuration for circuit breakers was not specified. This meant that Envoy Mobile was using the default of 3 concurrent retries. This is not appropriate, specially because all* active traffic goes through the same cluster. This PR changes the configuration and explicitly sets the `retry_budget` configuration.
Risk Level: low (more retries will actually be allowed)
Testing: local

### DIFF
--- a/library/common/config_template.cc
+++ b/library/common/config_template.cc
@@ -74,7 +74,7 @@ static_resources:
       thresholds:
         - priority: DEFAULT
           # n.b: with mobile clients there are scenarios where all concurrent requests might be
-          # retries, e.g when the phone goes offline. Therefore, Envoy Mobile allows all concurrent
+          # retries (e.g., when the phone goes offline). Therefore, Envoy Mobile allows all concurrent
           # requests to be retries by setting the concurrent retry budget to 100%.
           # This configuration uses the default for max concurrent requests (1024) because there is
           # no reasonable scenario where a mobile client should have even close to that many concurrent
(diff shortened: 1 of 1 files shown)


junr03:
I noticed through internal dashboards that most retry traffic was actually overflowing. This makes sense given the pretty low default of 3 concurrent retries, and given that all* concurrent Envoy Mobile traffic goes through the same cluster. I have set the retry budget to 50%, but I actually wonder if we might have situations were most active requests in mobile are retries. I'd like to hear @Augustyniak and @tonya11en opinions. 

* I say all, but it might be that some remnants might go through a different cluster on a network interface preference change, or if the stream was created for http/1.1 vs. h2.

library/common/config_template.cc
```diff
@@ -70,6 +70,12 @@ const char* config_template = R"(
         keepalive_interval: 10
         keepalive_probes: 1
         keepalive_time: 5
+    circuit_breakers: &circuit_breakers_settings
+      thresholds:
+        - priority: DEFAULT
+          retry_budget:
+            budget_percent:
+              value: 50
```
rebello95: Curious, why not set this to 100? Also looks like this defaults to 20%, how does that work with the 3 mentioned in the description? https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cluster/circuit_breaker.proto

library/common/config_template.cc
```diff
@@ -70,6 +70,18 @@ const char* config_template = R"(
         keepalive_interval: 10
         keepalive_probes: 1
         keepalive_time: 5
+    circuit_breakers: &circuit_breakers_settings
+      thresholds:
+        - priority: DEFAULT
+          # n.b: with mobile clients there are scenarios where all concurrent requests might be
+          # retries, e.g when the phone goes offline. Therefore, Envoy Mobile allows all concurrent
```
rebello95:
```suggestion
          # retries (e.g., when the phone goes offline). Therefore, Envoy Mobile allows all concurrent
```

PR #776, opened 2020-03-31
ci: cat iOS log to check when CI flakes

junr03:
Description: iOS CI greps for indicators of liveliness. Sometimes they fail in CI when local runs are ok. This additional step should show us if CI failures are flakes or real.
Risk Level: low
Testing: CI

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -51,6 +51,8 @@ jobs:
       # Wait for the app to start and get some requests/responses.
       - name: 'Sleep'
         run: sleep 120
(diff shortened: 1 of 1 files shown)


.github/workflows/ios.yml
```diff
@@ -51,6 +51,8 @@ jobs:
       # Wait for the app to start and get some requests/responses.
       - name: 'Sleep'
         run: sleep 120
+      - name: 'Cat log'
```
rebello95: Maybe provide more context in the name as to why this is happening?

PR #779, opened 2020-04-01
ci: prevent infinite retries of sonatype upload

rebello95:
The logic in the original script was incorrect, and we could end up infinitely retrying until the job times out.

### DIFF
--- a/dist/sonatype_nexus_upload.py
+++ b/dist/sonatype_nexus_upload.py
@@ -72,20 +72,20 @@ def _install_locally(version, files):
         shutil.copyfile(file, os.path.join(path, basename))
 
 
-def _urlopen_retried(request, max_retries=20, attempt=1, delay_sec=1):
+def _urlopen_retried(request, max_retries=20, attempt=1):
     """
     Retries a request via recursion. Retries happen with a default delay of 1 second. We do not exponentially back off.
     :param request: the request to be made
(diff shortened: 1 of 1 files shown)


dist/sonatype_nexus_upload.py
```diff
@@ -72,20 +72,20 @@ def _install_locally(version, files):
...
     :param attempt: The current attempt number for the request
-    :param delay_sec: The delay before making a retried request
     :return: the response if successful, raises error otherwise
     """
     try:
         return urlopen(request)
     except HTTPError as e:
-        if max_retries > 0 and e.code >= 500:
+        if max_retries > attempt and e.code >= 500:
+            delay_sec = 1
```
buildbreaker, line 88: Why not keep this in the params?

rebello95: It's never set to anything other than 1 and the docstrings in place say this function always retries after `1s`

buildbreaker: That's as a default value though. Would you be against keeping this in the parameter? It seems a little easier to see where this value is configured

rebello95: updated

buildbreaker: Thanks, @rebello95

PR #783, opened 2020-04-03
Update bazel

keith:


keith:
We might need some other changes here, we'll see what CI says

zachgrayio:
We are going to look at this tomorrow and may have a PR if no one beats us to it.

keith:
I think this is a flaky test and that we're actually good here

PR #785, opened 2020-04-06
Update bazel

keith:


keith:
We'll see if this works out of the box, I expect it to

rebello95:
Actually we should wait for Envoy to bump: https://github.com/envoyproxy/envoy/blob/master/.bazelversion

keith:
Attempting to update that here: https://github.com/envoyproxy/envoy/pull/10670

PR #791, opened 2020-04-07
coverage: add coverage for c++ code

junr03:
Description: add script for coverage report generation and CI test for coverage check.
Risk Level: low 
Testing: locally generated report, and new CI check

Fixes #786

Signed-off-by: Jose Nino <jnino@lyft.com>

PR #792, opened 2020-04-07
test: add testing for missing coverage

junr03:
Description: this PR adds tests based on local coverage runs to raise coverage. This PR adds all non-silly coverage, but still hits misses on some accessors and because `NOT_IMPLEMENTED_GCOVR_EXCL_LINE` _is_ being included (known problem in Envoy upstream since we stopped using gcovr).
Risk Level: lowest! we are increasing our code coverage :)
Testing: local coverage runs to see missing spots, and then CI


test/integration/dispatcher_integration_test.cc
```diff
@@ -229,5 +229,9 @@ TEST_P(DispatcherIntegrationTest, RaceDoesNotCauseDoubleDeletion) {
   http_dispatcher_.synchronizer().signal("dispatch_encode_final_data");
 }
 
+// TODO(junr03): test with envoy local reply with local stream not close, which causes a reset fired
```
rebello95, line 232:
```suggestion
// TODO(junr03): test with envoy local reply with local stream not closed, which causes a reset fired
```

test/common/http/dispatcher_test.cc
```diff
@@ -519,13 +520,123 @@ TEST_F(DispatcherTest, BasicStream) {
   ASSERT_EQ(cc.on_complete_calls, 1);
 }
 
+TEST_F(DispatcherTest, BasicStreamWithTrailers) {
```
buildbreaker, line 523: This test feels quite large and it's kind of difficult to understand what suite of behaviors we're trying to exercise here. Do you know if we can isolate the specific behaviors and have smaller individual tests?

test/common/http/dispatcher_test.cc
```diff
@@ -739,12 +850,189 @@ TEST_F(DispatcherTest, MultipleStreams) {
...
+  EXPECT_CALL(event_dispatcher_, isThreadSafe()).Times(1).WillRepeatedly(Return(true));
+  EXPECT_CALL(event_dispatcher_, post(_)).WillOnce(SaveArg<0>(&stream_deletion_post_cb));
+  Buffer::InstancePtr response_data{new Buffer::OwnedImpl("error message")};
+  response_encoder_->encodeData(*response_data, true);
+  ASSERT_EQ(cc.on_data_calls, 0);
+  stream_deletion_post_cb();
+
+  // Ensure that the callbacks on the bridge_callbacks were called.
+  ASSERT_EQ(cc.on_complete_calls, 0);
+  ASSERT_EQ(cc.on_error_calls, 1);
```
rebello95, line 1027: I agree with @buildbreaker's comment above regarding encapsulating testable pieces so we don't have large test functions. Another idea might be to break up this file with those smaller individual tests (it's ~1500LOC)

test/common/http/dispatcher_test.cc
junr03: Some of these tests can be condensed a little bit by taking the repeated parts into utility functions in the test class. I was thinking about this as I was adding more tests. Will do in a subsequent PR https://github.com/lyft/envoy-mobile/issues/794

test/common/http/dispatcher_test.cc
junr03: see https://github.com/lyft/envoy-mobile/pull/792#discussion_r405148553

test/integration/dispatcher_integration_test.cc
junr03: if you are ok with it, I will fix this in #791 to avoid re-running CI

test/common/http/dispatcher_test.cc
buildbreaker: I think that's great to do also (I am slightly opinionated about test utils but we can talk about that separately)

The confusion I have is more on the behavior exercised by this test (since it seems to be asserting on all the different handlers). Does this test primarily exercise a general end to end behavior of a basic stream?

junr03: yeah, it is an http stream with trailers. There are individual tests for a headers only request, one with a body, and now this one with trailers.

junr03:
Yeah open to more suggestions @buildbreaker. I opened https://github.com/lyft/envoy-mobile/issues/794 yesterday to track. We can chat, or you can drop an example in the issue

PR #797, opened 2020-04-20
bazel: Remove custom framework rule

keith:
With this change we can remove our custom swift_static_framework rule
and replace it with the rules_apple ios_static_framework rule. There are
a few behavioral differences here that required other changes.

1. ios_static_framework uses `-enable-library-evolution`, which means we
   get a `swiftinterface` file instead of a `swiftmodule`. This opts us
   in to having to guarantee some level of binary compatibility (once we
   hit 1.0 at least) but has the benefit of not requiring consumers to
   be on the exact same Xcode version as what we build with
2. To Support #1 we had to remove the bridging header use, this was
   previously required because if we imported the Objective-C dependency
   directly it also had to be shipped to consumers. Now we use
   `@_implementationOnly` imports with this, and consume it with
   `private_deps` so it's not exposed to consumers at all. This is a
   new-ish private feature in the Swift compiler to support this type of
   use case.
3. To propagate the correct header with the framework so that it can be
   imported by Objective-C (which isn't automatically done by
   rules_swift / rules_apple
   https://github.com/bazelbuild/rules_swift/issues/291) we need to
   access it from the swift_library target, which I wrote a tiny rule to
   do.


bazel/swift_header_collector.bzl
```diff
@@ -0,0 +1,23 @@
...
+Propagate the generated Swift header from a swift_library target
+
+This exists to work around https://github.com/bazelbuild/rules_swift/issues/291
+"""
+
+def _swift_header_collector(ctx):
+    dep = ctx.attr.deps[0]
+    return [
+        DefaultInfo(
+            files = dep[CcInfo].compilation_context.headers,
```
keith, line 11: AFAIK since this is a direct dep there cannot be more than 1 in this list, but I could filter by ones ending with `-Swift.h` if that wasn't the case.

examples/objective-c/hello_world/ViewController.m
```diff
@@ -1,4 +1,4 @@
-#import <Envoy/Envoy-Swift.h>
+#import <Envoy/Envoy.h>
```
keith, line 1: This header got renamed, and there is also now a modulemap in the framework so in theory we could also import with `@import Envoy;`. In theory we could attempt to keep the old name, but this seems nicer anyways

library/objective-c/BUILD
```diff
@@ -16,6 +16,7 @@ objc_library(
     hdrs = [
         "EnvoyEngine.h",
     ],
+    module_name = "EnvoyEngine",
```
keith, line 19: Previously this name didn't matter at all because we never imported it directly, it happened implicitly with a bridging header

keith:
Depends on https://github.com/lyft/envoy-mobile/pull/796

bazel/swift_header_collector.bzl
rebello95: From what I saw in the past we only had one here IIRC, so this is prob fine

examples/objective-c/hello_world/ViewController.m
rebello95: Yep this is nicer 👌

library/swift/src/BUILD
```diff
@@ -27,9 +29,19 @@ swift_static_framework(
...
     module_name = "Envoy",
-    objc_includes = [
-        "//library/objective-c:EnvoyEngine.h",
-    ],
+    private_deps = ["//library/objective-c:envoy_engine_objc_lib"],
+)
+
+swift_header_collector(
+    name = "collector",
+    deps = ["lib"],
```
rebello95, line 37: Does this change make it possible for us to split out the above code into different libraries and still collect them into one framework here?

bazel/swift_header_collector.bzl
kastiglione, line 11: I'm surprised it's not a `CcInfo` or `Objc`

bazel/swift_header_collector.bzl
```diff
@@ -0,0 +1,23 @@
...
+        ),
+    ]
+
+swift_header_collector = rule(
+    attrs = dict(
+        deps = attr.label_list(
```
kastiglione, line 17: it looks like it could be a single `label`, does it need to be a `label_list`? Maybe if it was named `library` or `root` or something?

library/swift/src/BUILD
keith: not particularly right here, but overall it depends on how you want to split things up. if you have public API in multiple modules I don't think that will end up working in a single framework, but if you have other internal modules like the EnvoyEngine objc is acting as here now, you could include that in private_deps and have as many as you want. But I think no matter what you'll have to be able to `@_implementationOnly` import them

bazel/swift_header_collector.bzl
```diff
@@ -0,0 +1,23 @@
...
+            mandatory = True,
+            allow_empty = False,
```
kastiglione, line 19: add `providers = [CcInfo],`?

bazel/swift_header_collector.bzl
keith: Surprised what isn't? What I'm propagating? It's just a file list used by `hdrs`

bazel/swift_header_collector.bzl
keith: Yea I was following the general convention but I agree it's weird, I'll replace with `library`

bazel/swift_header_collector.bzl
kastiglione: ah ok

library/swift/src/BUILD
```diff
@@ -27,9 +29,19 @@ swift_static_framework(
...
+)
+
+swift_header_collector(
+    name = "collector",
+    library = "lib",
+)
+
+ios_static_framework(
+    name = "ios_framework",
+    hdrs = ["collector"],
```
kastiglione, line 42: ah I see how this is used. Maybe name it `lib_headers`?

keith:
Might actually depend on https://github.com/bazelbuild/rules_apple/pull/753

keith:
Potential fix for the duplicate symbol issue I'm seeing currently https://github.com/bazelbuild/rules_swift/pull/420

keith:
Potential fix for the duplicate symbol issue I'm seeing currently https://github.com/bazelbuild/rules_swift/pull/420

keith:
Filed an issue for the blocker here https://github.com/bazelbuild/bazel/issues/11223

PR #798, opened 2020-04-20
platform: close streams explicitly with data or trailers

rebello95:
Replaces the `close(trailers: Map<String, List<String>>?)` function with two new functions (one for closing with data, and one for closing with trailers).

This removes the ambiguity of how a stream will be closed by making the consumer specify explicitly whether they'd like to close a stream using trailers or data.

The change is being done as a precursor for integrating L7 Swift/Kotlin filters, which will have interfaces in which trailers are not nullable.

### DIFF
--- a/library/kotlin/src/io/envoyproxy/envoymobile/EnvoyStreamEmitter.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/EnvoyStreamEmitter.kt
@@ -29,22 +29,14 @@ class EnvoyStreamEmitter(
     return this
   }
 
-  /**
-   * For ending an associated stream and sending trailers.
-   *
-   * @param trailers to send with ending a stream. If null, stream will be closed with an empty data frame.
-   */
-  override fun close(trailers: Map<String, List<String>>?) {
-    trailers?.let {
-      stream.sendTrailers(it)
-      return
-    }
-    stream.sendData(ByteBuffer.allocate(0), true)
+  override fun close(trailers: Map<String, List<String>>) {
+    stream.sendTrailers(trailers)
   }
 
-  /**
-   * For cancelling and ending an associated stream.
-   */
+  override fun close(byteBuffer: ByteBuffer) {
+    stream.sendData(byteBuffer, true)
+  }
+
   override fun cancel() {
     stream.cancel()
   }
--- a/library/kotlin/src/io/envoyproxy/envoymobile/GRPCStreamEmitter.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/GRPCStreamEmitter.kt
@@ -37,6 +37,6 @@ class GRPCStreamEmitter(
    * Close this connection.
    */
   fun close() {
-    emitter.close(null)
+    emitter.close(ByteBuffer.allocate(0))
   }
 }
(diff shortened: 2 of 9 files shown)


library/kotlin/src/io/envoyproxy/envoymobile/EnvoyStreamEmitter.kt
```diff
@@ -29,22 +29,14 @@ class EnvoyStreamEmitter(
...
-   *
-   * @param trailers to send with ending a stream. If null, stream will be closed with an empty data frame.
-   */
-  override fun close(trailers: Map<String, List<String>>?) {
-    trailers?.let {
-      stream.sendTrailers(it)
-      return
-    }
-    stream.sendData(ByteBuffer.allocate(0), true)
+  override fun close(trailers: Map<String, List<String>>) {
```
buildbreaker, line 32: Not going to keep the comments?

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyStreamEmitter.kt
```diff
@@ -29,22 +29,14 @@ class EnvoyStreamEmitter(
...
+    stream.sendTrailers(trailers)
+  }
+
+  override fun close(byteBuffer: ByteBuffer) {
```
buildbreaker, line 36: What was the rationale for making this required (instead of having something like `fun close(byteBuffer: ByteBuffer = ByteBuffer.allocate(0)`

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyStreamEmitter.kt
rebello95: They're specified in the superclass, so I figured it was better to have them in one place rather than copy-pasted to every implementation. WDYT?

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyStreamEmitter.kt
rebello95: It'd be ambiguous if we did that for trailers as well, right? Consumers calling `.close()` would no longer be able to differentiate between data and trailers

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyStreamEmitter.kt
buildbreaker: Could we also delete the other comments? If it's a lot to change, we can both do it in another PR

rebello95: I'll do it in a follow up PR yep, waiting on this to merge to avoid conflicts

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyStreamEmitter.kt
buildbreaker: Ah so the change here is to introduce the concept of closing with data versus closing with trailers

rebello95: Yes

rebello95:
> Related to closing the stream. Right now we don't have a convenience API for a headers only request. For instance a send method that does not return a StreamEmitter because you send with as an only headers request. Thoughts on fixing that here vs in a different PR?

Yep I'll go ahead and update some of that in a follow up PR

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyStreamEmitter.kt
rebello95: https://github.com/lyft/envoy-mobile/pull/799/files#diff-9652aa5519ef4ba01992a74aaa5cd9b1L10

PR #799, opened 2020-04-20
platform: remove sendMetadata/onMetadata

rebello95:
These are never called in practice today. Since we don't want to support them in the initial version of filters, we're removing them from the public interfaces now.

### DIFF
--- a/library/kotlin/src/io/envoyproxy/envoymobile/EnvoyStreamEmitter.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/EnvoyStreamEmitter.kt
@@ -7,25 +7,8 @@ class EnvoyStreamEmitter(
   private val stream: EnvoyHTTPStream
 ) : StreamEmitter {
 
-  /**
-   * For sending data to an associated stream.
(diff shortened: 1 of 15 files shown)


library/kotlin/src/io/envoyproxy/envoymobile/EnvoyStreamEmitter.kt
```diff
@@ -7,28 +7,11 @@ class EnvoyStreamEmitter(
   private val stream: EnvoyHTTPStream
 ) : StreamEmitter {
 
-  /**
```
rebello95, line 10: https://github.com/lyft/envoy-mobile/pull/798#discussion_r411730064

rebello95:
> Given that we are potentially swallowing metadata, should we document somewhere that this is currently supported behavior intentionally?

Not opposed to that, though I do think it's pretty clear since it's not in the public interface. Where would you like to document it?

junr03:
> pretty clear since it's not in the public interface

good point. I think this is fine, then.

PR #800, opened 2020-04-20
kotlin: clean up HTTPClient.kt formatting

rebello95:

### DIFF
--- a/library/kotlin/src/io/envoyproxy/envoymobile/HTTPClient.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/HTTPClient.kt
@@ -12,7 +12,6 @@ interface HTTPClient {
    */
   fun send(request: Request, responseHandler: ResponseHandler): StreamEmitter
 
-
   /**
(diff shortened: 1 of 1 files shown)


library/kotlin/src/io/envoyproxy/envoymobile/HTTPClient.kt
```diff
@@ -22,8 +21,7 @@ interface HTTPClient {
    * @param responseHandler the callback for receiving stream events.
    * @return CancelableStream, a cancelable request.
    */
-  fun send(request: Request, body: ByteBuffer?, trailers: Map<String, List<String>>,
-                responseHandler: ResponseHandler): CancelableStream
+  fun send(request: Request, body: ByteBuffer?, trailers: Map<String, List<String>>, responseHandler: ResponseHandler): CancelableStream
```
rebello95, line 24: Matches `EnvoyClient.kt` - at some point we should probably add a column length linter for kotlin though

PR #802, opened 2020-04-21
platform: update interfaces for unary/streamed requests

rebello95:
Currently, our platform-level unary interfaces have the following limitations:

- It's not very clear when sending a request if it's going to be unary or streamed
- It's not possible to send headers-only requests
- Trailers are non-optional

This PR introduces the following changes to improve these interfaces:

- `send(...)` is used for unary requests, and `start(...)` is used for streaming
- `send(...)` provides the ability to send headers-only requests
- Callers can specify whether they'd like to close with body data or trailers by explicitly passing one as `nil` (see inline documentation)

Additionally, this PR introduces a new set of `EnvoyClient` tests which are 1:1 on iOS and Android.

This change is being made as a precursor to L7 Swift/Kotlin filters.

### DIFF
--- a/library/swift/test/EnvoyClientTests.swift
+++ b/library/swift/test/EnvoyClientTests.swift
@@ -17,26 +17,163 @@ private final class MockEnvoyEngine: EnvoyEngine {
 }
 
 final class EnvoyClientTests: XCTestCase {
+  private var envoy: EnvoyClient!
+
+  override func setUp() {
+    super.setUp()
+    self.envoy = try! EnvoyClientBuilder()
+      .addEngineType(MockEnvoyEngine.self)
+      .build()
+  }
+
   override func tearDown() {
     super.tearDown()
     MockEnvoyHTTPStream.reset()
   }
 
-  func testNonStreamingExtensionSendsRequestDetailsThroughStream() throws {
-    let requestExpectation = self.expectation(description: "Sends request")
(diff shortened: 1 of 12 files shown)


library/swift/test/EnvoyClientTests.swift
```diff
@@ -17,26 +17,163 @@ private final class MockEnvoyEngine: EnvoyEngine {
...
 
+    let expectedData = Data([0x0, 0x1, 0x2])
+    MockEnvoyHTTPStream.onData = { data, closeStream in
+      XCTAssertEqual(expectedData, data)
+      XCTAssertFalse(closeStream)
+      expectation.fulfill()
+    }
+
+    let request = RequestBuilder(
+      method: .post, scheme: "https", authority: "www.envoyproxy.io", path: "/test")
```
buildbreaker, line 71:
```suggestion
      method: .post, scheme: "https", authority: "envoyproxy.io", path: "/test")
```

buildbreaker: ?

rebello95: If anything I think this would probably be `api.envoyproxy.io`, but really could be any string for these tests. I'd prefer to leave it as is unless you feel strongly 🤷‍♂️

PR #805, opened 2020-04-22
ci: run kotlin linter on library*

rebello95:
We don't currently run the Kotlin linter on anything other than `hello_world`, but we should.

This PR fixes all existing violations and enables the linter.

Resolves https://github.com/lyft/envoy-mobile/issues/599.


buildbreaker:
I purposefully didn't do this because there was some issue before I believe...

Issue #97, opened 2019-06-16
Add docker image for kotlin linting to dockerhub


buildbreaker:
Context: https://github.com/lyft/envoy-mobile/pull/96#discussion_r294037011

Currently, we're using scripts to install [`detekt`](https://github.com/arturbosch/detekt). The reason we went with an interim solution is because we wanted to get something on CI integrated first.

We would like to move this to a docker image and upload it to docker hub. This will allow us to improve the linting and format checks in the future

jpsim:
https://github.com/envoyproxy/envoy-mobile/pull/810

PR #810, opened 2020-04-23
kotlin: replace Detekt runner script with Bazel rule

arturdryomov:
Signed-off-by: Artur Dryomov <artur.dryomov@gmail.com>

Description:
Risk Level: low
Testing: run new Detekt targets without changes and with invalid changes to check Detekt output
Docs Changes: none
Release Notes: none


arturdryomov:
Probably can close #97.

CC @artem-zinnatullin

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
```diff
@@ -60,3 +61,10 @@ envoy_mobile_kt_library(
         "//library/java/src/io/envoyproxy/envoymobile/engine:envoy_base_engine_lib",
     ],
 )
+
+detekt(
+    name = "envoy_lib_lint",
+    srcs = glob(["*.kt"]),
+    build_upon_default_config = True,
+    cfgs = ["//:kotlin_lint_config"],
+)
```
buildbreaker, line 70: Maybe we can just do a global `detekt` rule and just add it to the top level `BUILD` file. What do you think about that?

buildbreaker:
@arturdryomov Thanks for helping out here! We really appreciate it ❤️. Just a question to strike a conversation between all of us

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
arturdryomov: Do you mean a single target with all Kotlin sources? Certainly will work. However, this gives Bazel a little room to run things in parallel. Thinking about — I probably should change the GitHub Actions workflow to query Detekt targets and run them in a single Bazel invocation.

buildbreaker: Gotcha. That sounds good. There might be a good way for us to hide this in a custom bazel macro in the future too.

Do you want to do the actions change in this pr as well or are you looking to do it as a follow up?

arturdryomov: Just tried to fight the CI. Seems like `bazel query` [evaluates a bit much and fails to resolve `swiftc` on a Linux machine](https://github.com/lyft/envoy-mobile/runs/613134683#step:4:50). Is it possible to resolve this somehow? I’ve looked over different GitHub workflows but saw no difference.

examples/kotlin/hello_world/BUILD
```diff
@@ -29,3 +30,10 @@ kt_android_library(
         "@androidsdk//com.android.support:recyclerview-v7-25.0.0",
     ],
 )
+
+detekt(
+    name = "hello_envoy_kt_lint",
```
artem-zinnatullin, line 35:
```suggestion
    name = "hello_envoy_kt_detekt",
```

.github/workflows/format.yml
```diff
@@ -49,11 +49,15 @@ jobs:
...
-      image: openjdk:8-jdk
     steps:
       - uses: actions/checkout@v1
         with:
           submodules: true
-      - name: 'Run Kotlin lint'
-        run: ./ci/kotlin_lint.sh
+      - name: 'Install dependencies'
+        run: ./ci/linux_ci_setup.sh
+      - name: 'Run Kotlin Lint'
```
artem-zinnatullin, line 58:
```suggestion
      - name: 'Run Detekt (Kotlin Lint)'
```

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
```diff
@@ -60,3 +61,10 @@ envoy_mobile_kt_library(
         "//library/java/src/io/envoyproxy/envoymobile/engine:envoy_base_engine_lib",
     ],
 )
+
+detekt(
+    name = "envoy_lib_lint",
```
artem-zinnatullin, line 66:
```suggestion
    name = "envoy_lib_detekt",
```

.github/workflows/format.yml
artem-zinnatullin: The more info about the actual tool we have in the names of things the easier it is to work with for external contributors, ie when they see a Detekt failure but don't know that it's Detekt it might be confusing

arturdryomov: Let’s hear @buildbreaker. I was under an impression that the Lint tool was intentionally abstracted — even the config file is called `kotlinlint.yml`.

buildbreaker: I think it'll be nice to abstract it away. We could do something like `Run Kotlin Lint (Detekt)`

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
buildbreaker: I'll have to dig deeper to figure out (I end up needing to relearn bazel + ci every time). We can move forward without that niceness change for now.

buildbreaker:
@arturdryomov Thanks for your help here! I'm going to let @artem-zinnatullin finish his review and I'll take a final pass after

.github/workflows/format.yml
artem-zinnatullin: `Run Kotlin Lint (Detekt)` lgtm, but I'm not blocking on that, use your judgement :)

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
arturdryomov: @artem-zinnatullin, do you have ideas how to resolve this? Or a better approach is just not to use `bazel query` for this at all?

.github/workflows/format.yml
arturdryomov: Replaced with `Run Kotlin Lint (Detekt)` and `Run Swift Lint (SwiftLint)` for consistency.

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
arturdryomov: All right, removed `bazel query` for now, we can tune it in the future with the growing number of Kotlin targets.

arturdryomov:
Huh, interesting. The rule [caught an issue](https://github.com/lyft/envoy-mobile/pull/810/checks?check_run_id=627092017) that [the Bash script missed](https://github.com/lyft/envoy-mobile/pull/813/checks?check_run_id=624347988). Seems like this happens because the Bash script doesn’t declare `set -eou pipefail`.

arturdryomov:
@buildbreaker, everything seems to be addressed and the `kotlin_lint` workflow is green.

arturdryomov:
Awesome, thanks! What are next steps to get this merged? Do we need Artem approval?

buildbreaker:
If @artem-zinnatullin doesn't have any more comments, we can just merge

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
artem-zinnatullin: hm, this is not good indeed if `query` fails depending on OS, there is probably a way to use [config_settings](https://docs.bazel.build/versions/master/configurable-attributes.html) to only configure swift if we're running on macOS and then use `cquery` or something like that

Maybe @justhecuke, @keith or @kastiglione have some ideas

keith: What's the specific query failure? I expect it's because some of the repository rule setup in rules_swift checks what features your swift version supports, and those fail. I don't think we've had this issue internally since likely even if we weren't using it directly our docker images always had swift installed. It might be worth looking for a related bazel issue, seems like there must be one about this. Note that you also hit problems like this on macOS if you don't have the Android SDK installed

PR #811, opened 2020-04-23
config: fix max concurrent retries circuit breaker

junr03:
Description: while investigating retry scenarios I noticed that some retries were not getting performed due to overflow. This PR actually increases the concurrency limit. More information can be found in the PR comments.
Risk Level: med - this change actually allows higher retry concurrency.
Testing: local testing

### DIFF
--- a/library/common/config_template.cc
+++ b/library/common/config_template.cc
@@ -79,7 +79,14 @@ static_resources:
           # This configuration uses the default for max concurrent requests (1024) because there is
           # no reasonable scenario where a mobile client should have even close to that many concurrent
           # requests.
-          max_retries: 1024
+          #
+          # `max_retries` could be used but maintainers advised against it, as there are plans to
+          # deprecate that setting.
+          # https://github.com/lyft/envoy-mobile/pull/811#issuecomment-619169529.
+          retry_budget:
+            budget_percent:
+              value: 100
+            min_retry_concurrency: 1024
   - name: base_wlan
     connect_timeout: {{ connect_timeout_seconds }}s
     lb_policy: CLUSTER_PROVIDED
--- a/library/common/config_template.cc
+++ b/library/common/config_template.cc
@@ -74,14 +74,12 @@ static_resources:
       thresholds:
         - priority: DEFAULT
           # n.b: with mobile clients there are scenarios where all concurrent requests might be
-          # retries (e.g., when the phone goes offline). Therefore, Envoy Mobile allows all concurrent
-          # requests to be retries by setting the concurrent retry budget to 100%.
+          # retries (e.g., when the phone goes offline). Therefore, Envoy Mobile allows for the same
+          # amount of concurrency for retries as the default value for concurrency for requests.
           # This configuration uses the default for max concurrent requests (1024) because there is
           # no reasonable scenario where a mobile client should have even close to that many concurrent
           # requests.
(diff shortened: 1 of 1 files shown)


junr03:
I had previously merged #774, with a configuration I believed would essentially allow "unlimited" retries for all practical purposes. However, I just actually read the retry budget code and realized that the budget is based on upstream requests. In the scenario where a mobile device goes offline, we expect all currently live upstream requests to fail and potentially retry. However, as requests start becoming retries the upstream request count diminishes thus making the retry budget smaller and smaller. For example:

t0: device has 6 concurrent upstream requests.
t1: device goes offline
t2: envoy starts terminating upstream requests with connection termination and checking if it can retry.
t3: for the 1st request being reset the retry budget is 5. The retry is allowed. Count is now 1.
t4: for the 2nd request being reset the retry budget is 4. The retry is allowed. Count is now 2.
t5: for the 3rd request being reset the retry budget is 3. The retry is allowed. Count is now 3.
t6: for the 4th request being reset the retry budget is 3. The retry is not allowed.

This is not the behavior we want for mobile devices because during an offline event we want all requests to "transform" to retries.

Internally at Lyft we have seen retry overflows since we started our production experiment, this PR should fix that.

cc @tonya11en

library/common/config_template.cc
```diff
@@ -74,14 +74,12 @@ const char* config_template = R"(
...
-          # requests to be retries by setting the concurrent retry budget to 100%.
+          # retries (e.g., when the phone goes offline). Therefore, Envoy Mobile allows for the same
+          # amount of concurrency for retries as the default value for concurrency for requests.
           # This configuration uses the default for max concurrent requests (1024) because there is
           # no reasonable scenario where a mobile client should have even close to that many concurrent
           # requests.
-          retry_budget:
-            budget_percent:
-              value: 100
+          max_retries: 1024
```
tonya11en, line 82: Why not just add a `min_retry_concurrency`? See:
https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/cluster/circuit_breaker.proto#cluster-circuitbreakers-thresholds-retrybudget

tonya11en: If you just say `min_retry_concurrency: 1024` you'll then just have that be the lower bound.

junr03: @tonya11en im not sure I see the benefit of having a retry budget if the min is the same as the max requests circuit breaker. In that case the concurrency would be capped at 1024 anyway you dice it. 

What would be interesting for this scenario would be a retry budget based on active _downstream_ requests. Because what we want here is to have active retries for potentially 100% of active downstream requests.

rebello95:
Would you mind including your explanatory comment in the final commit message as well?

tonya11en:
Mainly because we are looking to make the retry budgets default and it
might save you from having to change this down the line. They’re both
equivalent, so I don’t feel strongly about which I’d used.
-- 
Tony Allen
Software Engineer
<https://www.lyft.com/>

library/common/config_template.cc
junr03: Thanks for the reply below @tonya11en :).

Any thoughts about retry budgets based on number of concurrent downstream requests?

junr03:
I'm not going to update the commit description with the long form comment, as I think it is too verbose.

PR #813, opened 2020-04-24
error handling: add attempt counts to the library's error types

junr03:
Description: this PR updates the error objects in the library to contain an attempt count. Additionally it wires up the attempt count as reported via envoy headers to the error object used by the Http::Dispatcher.
Risk Level: low - no breaking API changes.
Testing: unit tests, plus local end-to-end tests.
Docs Changes: updated the types docstrings

### DIFF
--- a/examples/java/hello_world/MainActivity.java
+++ b/examples/java/hello_world/MainActivity.java
@@ -104,8 +104,8 @@ public class MainActivity extends Activity {
               return Unit.INSTANCE;
             })
             .onError((error) -> {
-              String msg = new String("failed with error after " + error.getAttemptCount() +
-                                      " attempts: " + error.getMessage());
+              String msg = "failed with error after " + error.getAttemptCount() +
+                           " attempts: " + error.getMessage();
               Log.d("MainActivity", msg);
               recyclerView.post(() -> viewAdapter.add(new Failure(msg)));
               return Unit.INSTANCE;
--- a/examples/java/hello_world/MainActivity.java
+++ b/examples/java/hello_world/MainActivity.java
@@ -104,10 +104,10 @@ public class MainActivity extends Activity {
               return Unit.INSTANCE;
             })
             .onError((error) -> {
-              String msg = new String("failed with error after " + error.getAttemptCount() + " attempts: " + error.getMessage());
+              String msg = new String("failed with error after " + error.getAttemptCount() +
+                                      " attempts: " + error.getMessage());
               Log.d("MainActivity", msg);
-              recyclerView.post(
-                  () -> viewAdapter.add(new Failure(msg)));
+              recyclerView.post(() -> viewAdapter.add(new Failure(msg)));
               return Unit.INSTANCE;
             });
 
--- a/examples/java/hello_world/MainActivity.java
+++ b/examples/java/hello_world/MainActivity.java
@@ -104,8 +104,10 @@ public class MainActivity extends Activity {
               return Unit.INSTANCE;
             })
             .onError((error) -> {
+              String msg = new String("failed with error after " + error.getAttemptCount() + " attempts: " + error.getMessage());
+              Log.d("MainActivity", msg);
               recyclerView.post(
-                  () -> viewAdapter.add(new Failure("failed with error " + error.getMessage())));
+                  () -> viewAdapter.add(new Failure(msg)));
               return Unit.INSTANCE;
             });
 
--- a/examples/kotlin/hello_world/MainActivity.kt
+++ b/examples/kotlin/hello_world/MainActivity.kt
@@ -103,7 +103,7 @@ class MainActivity : Activity() {
           Unit
         }
         .onError { error ->
-          val msg = String("failed with error after " + error.attemptCount + " attempts: " + error.message)
+          val msg = "failed with error after ${error.attemptCount ?: -1} attempts: ${error.message}"
           Log.d("MainActivity", msg)
           recyclerView.post { viewAdapter.add(Failure(msg)) }
           Unit
--- a/examples/kotlin/hello_world/MainActivity.kt
+++ b/examples/kotlin/hello_world/MainActivity.kt
@@ -103,7 +103,9 @@ class MainActivity : Activity() {
           Unit
         }
         .onError { error ->
-          recyclerView.post { viewAdapter.add(Failure("failed with error " + error.message)) }
+          val msg = String("failed with error after " + error.attemptCount + " attempts: " + error.message)
+          Log.d("MainActivity", msg)
+          recyclerView.post { viewAdapter.add(Failure(msg)) }
           Unit
         }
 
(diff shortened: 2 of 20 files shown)


examples/kotlin/hello_world/MainActivity.kt
```diff
@@ -103,7 +103,9 @@ class MainActivity : Activity() {
           Unit
         }
         .onError { error ->
-          recyclerView.post { viewAdapter.add(Failure("failed with error " + error.message)) }
+          val msg = String("failed with error after " + error.attemptCount + " attempts: " + error.message)
+          Log.d("MainActivity", msg)
```
buildbreaker, line 107:
```suggestion
          Log.d("MainActivity", "failed with error after ${error.attemptCount} attempts: ${error.message}")
```

examples/java/hello_world/MainActivity.java
```diff
@@ -104,8 +104,10 @@ private void makeRequest() {
               return Unit.INSTANCE;
             })
             .onError((error) -> {
-              recyclerView.post(
-                  () -> viewAdapter.add(new Failure("failed with error " + error.getMessage())));
+              String msg = new String("failed with error after " + error.getAttemptCount() +
+                                      " attempts: " + error.getMessage());
+              Log.d("MainActivity", msg);
```
buildbreaker, line 109:
```suggestion
              Log.d("MainActivity", "failed with error after " + error.getAttemptCount() +
                                      " attempts: " + error.getMessage());
```

library/kotlin/src/io/envoyproxy/envoymobile/GRPCResponseHandler.kt
```diff
@@ -123,7 +123,7 @@ class GRPCResponseHandler(
         val compressionFlag = byteArray[0]
         // TODO: Support gRPC compression https://github.com/lyft/envoy-mobile/issues/501.
         if (compressionFlag.compareTo(0) != 0) {
-          errorClosure(EnvoyError(0, "Unable to read compressed gRPC response message"))
+          errorClosure(EnvoyError(0, "Unable to read compressed gRPC response message", 0))
```
rebello95, line 126: Seems like this should probably be 1 technically 🤔 it's kind of arbitrary here I suppose since the request _could_ have been retried

library/java/src/io/envoyproxy/envoymobile/engine/JvmCallbackContext.java
```diff
@@ -127,14 +127,15 @@ public void run() {
   /**
    * Dispatches error received from the JNI layer up to the platform.
    *
-   * @param message,   the error message.
-   * @param errorCode, the envoy_error_code_t.
+   * @param errorCode,    the envoy_error_code_t.
```
rebello95, line 130: Thi
```suggestion
   * @param errorCode,    the error code.
```

examples/java/hello_world/MainActivity.java
junr03: I want to give the Failure object the same string, though

buildbreaker: uhh how about
```suggestion
              String msg = "failed with error after " + error.getAttemptCount() +
                                      " attempts: " + error.getMessage();
              Log.d("MainActivity", msg);
```
Since the `new String()` is redundant here

examples/kotlin/hello_world/MainActivity.kt
buildbreaker:
```suggestion
          val msg = "failed with error after ${error.attemptCount} attempts: ${error.message}"
          Log.d("MainActivity", msg)
```

junr03:
> LGTM overall, though in some cases `attemptCount` doesn't seem to be relevant on the error. Wonder if it'd make sense for this to be optional?

Good suggestion! The only rub is that we bridge through the c struct, which does not have "optionals". There are two ways this could be handled:
1. With a uint32_t pointer. I kinda want to avoid this to avoid managing memory for it.
2. With a int32_t and use `-1` as not present. While not semantically explicit (and kind of ugly by modern standards) this pattern is used in c code widely from my experience.

rebello95:
Ah, got it. In that case, I think leaving it as `-1` for now is probably fine (and we can assume that negative values are "invalid") 👍

examples/swift/hello_world/ViewController.swift
```diff
@@ -65,7 +65,13 @@ final class ViewController: UITableViewController {
         NSLog("Response data (\(requestID)): \(data.count) bytes")
       }
       .onError { [weak self] error in
-        let message = "failed within Envoy library: \(error.message)"
+        var message: String
```
rebello95, line 68:
```suggestion
        let message: String
```
Compiler enforces that this is set exactly once here

library/common/jni_interface.cc
```diff
@@ -209,7 +209,7 @@ static void jvm_on_error(envoy_error error, void* context) {
   jobject j_context = static_cast<jobject>(context);
 
   jclass jcls_JvmObserverContext = env->GetObjectClass(j_context);
-  jmethodID jmid_onError = env->GetMethodID(jcls_JvmObserverContext, "onError", "([BI)V");
+  jmethodID jmid_onError = env->GetMethodID(jcls_JvmObserverContext, "onError", "(I[BI)V");
```
rebello95, line 212: What does this change?

library/swift/src/ResponseHandler.swift
```diff
@@ -66,8 +66,9 @@ public final class ResponseHandler: NSObject {
     @escaping (_ error: EnvoyError) -> Void)
     -> ResponseHandler
   {
-    self.underlyingCallbacks.onError = { errorCode, message in
-      closure(EnvoyError(errorCode: errorCode, message: message, cause: nil))
+    self.underlyingCallbacks.onError = { errorCode, message, attemptCount in
+      closure(EnvoyError(errorCode: errorCode, message: message,
+                         attemptCount: attemptCount < 0 ? nil : UInt32(attemptCount), cause: nil))
```
rebello95, line 71: You need to use `UInt32(exactly: attemptCount)` here to prevent crashes in case `attemptCount` is out of bounds. If you want, you could then additionally remove the ternary, as `UInt32(exactly: attemptCount)` will also return `nil` if the value is negative, but you may also want to keep it for clarity 🤷

examples/swift/hello_world/ViewController.swift
junr03: I wonder why my local compilation worked?

rebello95: It'll compile, just doesn't need to be mutable

library/common/jni_interface.cc
junr03: The signature of the method. It went from being a void method that took a byte array and an integer to a void method that takes an integer, a byte array, and an integer.

PR #815, opened 2020-04-27
bazel: Update swift_static_framework to use swiftinterfaces

keith:
This is an in between to https://github.com/lyft/envoy-mobile/pull/797
which starts building the archive with swiftinterface files, but does
it with our existing swift_static_framework rule because of issues
linked on that PR.

### DIFF
--- a/bazel/swift_static_framework.bzl
+++ b/bazel/swift_static_framework.bzl
@@ -65,7 +65,10 @@ def _swift_static_framework_impl(ctx):
             fail("Expected exactly 1 '-Swift.h' header, got {}".format(", ".join(header_names)))
 
         swiftdoc = swift_info.direct_swiftdocs[0]
-        swiftmodule = swift_info.direct_swiftmodules[0]
+        swiftinterfaces = swift_info.transitive_swiftinterfaces.to_list()
+        if len(swiftinterfaces) != 1:
+            fail("Expected a single swiftinterface file, got: {}".format(swiftinterfaces))
+        swiftinterface = swiftinterfaces[0]
 
         libraries = archive[CcInfo].linking_context.libraries_to_link
         archives = []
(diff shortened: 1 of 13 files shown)


bazel/swift_static_framework.bzl
```diff
@@ -163,37 +166,28 @@ def swift_static_framework(
...
         copts = [],
-        swiftc_inputs = [],
         visibility = []):
     """Create a static library, and static framework target for a swift module
 
     Args:
         name: The name of the module, the framework's name will be this name
             appending Framework so you can depend on this from other modules
         srcs: Custom source paths for the swift files
-        objc_includes: Header files for any objective-c dependencies (required for linking)
```
rebello95, line 177: Can you add a docstring for `private_deps`?

keith:
This still includes the swiftmodule file in the framework because of this issue, https://github.com/bazelbuild/rules_apple/pull/753 this way people can continue to use it as before for now, but folks who apply this patch can use the swiftinterface instead.

keith:
This currently contains both the swiftinterface and the swiftmodule file, once https://github.com/bazelbuild/rules_apple/pull/753 is fixed we can remove the swiftmodule file

library/swift/test/EnvoyClientBuilderTests.swift
```diff
@@ -1,4 +1,5 @@
 @testable import Envoy
+import EnvoyEngine
```
rebello95, line 2: Out of curiosity, any reason not to use `@_implementationOnly` here for consistency?

keith: It doesn't really matter for tests since no one consumes their output swiftmodule, and the annoyance is that if you do it once in a single target you have to do it for all imports, so then everyone has to know about it.

PR #816, opened 2020-04-27
build: Enable opt for Swift compiles

keith:

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -19,6 +19,7 @@ build --xcode_version=11.3.1
 build --incompatible_bzl_disallow_load_after_statement=false
 build --javabase=@bazel_tools//tools/jdk:jdk
 build --define disable_known_issue_asserts=true
+build --features=swift.opt_uses_wmo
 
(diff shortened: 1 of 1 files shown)


keith:
> Does this change anything about the binary size?

I haven't tested, but no matter what it can't be significant given the percent of the total binary that's Swift vs C++

.bazelrc
```diff
@@ -41,6 +42,9 @@ build:release-ios --apple_bitcode=embedded
 build:release-ios --config=ios
 build:release-ios --config=release-common
 build:release-ios --copt=-fembed-bitcode
+# This is a workaround because we can't set --compilation_mode=opt because of size issues
+# this forces -O to be passed at least to Swift compiles
```
kastiglione, line 46: > we can't set --compilation_mode=opt because of size issues

is this for c++ code that it causes a problem? Is it because `opt` does flags other than `-Os`?

keith: not sure, but the resulting binary is too large for zip

PR #830, opened 2020-04-29
bazel: build releases with opt

keith:
Previously we could not do this because the archive ended up being too
large with debug info. I've added an option to disable this here https://github.com/envoyproxy/envoy/pull/10987


keith:
Need to update the submodule after the linked PR for this to work for real

keith:
Verified debugging still works as expected

PR #835, opened 2020-05-04
main: use custom main common to disable listening for termination signals

### LINKED ISSUE #831: crash: trying to use garbage/null pointer
Stacktrace:

```
# (sometimes a garbage pointer)
EXC_BAD_ACCESS: Attempted to dereference null pointer. 

0  Lyft                    Envoy::Http::Dispatcher::post(std::__1::function<void ()>) (Lyft)
1  Lyft                    Envoy::Http::Dispatcher::startStream(long, envoy_http_callbacks) (Lyft)
2  Lyft                    start_stream (Lyft)
3  Lyft                    -[EnvoyHTTPStreamImpl initWithHandle:callbacks:] (Lyft)
4  Lyft                    -[EnvoyEngineImpl startStreamWithCallbacks:] (Lyft)
```

The only pointer accessed in `Http::Dispatcher::post` is the `event_dispatcher_` pointer here: 

https://github.com/lyft/envoy-mobile/blob/5c01e711e70c80c101087e61722862cbfbe961c9/library/common/http/dispatcher.cc#L282

The pointer is correctly initialized to nullptr, set by an initialization callback, and guarded by synchronization mechanisms. So I believe the crash is happening because the `Event::Dispatcher` the pointer points to is getting deallocated.

That `Event::Dispatcher` is uniquely owned by the `Server::InstanceImpl`. Which in turn (through some layers of inner classes ends up uniquely owned by `main_common_` in the Envoy Mobile codebase. The only code path where that unique pointer is reset is when the event loop exits here: 

https://github.com/lyft/envoy-mobile/blob/5c01e711e70c80c101087e61722862cbfbe961c9/library/common/engine.cc#L84

The only place where Envoy Mobile explicitly exits the event loop is in the Engine destructor: 

https://github.com/lyft/envoy-mobile/blob/5c01e711e70c80c101087e61722862cbfbe961c9/library/common/engine.cc#L104

Currently envoy mobile has only a singleton engine that is statically instantiated. Therefore, Envoy Mobile should not be exiting the event loop unless the app is quitting, or has otherwise crashed causing static deallocation. However, from the stacktrace above it looks like:

a) The event loop has exited, causing` main_common_` to be reset and the `event_dispatcher_` pointer to become garbage
b) The exit is happening without the static Engine singleton being destroyed, as evidenced by the fact that the stack trace above is able to lock the Engine's weak pointer.

I have created a [test case](https://github.com/lyft/envoy-mobile/blob/set-null/test/common/engine_test.cc#L10) where I force event loop exit and subsequently use the engine, resulting in a repro where the `event_dispatcher_` is garbage.

## Open Question

My remaining unknown is: why is the event loop exiting without the Envoy Mobile code calling `exit()` on it. My only theory is that Envoy runs the event loop with `Event::Dispatcher::RunType::Block` 

https://github.com/envoyproxy/envoy/blob/c2a63735145ef1ce056ca2a50927e3e6682177eb/source/server/server.cc#L592

Which is described as `// Runs the event-loop until there are no pending events.`All of the crashes show that the app had been sent to the background (and afaict the crash is only happening in iOS). So I am wondering somehow the event loop is exiting prematurely I would assume that pending timers for stats flushes, timeouts, should keep the event loop running given that there are pending events. I am not a libevent or iOS expert, but I am wondering if on mobile libevent behaves differently, and exits under some conditions which does not happen on the server.

One idea I have is to allow users of the `Server::InstanceImpl` to dictate how the event loop is run, and run it with `Event::Dispatcher::RunType::Block`

@Reflejo @mattklein123 I investigated this yesterday afternoon, I would appreciate if you could give it a read, and let me know what you think. @Reflejo specially around the theory of iOS causing the libevent loop to exit prematurely.

junr03:
Description:  Disabling signal handling in the Server::Options makes it so that the server's event dispatcher [_does not_ listen for termination signals](https://github.com/envoyproxy/envoy/blob/048f4231310fbbead0cbe03d43ffb4307fff0517/source/server/server.cc#L519) such as SIGTERM, SIGINT, etc. Previous crashes in iOS were experienced due to out-of-band event loop exit as described in #831. Ignoring termination signals makes it more likely that the event loop will only exit due to [Engine destruction](https://github.com/lyft/envoy-mobile/blob/a72a51e64543882ea05fba3c76178b5784d39cdc/library/common/engine.cc#L105).

This PR introduces `Envoy::MobileMainCommon`, as this is the canonical way to customize how main runs, and options setup per https://github.com/envoyproxy/envoy/pull/3424. The new `Envoy::MobileMainCommon` also does away with other logic in `Envoy::MainCommon` that does not apply to Envoy Mobile.

Risk Level: med - low-level change in termination handling
Testing: unit test to assert option is correctly set. End to end test with iOS and Android devices to ensure clean exit when the app using envoy mobile exits (and thus destructs the engine). Moreover, there is no event loop exit any longer when the simulator app receives a SIGTERM, i.e., the event dispatcher is no longer listening to SIGTERM and exiting due to them.

Potentially fixes #831. Will need to verify with wider client release.

### DIFF
--- a/library/common/BUILD
+++ b/library/common/BUILD
@@ -16,11 +16,23 @@ envoy_cc_library(
     hdrs = ["main_interface.h"],
     repository = "@envoy",
     deps = [
+        ":envoy_mobile_main_common_lib",
         "//library/common/buffer:utility_lib",
         "//library/common/http:dispatcher_lib",
         "//library/common/http:header_utility_lib",
         "//library/common/types:c_types_lib",
         "@envoy//include/envoy/server:lifecycle_notifier_interface",
+    ],
+)
+
+envoy_cc_library(
+    name = "envoy_mobile_main_common_lib",
+    srcs = ["envoy_mobile_main_common.cc"],
+    hdrs = ["envoy_mobile_main_common.h"],
+    repository = "@envoy",
+    deps = [
+        "@envoy//source/common/runtime:runtime_lib",
+        "@envoy//source/exe:envoy_common_lib",
         "@envoy//source/exe:envoy_main_common_lib",
     ],
 )
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -44,7 +44,7 @@ envoy_status_t Engine::run(std::string config, std::string log_level) {
       char* envoy_argv[] = {strdup("envoy"), strdup("--config-yaml"),   strdup(config.c_str()),
                             strdup("-l"),    strdup(log_level.c_str()), nullptr};
 
-      main_common_ = std::make_unique<Envoy::MainCommon>(5, envoy_argv);
+      main_common_ = std::make_unique<MobileMainCommon>(5, envoy_argv);
       event_dispatcher_ = &main_common_->server()->dispatcher();
       cv_.notifyOne();
     } catch (const Envoy::NoServingException& e) {
--- a/library/common/engine.h
+++ b/library/common/engine.h
@@ -4,8 +4,6 @@
 
 #include "common/upstream/logical_dns_cluster.h"
 
-#include "exe/main_common.h"
-
 #include "extensions/clusters/dynamic_forward_proxy/cluster.h"
 #include "extensions/filters/http/dynamic_forward_proxy/config.h"
 #include "extensions/filters/http/router/config.h"
@@ -14,6 +12,7 @@
 #include "extensions/transport_sockets/tls/config.h"
 
 #include "absl/base/call_once.h"
+#include "library/common/envoy_mobile_main_common.h"
 #include "library/common/http/dispatcher.h"
 #include "library/common/types/c_types.h"
 
@@ -57,7 +56,7 @@ private:
   Thread::CondVar cv_;
   std::thread main_thread_;
   std::unique_ptr<Http::Dispatcher> http_dispatcher_;
-  std::unique_ptr<MainCommon> main_common_ GUARDED_BY(mutex_);
+  std::unique_ptr<MobileMainCommon> main_common_ GUARDED_BY(mutex_);
   Server::Instance* server_{};
   Server::ServerLifecycleNotifier::HandlePtr postinit_callback_handler_;
   Event::Dispatcher* event_dispatcher_;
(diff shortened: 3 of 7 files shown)


junr03:
@mattklein123, thanks for the ideas and pointers in #831. Based on my research online, there is a good chance that making it so that the event loop does not listen for termination signals will prevent exits outside of the Engine destructor. @Reflejo if you have some time, I'd appreciate your eyes here. Your expertise would definitely help verify my research.

Reflejo:
This looks good for the attempt you are trying to make. I'm still not convinced this is due to a signal by reading the other issue but if it is:

Signal handling is hard. I think it's entirely possible for that null pointer bug to be a red herring (ie. the app crashed for a different reason, sent a SIGTERM / SIGINT so things get shut down twice and the initial crash is lost). Bugsnag is trying to catch signals as well and the ordering is unpredictable. 

I don't think there are many other cases where iOS will send anything other than SIGKILL but maybe things changed on recent OSes. 

Can we send `ENVOY_LOG()`s to bugsnag as breadcrumbs? I think that'd be helpful. Or at least have a way to consume logs on crashes.

junr03:
> I think it's entirely possible for that null pointer bug to be a red herring (ie. the app crashed for a different reason, sent a SIGTERM / SIGINT.

I see, potentially. The scenario you described reinforces my wanting to make this change anyways, because the envoy event dispatcher should not be listening to termination signals, and thus leading to envoy's thread termination. The way we organized shutdown the only way to maintain predictable behavior is to have the event loop exit from the `exit()` call issued in the Engine's destructor. For that reason, this change should still be done to prevent "out of band" event loop exits.

> Can we send `ENVOY_LOG()`s to bugsnag as breadcrumbs? I think that'd be helpful. Or at least have a way to consume logs on crashes.

Yeah, I have thought about that before ([doc](https://docs.google.com/document/d/1TaiIiDsyZ-LmJvEFX7TC7xMvYuieYuhT4QctwHG3jGI/edit?usp=sharing)), but we have not gotten around to prioritize that work.

PR #795, opened 2020-04-09
swift: add L7 platform filter interfaces

rebello95:
This PR introduces a set of Swift interfaces which will be used to integrate L7 platform filters into Envoy Mobile.

**Goals:**
- Maintain close parity with [Envoy's C++ filter interfaces](https://github.com/envoyproxy/envoy/blob/master/include/envoy/http/filter.h)
- Add interfaces and types that will be used for the Swift filter chain, which will enable us to add Kotlin interfaces and begin implementation work

**Non-goals:**
- Implement the full filter chain functionality (will be done in separate PRs)
- Maintain exact parity with Envoy's filters (some are being simplified based on the initial set of functionality we expect to support)

Part of https://github.com/lyft/envoy-mobile/issues/545. 

Signed-off-by: Michael Rebello <me@michaelrebello.com>

Issue #545, opened 2019-10-24
ios/android: add support for native L7 filters


rebello95:
In addition to supporting the core Envoy C++ L7 filter chain (https://github.com/lyft/envoy-mobile/issues/543), we plan to add new interfaces in Swift/Kotlin that allow consumers to write and register L7 filters from the platform APIs.

This will decrease the barrier to entry for writing Envoy Mobile filters, and allow consumers to utilize platform-specific APIs with their filter implementations.

PR #840, opened 2020-05-12
filters: add kotlin filter interfaces

junr03:
Description: kotlin version of #795
Risk Level: low
Testing: builds. No current tests, as we are just checking in interfaces

progress for #545 


library/kotlin/src/io/envoyproxy/envoymobile/BUILD
```diff
@@ -40,17 +40,34 @@ envoy_mobile_kt_library(
...
+        "RequestHeaders.kt",
+        "RequestHeadersBuilder.kt",
         "RequestMapper.kt",
         "RequestMethod.kt",
+        "RequestTrailersBuilder.kt",
+        "RequestTrailers.kt",
+        "ResponseFilter.kt",
+        "ResponseFilterCallbacks.kt",
         "ResponseHandler.kt",
+        "ResponseHeaders.kt",
```
junr03, line 69: We are missing builders for ResponseHeaders/Trailers for both kotlin and swift.

library/kotlin/src/io/envoyproxy/envoymobile/Headers.kt
```diff
@@ -0,0 +1,30 @@
+package io.envoyproxy.envoymobile
+
+/*
+ * Base class that is used to represent header/trailer data structures.
+ * To instantiate new instances, see `{Request|Response}HeadersBuilder`.
+ */
+open class Headers {
+  @Suppress("MemberNameEqualsClassName")
```
buildbreaker, line 8: Can we add a TODO for fixing the linter from being this strict?

library/kotlin/src/io/envoyproxy/envoymobile/Filter.kt
```diff
@@ -0,0 +1,7 @@
+package io.envoyproxy.envoymobile
+
+/*
+ * Interface representing a filter. See `RequestFilter` and `ResponseFilter` for more details.
+ */
+@Suppress("EmptyClassBlock")
```
buildbreaker, line 6: Let's add a TODO for avoid linting on this

library/kotlin/src/io/envoyproxy/envoymobile/RequestMethod.kt
```diff
@@ -1,5 +1,5 @@
 package io.envoyproxy.envoymobile
-
+import java.lang.IllegalArgumentException
```
buildbreaker, line 2:
```suggestion

import java.lang.IllegalArgumentException
```

library/kotlin/src/io/envoyproxy/envoymobile/RequestTrailers.kt
```diff
@@ -0,0 +1,14 @@
+package io.envoyproxy.envoymobile
+
+/*
+ * Trailers representing an outbound request.
+ */
+@Suppress("EmptyClassBlock")
```
buildbreaker, line 6: Is this actually needed? We can add a TODO for removing this lint check

library/swift/src/RetryPolicyMapper.swift
```diff
@@ -36,8 +36,10 @@ extension RetryPolicy {
     self.init(
       maxRetryCount: maxRetryCount,
       retryOn: headers.value(forName: "x-envoy-retry-on")?.compactMap(RetryRule.init) ?? [],
+      // TODO: does the UInt.init blow up? or does it return nil?
```
buildbreaker, line 39: @rebello95 Could you give insight onto how we can find this out (or help find this out for us?)

library/swift/src/ResponseHeaders.swift
```diff
@@ -4,6 +4,7 @@ import Foundation
 @objcMembers
 public final class ResponseHeaders: Headers {
   /// HTTP status code received with the response.
+  /// TODO: in Kotlin this blows up if the value of status cannot be translated to an int.
```
buildbreaker, line 7: @rebello95 Same here, do we error out when a `:status` is not a number?

library/swift/src/ResponseFilterCallbacks.swift
```diff
@@ -20,5 +21,5 @@ public protocol ResponseFilterCallbacks {
   /// `endStream = true` in order to guarantee that the client will not send its own trailers.
   ///
   /// - parameter trailers: The trailers to add and pass to subsequent filters.
-  func addResponseTrailers() -> ResponseHeaders
+  func addResponseTrailers(_ trailers: ResponseTrailers)
```
buildbreaker, line 24: Nice catch!

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicyMapper.kt
```diff
@@ -30,6 +30,7 @@ internal fun RetryPolicy.outboundHeaders(): Map<String, List<String>> {
  *
  * @return The string representation of the retry rule
  */
+ //TODO: change to enum pattern in UpstreamHttpProtocol
```
buildbreaker, line 33: Is this still relevant since you've added the `enumValue()` method in the class?

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
```diff
@@ -25,6 +25,22 @@ data class RetryPolicy(
...
   }
+
+  /**
+   * Initialize the retry policy from a set of headers.
+   *
+   * @param headers: The headers with which to initialize the retry policy.
+   */
+  constructor(headers: Headers): this(
+      headers.value("x-envoy-max-retries")?.first()!!.toInt(),
+      // TODO: should we have ? after the map?
```
buildbreaker, line 36: Contingent on https://github.com/lyft/envoy-mobile/pull/840/files#r424029491

library/swift/src/RetryPolicyMapper.swift
rebello95: Just checked, and mapping `String` to `UInt*` here just returns `nil` if it's not a valid unsigned int

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
rebello95: Ah yea we should add those. Originally I thought they wouldn't be necessary, but since we'll allow filters to add/mutate response headers/trailers, we'll need to add those. Can do in a separate PR (maybe comment on the issue?)

library/kotlin/src/io/envoyproxy/envoymobile/RequestHeaders.kt
```diff
@@ -0,0 +1,55 @@
...
+   * Internal constructor used by builders.
+   *
+   * @param headers: Headers to set.
+   */
+  internal constructor(headers: Map<String, List<String>>) : super(headers)
+
+  /**
+   * Method for the request.
+   */
+  val method = RequestMethod.enumValue(value(":method")?.first()!!)
```
rebello95, line 18: In Swift these are computed and are only computed when they're accessed by the consumer. Is it possible to do that here @buildbreaker?

library/kotlin/src/io/envoyproxy/envoymobile/RequestHeaders.kt
```diff
@@ -0,0 +1,55 @@
...
+   * Retry policy to use for this request.
+   */
+  // TODO: retry policy mapper
+  val retryPolicy = RetryPolicy(this)
+
+  /**
+   * The protocol version to use for upstream requests.
+   */
+  val upstreamHttpProtocol =
+    UpstreamHttpProtocol.enumValue(value("x-envoy-mobile-upstream-protocol")?.first()!!)
```
rebello95, line 45: This is optional/nullable, so we probably should not force unwrap here

library/kotlin/src/io/envoyproxy/envoymobile/RequestHeaders.kt
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * The URL path for the request (i.e., "/foo").
+   */
+  val path = value(":path")?.first()!!
+
+  /**
+   * Retry policy to use for this request.
+   */
+  // TODO: retry policy mapper
```
rebello95, line 38: What's necessary to do here?

library/kotlin/src/io/envoyproxy/envoymobile/ResponseHeaders.kt
```diff
@@ -0,0 +1,18 @@
...
+   * Internal constructor used by builders.
+   *
+   * @param headers: Headers to set.
+   */
+  internal constructor(headers: Map<String, List<String>>) : super(headers)
+
+  /**
+   * HTTP status code received with the response.
+   */
+  val httpStatus = value(":status")?.first()?.toInt()
```
rebello95, line 17: I think we should probably avoid throwing an exception here if possible to match Swift

library/swift/src/RequestFilter.swift
```diff
@@ -36,5 +36,5 @@ public protocol RequestFilter: Filter {
   /// - parameter trailers: The outbound trailers.
   ///
   /// - returns: The trailer status containing body with which to continue or buffer.
-  func onRequestTrailers(_ trailers: RequestHeaders) -> FilterTrailerStatus<RequestHeaders>
+  func onRequestTrailers(_ trailers: RequestTrailers) -> FilterTrailerStatus<RequestHeaders>
```
rebello95, line 39:
```suggestion
  func onRequestTrailers(_ trailers: RequestTrailers) -> FilterTrailerStatus<RequestTrailers>
```

Missed one 😉

library/swift/src/RequestFilterCallbacks.swift
```diff
@@ -1,5 +1,6 @@
 import Foundation
 
+/// Interface for RequestFilters to access FilterManager state.
```
rebello95, line 3:
```suggestion
/// Interface for RequestFilters to access FilterManager state.
```

Might want to remove this (same in Kotlin) for now since we haven't yet confirmed if `FilterManager` will be a class or where that logic will live

library/swift/src/RequestTrailersBuilder.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
 
-/// Builder used for constructing instances of `RequestTrailers` types.
+/// Builder used for constructing instances of `RequestTrailers` type.
```
rebello95, line 3:
```suggestion
/// Builder used for constructing instances of the `RequestTrailers` type.
```

rebello95: This was "types" before since we have multiple subclasses, but no strong opinion (let's just keep them all consistent)

library/swift/src/ResponseHeaders.swift
rebello95: Commented above, but swift returns `nil` in the case of an invalid `:status` which I think makes sense, especially if at some point we have internal headers generated by Envoy/Envoy Mobile in the future that don't contain this header

junr03: update kotlin with an optional. Although the server-side bias in me wishes we blew up.

library/kotlin/src/io/envoyproxy/envoymobile/Headers.kt
junr03: My personal bias is to make it intentional to suppress the linter rather than making the linter less-strict. But I am draconian like that.

This way we don't fall into anti-patterns (according to the linter's convention) without giving it thought.

library/kotlin/src/io/envoyproxy/envoymobile/Filter.kt
junr03: same as other comment

library/kotlin/src/io/envoyproxy/envoymobile/RequestTrailers.kt
junr03: same as other comment

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicyMapper.kt
junr03: yep I mean I wanted to update the enum classes to be `enum class RetryPolicy(internal val stringValue: String) {...}` to avoid having the `stringValue()` method. I have updated RequestMethod and RetryPolicy.

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
junr03: I guess my question is what optional is the `?` after map talking about? The first one is because the key `x-envoy-retry-on` might not be in the headers map so the list is optional. But the return type of map is `List<T>`, so I am not sure what ? means there.

library/kotlin/src/io/envoyproxy/envoymobile/RequestHeaders.kt
junr03: stale

library/swift/src/RetryPolicyMapper.swift
```diff
@@ -38,6 +38,7 @@ extension RetryPolicy {
       retryOn: headers.value(forName: "x-envoy-retry-on")?.compactMap(RetryRule.init) ?? [],
       retryStatusCodes: headers.value(forName: "x-envoy-retriable-status-codes")?
         .compactMap(UInt.init) ?? [],
+      // TODO: is the flatmap to apply the constructor even if there is only one value?
```
junr03, line 41: https://www.hackingwithswift.com/articles/205/whats-the-difference-between-map-flatmap-and-compactmap

library/kotlin/src/io/envoyproxy/envoymobile/RequestHeaders.kt
buildbreaker: yeah we have to use the `lazy` keyword: https://stackoverflow.com/questions/36623177/kotlin-property-initialization-using-by-lazy-vs-lateinit

junr03:
@buildbreaker @rebello95 this is updated!

library/kotlin/src/io/envoyproxy/envoymobile/RequestHeaders.kt
junr03: thanks, done!

library/kotlin/src/io/envoyproxy/envoymobile/RequestHeaders.kt
```diff
@@ -0,0 +1,54 @@
...
+   * Internal constructor used by builders.
+   *
+   * @param headers: Headers to set.
+   */
+  internal constructor(headers: Map<String, List<String>>) : super(headers)
+
+  /**
+   * Method for the request.
+   */
+  val method = RequestMethod.enumValue(value(":method")?.first()!!)
```
buildbreaker:
```suggestion
  val method by lazy { RequestMethod.enumValue(value(":method")?.first()!!) }
```

library/kotlin/src/io/envoyproxy/envoymobile/RequestFilter.kt
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * Called at most once when the request is closed from the client with trailers.
+   *
+   * Filters may mutate or delay the trailers.
+   *
+   * @param trailers: The outbound trailers.
+   *
+   * @return: The trailer status containing body with which to continue or buffer.
+   */
+  fun onRequestTrailers(trailers: RequestTrailers): FilterTrailerStatus<RequestHeaders>
```
rebello95, line 51:
```suggestion
  fun onRequestTrailers(trailers: RequestTrailers): FilterTrailerStatus<RequestTrailers>
```

library/kotlin/src/io/envoyproxy/envoymobile/RequestHeaders.kt
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * The URL path for the request (i.e., "/foo").
+   */
+  val path: String by lazy { value(":path")?.first()!! }
+
+  /**
+   * Retry policy to use for this request.
+   */
+  val retryPolicy: RetryPolicy by lazy { RetryPolicy(this) }
```
rebello95, line 38:
```suggestion
  val retryPolicy: RetryPolicy? by lazy { RetryPolicy(this) }
```
Shouldn't this be optional?

library/kotlin/src/io/envoyproxy/envoymobile/RequestMapper.kt
```diff
@@ -8,7 +8,7 @@ internal fun Request.outboundHeaders(): Map<String, List<String>> {
     !entry.key.startsWith(":") && !entry.key.startsWith("x-envoy-mobile")
   })
   result.putAll(retryPolicyHeaders)
-  result[":method"] = listOf(method.stringValue())
+  result[":method"] = listOf(method.stringValue)
```
rebello95, line 11: 🤔 how'd this work before? Do they work the same way @buildbreaker?

library/kotlin/src/io/envoyproxy/envoymobile/RequestMethod.kt
```diff
@@ -1,15 +1,33 @@
...
+  GET("GET"),
+  HEAD("HEAD"),
+  OPTIONS("OPTIONS"),
+  PATCH("PATCH"),
+  POST("POST"),
+  PUT("PUT"),
+  TRACE("TRACE");
+
+  companion object {
+    fun enumValue(stringRepresentation: String): RequestMethod {
```
rebello95, line 19: I think this can be internal

library/kotlin/src/io/envoyproxy/envoymobile/ResponseFilter.kt
```diff
@@ -0,0 +1,70 @@
...
+   *
+   * Filters may mutate or delay the response headers.
+   *
+   * @param headers:   The current response headers.
+   * @param endStream: Whether this is a headers-only response.
+   *
+   * @return: The header status containing headers with which to continue or buffer.
+   */
+  fun onResponseHeaders(headers: ResponseHeaders, endStream: Boolean):
+    FilterHeaderStatus<ResponseTrailers>
```
rebello95, line 28:
```suggestion
    FilterHeaderStatus<ResponseHeaders>
```

library/kotlin/src/io/envoyproxy/envoymobile/ResponseFilter.kt
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * Called at most once when the response is closed from the server with trailers.
+   *
+   * Filters may mutate or delay the trailers.
+   *
+   * @param trailers: The inbound trailers.
+   *
+   * @return: The trailer status containing body with which to continue or buffer.
+   */
+  fun onResponseTrailers(trailers: RequestTrailers): FilterTrailerStatus<ResponseTrailers>
```
rebello95, line 51:
```suggestion
  fun onResponseTrailers(trailers: ResponseTrailers): FilterTrailerStatus<ResponseTrailers>
```

library/kotlin/src/io/envoyproxy/envoymobile/ResponseHeaders.kt
```diff
@@ -0,0 +1,18 @@
...
+   * Internal constructor used by builders.
+   *
+   * @param headers: Headers to set.
+   */
+  internal constructor(headers: Map<String, List<String>>) : super(headers)
+
+  /**
+   * HTTP status code received with the response.
+   */
+  val httpStatus = value(":status")?.first()?.toIntOrNull()
```
rebello95, line 17: This one should be lazy as well

library/kotlin/src/io/envoyproxy/envoymobile/ResponseTrailers.kt
```diff
@@ -0,0 +1,14 @@
...
+
+/*
+ * Trailers representing an inbound response.
+ */
+@Suppress("EmptyClassBlock")
+class ResponseTrailers: Headers {
+  /**
+   * Internal constructor used by builders.
+   *
+   * @param trailers: Headers to set.
```
rebello95, line 11:
```suggestion
   * @param trailers: Trailers to set.
```

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
```diff
@@ -25,18 +25,49 @@ data class RetryPolicy(
...
     }
   }
+
+  /**
+   * Initialize the retry policy from a set of headers.
+   *
+   * @param headers: The headers with which to initialize the retry policy.
+   */
+  constructor(headers: Headers): this(
+      headers.value("x-envoy-max-retries")?.first()!!.toInt(),
```
rebello95, line 35: I think we should avoid crashing here like we do in Swift (and early return)

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
rebello95: Yea I'm actually curious about this as well

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
```diff
@@ -25,18 +25,49 @@ data class RetryPolicy(
...
+  STATUS_5XX("5xx"),
+  GATEWAY_ERROR("gateway-error"),
+  CONNECT_FAILURE("connect-failure"),
+  REFUSED_STREAM("refused-stream"),
+  RETRIABLE_4XX("retriable-4xx"),
+  RETRIABLE_HEADERS("retriable-headers"),
+  RESET("reset");
+
+  companion object {
+    fun enumValue(stringRepresentation: String): RetryRule? {
```
rebello95, line 60: Can be internal

library/kotlin/src/io/envoyproxy/envoymobile/UpstreamHttpProtocol.kt
```diff
@@ -5,5 +5,15 @@ package io.envoyproxy.envoymobile
  */
 enum class UpstreamHttpProtocol(internal val stringValue: String) {
   HTTP1("http1"),
-  HTTP2("http2"),
+  HTTP2("http2");
+
+  companion object {
+    fun enumValue(stringRepresentation: String): UpstreamHttpProtocol {
```
rebello95, line 11: Internal

library/kotlin/src/io/envoyproxy/envoymobile/UpstreamHttpProtocol.kt
```diff
@@ -5,5 +5,15 @@ package io.envoyproxy.envoymobile
...
+      return when (stringRepresentation) {
+        "http1" -> UpstreamHttpProtocol.HTTP1
+        "http2" -> UpstreamHttpProtocol.HTTP2
+        else -> throw IllegalArgumentException("Unable to find value for $stringRepresentation")
```
rebello95, line 15: I noticed that some of these throw and others are nullable. Should we keep them consistent?

library/kotlin/src/io/envoyproxy/envoymobile/RequestMapper.kt
buildbreaker: Was a method before: https://github.com/lyft/envoy-mobile/pull/840/files#diff-5075eb53390be0c0b1fc060ee9f6f359L23-L33

library/kotlin/src/io/envoyproxy/envoymobile/RequestHeaders.kt
buildbreaker: Constructors cannot return null

junr03: The constructor is not a "failable" constructor, no such concept in Kotlin afaict.

library/kotlin/src/io/envoyproxy/envoymobile/RequestMapper.kt
junr03: I homogenized all of the enums to use the same pattern

library/kotlin/src/io/envoyproxy/envoymobile/RequestMethod.kt
junr03: I see where you are coming from. Everything outside of the module should deal with only enum types, not strings. I like that.

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
buildbreaker: Are you talking about the elvis symbol `?:` here rather than the `?` symbol?

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
junr03: Same as above, there is no concept of a failable constructor. So our option is to have a default value, or crash.

library/kotlin/src/io/envoyproxy/envoymobile/UpstreamHttpProtocol.kt
junr03: We had it that way because of the automatic addition of "retriable-status-codes", although I'd rather throw in all of them. I have updated to something I think is better.

library/swift/src/ResponseHeaders.swift
buildbreaker: I feel we should blow up but let's keep it consistent for now and we can have a separate discussion about these details after. What do you think?

library/kotlin/src/io/envoyproxy/envoymobile/RequestHeaders.kt
buildbreaker: @junr03 I think we should keep consistency here for this PR. It's a little weird/strange of a thing to do but we'll need to do something like this:
```suggestion
  val retryPolicy: RetryPolicy? by lazy {
    try {
      RetryPolicy(this)
    } catch (t: Throwable) {
      null
    }
  }
```

buildbreaker:
From some of the discussions in this PR, it seems there's some key paradigm differences between Kotlin and Swift with respect to the fallible constructors/initializers. We should probably discuss what we want to do for these cases. Additionally, Kotlin has a concept of throwing/crashing with unexpected behaviors and Swift has a similar but different concept which I am assuming is leading to a lot optionality/nullability patterns in Swift (which is mirrored by the exception patterns in Kotlin).

Second is a more philosophical question regarding what we should do with unexpected behaviors such as what should be the expected behavior for someone setting a retry policy without specifying a max retry.

If we can agree that these are worthwhile topics, I feel we can assume consistency with what is checked in as our goal for this PR and we can have a more thorough conversation about the above two points after. What do you think, @junr03 @rebello95 ?

library/kotlin/src/io/envoyproxy/envoymobile/RequestHeaders.kt
rebello95: Yea +1 to this. Another option would be to make a static function that returns an optional `ReturnPolicy?` instead of having a throwing constructor. Not sure which is more "acceptable" 🤷

library/kotlin/src/io/envoyproxy/envoymobile/RequestMapper.kt
rebello95: Nice

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
rebello95: https://github.com/lyft/envoy-mobile/pull/840#discussion_r425429624

rebello95:
> From some of the discussions in this PR, it seems there's some key paradigm differences between Kotlin and Swift with respect to the fallible constructors/initializers. We should probably discuss what we want to do for these cases. Additionally, Kotlin has a concept of throwing/crashing with unexpected behaviors and Swift has a similar but different concept which I am assuming is leading to a lot optionality/nullability patterns in Swift (which is mirrored by the exception patterns in Kotlin).

I think in the spirit of reliability and consistency, we should try to avoid crashing at runtime where possible. That said, since Kotlin doesn't support failable initializers, WDYT about instead having `static` functions that return optional types instead? Might be a reasonable workaround which is supported by both languages.

> Second is a more philosophical question regarding what we should do with unexpected behaviors such as: what should be the expected behavior for someone setting a retry policy without specifying a max retry?

We require it in the initializer, and if we do that I think it makes sense to require it in the mapping as well.

> If we can agree that these are worthwhile topics, I feel we can assume consistency with what is checked in as our goal for this PR and we can have a more thorough conversation about the above two points after. What do you think, @junr03 @rebello95 ?

Curious to hear @junr03's thoughts, but I'd prefer to make a decision on the above as part of merging this PR. If there are more nontrivial changes that are required as a result, totally fine with making those on master after

junr03:
> WDYT about instead having `static` functions that return optional types instead

If we want to have failable construction, this is how I would do it. Static "factories" that return optional types. That way the factory does checks, before constructing the object.

> I think it makes sense to require it in the mapping as well

Agreed.

So I think the question boils down to what we want to do on error conditions. Because even if we return optional types out of factories, what should the library do if we get an empty value?

junr03:
@buildbreaker @rebello95 updated!

library/kotlin/src/io/envoyproxy/envoymobile/RequestMethod.kt
```diff
@@ -1,15 +1,33 @@
...
+      return when (stringRepresentation) {
+        "DELETE" -> RequestMethod.DELETE
+        "GET" -> RequestMethod.GET
+        "HEAD" -> RequestMethod.HEAD
+        "OPTIONS" -> RequestMethod.OPTIONS
+        "PATCH" -> RequestMethod.PATCH
+        "POST" -> RequestMethod.POST
+        "PUT" -> RequestMethod.PUT
+        "TRACE" -> RequestMethod.TRACE
+        else -> throw IllegalArgumentException("invalid value $stringRepresentation")
```
buildbreaker, line 29: Non-blocking: Do we return nil in swift?

library/swift/src/RequestMethod.swift
```diff
@@ -56,7 +56,7 @@ public enum RequestMethod: Int, CaseIterable {
     case "TRACE":
       self = .trace
     default:
-      return nil
+      fatalError("invalid value '\(stringValue)'")
```
buildbreaker, line 59: We do now!

library/kotlin/src/io/envoyproxy/envoymobile/RequestMethod.kt
rebello95: We do now 🙃

PR #858, opened 2020-05-20
swift: refactor public functionality as precursor to L7 filters

rebello95:
## Context

As part of the work to support platform (Swift/Kotlin) L7 filters, we are updating the public interfaces/functionality of the Envoy Mobile library to combine the concepts of a `StreamEmitter` and `ResponseHandler` into a "stream".

## Goals

- Unify the existing response handler and stream emitter types into a single "stream" type that will make it easier to communicate with an underlying filter manager for both outbound requests and inbound responses
- Provide guard rails that prevent the consumer from making mistakes, such as setting callback handlers after starting a stream and thus missing updates
- Provide a clear interface to the user for creating and starting streams
- Focus on the basics (bi-directional streaming support), which will allow for other wrappers to be written in the future (i.e., GRPC, unary, Rx, etc.)
- Update interfaces to utilize the new wrapper types introduced on master (`ResponseHeaders`, `RequestHeaders`, etc.)

## Types

- `StreamClient`: The publicly exposed interface of `EnvoyEngine`. Provides a function for starting a new stream using Envoy Mobile, and is built using `StreamClientBuilder`
- `StreamPrototype`: Type that allows users to set up a stream (providing callbacks, etc.) prior to starting it
- `Stream`: Produced by calling `start()` on a `StreamPrototype`, and allows the consumer to send data over the stream and eventually close it
- `GRPCClient` / `GRPCStreamPrototype` / `GRPCStream`: Types wrapping the above 3 types, converting the existing gRPC types to utilize the new functionality

To start a stream, consumers now do something like this:

```swift
let streamClient = try StreamClientBuilder().build()
...
```
```swift
let headers = RequestHeadersBuilder(method: .get, authority: "envoyproxy.io", path: "/test")
  .addUpstreamHttpProtocol(.http2)
  .build()

streamClient
  .newStream()
  .setOnResponseHeaders { ... }
  .setOnResponseData { ... }
  .start()
  .sendHeaders(headers, endStream: false)
  .sendData(data)
  .close()
```

## Tests

- Unit tests have been added / updated accordingly.
- Sample apps have also been updated (CI validates via "hello world").

## Notes

- A follow-up PR will migrate Android to mirror these
- Documentation will be updated separately


examples/objective-c/hello_world/ViewController.m
```diff
@@ -37,7 +37,7 @@ - (void)startEnvoy {
   NSLog(@"Starting Envoy...");
   NSError *error;
   EnvoyClientBuilder *builder = [[EnvoyClientBuilder alloc] init];
```
goaway, line 39: Should we change the name of the builder, since what it builds has changed?

examples/objective-c/hello_world/ViewController.m
```diff
@@ -65,38 +65,38 @@ - (void)performRequest {
...
+                                                                            path:_REQUEST_PATH];
+  [builder addUpstreamHttpProtocol:UpstreamHttpProtocolHttp1];
+  RequestHeaders *headers = [builder build];
 
   __weak ViewController *weakSelf = self;
-  [handler onHeaders:^(NSDictionary<NSString *, NSArray<NSString *> *> *headers,
-                       NSInteger statusCode, BOOL endStream) {
-    NSLog(@"Response status (%i): %ld\n%@", requestID, statusCode, headers);
-    NSString *body = [NSString stringWithFormat:@"Status: %ld", statusCode];
+  InactiveStream *inactiveStream = [self.client newStream];
```
goaway, line 80: `StreamPrototype`?

library/kotlin/src/io/envoyproxy/envoymobile/filters/ResponseFilter.kt
```diff
@@ -66,5 +66,5 @@ interface ResponseFilter: Filter {
    * This should be considered a terminal state, and invalidates any previous attempts to
    * `stopIteration{...}`.
    */
-  fun onCanceled()
+  fun onCancel()
```
goaway, line 69: I feel like `onCancellation` is most grammatically in line with the other callbacks, but ¯\\\_(ツ)\_/¯

examples/swift/hello_world/ViewController.swift
```diff
@@ -38,33 +38,36 @@ final class ViewController: UITableViewController {
...
-      .onHeaders { [weak self] headers, statusCode, _ in
+    let requestID = self.requestCount
+    let headers = RequestHeadersBuilder(method: .get, scheme: kRequestScheme,
+                                        authority: kRequestAuthority, path: kRequestPath)
+      .addUpstreamHttpProtocol(.http2)
+      .build()
+
+    client
+      .newStream()
+      .setOnResponseHeaders { [weak self] headers, _ in
```
goaway, line 60: I sort of prefer the old naming scheme (`onHeaders`) since `setOnResponseHeaders` to a degree suggests the implementation detail that this is a functor captured as a member, whereas I feel `onHeaders` (or `onResponseHeaders`) to me simply suggests "on headers, _do x_". But this is subjective.

examples/objective-c/hello_world/ViewController.m
rebello95: `StreamClientBuilder` would be fine yeah

examples/objective-c/hello_world/ViewController.m
rebello95: No strong opinion, @junr03 do you have thoughts?

library/kotlin/src/io/envoyproxy/envoymobile/filters/ResponseFilter.kt
rebello95: I'd like to keep it as-is for this PR since this is how it's named further down as well. We can open PRs after to rename everywhere if necessary

examples/swift/hello_world/ViewController.swift
rebello95: Discussed this a bit with @junr03 and @buildbreaker when we were first looking at this, and had elected to go this route to avoid confusion between _setting the callbacks_ and _calling the callbacks_. That said, I think `set` might be clearer but don't feel that strongly

examples/objective-c/hello_world/ViewController.m
junr03: active/inactive parallels envoy a bit more, but if you prefer Stream and StreamPrototype I don't care too much.

examples/swift/hello_world/ViewController.swift
goaway: I get the idea behind distinguishing them, but the user of the public API will never actually be the one calling the callbacks. If anything, I'd be slightly more on board with naming the closure-type something more verbose internally, since it's never really exposed. e.g., `onHeadersCallback`.

examples/objective-c/hello_world/ViewController.m
goaway: I mean Envoy has an `ActiveStream` type. It doesn't have an `InactiveStream` type, nor does it use streams classified as inactive them as a mechanism to produce active streams.

goaway: I'm open to other possibilities, too, but `InactiveStream` really suggests something other than a suite of callbacks along with a factory mechanism to me. `StreamBase`?

rebello95: `StreamBase` sounds more like a superclass which doesn't really fit here IMO. I'm not opposed to `StreamPrototype`/`Stream` if you and Jose are in agreement

junr03: lets do StreamPrototype/Stream

examples/objective-c/hello_world/ViewController.m
rebello95: Updated

examples/objective-c/hello_world/ViewController.m
rebello95: Done

examples/swift/hello_world/ViewController.swift
rebello95: That's true. However, if we're going for consistency, our other builders expose `addXYZ` and `setXYZ` for our other builders (i.e., `addRetryPolicy`, `set(name:value:)`, so this maintains consistency there. If you feel strongly we can change, but I'd lean towards keeping this for the time being for the sake of consistency

rebello95:
Thanks for the review @goaway. Pushed changes to address your comments

library/swift/src/RequestHeaders.swift
```diff
@@ -2,7 +2,7 @@ import Foundation
 
 /// Headers representing an outbound request.
 @objcMembers
-public final class RequestHeaders: Headers {
+public class RequestHeaders: Headers {
```
buildbreaker, line 5: No need for `final`? 

(`RequestTrailers` is `final` btw)

rebello95: `GRPCRequestHeaders` subclasses this, so we can't use `final` here

rebello95: For context, I had considered replicating `RequestHeaders` entirely for gRPC, but that would be problematic for filters if we're passing a `<T: RequestHeaders>` into filters, since gRPC requests wouldn't conform to that. Granted, we may very well be doing that at a much lower layer in which case this wouldn't be a problem, but this approach seemed like the least overhead for now

buildbreaker: Ah

examples/objective-c/hello_world/ViewController.m
```diff
@@ -65,38 +65,38 @@ - (void)performRequest {
...
+                                                                            path:_REQUEST_PATH];
+  [builder addUpstreamHttpProtocol:UpstreamHttpProtocolHttp1];
+  RequestHeaders *headers = [builder build];
 
   __weak ViewController *weakSelf = self;
-  [handler onHeaders:^(NSDictionary<NSString *, NSArray<NSString *> *> *headers,
-                       NSInteger statusCode, BOOL endStream) {
-    NSLog(@"Response status (%i): %ld\n%@", requestID, statusCode, headers);
-    NSString *body = [NSString stringWithFormat:@"Status: %ld", statusCode];
+  StreamPrototype *prototype = [self.client newStream];
```
goaway, line 80: I'd be fine with `newStream` or `newStreamPrototype`, but if you're going with the latter, looks like you missed this.

goaway:
I'd actually be okay with `ActiveStream` still being the name of the active streams (this actually is how Envoy's types are set up). But it's not that big a deal to me either way.

goaway:
For consistency you should probably rename the grpc stream files one way or the other though.

PR #860, opened 2020-05-21
extensions: move factory registration out of the engine

junr03:
Description: move factory registration out of the engine file. This will eventually allow third party builds to replace the extensions installed by the build system. The directory name and location is not necessarily final; this is just the first step in separating this out to its own target.
Risk Level: low
Testing: existing CI


library/common/extensions/registry.h
```diff
@@ -0,0 +1,17 @@
...
+#include "extensions/stat_sinks/metrics_service/config.h"
+#include "extensions/transport_sockets/tls/config.h"
+
+#include "common/upstream/logical_dns_cluster.h"
+
+namespace Envoy {
+class ExtensionRegistry {
+public:
+  static void registerFactories();
+};
```
rebello95, line 16: Can we add some documentation here?

junr03: It's in the cc file, let me move it here.

PR #861, opened 2020-05-26
compression: add response decompression

junr03:
Description: this PR pulls in the new decompressor filter in Envoy and configures response decompression in Envoy Mobile for gzip encoding.
Risk Level: med - new filter in envoy, not tested in production environments.
Testing: local build of the Lyft app. 


junr03:
perf compare will fail due to a >1% increase. Will merge

PR #870, opened 2020-05-30
bazel: lift dependencies out of WORKSPACE

buildbreaker:
`//bazel:envoy_mobile_repositories.bzl::envoy_mobile_repositories()`
Initializes all the `http_archive`, `http_file`, `http_jar` required by envoy mobile

`//bazel:envoy_mobile_swift_bazel_support.bzl::swift_support()`
Initializes the `@rules_foreign_cc//:workspace_definitions.bzl::rules_foreign_cc_dependencies()` and installs `build_bazel_apple_support` _after_ all the upstream envoy initializations

`//bazel:envoy_mobile_dependencies.bzl::envoy_mobile_dependencies()`
Initializes the dependencies from dependent repositories

`//bazel:envoy_mobile_toolchains.bzl::envoy_mobile_toolchains()`
Initializes toolchains required

Description: bazel: lift dependencies out of WORKSPACE
Risk Level: low
Testing: CI
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


bazel/envoy_mobile_repositories.bzl
```diff
@@ -0,0 +1,127 @@
...
+        strip_prefix = "rules_kotlin-%s" % rules_kotlin_version,
+        type = "zip",
+        urls = ["https://github.com/bazelbuild/rules_kotlin/archive/%s.zip" % rules_kotlin_version],
+    )
+
+    rules_detekt_version = "0.3.0"
+
+    rules_detekt_sha = "b1b4c8a3228f880a169ab60a817619bc4cf254443196e7e108ece411cb9c580e"
+
+    http_archive(
```
junr03, line 103: I am trying to understand the difference between the kotlin_repos and the the kotlin_dependencies (same for swift)

WORKSPACE
```diff
@@ -1,45 +1,5 @@
...
-# This should be kept in sync with Envoy itself, we just need to apply this patch
-# Remove this once https://boringssl-review.googlesource.com/c/boringssl/+/37804 is in master-with-bazel
-http_archive(
-    name = "boringssl",
-    patches = ["//bazel:boringssl.patch"],
-    sha256 = "36049e6cd09b353c83878cae0dd84e8b603ba1a40dcd74e44ebad101fc5c672d",
-    strip_prefix = "boringssl-37b57ed537987f1b4c60c60fa1aba20f3a0f6d26",
-    urls = ["https://github.com/google/boringssl/archive/37b57ed537987f1b4c60c60fa1aba20f3a0f6d26.tar.gz"],
-)
+load("//bazel:envoy_mobile_repositories.bzl", "envoy_mobile_repositories")
```
junr03, line 1: should we move this below all the upstream envoy rules?

buildbreaker: We have upstream envoy overrides here. We can do the non-overrides below the upstream envoy rules

junr03: oh so whoever declares first takes precedence?

buildbreaker: Yeah, whoever defines these first will take precedence because envoy has this macro which we leverage:  https://github.com/envoyproxy/envoy/blob/48a5b21d9483e7eddac79aeff7daac178d7b7462/api/bazel/envoy_http_archive.bzl#L4-L7

bazel/envoy_mobile_repositories.bzl
buildbreaker: There's a few concepts here: `http_archive`, `load()`, and actually calling the initialization method.

The `http_archive` is needed to download the remote repository, to call the initialization methods in the repository, we have to call the `load()` function to import the initialization method.

Here's the nuance of a `WORKSPACE` and a `bzl` file:
1. `load()` methods have to be declared at the very top of every `bzl` file
2. Some `load()` methods depend on other repositories which means we need to download these repositories first before we call `load()`
3. `load()` cannot be called within a `bzl` method so when we load a specific `bzl` file, we essentially call all the `load()` methods defined for the `bzl` file

This means that in the `WORKSPACE` we have more freedom to define things in a specific fragile order but when we move these to a `bzl` file, it requires some thoughtful organization (I did my best!). It's also the reason we have so many different files to replace whatever we had in `WORKSPACE`

WORKSPACE
goaway: Would it be worth splitting out overrides and including them first-class here, just to make it super clear the reason behind the ordering?

goaway: i.e. have a separate `envoy_mobile_overrides.bzl` or similar

buildbreaker: There's an implicit rationale for having this declared up top (and group these declarations before envoy). These are dependencies which are not dependent on any envoy upstream dependencies. Having this up top here guarantees that if we add new dependencies, we know it's free from any upstream envoy dependencies

We can separate this out but I'm not entirely sure if an explicit call out is needed. How about we keep this for now and if we end up deciding to split it out, we can do it then? @goaway

bazel/envoy_mobile_repositories.bzl
junr03: > This means that in the `WORKSPACE` we have more freedom to define things in a specific fragile order but when we move these to a `bzl` file, it requires some thoughtful organization

Ah this is the key point. I see! Thanks

WORKSPACE
buildbreaker: I'm going to merge this because I have other patches before we can depend on this from a separate repo. If you want this change, we can talk about it tomorrow @goaway

PR #871, opened 2020-06-01
observability: add stream stats to the http dispatcher

junr03:
Description: stream level stats for the http dispatcher
Risk Level:  low, new stats 
Testing: added new integration tests
Docs Changes: I want to have tables similar to envoy's, e.g., https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter.html?highlight=virtual%20cluster#virtual-clusters in an observability landing page https://github.com/lyft/envoy-mobile/issues/874


library/common/http/dispatcher.cc
```diff
@@ -206,6 +219,9 @@ void Dispatcher::DirectStreamCallbacks::onReset() {
   if (direct_stream_.dispatchable(true)) {
     ENVOY_LOG(debug, "[S{}] dispatching to platform remote reset stream",
               direct_stream_.stream_handle_);
+    // only count the on error if it is dispatchable. Otherwise, the onReset was caused due to a
```
buildbreaker, line 222:
```suggestion
    // Only count the on error if it is dispatchable. Otherwise, the onReset was caused due to a
```

library/common/http/dispatcher.cc
```diff
@@ -33,6 +34,13 @@ void Dispatcher::DirectStreamCallbacks::encodeHeaders(const ResponseHeaderMap& h
   ENVOY_LOG(debug, "[S{}] response headers for stream (end_stream={}):\n{}",
             direct_stream_.stream_handle_, end_stream, headers);
 
+  uint64_t response_status = Utility::getResponseStatus(headers);
+
+  // Record received status to report success or failure stat in
+  // Dispatcher::DirectStreamCallbacks::closeRemote. Not reported here to avoid more complexity with
+  // checking end stream, and the potential for stream reset, thus resulting in mis-reporting.
+  observed_success_ = CodeUtility::is2xx(response_status);
```
goaway, line 42: Why not just emit the stat here?

rebello95: Looks like this is explained via the docs directly above?

rebello95:
Can we also create an issue to add docs for this?

library/common/http/dispatcher.cc
buildbreaker: @junr03 Can correct me if I'm wrong (maybe that'll be best anyway!) but the intent here is to get a high level stat on when the application side is able to receive a success or failure. 

The reason we don't track it here is because we could run into other stream events where the application side wouldn't necessarily get a "complete request" but we'll be tracking a success here.

test/integration/dispatcher_integration_test.cc
```diff
@@ -177,13 +177,107 @@ TEST_P(DispatcherIntegrationTest, Basic) {
...
+  ConditionalInitializer ready_ran;
+  test_server_->server().dispatcher().post([this, &ready_ran]() -> void {
+    http_dispatcher_.ready(
+        test_server_->server().dispatcher(), test_server_->statStore(),
+        test_server_->server().listenerManager().apiListener()->get().http()->get());
+    ready_ran.setReady();
+  });
+  ready_ran.waitReady();
+
+  // Set response header status to be non-2xx to test correct stats get charged.
```
buildbreaker, line 194:
```suggestion
  // Set response header status to be non-2xx to test that correct stats get charged.
```

test/integration/dispatcher_integration_test.cc
```diff
@@ -114,7 +114,7 @@ TEST_P(DispatcherIntegrationTest, Basic) {
   ConditionalInitializer ready_ran;
   test_server_->server().dispatcher().post([this, &ready_ran]() -> void {
     http_dispatcher_.ready(
-        test_server_->server().dispatcher(),
+        test_server_->server().dispatcher(), test_server_->statStore(),
```
buildbreaker, line 117:
```suggestion
        test_server_->server().dispatcher(), 
        test_server_->statStore(),
```

test/integration/dispatcher_integration_test.cc
```diff
@@ -177,13 +177,107 @@ TEST_P(DispatcherIntegrationTest, Basic) {
   ASSERT_EQ(cc.on_headers_calls, 1);
   ASSERT_EQ(cc.on_data_calls, 2);
   ASSERT_EQ(cc.on_complete_calls, 1);
+
+  test_server_->waitForCounterEq("http.dispatcher.stream_success", 1);
+}
+
+TEST_P(DispatcherIntegrationTest, BasicNon2xx) {
```
buildbreaker, line 184: Maybe we should have a positive case also: `Basic2xx`?

test/integration/dispatcher_integration_test.cc
```diff
@@ -177,13 +177,107 @@ TEST_P(DispatcherIntegrationTest, Basic) {
...
+  terminal_callback.waitReady();
+
+  ASSERT_EQ(cc.on_error_calls, 0);
+  ASSERT_EQ(cc.on_headers_calls, 1);
+  ASSERT_EQ(cc.on_complete_calls, 1);
+
+  test_server_->waitForCounterEq("http.dispatcher.stream_failure", 1);
+}
+
+TEST_P(DispatcherIntegrationTest, BasicReset) {
```
buildbreaker, line 237: What do you think about the naming `BasicResetsIncrementFailures` to help describe what behavior we are trying to exercise here?

test/integration/dispatcher_integration_test.cc
buildbreaker: Do we want a test case to ensure we are incrementing the stats at the appropriate time based on: https://github.com/lyft/envoy-mobile/pull/871/files#r433531576

junr03: > Maybe we should have a positive case also: `Basic2xx`?

`Basic` covers that.

junr03: > Do we want a test case to ensure we are incrementing the stats at the appropriate time based on

I believe the race test below does that. Let me double check

library/common/http/dispatcher.cc
junr03: Yep, the intention to not increase here is that we could still receive a reset or a cancel if the on headers call is not final. So instead of increasing in this function in the dispatch block for the terminal case, and in on complete for the other cases I decided to just increase when on complete is dispatched.

test/integration/dispatcher_integration_test.cc
junr03: https://github.com/lyft/envoy-mobile/pull/871/files/bee32cbaa7147d723adc4ccabd810f65192efa65#diff-bb212e49f4cf7a1c832af632465e1440R339 this test shows that even if we received headers if cancelation won out the race we get the right stat

test/integration/dispatcher_integration_test.cc
junr03: letting clang-format call the shots

library/common/http/dispatcher.cc
goaway: I guess what I was getting at is you could emit stream_opened or stream_failed here. If canceled happens later that can still be reported and you (arguably?) have more insight, rather than less, with (maybe?) less complexity since you don't need state. But I guess it all depends on what you actually want captured.

I missed the comment. :)

goaway: I probably should have been more clear also - it's not that big a deal, and I am fine with the current approach. was mostly just wondering if the extra state was really necessary.

junr03:
It's green now!

PR #875, opened 2020-06-02
build: Downgrade xctestrunner

keith:

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -51,11 +51,12 @@ local_repository(
     path = "envoy_build_config",
 )
 
+# TODO: Remove once https://github.com/google/xctestrunner/issues/18 is fixed
 http_file(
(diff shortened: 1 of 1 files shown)


keith:
CI runs on a version of Xcode this doesn't support, so we're kinda stuck here. Locally you can `brew install python@2` and `brew link python@2`

buildbreaker:
Thanks for the workaround @keith

buildbreaker:
I seem to be having an issue locating `python@2` but I ended up doing this: https://stackoverflow.com/questions/60298514/brew-reinstalling-python2

buildbreaker:
Work around works!

keith:
probably `ln -s /usr/bin/python2.7 /usr/local/bin/python2` would be fine too

PR #877, opened 2020-06-04
bazel: ensure envoy-mobile is able to be built from an external repo

buildbreaker:
1. Add `@envoy_mobile` to loading bazel files which live in envoy mobile
1. Clean up the `aar_with_jni` macro and include sources and javadoc generation as part of the macro
1. Ensure we use java8 for executing mac os (dokka requires java8 rather than java14 which is what is used by default)

Description: bazel: ensure envoy-mobile is able to be built from an external repo
Risk Level: low
Testing: ci + local
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -4,12 +4,13 @@ licenses(["notice"])  # Apache 2
 
 alias(
     name = "ios_framework",
-    actual = "//library/swift/src:ios_framework",
+    actual = "@envoy_mobile//library/swift/src:ios_framework",
+    visibility = ["//visibility:public"],
 )
 
 genrule(
     name = "ios_dist",
-    srcs = ["//:ios_framework"],
+    srcs = [":ios_framework"],
     outs = ["ios_out"],
     cmd = """
 unzip -o $< -d dist/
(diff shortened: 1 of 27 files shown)


buildbreaker:
👀 @keith  -- most of this is pretty vanilla: find+replace or some small tweak to a target. 

The main/big change is the `bazel/aar_with_jni.bzl` file which I hope organizes things a little better

junr03:
>     1. Add `@envoy_mobile` to all absolute bazel references (target dependencies which start with `//`)

Can you explain why this is needed? I have not seen targets pertaining to a workspace need to reference the project with an `@project` prefix before. For instance, envoy does not have this.

BUILD
```diff
@@ -18,64 +19,37 @@ touch $@
...
-    cmd = """
-cp $(location :android_aar) dist/envoy.aar
-chmod 755 dist/envoy.aar
-touch $@
-""",
+    outs = ["envoy_mobile.zip"],
+    cmd = "$(location @envoy_mobile//bazel:zipper) fc $@ $(SRCS)",
     stamp = True,
+    tools = ["@envoy_mobile//bazel:zipper"],
+    visibility = ["//visibility:public"],
```
keith, line 37: Is an aar not already output as a zip? if it is you could probably remove this

BUILD
```diff
@@ -18,64 +19,37 @@ touch $@
...
-popd
-touch $@
-""",
+    for artifact in $(SRCS); do
+        chmod 755 $$artifact
+        cp $$artifact dist/
+    done
+    touch $@
+    """,
     stamp = True,
```
keith, line 53: I thought you were going to toss this one entirely?

BUILD
```diff
@@ -18,64 +19,37 @@ touch $@
...
     ],
-    outs = ["stub_android_dist_output"],
-    cmd = """
-cp $(location :android_aar) dist/envoy.aar
-chmod 755 dist/envoy.aar
-touch $@
-""",
+    outs = ["envoy_mobile.zip"],
+    cmd = "$(location @envoy_mobile//bazel:zipper) fc $@ $(SRCS)",
     stamp = True,
```
keith, line 35: Seems like you probably shouldn't need stamp here 🤔

bazel/BUILD
```diff
@@ -19,3 +19,8 @@ kt_jvm_library(
 exports_files([
     "pom_template.xml",
 ])
+
+alias(
+    name = "zipper",
+    actual = "@bazel_tools//tools/zip:zipper",
```
keith, line 25: I'd probably just inline this

bazel/aar_with_jni.bzl
```diff
@@ -4,82 +4,214 @@ load("@google_bazel_common//tools/maven:pom_file.bzl", "pom_file")
...
+        echo "Constructing aar..."
+        final_dir=$$(mktemp -d)
+        cp $$src_classes_jar $$final_dir/classes.jar
+        cd $$final_dir
+        unzip $$original_directory/$$src_jni_archive_apk > /dev/null
+        if [[ -d lib ]]; then
+            mv lib jni
+        else
+            echo "No jni directory found"
+        fi
```
keith, line 157:
```suggestion
            echo "No jni directory found"
            exit 1
        fi
```

might want to fail here

bazel/aar_with_jni.bzl
```diff
@@ -4,82 +4,214 @@ load("@google_bazel_common//tools/maven:pom_file.bzl", "pom_file")
...
+        cp $$original_directory/$$src_proguard_txt ./proguard.txt
+        cp $$original_directory/$$src_manifest_xml AndroidManifest.xml
+        zip -r tmp.aar * > /dev/null
+        cp tmp.aar $$original_directory/$@
```
keith, line 161: at some point with these scripts you might be better off splitting them into separate files

BUILD
buildbreaker: I...need to do more work to get that

BUILD
buildbreaker: What does that do actually?

bazel/BUILD
buildbreaker: buildifier (or an envoy format tool) complains that I'm using `@bazel` directly if I do

bazel/aar_with_jni.bzl
buildbreaker: It was to allow for the optionality of not having a native dependency (helped with testing this bazel rule)

bazel/aar_with_jni.bzl
buildbreaker: Would it be better to split each private method into a separate file now?

buildbreaker:
@junr03 
> Can you explain why this is needed? I have not seen targets pertaining to a workspace need to reference the project with an @project prefix before. For instance, envoy does not have this.

I think I misunderstood a bazel error and I'll need to load the appropriate rules. Let me fix this up.

PR #881, opened 2020-06-05
build: Add java_binary to produce sources jar

keith:
Every `java_library` rule (but maybe not kt_library rules) has a
implicit sources jar target that  can be requested. Unfortunately these
only contain the direct sources of the library. `java_binary` on the
other hand can produce one with all transitive sources. This adds the
smallest possible `java_binary` rule so we can produce this.

This behavior is lightly documented here: https://docs.bazel.build/versions/master/be/java.html

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -35,7 +35,7 @@ alias(
 
 alias(
     name = "android_sources",
-    actual = "//library:sources_jar",
+    actual = "//library/kotlin/src/io/envoyproxy/envoymobile:sources_jar_deploy-src.jar",
 )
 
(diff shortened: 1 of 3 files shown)


keith:
we'll have to rebase this once CI is fixed

buildbreaker:
@keith I actually took your idea and moved it into the `aar_with_jni` macro: https://github.com/lyft/envoy-mobile/pull/877. Would you be opposed to just going with the PR and closing this one?

keith:
sounds good

PR #883, opened 2020-06-05
wip: platform extension filter poc

goaway:
Description: Introduces a mechanism to write platform-native filters that can be linked into Envoy's HTTP filter chain.


goaway:
@junr03 format checks pass for me locally - any idea how I can debug what's going on?

rebello95:
Looks like buildifier failed based on the CI log. Do you have a diff when you do:

> library/common/extensions/filter/BUILD

junr03:
It might be the case that the version of buildifier you have is different than that in the build container https://github.com/envoyproxy/envoy-build-tools/blob/master/build_container/build_container_common.sh#L31. I would double check that.

library/common/extensions/filter/BUILD
```diff
@@ -0,0 +1,22 @@
+licenses(["notice"])  # Apache 2
+
+load("@envoy//bazel:envoy_build_system.bzl", "envoy_cc_library", "envoy_package")
+
+envoy_package()
+
+envoy_cc_library(
```
junr03, line 7: eventually I believe we want this to be `envoy_cc_extension`.

library/common/extensions/filter/c_types.h
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Return codes for header filter invocations. See envoy/http/filter.h
+ */
+typedef enum {
+  ENVOY_FILTER_HEADERS_STATUS_CONTINUE,
+  ENVOY_FILTER_HEADERS_STATUS_STOP_ITERATION,
+  ENVOY_FILTER_HEADERS_STATUS_CONTINUE_AND_END_STREAM,
+  ENVOY_FILTER_HEADERS_STATUS_STOP_ALL_ITERATION_AND_BUFFER,
+} envoy_filter_headers_status_t;
```
junr03, line 15: we discussed having a subset of these enums. Do you want to do that here in the declarations of in the mapping functions?

library/common/extensions/filter/c_types.h
```diff
@@ -0,0 +1,48 @@
...
+ * Raw datatype containing dispatch functions for a platform-native HTTP filter. Leveraged by the
+ * PlatformHarnessFilter
+ */
+typedef struct {
+  envoy_filter_on_headers_f on_request_headers;
+  envoy_filter_on_data_f on_request_data;
+  envoy_filter_on_headers_f on_response_headers;
+  envoy_filter_on_data_f on_response_data;
+  void* context;
+} envoy_http_filter;
```
junr03, line 48: Do we want to create a distinction between a request(decode) and response(encode) filter? And a sub-type that is able to do both? What would be a good pattern to do that in c?

library/common/extensions/filter/platform_filter.h
```diff
@@ -0,0 +1,36 @@
...
+#include "extensions/filters/http/common/pass_through_filter.h"
+
+#include "library/common/extensions/filter/c_types.h"
+
+namespace Envoy {
+namespace Http {
+/**
+ * Harness to bridge Envoy filter invocations up to the platform layer.
+ */
+class PlatformExtensionFilter final : public PassThroughFilter {
```
junr03, line 14: nit: above in c_types this was called `PlatformHarnessFilter`

library/common/extensions/filter/platform_filter.h
```diff
@@ -0,0 +1,36 @@
...
+public:
+  PlatformExtensionFilter(envoy_http_filter platform_filter) : platform_filter_(platform_filter) {}
+
+  // StreamDecoderFilter
+  FilterHeadersStatus decodeHeaders(RequestHeaderMap& headers, bool end_stream) override;
```
junr03, line 19: For now we don't need any of the other overrides except this one. We want to attempt to  harness a call to decode headers to modify the request header, right? The `PassThroughFilter` takes care of the rest.

library/common/extensions/filter/platform_filter.cc
```diff
@@ -0,0 +1,71 @@
...
+    return FilterHeadersStatus::StopAllIterationAndBuffer;
+  default:
+    ASSERT(false, "unrecognized filter status from platform: {}");
+    return FilterHeadersStatus::Continue;
+  }
+}
+
+FilterHeadersStatus PlatformExtensionFilter::decodeHeaders(RequestHeaderMap& headers,
+                                                           bool end_stream) {
+  return mapStatus(platform_filter_.on_request_headers(Utility::toBridgeHeaders(headers),
```
junr03, line 30: `Utility::toBridgeHeaders` copies the headers. If the platform modifies the values then it would not be reflected in the `RequestHeaderMap& headers`. We can solve this a few ways (as of course you know). I might bias toward on_request_headers returning the value? It would cause a second copy but allows us to keep the memory contract were the caller of the function owns memory cleanup, i.e the platform code would own cleaning the passed in value, and this code would on clean up of the return value after we update `RequestHeaderMap& headers` here.

wdyt?

library/common/extensions/filter/c_types.h
goaway: I decided to expose them all at this layer because there's no real reason not to.

library/common/extensions/filter/c_types.h
goaway: I mean you could use structs that have identical layouts but reserve the space for the missing callbacks. Or you could just use this one struct and set callbacks you're not using to null. I think the latter simplifies things.

library/common/extensions/filter/platform_filter.h
goaway: Yeah, I renamed a few times and missed that one. Settled on PlatformExtensionFilter, as in "the filter that enables platform extensions".

library/common/extensions/filter/platform_filter.h
goaway: We don't need them for a prototype, but I left them here for reference (and if we go this route, we'll definitely have them).

library/common/extensions/filter/c_types.h
junr03: Even if we don't want to expose all of that functionality to platform filters?

library/common/extensions/filter/c_types.h
junr03: But then you would have to do null checks or assertions, vs. having the type dictate what functions you should implement.

library/common/extensions/filter/platform_filter.h
junr03: Yeah, but given that the pass through filter implements, I was wondering if we should delete to make this PR easier to read TIOLI

library/common/extensions/filter/platform_filter.cc
goaway: discussed offline, will add more context to the PR/comments

library/common/extensions/filter/c_types.h
goaway: I think it's fine to expose them here. To explain myself better, my impulse is to not make assumptions at this layer.

library/common/extensions/filter/c_types.h
goaway: My plan was to default all callbacks to null and handle that gracefully in the base (bridge) implementation. I think it's more defensive, given the type variation.

library/common/extensions/filter/platform_filter.h
goaway: Either way.

junr03:
@goaway looks like there is currently an assert being triggered: [`[2020-06-18 06:32:57.068][81354][critical][assert] [library/common/api/external.cc:17] assert failure: api.`](https://github.com/lyft/envoy-mobile/pull/883/checks?check_run_id=783211948#step:10:73)

library/common/api/external.h
```diff
@@ -0,0 +1,22 @@
...
+
+#include <string>
+
+namespace Envoy {
+namespace Api {
+namespace External {
+
+/**
+ * Register an external runtime API for usage (most likely in extensions).
+ * NOTE: This is a proof of concept implementation and a HACK. Registration is NOT thread-safe.
```
rebello95, line 11: Do we intend to fix this before merging? Seems like both of these should be thread-safe now that we've proven it works

library/common/config_template.cc
```diff
@@ -32,6 +32,10 @@ const char* config_template = R"(
                         base_interval: 0.25s
                         max_interval: 60s
         http_filters:
+          - name: envoy.filters.http.platform_extension
+            typed_config:
+              "@type": type.googleapis.com/envoymobile.extensions.filters.http.platform_extension.Bridging
+              name: PlatformExample
```
rebello95, line 38: It'd be great to document how this works since it'll be removed before merging. Can we open an issue to document it?

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
```diff
@@ -0,0 +1,102 @@
...
+  case ENVOY_FILTER_HEADERS_STATUS_CONTINUE:
+    return Http::FilterHeadersStatus::Continue;
+  case ENVOY_FILTER_HEADERS_STATUS_STOP_ITERATION:
+    return Http::FilterHeadersStatus::StopIteration;
+  case ENVOY_FILTER_HEADERS_STATUS_CONTINUE_AND_END_STREAM:
+    return Http::FilterHeadersStatus::ContinueAndEndStream;
+  case ENVOY_FILTER_HEADERS_STATUS_STOP_ALL_ITERATION_AND_BUFFER:
+    return Http::FilterHeadersStatus::StopAllIterationAndBuffer;
+  default:
+    ASSERT(false, "unrecognized filter status from platform: {}");
```
rebello95, line 28: Looks like this is missing the format argument?

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
```diff
@@ -0,0 +1,102 @@
...
+    const envoymobile::extensions::filters::http::platform_extension::Bridging& proto_config)
+    : name_(proto_config.name()) {}
+
+BridgingFilter::BridgingFilter(BridgingFilterConfigSharedPtr config)
+    : platform_filter_(
+          static_cast<envoy_http_filter*>(Api::External::retrieveApi(config->name()))) {}
+
+Http::FilterHeadersStatus BridgingFilter::decodeHeaders(Http::RequestHeaderMap& headers,
+                                                        bool end_stream) {
+  // This is a hack that results in a full copy of the header map every time.
```
rebello95, line 43: Nit: this isn't really so much of a hack as an area for optimization, right? Can we file an issue and label it as tech debt?

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
```diff
@@ -0,0 +1,102 @@
...
+Http::FilterTrailersStatus BridgingFilter::decodeTrailers(Http::RequestTrailerMap& /*trailers*/) {
+  return Http::FilterTrailersStatus::Continue;
+}
+
+Http::FilterMetadataStatus BridgingFilter::decodeMetadata(Http::MetadataMap& /*metadata*/) {
+  return Http::FilterMetadataStatus::Continue;
+}
+
+Http::FilterHeadersStatus
+BridgingFilter::encode100ContinueHeaders(Http::ResponseHeaderMap& /*headers*/) {
```
rebello95, line 78: When you implement the rest of these functions, I assume we won't expose this one or metadata

library/common/extensions/filters/http/platform_extension/bridging_filter.h
```diff
@@ -0,0 +1,58 @@
...
+  Http::FilterHeadersStatus encode100ContinueHeaders(Http::ResponseHeaderMap& headers) override;
+  Http::FilterHeadersStatus encodeHeaders(Http::ResponseHeaderMap& headers,
+                                          bool end_stream) override;
+  Http::FilterDataStatus encodeData(Buffer::Instance& data, bool end_stream) override;
+  Http::FilterTrailersStatus encodeTrailers(Http::ResponseTrailerMap& trailers) override;
+  Http::FilterMetadataStatus encodeMetadata(Http::MetadataMap& metadata) override;
+
+private:
+  // FIXME: this leaks
+  const envoy_http_filter* platform_filter_;
```
rebello95, line 52: What does this leak do in practice? If it results in the Swift/Kotlin filter being retained indefinitely, that will have a very large impact on footprint since these are instantiated on every stream

library/common/extensions/filters/http/platform_extension/c_types.h
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Raw datatype containing dispatch functions for a platform-native HTTP filter. Leveraged by the
+ * BridgingFilter
+ */
+typedef struct {
+  envoy_filter_on_headers_f on_request_headers;
+  envoy_filter_on_data_f on_request_data;
+  envoy_filter_on_headers_f on_response_headers;
+  envoy_filter_on_data_f on_response_data;
```
rebello95, line 46: Looks like we're missing trailers?

library/common/extensions/filters/http/platform_extension/filter.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+package envoymobile.extensions.filters.http.platform_extension;
+
+import "validate/validate.proto";
+
+message Bridging {
```
rebello95, line 7: Would be nice to add a docstring

library/swift/src/filters/RequestFilter.swift
```diff
@@ -17,7 +26,7 @@ public protocol RequestFilter: Filter {
   ///
   /// - returns: The header status containing headers with which to continue or buffer.
   func onRequestHeaders(_ headers: RequestHeaders, endStream: Bool)
-    -> FilterHeaderStatus<RequestHeaders>
+    -> UInt32
```
rebello95, line 29: Is this the intended end state for this interface? I assume not since it doesn't actually allow the filter to mutate the headers (the existing interface requires the filter to return a status containing headers, which also has the benefit of ensuring the filter cannot mutate the headers after the function returns)

library/objective-c/EnvoyEngine.h
```diff
@@ -122,6 +133,7 @@ typedef NSDictionary<NSString *, NSArray<NSString *> *> EnvoyHeaders;
 @property (nonatomic, assign) UInt32 dnsRefreshSeconds;
 @property (nonatomic, assign) UInt32 dnsFailureRefreshSecondsBase;
 @property (nonatomic, assign) UInt32 dnsFailureRefreshSecondsMax;
+@property (nonatomic, strong) NSArray *httpFilters;
```
rebello95, line 136: Should these be `NSArray<EnvoyHTTPFilter *> *` types?

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -10,6 +10,72 @@ static void ios_on_exit() {
...
+    if (headerValueList == nil) {
+      headerValueList = [NSMutableArray new];
+      headerDict[headerKey] = headerValueList;
+    }
+    [headerValueList addObject:headerValue];
+  }
+  // The C envoy_headers struct can be released now because the headers have been copied.
+  release_envoy_headers(headers);
+  return headerDict;
+}
```
rebello95, line 47: Would be great to add tests for these bridging functions before merging

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -10,6 +10,72 @@ static void ios_on_exit() {
...
+// TODO: move impl
+static envoy_data toManagedNativeString(NSString *s) {
+  size_t length = s.length;
+  uint8_t *native_string = (uint8_t *)safe_malloc(sizeof(uint8_t) * length);
+  memcpy(native_string, s.UTF8String, length);
+  envoy_data ret = {length, native_string, free, native_string};
+  return ret;
+}
+
+static EnvoyMutableHeaders *to_ios_mutable_headers(envoy_headers headers) {
```
rebello95, line 27: Out of curiosity, why does this function use snake case when the other helpers use camel case?

library/swift/src/StreamClientBuilder.swift
```diff
@@ -107,6 +108,17 @@ public final class StreamClientBuilder: NSObject {
     return self
   }
 
+  /// Add HTTP filter chain for requests and responses (i.e. streams) managed by this client.
+  ///
+  /// - parameter filterChain: Filter chain to be invoked for streams.
+  ///
+  /// - returns: This builder.
+  @discardableResult
+  public func addFilterChain(_ filterChain: [RequestFilter]) -> StreamClientBuilder {
```
rebello95, line 117: This should probably take an individual filter, which gets appended to the underlying chain (rather than replacing)

rebello95: Also what are your thoughts on how inter-op ordering will work here with regard to C++ filters?

goaway:
Closing in favor of https://github.com/lyft/envoy-mobile/pull/940 and upcoming PRs.

library/objective-c/EnvoyEngineImpl.m
goaway: Our convention has been to use snake case for functions called from the c pathway; camel case for functions called from the Objective-C pathway.

library/swift/src/StreamClientBuilder.swift
goaway: We have a few options:
1. platform filters always come first on the request path / last on the response path - straightforward and what we'll do initially.
2. platform filters can be appended _or_ prepended - also pretty straightforward.
3. represent c++ filters as part of a platform-exposed chain that can be interacted with for arbitrary ordering - more complicated, and something to consider later, if necessary. possible, though.

library/objective-c/EnvoyEngineImpl.m
goaway: We don't yet have a framework in place for testing at the bridge layer, though @junr03 was talking about exploring this.

library/objective-c/EnvoyEngine.h
goaway: Sure 👍

library/swift/src/filters/RequestFilter.swift
goaway: Fixed in other PR.

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
goaway: Probably not initially, though probably both ultimately.

library/objective-c/EnvoyEngineImpl.m
rebello95: Got it. Maybe we should document that in our contribution guide?

library/swift/src/StreamClientBuilder.swift
rebello95: I'm up for doing #1 for MVP 👍

PR #885, opened 2020-06-05
build: add ENVOY_BAZEL_PREFIX to check_format script

goaway:
Description: Always needed for this project, so it might as well be baked in.


buildbreaker:
What does the `ENVOY_BAZEL_PREFIX` env var do in this script?

Can we leverage it to enforce that all the declarations within our repo uses `@envoy_mobile` as a prefix?

goaway:
It tells the build fixer that the envoy project uses the `@envoy` prefix, so it understands it doesn't need introduce additional includes at the top of every file.

PR #890, opened 2020-06-07
android: update to use androidx and fixate sdk version to be 29

buildbreaker:
Getting a bunch of strange build errors with using the legacy support v7 library from Android.

This change is to update to use the supported `androidx` libraries instead. The fixation of sdk api 29 is so we can target newer sdk apis. It shouldn't impact the NDK target version (goal is 21 but is currently 24)

Description: android: update to use androidx and fixate sdk version to be 29
Risk Level: low
Testing: ci
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


bazel/envoy_mobile_dependencies.bzl
```diff
@@ -22,7 +22,7 @@ def kotlin_dependencies():
         artifacts = [
             # Kotlin
             "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.3.11",
-
+            "androidx.recyclerview:recyclerview:1.0.0",
```
alexjlockwood, line 25: I recommend using the latest Android X releases listed under the "stable" column here: https://developer.android.com/jetpack/androidx/versions

The 1.0.0 versions are also super (like around September 2018 or something).

WORKSPACE
```diff
@@ -37,8 +37,8 @@ envoy_mobile_dependencies()
 load("@envoy_mobile//bazel:envoy_mobile_toolchains.bzl", "envoy_mobile_toolchains")
 envoy_mobile_toolchains()
 
-# Fixing to API 24. Need to lower to API 21: https://github.com/lyft/envoy-mobile/issues/936
 # Note: proguard is failing for API 30+
-android_sdk_repository(name = "androidsdk", api_level = 24)
+android_sdk_repository(name = "androidsdk", api_level = 29)
```
rebello95, line 41: Revert?

bazel/envoy_mobile_dependencies.bzl
```diff
@@ -22,7 +22,7 @@ def kotlin_dependencies():
         artifacts = [
             # Kotlin
             "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.3.11",
-
+            "androidx.recyclerview:recyclerview:1.1.0",
```
goaway, line 25: If this is only used in the example app, can it be removed from these dependencies?

WORKSPACE
rebello95: Ah I guess you still have it set for the NDK

buildbreaker: The NDK is more important than the SDK here. We _should_ be safe with this change

bazel/envoy_mobile_dependencies.bzl
buildbreaker: I can pull it to be a parameter if that helps but we'll have to declare it

goaway: I guess my vague misgiving is that it adds a new external dependency to the build (which could break), which isn't strictly necessary to build the library.

That said, if it's maintained by Google, it's probably fairly safe.

You could move this into the BUILD file of the example app though, couldn't you?

goaway: (Or to a function that only gets called when building those targets.)

buildbreaker: We spoke offline. I do agree that we should keep this as minimal as possible. @goaway is fine with making the change to clean this up as a follow up PR

buildbreaker: https://github.com/lyft/envoy-mobile/issues/950

PR #901, opened 2020-06-11
android: wire up on cancel from JNI layer

junr03:
Description: wire up missing on cancel call from JNI up to platform layer
Risk Level: low
Testing: build


buildbreaker:
`jvm_on_complete` is also like this i think

junr03:
On complete does not have a platform counterpart. It’s only used up to this layer to delete the ref.

buildbreaker:
Actually...let me verify this. It's building...

PR #904, opened 2020-06-12
swift / grpc: add grpc-status and grpc-message accessors

rebello95:
Adds a `GRPCResponseTrailers` and associated builder, which expose the `grpc-status` and `grpc-message` values - similarly to how we expose the `:status` from `HTTPResponseHeaders`.

I'll follow up with the Android changes when I port over the existing Android interfaces to match up with those introduced in https://github.com/lyft/envoy-mobile/pull/858.

Docs on these headers: https://chromium.googlesource.com/external/github.com/grpc/grpc/+/HEAD/doc/PROTOCOL-HTTP2.md


rebello95:
I need to think about this a bit more - looks like `grpc-status` can be returned in both/either headers or trailers 😬

PR #912, opened 2020-06-16
third-party build: update targets to allow extension registry overrides by a third-party build system

junr03:
Description: this PR moves the extension registry files out to top-level directories with their own Bazel workspaces. This allows us to declare it as local_repository. Once treated as a external repository, a third-party build system (i.e, a consumer of envoy-mobile interested in building their own library with private extensions) is able to do so using local_repository declarations in _their_ own repo.
Risk Level: low nothing changes for the project. However, the ergonomics for consumers to extend oss envoy mobile are being created here so we should make sure it is easy to do so.
Testing: local build with a third party filter.
Docs Changes: added docs
Release Notes: I propose we bump v0.3.1 when this override system lands.


WORKSPACE
```diff
@@ -13,6 +13,16 @@ local_repository(
     path = "envoy_build_config",
 )
 
+local_repository(
+    name = "envoy_mobile_bootstrap_config",
+    path = "envoy_mobile_bootstrap_config",
+)
+
+local_repository(
+    name = "extension_registry",
```
junr03, line 22: right now I have created just a single extension registry, but my other idea was to perhaps have three extension registries:
1. extensions declared in envoy
2. extensions declared in envoy-mobile
3. extensions declared in envoy-mobile private

This way if a consumer is not interested in overriding 1, and 2, they don't need to.

envoy_mobile_bootstrap_config/config_template.cc
```diff
@@ -77,6 +77,7 @@ const char* config_template = R"(
               inline_string: |
 )"
 #include "certificates.inc"
+
```
rebello95, line 80: Is this deliberate?

WORKSPACE
goaway: 2. I think is a special case that can be omitted - the only reason envoy-mobile declare an extension is if its integral to the project, right?

1. and 3. seem like they can be combined.

junr03: yep, for simplicity, I am going to leave just the one override point then.

junr03:
@goaway could you give another look?

junr03:
@goaway I decided to remove the config override, and make this PR only about extensions. Let me know if you have any remaining comments!

envoy_build_config/BUILD
```diff
@@ -1 +1,26 @@
+load("@envoy//bazel:envoy_build_system.bzl", "envoy_cc_library", "envoy_package")
+
 licenses(["notice"])  # Apache 2
+
+envoy_package()
+
+envoy_cc_library(
+    name = "static_registry",
```
goaway, line 8: would "extension_regsitry" be clearer?

goaway: (not rhetorical)

PR #913, opened 2020-06-16
kotlin: refactor public functionality as precursor to L7 filters

rebello95:
## Context

This is a mirror of the iOS interface refactors done in https://github.com/lyft/envoy-mobile/pull/858. Additionally, it introduces a set of mocks to match the iOS mocks introduced in https://github.com/lyft/envoy-mobile/pull/900.

Resolves https://github.com/lyft/envoy-mobile/issues/905 and resolves https://github.com/lyft/envoy-mobile/issues/907.

## Goals

- Unify the existing response handler and stream emitter types into a single "stream" type that will make it easier to communicate with an underlying filter manager for both outbound requests and inbound responses
- Provide guard rails that prevent the consumer from making mistakes, such as setting callback handlers after starting a stream and thus missing updates
- Provide a clear interface to the user for creating and starting streams
- Focus on the basics (bi-directional streaming support), which will allow for other wrappers to be written in the future (i.e., GRPC, unary, Rx, etc.)
- Update interfaces to utilize the new wrapper types introduced on master (`ResponseHeaders`, `RequestHeaders`, etc.)

Goals of mocks:

- Avoid adding logic to the mocked types. Instead, provide types that utilize the existing underlying logic of the library (`MockStreamPrototype` is a good example of this: it does not add a custom set of mappings for `StreamCallbacks` to the engine's `EnvoyHTTPCallbacks`, but instead uses the existing implementation to bridge simulated response calls)
- Ensure that no Java types/interfaces are required to mock requests/responses end-to-end
- Make it easy for end users to consume the mocks in the same way that they would the production library

## Types

- `StreamClient`: The publicly exposed interface of `EnvoyEngine`. Provides a function for starting a new stream using Envoy Mobile, and is built using `StreamClientBuilder`
- `StreamPrototype`: Type that allows users to set up a stream (providing callbacks, etc.) prior to starting it
- `Stream`: Produced by calling `start()` on a `StreamPrototype`, and allows the consumer to send data over the stream and eventually close it
- `GRPCClient` / `GRPCStreamPrototype` / `GRPCStream`: Types wrapping the above 3 types, converting the existing gRPC types to utilize the new functionality

## Tests

- I went through all of the Swift tests and audited the Kotlin tests to:
  - **Add/update to ensure that we now have 1:1 parity on tests between the two languages (even the test names are the same!)**
  - Fixed a bug I noticed while adding the tests where the retryable status code enum couldn't be properly serialized to headers and deserialized from those same headers
  - Migrate them to the new interfaces
- Sample apps have also been updated (CI validates via "hello world")

## Notes

- I noticed some Kotlin types that were unnecessarily public. Those have been made internal in this PR
- Logic and tests were added for setting headers with restricted prefixes (`:` and `x-envoy-mobile`), for parity with iOS
- Documentation will be updated separately in https://github.com/lyft/envoy-mobile/issues/906


library/java/src/io/envoyproxy/envoymobile/engine/EnvoyHTTPStream.java
```diff
@@ -9,14 +9,22 @@
...
 public class EnvoyHTTPStream {
-
   private final long streamHandle;
   private final JvmCallbackContext callbacksContext;
 
-  EnvoyHTTPStream(long streamHandle, EnvoyHTTPCallbacks callbacks) {
+  /**
+   * Start the stream via the JNI library.
+   */
+  void start() { JniLibrary.startStream(streamHandle, callbacksContext); }
```
rebello95, line 18: This change was necessary in order to ensure that mock subclasses of `EnvoyHTTPStream` don't end up calling JNI at all (doing so caused tests to crash). Instead, the initializer now no longer has the side effect of calling the JNI, and the engine starts the stream via this method (similarly to the ObjC engine [here](https://github.com/lyft/envoy-mobile/blob/99fdb261eb614685825c32ab71d4e98f4e579d56/library/objective-c/EnvoyEngineImpl.m#L59-L62))

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyClient.kt
```diff
@@ -0,0 +1,37 @@
+package io.envoyproxy.envoymobile
```
rebello95, line 1: File was renamed from `Envoy`, but git didn't recognize this as a rename because there were enough changes in the file

library/kotlin/src/io/envoyproxy/envoymobile/RequestHeadersBuilder.kt
```diff
@@ -35,6 +35,26 @@ class RequestHeadersBuilder : HeadersBuilder {
...
+  override fun remove(name: String): RequestHeadersBuilder {
+    super.remove(name)
+    return this
+  }
+
+  override fun internalSet(name: String, value: MutableList<String>): RequestHeadersBuilder {
+    super.internalSet(name, value)
+    return this
+  }
+
```
rebello95, line 57: These are necessary in order to return a `RequestHeadersBuilder` type rather than a `HeadersBuilder`

library/kotlin/src/io/envoyproxy/envoymobile/RetryPolicy.kt
```diff
@@ -68,7 +67,7 @@ enum class RetryRule(internal val stringValue: String) {
   companion object {
     internal fun enumValue(stringRepresentation: String): RetryRule? {
       return when (stringRepresentation) {
-        "status-5xx" -> STATUS_5XX
+        "5xx" -> STATUS_5XX
```
rebello95, line 70: This was a bug caught by the new tests I added 🙃

library/kotlin/src/io/envoyproxy/envoymobile/StreamCallbacks.kt
```diff
@@ -0,0 +1,52 @@
...
+  var onTrailers: ((trailers: ResponseTrailers) -> Unit)? = null
+  var onCancel: (() -> Unit)? = null
+  var onError: ((error: EnvoyError) -> Unit)? = null
+}
+
+/**
+ * Class responsible for bridging between the platform-level `StreamCallbacks` and the
+ * engine's `EnvoyHTTPCallbacks`.
+ */
+internal class EnvoyHTTPCallbacksAdapter(
```
rebello95, line 25: Partly moved from a different file

library/swift/src/grpc/GRPCRequestHeadersBuilder.swift
```diff
@@ -21,14 +21,14 @@ public final class GRPCRequestHeadersBuilder: HeadersBuilder {
 
   /// Add a specific timeout for the gRPC request. This will be sent in the `grpc-timeout` header.
   ///
-  /// - parameter timeoutMS: Timeout, in milliseconds.
+  /// - parameter timeoutMs: Timeout, in milliseconds.
   ///
   /// - returns: This builder.
   @discardableResult
-  public func addTimeoutMS(_ timeoutMS: UInt?) -> GRPCRequestHeadersBuilder {
+  public func addTimeoutMs(_ timeoutMs: UInt?) -> GRPCRequestHeadersBuilder {
```
rebello95, line 28: Renamed for consistency on both platforms

library/swift/src/grpc/GRPCStreamPrototype.swift
```diff
@@ -96,6 +96,20 @@ public final class GRPCStreamPrototype: NSObject {
     self.underlyingStream.setOnError(closure: closure)
     return self
   }
+
+  /// Specify a callback for when the stream is canceled.
```
rebello95, line 100: This was missing on master

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyClient.kt
```diff
@@ -0,0 +1,37 @@
...
+    if (envoyConfiguration == null) {
+      engine.runWithConfig(configurationYAML, logLevel.level)
+    } else {
+      engine.runWithConfig(envoyConfiguration, logLevel.level)
+    }
+  }
+
+  override fun newStreamPrototype(): StreamPrototype {
+    return StreamPrototype(engine)
+  }
```
johnpaulmckean, line 36: can just be:
```
override fun newStreamPrototype() = StreamPrototype(engine)
```

library/kotlin/src/io/envoyproxy/envoymobile/HeadersBuilder.kt
```diff
@@ -49,8 +55,28 @@ open class HeadersBuilder {
...
+   * @return HeadersBuilder, This builder.
+   */
+  internal open fun internalSet(name: String, value: MutableList<String>): HeadersBuilder {
+    headers[name] = value
+    return this
+  }
+
+  private fun isRestrictedHeader(name: String): Boolean {
+    return name.startsWith(":") || name.startsWith("x-envoy-mobile")
+  }
```
johnpaulmckean, line 81: can be:
```
private fun isRestrictedHeader(name: String) = name.startsWith(":") || name.startsWith("x-envoy-mobile")
```

library/kotlin/src/io/envoyproxy/envoymobile/RequestTrailers.kt
```diff
@@ -11,4 +11,16 @@ class RequestTrailers : Headers {
...
+   * Convert the trailers back to a builder for mutation.
+   *
+   * @return RequestTrailersBuilder, The new builder.
+   */
+  fun toRequestTrailersBuilder(): RequestTrailersBuilder {
+    return RequestTrailersBuilder(
+      this.headers.mapValues { it.value.toMutableList() }
+        .toMutableMap()
+    )
+  }
```
johnpaulmckean, line 25: can be:
```
fun toRequestTrailersBuilder() = RequestTrailersBuilder(this.headers.mapValues { 
    it.value.toMutableList() 
}.toMutableMap())
```

library/kotlin/src/io/envoyproxy/envoymobile/grpc/GRPCClient.kt
```diff
@@ -1,27 +1,23 @@
...
-   * @returns An emitter that can be used for sending more traffic over the stream.
+   * @return The new gRPC stream prototype.
    */
-  fun start(request: Request, grpcResponseHandler: GRPCResponseHandler): GRPCStreamEmitter {
-    val emitter = httpClient.start(request, grpcResponseHandler.underlyingHandler)
-    return GRPCStreamEmitter(emitter)
+  fun newGRPCStreamPrototype(): GRPCStreamPrototype {
+    val prototype = streamClient.newStreamPrototype()
+    return GRPCStreamPrototype(prototype)
   }
```
johnpaulmckean, line 22: would be pretty:
```
  fun newGRPCStreamPrototype() = GRPCStreamPrototype(
    streamClient.newStreamPrototype()
 )
```

library/kotlin/src/io/envoyproxy/envoymobile/grpc/GRPCRequestHeaders.kt
```diff
@@ -0,0 +1,22 @@
...
+  internal constructor(headers: Map<String, List<String>>) : super(headers)
+
+  /**
+   * Convert the headers back to a builder for mutation.
+   *
+   * @return GRPCRequestHeadersBuilder, The new builder.
+   */
+  fun toGRPCRequestHeadersBuilder(): GRPCRequestHeadersBuilder {
+    return GRPCRequestHeadersBuilder(this.headers.mapValues { it.value.toMutableList() }.toMutableMap())
+  }
```
johnpaulmckean, line 21: you know ;)

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyClient.kt
```diff
@@ -0,0 +1,37 @@
...
+  constructor(
+    engine: EnvoyEngine,
+    envoyConfiguration: EnvoyConfiguration,
+    logLevel: LogLevel = LogLevel.INFO
+  ) : this(engine, envoyConfiguration, null, logLevel)
+  constructor(
+    engine: EnvoyEngine,
+    configurationYAML: String,
+    logLevel: LogLevel = LogLevel.INFO
+  ) : this(engine, null, configurationYAML, logLevel)
```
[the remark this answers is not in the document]
rebello95: Fair callout. This was moved from a different file, but if I recall correctly we elected to keep these two constructors to make it clearer that _either_ a yaml configuration or a typed configuration could be specified (i.e., preventing them from passing both values and ignoring one of them). iOS for example: [here](https://github.com/lyft/envoy-mobile/blob/d4e0d0a61c3a05ef377e10ebbda94da17491b80b/library/swift/src/EnvoyClient.swift#L26-L43)

library/kotlin/src/io/envoyproxy/envoymobile/EnvoyClient.kt
rebello95: nice

library/kotlin/src/io/envoyproxy/envoymobile/mocks/MockEnvoyEngine.kt
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  override fun runWithConfig(configurationYAML: String?, logLevel: String?): Int {
+    return 0
+  }
+
+  override fun startStream(callbacks: EnvoyHTTPCallbacks?): EnvoyHTTPStream {
+    return MockEnvoyHTTPStream(callbacks!!)
+  }
+}
```
johnpaulmckean, line 23: this can all be just a few lines.

library/kotlin/src/io/envoyproxy/envoymobile/HeadersBuilder.kt
rebello95: TIL 👍

library/kotlin/test/io/envoyproxy/envoymobile/GRPCRequestHeadersBuilderTest.kt
```diff
@@ -0,0 +1,69 @@
+package io.envoyproxy.envoymobile
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+class GRPCRequestHeadersBuilderTest {
```
johnpaulmckean, line 6: would highly recommend using kluent:
https://github.com/MarkusAmshove/Kluent

rebello95: Might be good for future reference (not sure if @buildbreaker has thoughts, I'm not familiar with it 🤷)

rebello95:
Thanks for the review @johnpaulmckean! I addressed your comments

library/kotlin/test/io/envoyproxy/envoymobile/GRPCRequestHeadersBuilderTest.kt
buildbreaker: Certainly open to new frameworks which would help. Would you like to file an issue for us to have a conversation about this @johnpaulmckean?

johnpaulmckean: done.

buildbreaker: Thank you @johnpaulmckean!

rebello95: Thanks! https://github.com/lyft/envoy-mobile/issues/930 for future reference

PR #915, opened 2020-06-17
swift: Add another explicit init for Swift 5.2

keith:
Without this we see this crash:

```
Fatal error: Use of unimplemented initializer 'init()' for class 'Envoy.MockStreamClient'
```


keith:
It would if we were validating this here, but we can't right now because we're waiting on https://github.com/bazelbuild/rules_apple/pull/753

In theory we could add the patch but that would force any bazel users to also add the patch (probably no one tho?)

PR #927, opened 2020-06-23
ci: fix liveliness flakiness

rebello95:
We've seen some flakiness on the iOS liveliness CI jobs which appear to be caused by the fact that the first response log from the hello world projects is not showing up when the script greps for `Response status (1): 200`. This PR aims to:

- Fix these flakes by grepping for any 200 response in the logs (rather than specifically the first one)
- Unify the logging and CI behavior for all 4 languages
- Fix error messages not being displayed in the Objective-C hello world app


junr03:
@rebello95 looks like there is some linting and building issues with the android side of things

rebello95:
Green now @buildbreaker @junr03

PR #940, opened 2020-07-07
filters: add initial support for bridged filters

goaway:
Description: Adds support for platform-bridged filter implementations at the common and bridge layers of the library. Includes a stub implementation hardcoded into configuration. Dynamic configuration and data support, with platform-specific presentations, to come.
Risk Level: Moderate
Testing: Local end-to-end


goaway:
Note: this requires upstream Envoy update, and will need to be rebased when that merges. See: https://github.com/lyft/envoy-mobile/pull/923

goaway:
Note: pending completed stub implementation.

library/common/extensions/filters/http/platform_extension/c_types.h
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Raw datatype containing dispatch functions for a platform-native HTTP filter. Leveraged by the
+ * BridgingFilter
+ */
+typedef struct {
+  envoy_filter_on_headers_f on_request_headers;
+  envoy_filter_on_data_f on_request_data;
+  envoy_filter_on_trailers_f on_request_trailers;
+  envoy_filter_on_headers_f on_response_headers;
```
rebello95, line 93: In the past we've had bugs with C structs that have callback properties with the same types (i.e., `envoy_filter_on_headers_f` for request and response). Would it be possible to either a) add tests for the bridging (@junr03 maybe has thoughts?) or b) use different type signatures so they're enforced by the compiler?

library/common/extensions/filters/http/platform_extension/config.h
```diff
@@ -0,0 +1,33 @@
...
+#include "library/common/extensions/filters/http/platform_extension/filter.pb.h"
+#include "library/common/extensions/filters/http/platform_extension/filter.pb.validate.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace PlatformExtension {
+
+/**
+ * Config registration for the decompressor filter. @see NamedHttpFilterConfigFactory.
```
rebello95, line 14: Bad copy/paste

library/common/extensions/filters/http/platform_extension/filter.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+package envoymobile.extensions.filters.http.platform_extension;
+
+import "validate/validate.proto";
+
+message Bridging {
+  string name = 1 [(validate.rules).string.min_bytes = 1];
+}
```
rebello95, line 9: Can we document how this works/is intended to be used?

envoy_build_config/BUILD
```diff
@@ -22,5 +22,6 @@ envoy_cc_library(
         "@envoy//source/extensions/filters/network/http_connection_manager:config",
         "@envoy//source/extensions/stat_sinks/metrics_service:config",
         "@envoy//source/extensions/transport_sockets/tls:config",
+        "@envoy_mobile//library/common/extensions/filters/http/platform_extension:config",
```
junr03, line 25: nit: the name extension suffix is a little odd, because the extension is a filter, not a generic "PlatformExtension".

envoy_build_config/extensions_build_config.bzl
```diff
@@ -5,5 +5,6 @@ EXTENSIONS = {
     "envoy.filters.network.http_connection_manager":    "//source/extensions/filters/network/http_connection_manager:config",
     "envoy.stat_sinks.metrics_service":                 "//source/extensions/stat_sinks/metrics_service:config",
     "envoy.transport_sockets.tls":                      "//source/extensions/transport_sockets/tls:config",
+    "envoy.filters.http.platform_extension":            "@envoy_mobile//library/common/extensions/filters/http/platform_extension:config",
```
junr03, line 8: should we do alpha ordering here?

envoy_build_config/BUILD
junr03: Ok, after reading more of the PR I think I see your thought here. The `PlatformExtension` namespace contains types that bridge to the platform. In this case it is the bridge filter. Usually the directory and the namespace share the filter's name, but I see how here is different.

library/common/api/external.cc
```diff
@@ -0,0 +1,25 @@
+#include "external.h"
+
+#include <unordered_map>
+
+#include "common/common/assert.h"
+
+namespace Envoy {
+namespace Api {
+namespace External {
```
junr03, line 9:
```suggestion
namespace PlatformExtension {
```

I think this meshes well with the rest of the PR, because it tells us that these function signatures allow us to register _any_ platform extension api?

library/common/api/external.cc
```diff
@@ -0,0 +1,25 @@
...
+
+static std::unordered_map<std::string, void*> registry_{};
```
junr03, line 11: How will this work with instance-based engines? Will the registry be per instance, or do you see this always being static?

library/common/api/external.h
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Envoy {
+namespace Api {
+namespace External {
+
+/**
+ * Register an external runtime API for usage (most likely in extensions).
+ * NOTE: This is a proof of concept implementation and a HACK. Registration is NOT thread-safe.
+ */
+void registerApi(std::string name, void* api);
```
junr03, line 13: Should probably also note that name collisions update the stored api?

library/common/api/external.cc
```diff
@@ -0,0 +1,25 @@
...
+
+void registerApi(std::string name, void* api) {
+  registry_[name] = api;
+}
+
+void* retrieveApi(std::string name) {
+  void* api = registry_[name];
+  ASSERT(api);
```
junr03, line 19: Interesting, crash if not present?

library/common/extensions/filters/http/platform_extension/filter.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+package envoymobile.extensions.filters.http.platform_extension;
+
+import "validate/validate.proto";
+
+message Bridging {
+  string name = 1 [(validate.rules).string.min_bytes = 1];
```
junr03, line 8:
```suggestion
  string platform_api_name = 1 [(validate.rules).string.min_bytes = 1];
```

library/common/extensions/filters/http/platform_extension/config.h
junr03, line 14:
```suggestion
 * Config registration for the bridging filter. @see NamedHttpFilterConfigFactory.
```

library/common/extensions/filters/http/platform_extension/config.h
```diff
@@ -0,0 +1,33 @@
...
+ */
+class BridgingFilterFactory
+    : public Common::FactoryBase<
+          envoymobile::extensions::filters::http::platform_extension::Bridging> {
+public:
+  BridgingFilterFactory() : FactoryBase("platform_extension") {}
```
junr03, line 20:
```suggestion
  BridgingFilterFactory() : FactoryBase("bridging") {}
```

library/common/extensions/filters/http/platform_extension/c_types.h
```diff
@@ -0,0 +1,97 @@
+#pragma once
+
+#include "library/common/types/c_types.h"
+
+// NOLINT(namespace-envoy)
+
+/**
+ * Return codes for on-headers filter invocations. See envoy/http/filter.h
```
junr03, line 8:
```suggestion
 * Return codes for on-headers filter invocations. @see envoy/http/filter.h
```

same elsewhere

library/common/extensions/filters/http/platform_extension/c_types.h
```diff
@@ -0,0 +1,97 @@
...
+                                                                   bool end_stream,
+                                                                   void* context);
+
+#ifdef __cplusplus
+} // function pointers
+#endif
+
+/**
+ * Raw datatype containing dispatch functions for a platform-native HTTP filter. Leveraged by the
+ * BridgingFilter
```
junr03, line 87:
```suggestion
 * BridgingFilter.
```

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
```diff
@@ -0,0 +1,115 @@
...
+    return Http::FilterHeadersStatus::Continue;
+  }
+}
+
+BridgingFilterConfig::BridgingFilterConfig(
+    const envoymobile::extensions::filters::http::platform_extension::Bridging& proto_config)
+    : name_(proto_config.name()) {}
+
+BridgingFilter::BridgingFilter(BridgingFilterConfigSharedPtr config)
+    : platform_filter_(
```
junr03, line 40: I wonder if this should live in the BridgingFilterConfig class. That way if we crash/throw an exception we can do so at config time and not when the first stream creates the filter.

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
```diff
@@ -0,0 +1,115 @@
...
+Http::FilterHeadersStatus BridgingFilter::decodeHeaders(Http::RequestHeaderMap& headers,
+                                                        bool end_stream) {
+  envoy_headers in_headers = Http::Utility::toBridgeHeaders(headers);
+  envoy_filter_headers_status result = platform_filter_->on_request_headers(in_headers, end_stream, platform_filter_->context);
+  Http::FilterHeadersStatus status = mapStatus(result.status);
+  // Current platform implementations expose immutable headers, thus any modification necessitates a full copy.
+  // If the returned pointer is identical, we assume no modification was made and elide the copy here.
+  if (in_headers.headers != result.headers.headers) {
+    headers.clear();
+    for (envoy_header_size_t i = 0; i < result.headers.length; i++) {
```
junr03, line 52: TODO for optimization

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
```diff
@@ -0,0 +1,115 @@
...
+Http::FilterMetadataStatus BridgingFilter::decodeMetadata(Http::MetadataMap& /*metadata*/) {
+  return Http::FilterMetadataStatus::Continue;
+}
+
+Http::FilterHeadersStatus
+BridgingFilter::encode100ContinueHeaders(Http::ResponseHeaderMap& /*headers*/) {
+  return Http::FilterHeadersStatus::Continue;
+}
+
+Http::FilterHeadersStatus BridgingFilter::encodeHeaders(Http::ResponseHeaderMap& headers,
```
junr03, line 80: we can probably extract all this logic to a helper function that takes a function pointer, that way decode/encodeHeaders don't need to have all this repeated code?

library/common/extensions/filters/http/platform_extension/filter.proto
goaway: This particular part won't ever really be publicly exposed/used. But I do intend to make a documentation PR that fully describes how to create and use platform extensions.

library/common/extensions/filters/http/platform_extension/config.h
goaway: Yep, thank you.

library/common/extensions/filters/http/platform_extension/c_types.h
goaway: The compiler won't actually save us here even if we create a separate definition, because at the end of the day they are just `typedef`s that will compile down to the same thing.

I agree about testing and once we have a framework in place for bridge testing, this will be an area to cover.

library/common/extensions/filters/http/platform_extension/bridging_filter.h
```diff
@@ -0,0 +1,58 @@
...
+  Http::FilterHeadersStatus encode100ContinueHeaders(Http::ResponseHeaderMap& headers) override;
+  Http::FilterHeadersStatus encodeHeaders(Http::ResponseHeaderMap& headers,
+                                          bool end_stream) override;
+  Http::FilterDataStatus encodeData(Buffer::Instance& data, bool end_stream) override;
+  Http::FilterTrailersStatus encodeTrailers(Http::ResponseTrailerMap& trailers) override;
+  Http::FilterMetadataStatus encodeMetadata(Http::MetadataMap& metadata) override;
+
+private:
+  // FIXME: this leaks
+  const envoy_http_filter* platform_filter_;
```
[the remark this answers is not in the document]
goaway: This won't cause an issue right now because it's effectively static (as is the engine itself). It's just a note to myself so I don't forget about it later.

I will add a more detailed comment and link it to the issue for a fully-instantiated engine.

envoy_build_config/BUILD
goaway: I'm not married to the naming, but yeah, that was my thinking at the time.

library/common/api/external.cc
goaway: Per-instance. I'm going to do a pass and flag and link anything that needs to be updated once the engine is instantiated.

library/common/api/external.h
goaway: I considered asserting on collisions. This is really just a quick fix solution that could be improved to be safe.

Since this will be handled behind the scenes when I update with the dynamic configuration PR, I decided to wait to update it until then.

library/common/api/external.cc
goaway: Sure - it may be obvious that I waffled a bit on naming and "external" was just a possibility I considered, too.

envoy_build_config/extensions_build_config.bzl
goaway: Sounds good.

library/common/config_template.cc
```diff
@@ -32,6 +32,10 @@ const char* config_template = R"(
                         base_interval: 0.25s
                         max_interval: 60s
         http_filters:
+          - name: envoy.filters.http.platform_extension
+            typed_config:
+              "@type": type.googleapis.com/envoymobile.extensions.filters.http.platform_extension.Bridging
+              name: PlatformStub
```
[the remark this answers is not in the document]
goaway: (Same as below)
This particular part won't ever really be publicly exposed/used. But I do intend to make a documentation PR that fully describes how to create and use platform extensions.

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
```diff
@@ -0,0 +1,115 @@
+#include "library/common/extensions/filters/http/platform_extension/bridging_filter.h"
+
+//#include <iostream>
```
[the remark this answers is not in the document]
goaway: Thanks. :)

library/common/api/external.h
```diff
@@ -0,0 +1,22 @@
...
+
+#include <string>
+
+namespace Envoy {
+namespace Api {
+namespace External {
+
+/**
+ * Register an external runtime API for usage (most likely in extensions).
+ * NOTE: This is a proof of concept implementation and a HACK. Registration is NOT thread-safe.
```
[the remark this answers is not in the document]
goaway: Yes, but improvement will come in a separate PR (see below).

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
goaway: Remember that for the current use case, with dynamic filter configuration, this will be handled correctly behind the scenes, at config time.

For the general case I don't really know how to make that happen without significant new plumbing, because the platform API is a runtime-defined memory location, that must be passed by the platform down into Envoy. This would require threading that into Envoy's startup sequencing. I'd suggest it could maybe be an item for future work, but a crash an API lookup seemed good enough for now (again, especially since dynamic configuration will handle this correctly without user intervention).

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
goaway: This already elides the copy in the unmodified case.

@rebello95 and @buildbreaker felt somewhat strongly about using immutable objects in the filter callbacks, so a full copy is more or less necessary if any modification was made (and in fact, has already happened). (We could filter for a diff in linear time, but then the copy is linear, too). If we created a _custom_ builder in the callback that accumulated a diff and passed back the builder _instead_ of a realized collection, we could operate on the diff directly, but I think that's introducing quite a lot of complexity at that point. After some more reflection, I basically think there are two reasonable approaches:

1. Accept that using immutable objects implies copies will be made.
2. Switch to mutable objects like Envoy does in its own filters, opening the door for fully-bridged collections.

1) seems good enough to me for now, though I'd somewhat prefer 2) long-term. Either way I think further discussion would need to happen were we to change course.

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
goaway: Yep I actually planned to do exactly that, but was holding off while things were in flux.

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
```diff
@@ -0,0 +1,115 @@
...
+  case ENVOY_FILTER_HEADERS_STATUS_CONTINUE:
+    return Http::FilterHeadersStatus::Continue;
+  case ENVOY_FILTER_HEADERS_STATUS_STOP_ITERATION:
+    return Http::FilterHeadersStatus::StopIteration;
+  case ENVOY_FILTER_HEADERS_STATUS_CONTINUE_AND_END_STREAM:
+    return Http::FilterHeadersStatus::ContinueAndEndStream;
+  case ENVOY_FILTER_HEADERS_STATUS_STOP_ALL_ITERATION_AND_BUFFER:
+    return Http::FilterHeadersStatus::StopAllIterationAndBuffer;
+  default:
+    ASSERT(false, "unrecognized filter status from platform: {}");
```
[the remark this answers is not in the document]
goaway: It's coming from the platform layer. Thank you for the catch on the interpolation. :)

library/common/api/external.cc
goaway: Yes, because I don't really see a readily-available better option, and at least for this case, this won't be exposed for direct user interaction (see the comment for more of my thinking on this).

If the API isn't located, a crash is probably about to happen anyways; this just does the operator of the library the courtesy of telling them why.

library/common/extensions/filters/http/platform_extension/config.h
goaway: Thanks, good catch. :)

library/common/extensions/filters/http/platform_extension/filter.proto
goaway: I sort of think `name` is fine here. There's precedent, and recall that we'll be generating this. If we were to change it though, I would consider `platform_filter_name`, since not just any platform api will cut it.

goaway:
This update adds an optimized path for no-op filter callbacks, as well as allowing calloc to trivially create a no-op filter - utilized here as a PlatformStub implementation.

It also factors out common callback logic on the request and response path for headers.

goaway:
With this the PR should be functionally green, once upstream can be updated.

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
goaway: Done in the latest update. :)

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
buildbreaker: Yeah, I feel 1 is an acceptable approach right now. We know of this could be optimized (we also know how we want to optimize it) and we can tackle this problem later on.

library/common/api/external.h
junr03: alright, I'd leave a todo then

library/common/extensions/filters/http/platform_extension/filter.proto
junr03: I vote for `platform_filter_name`. Usually, `name` referred to the name of the filter itself, but here it refers to the name of the _platform_ filter. So even though this is not public, I think a more specific name is good.

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
junr03: Ok, my thinking was that a user would register platform filters after engine instantiation but before initialization. So then it would make sense to have the lookup in the config. But if the thinking is that a user might register filters after engine initialization this makes sense. 

So then, I guess the larger question is how long do we want to support filter registration.

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
junr03: Right, so I think we should leave some trail of that long term thinking. Perhaps actually an issue is better.

envoy_build_config/BUILD
goaway: See what you think of the updated naming.

library/common/api/external.cc
goaway: Now that I've updated everything else to standardize on `PlatformBridge`, I actually think this should remain `External`. _Any_ API external to the library could be registered here, but everything labeled `PlatformBridge` relates specifically to bridging up to a platform runtime environment.

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
goaway: I think we're misunderstanding each other. I am indeed planning to register filters prior to Engine initialization.

But I think I understand you now. You're suggesting the config could perform the API lookup during initialization. Sure - in fact, I agree, I like that better. (Is that what you mean?)

library/common/extensions/filters/http/platform_extension/bridging_filter.cc
goaway: To clarify, I don't think there is really room for further optimization if we stick with immutable types. (At least anything that won't be terribly convoluted.)

So, I was trying to say is that there's maybe only a TODO for further optimization if we expect to switch to mutable types at some point in the future.

But I agree with your point about capturing context, and will create an issue and link it here.

goaway: Done.

library/common/extensions/filters/http/platform_bridge/c_types.h
```diff
@@ -0,0 +1,96 @@
...
+ * Function signature for on-data filter invocations.
+ */
+typedef envoy_filter_data_status (*envoy_filter_on_data_f)(envoy_data data, bool end_stream,
+                                                           void* context);
+
+/**
+ * Function signature for on-trailers filter invocations.
+ */
+typedef envoy_filter_trailers_status (*envoy_filter_on_trailers_f)(envoy_headers trailers,
+                                                                   bool end_stream, void* context);
```
rebello95, line 78: `end_stream` should be omitted since trailers always indicate an end-of-stream

library/common/types/c_types.h
```diff
@@ -103,6 +103,14 @@ extern "C" { // utility functions
  */
 void* safe_malloc(size_t size);
 
+/**
+ * calloc wrapper that asserts that the returned pointer is valid. Otherwise, the program exits.
```
rebello95, line 107: This isn't necessarily true since we compile out release asserts, right? Maybe worth clarifying

library/common/extensions/filters/http/platform_bridge/c_types.h
goaway: Yep, good catch.

library/common/types/c_types.h
goaway: It's a little confusing, but actually the other way around. `RELEASE_ASSERT` is never compiled out, regular `ASSERT` can be.

goaway:
Finally rebased on upstream Envoy update.

goaway:
CI is green; proceeding with merge (with previous approvals).

library/common/extensions/filters/http/platform_extension/filter.proto
goaway: https://github.com/lyft/envoy-mobile/pull/955

PR #943, opened 2020-07-08
filters: add iOS header support

goaway:
Description: Adds initial support for Swift presentation of HTTP filter on-headers invocations.
Risk Level: Moderate
Testing: Local end-to-end


library/objective-c/EnvoyEngineImpl.m
```diff
@@ -29,13 +117,33 @@ - (void)dealloc {
...
+  ios_http_filter_context *context = safe_malloc(sizeof(ios_http_filter_context));
+  CFBridgingRetain(filter);
+  context->filter = filter;
+  envoy_http_filter *api = safe_malloc(sizeof(envoy_http_filter));
+  api->on_request_headers = ios_http_filter_on_request_headers;
+  api->on_request_data = NULL;
+  api->on_response_headers = ios_http_filter_on_response_headers;
+  api->on_response_data = NULL;
+  api->context = context;
+  register_platform_api(filter.name.UTF8String, api);
```
buildbreaker, line 132: Are we able to generate a filter name here (for the user) instead of requiring the implementer to handle the name for the registration?

goaway: Yep, that's covered in the dynamic configuration task in the linked issue. @junr03 suggested, and I agreed that we split that into a separate PR.

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -10,6 +11,93 @@ static void ios_on_exit() {
...
+    NSArray *headerList = headers[headerKey];
+    for (id headerValue in headerList) {
+      envoy_header new_header = {toManagedNativeString(headerKey),
+                                 toManagedNativeString(headerValue)};
+      header_array[header_index++] = new_header;
+    }
+  }
+  // TODO: ASSERT(header_index == length);
+  envoy_headers ret = {length, header_array};
+  return ret;
```
rebello95, line 66:
```suggestion
  return envoy_headers{length, header_array};
```

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -29,13 +117,33 @@ - (void)dealloc {
   [[NSNotificationCenter defaultCenter] removeObserver:self];
 }
 
+- (int)registerFilter:(EnvoyHTTPFilter *)filter {
+  // TODO(goaway): Everything here leaks, but it's all be tied to the life of the engine.
+  // This will need to be updated for https://github.com/lyft/envoy-mobile/issues/332
```
rebello95, line 122: Assuming my comment above about providing an instance of each filter for each request, I think this will become more problematic, no?

library/swift/test/StreamClientBuilderTests.swift
```diff
@@ -172,6 +172,7 @@ final class StreamClientBuilderTests: XCTestCase {
                                     dnsRefreshSeconds: 300,
                                     dnsFailureRefreshSecondsBase: 400,
                                     dnsFailureRefreshSecondsMax: 500,
+                                    filterChain: [],
```
rebello95, line 175: We should add tests for this new functionality

library/swift/src/filters/ResponseFilter.swift
```diff
@@ -1,7 +1,28 @@
+@_implementationOnly import EnvoyEngine
 import Foundation
 
+extension EnvoyHTTPFilter {
+  /// Initializer
```
rebello95, line 5: I'd either remove this or provide an actual description

library/swift/src/filters/ResponseFilter.swift
```diff
@@ -1,7 +1,28 @@
...
+  convenience init(responseFilter: ResponseFilter) {
+    self.init()
+    self.name = responseFilter.name
+    self.onResponseHeaders = {
+      let result = responseFilter.onResponseHeaders(ResponseHeaders(headers: $0), endStream: $1)
+      switch result {
+      case .continue(let headers):
+        return [0, headers.headers]
+      case .stopIteration(let headers):
+        return [1, headers.headers]
```
rebello95, line 15: Can we name these instead of using 0/1?

library/swift/src/filters/ResponseFilter.swift
```diff
@@ -1,7 +1,28 @@
+@_implementationOnly import EnvoyEngine
 import Foundation
 
+extension EnvoyHTTPFilter {
+  /// Initializer
+  convenience init(responseFilter: ResponseFilter) {
+    self.init()
+    self.name = responseFilter.name
+    self.onResponseHeaders = {
+      let result = responseFilter.onResponseHeaders(ResponseHeaders(headers: $0), endStream: $1)
```
rebello95, line 10: nit: let's name these parameters in the opening of the closure

library/swift/src/filters/RequestFilter.swift
```diff
@@ -1,7 +1,28 @@
...
+      let result = requestFilter.onRequestHeaders(RequestHeaders(headers: $0), endStream: $1)
+      switch result {
+      case .continue(let headers):
+        return [0, headers.headers]
+      case .stopIteration(let headers):
+        return [1, headers.headers]
+      }
+    }
+  }
+}
```
rebello95, line 19: Same comments here as below

library/swift/src/filters/RequestFilter.swift
```diff
@@ -1,7 +1,28 @@
+@_implementationOnly import EnvoyEngine
 import Foundation
 
+extension EnvoyHTTPFilter {
+  /// Initializer
+  convenience init(requestFilter: RequestFilter) {
```
rebello95, line 6: How will this work for filters that want to act as both request and response filters? As-is, it doesn't look like that's possible since the callbacks are only set on `EnvoyHTTPFilter.init`

library/swift/src/StreamClientBuilder.swift
```diff
@@ -107,6 +108,28 @@ public final class StreamClientBuilder: NSObject {
...
+  /// Add HTTP filter for resonses received by this client.
+  ///
+  /// - parameter filter: RequestFilter to be invoked for streams.
+  ///
+  /// - returns: This builder.
+  @discardableResult
+  public func addFilter(_ filter: ResponseFilter) -> StreamClientBuilder {
+    self.filterChain.append(EnvoyHTTPFilter(responseFilter: filter))
+    return self
+  }
```
rebello95, line 131: 1. It looks like this prevents filters from being able to act as both `RequestFilter` and `ResponseFilter` conformers. The idea behind the empty `Filter` protocol in place is to allow registration of a filter, then call its request/response functionality depending on what it conforms to. Should we be doing something like that here?
2. Per comments above, I think these should accept closures that create the filters on each request/stream, rather than instances of the filters.

examples/swift/hello_world/ViewController.swift
```diff
@@ -6,6 +6,29 @@ private let kRequestAuthority = "api.lyft.com"
 private let kRequestPath = "/ping"
 private let kRequestScheme = "https"
 
+struct PlatformDemoFilter: ResponseFilter {
```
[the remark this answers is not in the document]
goaway: Sure, sounds good.

examples/swift/hello_world/ViewController.swift
```diff
@@ -15,7 +38,7 @@ final class ViewController: UITableViewController {
     super.viewDidLoad()
     do {
       NSLog("starting Envoy...")
-      self.client = try StreamClientBuilder().build()
+      self.client = try StreamClientBuilder().addFilter(PlatformDemoFilter()).build()
```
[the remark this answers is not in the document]
goaway: You're right, the current implementation uses a single long-lived filter instance, which is a departure from upstream Envoy. This approach is definitely simpler to map, but I agree that aligning with upstream Envoy would be nice. Let me think a little bit about what it would take to make filter instances per-stream.

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -10,6 +11,93 @@ static void ios_on_exit() {
...
+    [headerValueList addObject:headerValue];
+  }
+  // TODO(goaway): consider solution that doesn't violate release convention
+  // Note: We don't call release_envoy_headers because they may not be modified by the filter
+  return headerDict;
+}
+
+static envoy_headers toNativeHeaders(EnvoyHeaders *headers) {
+  envoy_header_size_t length = 0;
+  for (id headerKey in headers) {
```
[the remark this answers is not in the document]
goaway: Yep, you're totally right.

examples/swift/hello_world/ViewController.swift
```diff
@@ -6,6 +6,29 @@ private let kRequestAuthority = "api.lyft.com"
 private let kRequestPath = "/ping"
 private let kRequestScheme = "https"
 
+struct PlatformDemoFilter: ResponseFilter {
+  let name = "PlatformStub"
```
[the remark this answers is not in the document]
goaway: Sure thing.

library/swift/src/filters/ResponseFilter.swift
goaway: Yeah, we should. We don't have a great solution from enums across the bridge layer yet, and I had this in place while I thought about it more. But we should at least make them some kind of named constant.

library/swift/src/filters/ResponseFilter.swift
goaway: sgtm

library/swift/src/filters/RequestFilter.swift
goaway: Already by the objective-c mapping layer (and everything below) it's actually unified into a bi-directional solution. It's only bifurcated here because it wasn't clear to me here how we'd work with the two disparate interfaces, and I decided to punt on the problem since it wasn't clear to me that the intent was that we'd have filters that implemented both.

I think I can make things work to allow bi-directional filters with a little tweaking. Since we don't have any usage that necessitates it yet, my slight inclination is to punt on it so we can get something functional onto main and work from there, but I'm open to your thoughts.

library/swift/src/StreamClientBuilder.swift
goaway: 1. Agreed, see my response above. With a little bit of reflection I think we can achieve that, so I'm open to working that out here, or in a subsequent PR, though my slight preference is for the latter.
2. I agree, though I have to think a little bit more about the ramifications to the API and mapping layer. I'm forming an idea of how to do it though - it should be possible.

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -10,6 +11,93 @@ static void ios_on_exit() {
...
+  envoy_header_size_t header_index = 0;
+  for (id headerKey in headers) {
+    NSArray *headerList = headers[headerKey];
+    for (id headerValue in headerList) {
+      envoy_header new_header = {toManagedNativeString(headerKey),
+                                 toManagedNativeString(headerValue)};
+      header_array[header_index++] = new_header;
+    }
+  }
+  // TODO: ASSERT(header_index == length);
```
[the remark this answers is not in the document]
goaway: We haven't actually (at present) introduced a dependency on Envoy's assertion library at this layer.

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -10,6 +11,93 @@ static void ios_on_exit() {
   NSLog(@"[Envoy] library is exiting");
 }
 
+typedef struct {
+  __unsafe_unretained EnvoyHTTPFilter *filter;
+} ios_http_filter_context;
+
+// TODO(goaway): refactor and relocate translation
```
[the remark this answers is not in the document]
goaway: There's some (egregious) duplication with some of our other mapping code here. It's a little bit tricky to reconcile, because the retain semantics are slightly different. It sure would be nice to though.

library/objective-c/EnvoyEngineImpl.m
goaway: Sure, can't remember why I didn't inline this.

library/objective-c/EnvoyEngineImpl.m
goaway: Yes. :)

It will need to be solved if/when we go that route.

library/swift/test/StreamClientBuilderTests.swift
goaway: Agreed. To some degree the example app acts as an end-to-end integration test (and I actually plan to update it with n explicit check on the filter's effect). But I can look into what level of testing we can do at the unit layer here as well.

library/swift/src/filters/ResponseFilter.swift
goaway: Sure, I'll elaborate.

examples/swift/hello_world/ViewController.swift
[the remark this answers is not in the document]
rebello95: 👍 I think this is actually a hard requirement for filters in Envoy Mobile. I know a lot of the use cases Lyft has for platform filters today require filters to be per-stream and to act as both request and response filters

library/objective-c/EnvoyEngineImpl.m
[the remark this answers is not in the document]
rebello95: Sounds good, maybe make this TODO a bit more descriptive so it doesn't lose context over time

library/objective-c/EnvoyEngineImpl.m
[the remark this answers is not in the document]
rebello95: Got it. If we want, we could use the native `assert`, but no strong opinion for now I guess

library/swift/src/filters/RequestFilter.swift
rebello95: I don't necessarily want to block the PR on this, but I do think it's required functionality for using these filters in production, and am a little concerned that if we don't fix it early on it'll be more difficult to do as we go further down this route. Thoughts?

library/objective-c/EnvoyEngineImpl.m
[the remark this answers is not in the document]
goaway: For sure, we definitely can - we just haven't decided on a convention for this layer yet.

library/objective-c/EnvoyEngineImpl.m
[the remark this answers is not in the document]
junr03: Isn't it the same as https://github.com/lyft/envoy-mobile/blob/2d86a1c052ec97852ac8161c470649290310fde7/library/objective-c/EnvoyHTTPStreamImpl.m#L21?

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -10,6 +11,91 @@ static void ios_on_exit() {
...
+                                                     length:header.value.length
+                                                   encoding:NSUTF8StringEncoding];
+    NSMutableArray *headerValueList = headerDict[headerKey];
+    if (headerValueList == nil) {
+      headerValueList = [NSMutableArray new];
+      headerDict[headerKey] = headerValueList;
+    }
+    [headerValueList addObject:headerValue];
+  }
+  // TODO(goaway): consider solution that doesn't violate release convention
```
junr03, line 43: I feel like the user can determine if they are going to modify or not, and this function should release on copy? Or conversely, change the existing API and have the caller bear the responsibility of releasing? In any case, I think we can unify these helpers with the ones in EnvoyHTTPStreamImpl

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -10,6 +11,91 @@ static void ios_on_exit() {
...
+  // TODO(goaway): optimize unmodified case
+  ios_http_filter_context *c = (ios_http_filter_context *)context;
+  if (c->filter.onRequestHeaders == nil) {
+    return (envoy_filter_headers_status){/*status*/ 0, /*headers*/ headers};
+  }
+
+  EnvoyHeaders *platformHeaders = to_ios_headers(headers);
+  release_envoy_headers(headers);
+  // TODO(goaway): consider better solution for compound return
+  NSArray *result = c->filter.onRequestHeaders(platformHeaders, end_stream);
```
junr03, line 78: Is there no tuple type?

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -29,13 +115,33 @@ - (void)dealloc {
   [[NSNotificationCenter defaultCenter] removeObserver:self];
 }
 
+- (int)registerFilter:(EnvoyHTTPFilter *)filter {
+  // TODO(goaway): Everything here leaks, but it's all be tied to the life of the engine.
+  // This will need to be updated for https://github.com/lyft/envoy-mobile/issues/332
+  ios_http_filter_context *context = safe_malloc(sizeof(ios_http_filter_context));
```
junr03, line 121: following on @rebello95's comment about lifetime. We definitely want a per stream instance of the full chain, like Envoy. 

I wonder if what we can do is: have the api retain a filter factory in its context, and register that. Then the registry can instantiate a filter chain, that the stream retains, and passes that in the context to the static `ios_http_filter_on_*_*` functions. So we would have two different contexts, and the one here (which retains the factory) leaks, but will eventually be tied/retained to/by the engine instance.

library/swift/src/StreamClientBuilder.swift
junr03: re: 2. we should definitely do this per stream instance. And re: 1. do we really need reflection? Can't we have multiple inheritance and have a filter class that is both, and thus calls both request and response addFilter functions as in envoy: 

https://github.com/envoyproxy/envoy/blob/bc02b7748a69fd3854f74b9e5bec236b5499f7fd/include/envoy/http/filter.h#L772

https://github.com/envoyproxy/envoy/blob/0c6523c0925d260de41d89002044d0d6271550d0/source/common/http/conn_manager_impl.h#L553

library/objective-c/EnvoyEngineImpl.m
goaway: The convention I feel we should maintain is that the callee always releases or passes that responsibility. I don't think it's really possible to invert that, because there's no way for the caller (or returner) to otherwise determine what the callee does with it. This is true for calls going in either direction across the bridge layer.

While I do think we can reconcile the differences here, I added it as a TODO because it felt a little bit extraneous to getting the core functionality in place.

library/objective-c/EnvoyEngineImpl.m
goaway: There is at the core layer, and at the platform layer. I skipped making yet another implementation in between. - While we could re-use the bridge/core type on iOS technically, it will then deviate from what we'll need on Android. It also comes back down to that we don't really have a great way to share enums in a consistent way either - something I would like to come up with a better solution for as well.

library/objective-c/EnvoyEngineImpl.m
goaway: That's one of the options that I've been considering, but there's some added complexity with the memory management of the instance. We'll need a dealloc hook as well, and we need to guard against the platform implementation living longer than the underlying core instantiation (and more specifically, trying to interact with it after it's gone).

I've actually been leaning towards a more integrated approach - with the API presenting both an `init()` and a `release()` function, the former of which will populate the rest of the function pointers on the struct, and the latter of which will nullify them (and trigger memory-safe cleanup).

goaway: Per our discussion Friday, my current plan is to finish addressing other issues here, land this and build per-stream instantiation on top.

library/swift/src/StreamClientBuilder.swift
goaway: I believe reflection is the way to go, yes. It's true a class can implement multiple interfaces, but even if we declare an interface that conforms to both `RequestFilter` and `ResponseFilter`, Swift will not implicitly or automatically cause classes that implement both interfaces to conform to it. So when we go to instantiate bridge type from a generic `Filter`, we will need to check which interfaces it implements. If we create type-specialized constructors (like I had here), it will be ambiguous to the compiler which it should call.

That's the best explanation I can provide here. Happy to discuss more offline.

library/swift/src/filters/RequestFilter.swift
goaway: Just to capture discussion from Friday here, I will be adding support for bi-directionality prior to merging.

library/swift/src/StreamClientBuilder.swift
goaway: For bi-directionality, see the solution I've pushed, and let me know if you have additional thoughts. :)

library/swift/src/filters/RequestFilter.swift
goaway: Pushed a solution, let me know if you have further thoughts.

library/objective-c/EnvoyEngineImpl.m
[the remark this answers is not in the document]
goaway: The bridge semantics are a little different; see my comments below. Happy to talk about it more. My inclination is to leave this as a TODO (with extra description per @rebello95's suggestion, and work on a better solution separately. To provide a little bit more context, I'm considering if we need an actual refcount on our structs. It's kind of a bridge I'd like to avoid crossing if possible, but if we're going to avoid it, I do feel it's important to be strict about our release expectations - which we can't yet reconcile here.

library/objective-c/EnvoyEngineImpl.m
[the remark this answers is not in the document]
goaway: Note a native `assert` won't get compiled out by Envoy's build directives, so it will always crash the app if we hit it. Not saying it's the right or wrong thing to do, but it's just something to consider.

examples/swift/hello_world/ViewController.swift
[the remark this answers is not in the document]
goaway: Per discussion Friday, we'll do this on top of this PR.

library/objective-c/EnvoyEngineImpl.m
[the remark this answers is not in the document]
goaway: Updated TODO. Note that some of these TODOs I'd like to address pretty much immediately after core functionality ships. However, as long as they don't affect the public interfaces that Android and iOS applications adhere to, I'm biasing towards getting that functionality shipped so that the apps can be updated independent of under-the-hood improvements. As always, open to discussing.

library/swift/src/filters/ResponseFilter.swift
goaway: Okay, I came up with a solution, which is "don't use enums." Using enums, we'd have to declare and map this set of options 4 times over: first, in Envoy, next at the bridge layer (because Envoy's type is c++ and can't be directly exposed), next in objective-c (because the EnvoyEngine header can't have includes due to the clang issue), next in Swift (because the objective-c type doesn't support associated values). We'd also have to create mapping functions between all 4 variants.

`extern  const`s can be declared in terms of Envoy's implicit internal values and eliminate all but one mapping.

goaway: (updated)

library/objective-c/EnvoyEngine.h
```diff
@@ -7,6 +7,9 @@ NS_ASSUME_NONNULL_BEGIN
 /// A set of headers that may be passed to/from an Envoy stream.
 typedef NSDictionary<NSString *, NSArray<NSString *> *> EnvoyHeaders;
 
+/// A mutable set of headers that may be passed to/from an Envoy stream.
+typedef NSMutableDictionary<NSString *, NSArray<NSString *> *> EnvoyMutableHeaders;
```
rebello95, line 11: Shouldn't this be an `NSMutableArray` as the values?

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -29,13 +119,33 @@ - (void)dealloc {
...
+  CFBridgingRetain(filter);
+  context->filter = filter;
+  envoy_http_filter *api = safe_malloc(sizeof(envoy_http_filter));
+  api->on_request_headers = ios_http_filter_on_request_headers;
+  api->on_request_data = NULL;
+  api->on_response_headers = ios_http_filter_on_response_headers;
+  api->on_response_data = NULL;
+  api->context = context;
+  register_platform_api(filter.name.UTF8String, api);
+  return 0;
```
rebello95, line 135: `ENVOY_SUCCESS`?

library/swift/src/StreamClientBuilder.swift
```diff
@@ -107,6 +108,17 @@ public final class StreamClientBuilder: NSObject {
     return self
   }
 
+  /// Add HTTP filter for requests sent by this client.
+  ///
+  /// - parameter filter: RequestFilter to be invoked for streams.
+  ///
+  /// - returns: This builder.
+  @discardableResult
+  public func addFilter(_ filter: Filter) -> StreamClientBuilder {
```
rebello95, line 117: Maybe add a todo here since we'll be switching from instance-based filters in a future PR?

library/objective-c/EnvoyEngineImpl.m
goaway: We don't have the type available on the header here, so we've only used `0` and `1` in this file so far. The extern-based solution I propose in this PR for filter status would work for this as well though, so maybe I can fix this for all the outstanding cases in a subsequent PR?

library/objective-c/EnvoyEngine.h
goaway: Sure, sounds good. I was trying to avoid unnecessary reallocations when converting from the immutable type, but honestly I wouldn't be surprised if Apple had a copy-on-write optimization under the hood anyways.

library/swift/src/StreamClientBuilder.swift
goaway: Sure thing.

library/objective-c/EnvoyEngineImpl.m
rebello95: SGTM

PR #947, opened 2020-07-09
android: only build aar for android_dist target

buildbreaker:

Description: android: only build aar for android_dist target
Risk Level: low
Testing: ci
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -26,12 +26,18 @@ alias(
 )
 
 genrule(
-    name = "android_zip",
+    name = "android_dist_ci",
     srcs = [
         "//library/kotlin/src/io/envoyproxy/envoymobile:android_aar",
     ],
     outs = ["envoy_mobile.zip"],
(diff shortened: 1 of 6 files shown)


BUILD
```diff
@@ -40,14 +46,12 @@ genrule(
 genrule(
     name = "android_dist",
     srcs = [
-        "//library/kotlin/src/io/envoyproxy/envoymobile:android_aar",
+        "//library/kotlin/src/io/envoyproxy/envoymobile:android_aar_only_aar",
```
junr03, line 49: I think this name is a little funny. I would presume that an `android_aar` target would _only_ build the aar. i.e, I would expect the target that builds more to be the one that has a suffix: `envoy_aar_with_javadocs`?

buildbreaker: Hmm...How about `android_artifacts` since it generates other things too

PR #952, opened 2020-07-14
Test updating Envoy SHA

rebello95:


rebello95:
Android built successfully on macOS and Linux

PR #963, opened 2020-07-17
filters: remove unsafe optimization

goaway:
Description: The assumption underpinning the optimization is unsound and results in non-determinism. Added TODO for safe approach.
Risk Level: Low
Testing: Local end-to-end


goaway:
Thanks @rebello95!

PR #967, opened 2020-07-22
filters: instantiate filters on a per-stream basis

goaway:
Description: Changes filter instantiation to be a per-stream basis, across the core, bridge, and iOS platform layers. Also improves usage of CFBridge* functions to be better in line with their intent.
Risk Level: Moderate
Testing: Local end-to-end

### DIFF
--- a/library/common/extensions/filters/http/platform_bridge/filter.cc
+++ b/library/common/extensions/filters/http/platform_bridge/filter.cc
@@ -28,7 +28,7 @@ PlatformBridgeFilter::PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr c
 }
 
 void PlatformBridgeFilter::onDestroy() {
-  ASSERT(platform_filter_.instance_context, "init_filter must be called initially");
+  ASSERT(platform_filter_.instance_context, "expected initialized filter context");
 
   platform_filter_.release_filter(platform_filter_.instance_context);
   platform_filter_.instance_context = nullptr;
@@ -36,7 +36,7 @@ void PlatformBridgeFilter::onDestroy() {
 
 Http::FilterHeadersStatus PlatformBridgeFilter::onHeaders(Http::HeaderMap& headers, bool end_stream,
                                                           envoy_filter_on_headers_f on_headers) {
-  ASSERT(platform_filter_.instance_context, "init_filter must be called initially");
+  ASSERT(platform_filter_.instance_context, "expected initialized filter context");
 
   // Allow nullptr to act as (optimized) no-op.
   if (on_headers == nullptr) {
--- a/library/common/extensions/filters/http/platform_bridge/filter.cc
+++ b/library/common/extensions/filters/http/platform_bridge/filter.cc
@@ -21,10 +21,23 @@ PlatformBridgeFilterConfig::PlatformBridgeFilterConfig(
           Api::External::retrieveApi(proto_config.platform_filter_name()))) {}
 
 PlatformBridgeFilter::PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr config)
-    : platform_filter_(config->platform_filter()) {}
+    : platform_filter_(*config->platform_filter()) {
+  platform_filter_.instance_context = platform_filter_.init_filter(platform_filter_.static_context);
+
+  ASSERT(platform_filter_.instance_context, "init_filter unsuccessful");
+}
+
+void PlatformBridgeFilter::onDestroy() {
+  ASSERT(platform_filter_.instance_context, "init_filter must be called initially");
+
+  platform_filter_.release_filter(platform_filter_.instance_context);
+  platform_filter_.instance_context = nullptr;
+}
 
 Http::FilterHeadersStatus PlatformBridgeFilter::onHeaders(Http::HeaderMap& headers, bool end_stream,
                                                           envoy_filter_on_headers_f on_headers) {
+  ASSERT(platform_filter_.instance_context, "init_filter must be called initially");
+
   // Allow nullptr to act as (optimized) no-op.
   if (on_headers == nullptr) {
     return Http::FilterHeadersStatus::Continue;
@@ -32,7 +45,7 @@ Http::FilterHeadersStatus PlatformBridgeFilter::onHeaders(Http::HeaderMap& heade
 
   envoy_headers in_headers = Http::Utility::toBridgeHeaders(headers);
   envoy_filter_headers_status result =
-      on_headers(in_headers, end_stream, platform_filter_->context);
+      on_headers(in_headers, end_stream, platform_filter_.instance_context);
   Http::FilterHeadersStatus status = static_cast<Http::FilterHeadersStatus>(result.status);
   // TODO(goaway): Current platform implementations expose immutable headers, thus any modification
   // necessitates a full copy. Add 'modified' bit to determine when we can elide the copy. See also
@@ -73,7 +86,7 @@ Http::FilterDataStatus PlatformBridgeFilter::onData(Buffer::Instance& data, bool
 Http::FilterHeadersStatus PlatformBridgeFilter::decodeHeaders(Http::RequestHeaderMap& headers,
                                                               bool end_stream) {
   // Delegate to shared implementation for request and response path.
-  return onHeaders(headers, end_stream, platform_filter_->on_request_headers);
+  return onHeaders(headers, end_stream, platform_filter_.on_request_headers);
 }
 
 Http::FilterDataStatus PlatformBridgeFilter::decodeData(Buffer::Instance& data, bool end_stream) {
@@ -98,7 +111,7 @@ PlatformBridgeFilter::encode100ContinueHeaders(Http::ResponseHeaderMap& /*header
 Http::FilterHeadersStatus PlatformBridgeFilter::encodeHeaders(Http::ResponseHeaderMap& headers,
                                                               bool end_stream) {
   // Delegate to shared implementation for request and response path.
-  return onHeaders(headers, end_stream, platform_filter_->on_response_headers);
+  return onHeaders(headers, end_stream, platform_filter_.on_response_headers);
 }
 
 Http::FilterDataStatus PlatformBridgeFilter::encodeData(Buffer::Instance& data, bool end_stream) {
--- a/library/common/extensions/filters/http/platform_bridge/filter.cc
+++ b/library/common/extensions/filters/http/platform_bridge/filter.cc
@@ -22,13 +22,31 @@ PlatformBridgeFilterConfig::PlatformBridgeFilterConfig(
 
 PlatformBridgeFilter::PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr config)
     : platform_filter_(*config->platform_filter()) {
+  // The initialization above sets platform_filter_ to a copy of the struct stored on the config.
+  // In the typical case, this will represent a filter implementation that needs to be intantiated.
+  // static_context will contain the necessary platform-specific mechanism to produce a filter
+  // instance. instance_context will initially be null, but after initialization, set to the
+  // context needed for actual filter invocations.
+
+  // We treat a null initializer as a static implementation.
+  if (platform_filter_.init_filter == nullptr) {
+    // Make static context available to invocations, in lieu of instatiated context.
+    platform_filter_.instance_context = platform_filter_.static_context;
+    return;
+  }
+
+  // Set the instance_context to the result of the initialization call. Cleanup will ultimately
+  // occur during in the onDestroy() invocation below.
   platform_filter_.instance_context = platform_filter_.init_filter(platform_filter_.static_context);
 
   ASSERT(platform_filter_.instance_context, "init_filter unsuccessful");
 }
 
 void PlatformBridgeFilter::onDestroy() {
-  ASSERT(platform_filter_.instance_context, "expected initialized filter context");
+  // Allow nullptr to act as no-op. Also return if nothing was initialized.
+  if (platform_filter_.release_filter == nullptr || platform_filter_.init_filter == nullptr) {
+    return;
+  }
 
   platform_filter_.release_filter(platform_filter_.instance_context);
   platform_filter_.instance_context = nullptr;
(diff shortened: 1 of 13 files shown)


library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -20,18 +20,31 @@ PlatformBridgeFilterConfig::PlatformBridgeFilterConfig(
...
 PlatformBridgeFilter::PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr config)
-    : platform_filter_(config->platform_filter()) {}
+    : platform_filter_(*config->platform_filter()) {
+  platform_filter_.instance_context = platform_filter_.init_filter(platform_filter_.static_context);
+
+  ASSERT(platform_filter_.instance_context, "init_filter unsuccessful");
+}
+
+void PlatformBridgeFilter::onDestroy() {
+  ASSERT(platform_filter_.instance_context, "init_filter must be called initially");
```
rebello95, line 30:
```suggestion
  ASSERT(platform_filter_.instance_context, "expected initialized filter context");
```

rebello95: Same below. I think this is a bit clearer since technically `init_filter` can fail per the above assert

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -66,40 +62,51 @@ static envoy_headers toNativeHeaders(EnvoyHeaders *headers) {
...
   // TODO(goaway): consider better solution for compound return
-  NSArray *result = c->filter.onResponseHeaders(platformHeaders, end_stream);
+  NSArray *result = filter.onResponseHeaders(platformHeaders, end_stream);
   return (envoy_filter_headers_status){/*status*/ [result[0] intValue],
                                        /*headers*/ toNativeHeaders(result[1])};
 }
 
+static void ios_http_filter_release(const void *context) {
+  CFRelease(context);
+  return;
```
rebello95, line 107: Is this necessary?

library/swift/src/filters/Filter.swift
```diff
@@ -4,14 +4,33 @@ import Foundation
...
+
+  /// Required initializer for internal creation.
+  init()
+}
+
+extension Filter {
+  static func create() -> EnvoyHTTPFilter {
+    return EnvoyHTTPFilter(filter: Self.init())
+  }
+}
```
rebello95, line 17: I'd probably remove this and just it below using `filterType.init()`

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: Sounds good.

library/objective-c/EnvoyEngineImpl.m
goaway: The `return` statement? Technically no, but I guess I tend to err on the side of being explicit.

library/swift/src/filters/Filter.swift
goaway: Yep, good catch - this is leftover from an earlier approach.

library/objective-c/EnvoyEngine.h
```diff
@@ -68,14 +68,22 @@ extern const int kEnvoyFilterHeadersStatusStopAllIterationAndBuffer;
...
 
 @end
 
+#pragma mark - EnvoyHTTPFilterFactory
+
+@interface EnvoyHTTPFilterFactory : NSObject
+
+@property (nonatomic, strong) NSString *filterName;
+
+@property (nonatomic, strong) EnvoyHTTPFilter * (^create)();
```
junr03, line 83: should we use similar verbs between here and the bridge (init)?

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -66,40 +62,51 @@ static envoy_headers toNativeHeaders(EnvoyHeaders *headers) {
...
+  return CFBridgingRetain(filter);
+}
+
 static envoy_filter_headers_status
-ios_http_filter_on_request_headers(envoy_headers headers, bool end_stream, void *context) {
+ios_http_filter_on_request_headers(envoy_headers headers, bool end_stream, const void *context) {
   // TODO(goaway): optimize unmodified case
-  ios_http_filter_context *c = (ios_http_filter_context *)context;
-  if (c->filter.onRequestHeaders == nil) {
+  EnvoyHTTPFilter *filter = (__bridge EnvoyHTTPFilter *)context;
```
junr03, line 74: what does the `__bridge` in the cast mean?

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -66,40 +62,51 @@ static envoy_headers toNativeHeaders(EnvoyHeaders *headers) {
...
+  if (filter.onRequestHeaders == nil) {
```
junr03, line 75: how come the pointer doen't need dereference here?

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -66,40 +62,51 @@ static envoy_headers toNativeHeaders(EnvoyHeaders *headers) {
...
   // TODO(goaway): consider better solution for compound return
-  NSArray *result = c->filter.onRequestHeaders(platformHeaders, end_stream);
+  NSArray *result = filter.onRequestHeaders(platformHeaders, end_stream);
   return (envoy_filter_headers_status){/*status*/ [result[0] intValue],
                                        /*headers*/ toNativeHeaders(result[1])};
 }
 
 static envoy_filter_headers_status
-ios_http_filter_on_response_headers(envoy_headers headers, bool end_stream, void *context) {
+ios_http_filter_on_response_headers(envoy_headers headers, bool end_stream, const void *context) {
```
junr03, line 89: I think we talked about it in one of the earlier PRs, but could we delete the duplication and dispatch to common code between request and reply?

goaway: I did delete the (more significant) duplication on that PR (in filter.cc). Here, I'd suggest it's pretty limited. We check if the function's present, map the headers, call the function. I almost feel the extra indirection from factoring here would serve only to complicate the code.

goaway: But I mean, only just almost? If you feel strongly, I can still do it.

library/objective-c/EnvoyEngineImpl.m
goaway: Yeah, seems weird if you've been looking at C++ for a while. Technically speaking we're using shorthand to send an Objective-C message to `filter` - the expanded syntax would be `[filter onRequestHeaders]`. This just happens to be what the shorthand looks like.

library/objective-c/EnvoyEngineImpl.m
goaway: `__bridge` tells ARC that we're pulling out a retainable object here, but not to actually do anything with the retain count at this point.

library/objective-c/EnvoyEngine.h
goaway: I presented the interface to the bridge layer as something like a class/instance, so `init` made sense. I implemented it at this layer as a split between a factory and an implementation, so `create` made sense.

At least, to me.

library/common/extensions/filters/http/platform_bridge/filter.h
```diff
@@ -50,7 +53,7 @@ class PlatformBridgeFilter final : public Http::PassThroughFilter {
 private:
   Http::FilterHeadersStatus onHeaders(Http::HeaderMap& headers, bool end_stream,
                                       envoy_filter_on_headers_f on_headers);
-  const envoy_http_filter* platform_filter_;
+  envoy_http_filter platform_filter_;
```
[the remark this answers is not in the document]
goaway: Good call.

junr03:
lgtm, pending CI

goaway:
The objective-c app was failing because the no-op filter couldn't be instantiated (of course). I realized this was effectively a one-off case of a static/stateless filter, and provided a simple mechanism to continue to support that option. I also added additional comments generally describing the core implementation, per your suggestion @junr03.

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -21,18 +21,47 @@ PlatformBridgeFilterConfig::PlatformBridgeFilterConfig(
...
 PlatformBridgeFilter::PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr config)
-    : platform_filter_(config->platform_filter()) {}
+    : platform_filter_(*config->platform_filter()) {
+  // The initialization above sets platform_filter_ to a copy of the struct stored on the config.
+  // In the typical case, this will represent a filter implementation that needs to be intantiated.
+  // static_context will contain the necessary platform-specific mechanism to produce a filter
+  // instance. instance_context will initially be null, but after initialization, set to the
+  // context needed for actual filter invocations.
+
+  // We treat a null initializer as a static implementation.
```
junr03, line 31: Is this a feature we want to support? i.e do we want it to be possible to have static filters? While potentially useful in some cases I worry that giving options might make the entire affair a little more confusing.

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -21,18 +21,47 @@ PlatformBridgeFilterConfig::PlatformBridgeFilterConfig(
...
+    // Make static context available to invocations, in lieu of instatiated context.
+    platform_filter_.instance_context = platform_filter_.static_context;
+    return;
+  }
+
+  // Set the instance_context to the result of the initialization call. Cleanup will ultimately
+  // occur during in the onDestroy() invocation below.
+  platform_filter_.instance_context = platform_filter_.init_filter(platform_filter_.static_context);
+
+  ASSERT(platform_filter_.instance_context, "init_filter unsuccessful");
```
junr03, line 42: can we interpolate the filter name here? Makes for easier crash debugging

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -21,18 +21,47 @@ PlatformBridgeFilterConfig::PlatformBridgeFilterConfig(
...
+}
+
+void PlatformBridgeFilter::onDestroy() {
+  // Allow nullptr to act as no-op. Also return if nothing was initialized.
```
junr03, line 46: I liked being able to tell that reaching a nullptr here was a programming mistake.

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -21,18 +21,47 @@ PlatformBridgeFilterConfig::PlatformBridgeFilterConfig(
           Api::External::retrieveApi(proto_config.platform_filter_name()))) {}
 
 PlatformBridgeFilter::PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr config)
-    : platform_filter_(config->platform_filter()) {}
+    : platform_filter_(*config->platform_filter()) {
+  // The initialization above sets platform_filter_ to a copy of the struct stored on the config.
```
junr03, line 25: Awesome this is great.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: The thing is that as long as Envoy adheres to its contract about filter callbacks, the assertion in `init_filter` will have already been hit.

I decided the extra assertions really amounted to asserting Envoy adheres to its contract, which is better done by Envoy's own coverage.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: I'm primarily interested in supporting it right now because I like the convention of being able to `calloc` these structs as a safety convention. I wasn't planning on exposing it above this level at this point.

I agree about not introducing extra complexity, but I also realized it didn't take very much to support it generally here. Assuming we move forward, I will ensure a test covers this case.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: Sure.

library/common/extensions/filters/http/platform_bridge/filter.cc
rebello95: I agree with @junr03 that I'd prefer we stick with one approach (instance-based filters) rather than supporting static implementations (even if it's minimal work to add that support). Even if it's not something we bubble up to the public interface, it's still something we'd be implicitly supporting internally and subject to breakage, and I'm not sure I see much value in providing that functionality

rebello95: I'm also a little concerned that having this in place might lead us to miss actual problems with instance-based filters (i.e., `init_filter` being accidentally unset)

library/common/extensions/filters/http/platform_bridge/filter.cc
rebello95: As I alluded above, I also think that it'd be better to treat `nullptr` as a programming error here rather than allowing for static filters

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: Talked to @rebello95 and upon further consideration, I think the two of you make a good point and we should prefer simplicity. I'll switch this over to logging if init_filter isn't provided, and nulling out the rest of the struct for safety.

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -17,34 +17,37 @@ namespace PlatformBridge {
...
   // static_context will contain the necessary platform-specific mechanism to produce a filter
   // instance. instance_context will initially be null, but after initialization, set to the
   // context needed for actual filter invocations.
 
-  // We treat a null initializer as a static implementation.
+  // If init_filter is missing, zero out the rest of the struct for safety.
   if (platform_filter_.init_filter == nullptr) {
-    // Make static context available to invocations, in lieu of instatiated context.
-    platform_filter_.instance_context = platform_filter_.static_context;
+    ENVOY_LOG(debug, "platform bridge filter: missing initializer for {}", filter_name_);
```
rebello95, line 34: Any reason not to make this an `ASSERT`?

goaway: Yeah, calloc as no-op still has way less overhead than an empty implementation. This restricts handling to only that case though (as opposed to the previous general notion of static support).

PR #971, opened 2020-07-23
filters: wire data and trailers for iOS

goaway:
Description: Completes end-to-end functionality for bridged platform filters on iOS. Factors out duplication in objective-c bridging functions.
Risk Level: Moderate
Testing: Local end-to-end

### DIFF
--- a/examples/swift/hello_world/DemoFilter.swift
+++ b/examples/swift/hello_world/DemoFilter.swift
@@ -17,6 +17,7 @@ struct DemoFilter: ResponseFilter {
   func setResponseFilterCallbacks(_ callbacks: ResponseFilterCallbacks) {}
 
   func onResponseData(_ body: Data, endStream: Bool) -> FilterDataStatus {
+    // TODO(goaway): Can remove this when we have better integration coverage in place.
     NSLog("Saw data chunk of length \(body.count)")
     return .continue(body)
   }
--- a/examples/swift/hello_world/DemoFilter.swift
+++ b/examples/swift/hello_world/DemoFilter.swift
@@ -17,6 +17,7 @@ struct DemoFilter: ResponseFilter {
   func setResponseFilterCallbacks(_ callbacks: ResponseFilterCallbacks) {}
 
   func onResponseData(_ body: Data, endStream: Bool) -> FilterDataStatus {
+    NSLog("Saw data chunk of length \(body.count)")
     return .continue(body)
   }
 
--- a/library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
@@ -43,5 +43,6 @@ sealed class FilterDataStatus {
    *
    * This may be called by filters which must parse a larger block of the incoming data before
    * continuing processing, and will handle their own buffering.
+   */
   class StopIterationNoBuffer : FilterDataStatus()
 }
--- a/library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
@@ -30,4 +30,18 @@ sealed class FilterDataStatus {
    * continuing processing.
    */
   class StopIterationAndBuffer(val data: ByteBuffer) : FilterDataStatus()
+
+  /**
+   * Do not iterate to any of the remaining filters in the chain, and do not internally buffer
+   * data.
+   *
+   * This filter will continue to be called with new chunks of data.
+   *
+   * Returning `Continue` from `onRequestData()`/`onResponseData()` or calling
+   * `continueRequest()`/`continueResponse()` MUST be called when continued filter iteration is
+   * desired.
+   *
+   * This may be called by filters which must parse a larger block of the incoming data before
+   * continuing processing, and will handle their own buffering.
+  class StopIterationNoBuffer : FilterDataStatus()
 }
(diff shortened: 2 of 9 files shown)


examples/swift/hello_world/DemoFilter.swift
```diff
@@ -17,6 +17,7 @@ struct DemoFilter: ResponseFilter {
   func setResponseFilterCallbacks(_ callbacks: ResponseFilterCallbacks) {}
 
   func onResponseData(_ body: Data, endStream: Bool) -> FilterDataStatus {
+    NSLog("Saw data chunk of length \(body.count)")
```
goaway, line 20: I intend to eventually remove this when we have more direct integration coverage of this functionality in CI, based on the approach @junr03 has proposed with the ValidationFilter (name TBD? ExpectationFilter? RequirementsFilter? SatisfactionFilter.).

junr03: sg, do you mind leaving a TODO? 

AssertionFilter? So many options!

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -95,13 +44,68 @@ static envoy_headers toNativeHeaders(EnvoyHeaders *headers) {
   }
 
   EnvoyHeaders *platformHeaders = to_ios_headers(headers);
-  release_envoy_headers(headers);
```
junr03, line 98: we were always releasing right after the transformation call, right? Now I can't remember why https://github.com/lyft/envoy-mobile/pull/971/files#diff-0fdd94f3084afc47d1b505544703ca0dL41 existed in the first place? i.e there was not a site (is there one in this PR and I am blind to it) where we conditionally released?

library/objective-c/EnvoyBridgeUtility.h
```diff
@@ -0,0 +1,68 @@
...
+static inline envoy_headers toNativeHeaders(EnvoyHeaders *headers) {
+  envoy_header_size_t length = 0;
+  for (id headerKey in headers) {
+    length += [headers[headerKey] count];
+  }
+  envoy_header *header_array = (envoy_header *)safe_malloc(sizeof(envoy_header) * length);
+  envoy_header_size_t header_index = 0;
+  for (id headerKey in headers) {
+    NSArray *headerList = headers[headerKey];
+    for (id headerValue in headerList) {
```
rebello95, line 29: Can we use typed keys here instead of `id`?

library/swift/src/filters/FilterDataStatus.swift
```diff
@@ -21,4 +21,17 @@ public enum FilterDataStatus {
...
+  ///
+  /// This filter will continue to be called with new chunks of data.
+  ///
+  /// Returning `continue` from `onRequestData()`/`onResponseData()` or calling
+  /// `continueRequest()`/`continueResponse()` MUST be called when continued filter iteration is
+  /// desired.
+  ///
+  /// This may be called by filters which must parse a larger block of the incoming data before
+  /// continuing processing, and will handle their own buffering.
+  case stopIterationNoBuffer
```
rebello95, line 36: We had actually dropped this deliberately since it wasn't something we thought was needed yet. What was the reasoning for adding it back? If we want to keep it, we should also add it to the Android interfaces

library/swift/src/filters/Filter.swift
```diff
@@ -50,6 +75,32 @@ extension EnvoyHTTPFilter {
...
+      self.onResponseTrailers = { envoyTrailers in
+        let result = responseFilter.onResponseTrailers(ResponseTrailers(headers: envoyTrailers))
+        switch result {
+        case .continue(let trailers):
+          return [kEnvoyFilterTrailersStatusContinue, trailers.headers]
+        case .stopIteration(let trailers):
+          return [kEnvoyFilterTrailersStatusStopIteration, trailers.headers]
+        }
+      }
+
```
rebello95, line 103:
```suggestion
```

examples/swift/hello_world/DemoFilter.swift
goaway: StipulationsFilter.

library/objective-c/EnvoyEngineImpl.m
goaway: It's already gone. I removed the optimization and came up with the flag-based way (that we discussed earlier) to do it (later).

library/objective-c/EnvoyBridgeUtility.h
goaway: Yes. In fact, I feel like I already had changed this, but it might have gotten lost in the chain of branches.

library/swift/src/filters/FilterDataStatus.swift
goaway: We discussed earlier, but posting here for posterity - leaving this in potentially eases the path for migrating over platform logic that already does its own buffering, since less code needs to change initially. It also still provides for application logic that for one reason or another must to do its own buffering.

library/objective-c/EnvoyBridgeUtility.h
goaway: Ah, I did in the new location, but not the old one where I factored from.

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
```diff
@@ -30,4 +30,18 @@ sealed class FilterDataStatus {
...
+   * data.
+   *
+   * This filter will continue to be called with new chunks of data.
+   *
+   * Returning `Continue` from `onRequestData()`/`onResponseData()` or calling
+   * `continueRequest()`/`continueResponse()` MUST be called when continued filter iteration is
+   * desired.
+   *
+   * This may be called by filters which must parse a larger block of the incoming data before
+   * continuing processing, and will handle their own buffering.
```
rebello95, line 45: Missing closing comment `*/` 🙃

goaway: 🙃

PR #974, opened 2020-07-24
assertion filter: filter to setup e2e testing for multihop http expectations

junr03:

PENDING DESCRIPTION


library/kotlin/test/io/envoyproxy/envoymobile/BridgeTest.kt
```diff
@@ -50,7 +51,7 @@ static_resources:
                 http_request_headers_match:
                   headers:
                     - name: ":authority"
-                      exact_match: not_my_name
+                      exact_match: api.lyft.com
```
goaway, line 54: Let's use `example.com`.

PR #977, opened 2020-07-28
stats: clean up default inclusion list

goaway:
Description: The previous list of stats would result in redundant checks and larger state machines. This change slims things down.

### DIFF
--- a/library/common/config_template.cc
+++ b/library/common/config_template.cc
@@ -197,34 +197,10 @@ stats_config:
       patterns:
         - safe_regex:
             google_re2: {}
-            regex: 'cluster\.[\w]+?\.upstream_cx_active'
+            regex: '^cluster\.[\w]+?\.upstream_cx_active'
         - safe_regex:
             google_re2: {}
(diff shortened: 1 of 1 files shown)


goaway:
Note, I'm assuming we're including all the retry stats.

library/common/config_template.cc
```diff
@@ -233,33 +209,10 @@ stats_flush_interval: {{ stats_flush_interval_seconds }}s
...
-        - safe_regex:
-            google_re2: {}
-            regex: 'vhost.api.vcluster\.[\w]+?\.upstream_rq_retry_success'
-        - safe_regex:
-            google_re2: {}
-            regex: 'vhost.api.vcluster\.[\w]+?\.upstream_rq_time'
-        - safe_regex:
-            google_re2: {}
-            regex: 'vhost.api.vcluster\.[\w]+?\.upstream_rq_timeout'
+            regex: '^http.hcm.downstream_rq_(?:[12345]xx|total|completed)'
```
junr03, line 212: should total and completed be kept as exact?

goaway: It'll still be cheaper to check the shared prefix once via the single regex.

PR #979, opened 2020-07-28
filters: rework callbacks for shared signatures

goaway:
Description: Rework callback signatures by allowing a type-erased return context similar to the context parameter that gets included in the call argument. This opens the door to filter callbacks and stream callbacks sharing streamlined bridge code, especially in the case of the JNI. Also improves upon the current header passing mechanism for the JNI.
Risk Level: Moderate
Testing: Local and CI

### DIFF
--- a/library/common/jni/jni_interface.cc
+++ b/library/common/jni/jni_interface.cc
@@ -146,6 +146,7 @@ static JNIEnv* get_env() {
 static void jvm_on_headers(envoy_headers headers, bool end_stream, void* context) {
   JNIEnv* env = get_env();
   jobject j_context = static_cast<jobject>(context);
+  pass_headers(env, headers, j_context);
 
   jclass jcls_JvmCallbackContext = env->GetObjectClass(j_context);
   jmethodID jmid_onHeaders = env->GetMethodID(jcls_JvmCallbackContext, "onHeaders", "(JZ)V");
@@ -155,7 +156,6 @@ static void jvm_on_headers(envoy_headers headers, bool end_stream, void* context
                       end_stream ? JNI_TRUE : JNI_FALSE);
 
   env->DeleteLocalRef(jcls_JvmCallbackContext);
-  pass_headers(env, headers, j_context);
 }
 
 static void jvm_on_data(envoy_data data, bool end_stream, void* context) {
@@ -192,6 +192,7 @@ static void jvm_on_trailers(envoy_headers trailers, void* context) {
 
   JNIEnv* env = get_env();
   jobject j_context = static_cast<jobject>(context);
+  pass_headers(env, trailers, j_context);
 
   jclass jcls_JvmCallbackContext = env->GetObjectClass(j_context);
   jmethodID jmid_onTrailers = env->GetMethodID(jcls_JvmCallbackContext, "onTrailers", "(J)V");
@@ -200,7 +201,6 @@ static void jvm_on_trailers(envoy_headers trailers, void* context) {
   env->CallVoidMethod(j_context, jmid_onTrailers, (jlong)trailers.length);
 
   env->DeleteLocalRef(jcls_JvmCallbackContext);
-  pass_headers(env, trailers, j_context);
 }
 
 static void jvm_on_error(envoy_error error, void* context) {
--- a/library/common/jni/jni_interface.cc
+++ b/library/common/jni/jni_interface.cc
@@ -143,7 +143,7 @@ static JNIEnv* get_env() {
   return env;
 }
 
-static void jvm_on_headers(envoy_headers headers, bool end_stream, void* context) {
+static void* jvm_on_headers(envoy_headers headers, bool end_stream, void* context) {
   JNIEnv* env = get_env();
   jobject j_context = static_cast<jobject>(context);
   pass_headers(env, headers, j_context);
@@ -152,13 +152,14 @@ static void jvm_on_headers(envoy_headers headers, bool end_stream, void* context
   jmethodID jmid_onHeaders = env->GetMethodID(jcls_JvmCallbackContext, "onHeaders", "(JZ)V");
   // Note: be careful of JVM types. Before we casted to jlong we were getting integer problems.
   // TODO: make this cast safer.
-  env->CallObjectMethod(j_context, jmid_onHeaders, (jlong)headers.length,
-                        end_stream ? JNI_TRUE : JNI_FALSE);
+  jobject result = env->CallObjectMethod(j_context, jmid_onHeaders, (jlong)headers.length,
+                                         end_stream ? JNI_TRUE : JNI_FALSE);
 
   env->DeleteLocalRef(jcls_JvmCallbackContext);
+  return result;
 }
 
-static void jvm_on_data(envoy_data data, bool end_stream, void* context) {
+static void* jvm_on_data(envoy_data data, bool end_stream, void* context) {
   __android_log_write(ANDROID_LOG_VERBOSE, "[Envoy]", "jvm_on_data");
   JNIEnv* env = get_env();
   jobject j_context = static_cast<jobject>(context);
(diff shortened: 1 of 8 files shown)


goaway:
Splitting this out as it's getting fairly large and can be reviewed independently.

goaway:
Thanks @rebello95!

library/java/src/io/envoyproxy/envoymobile/engine/JvmBridgeUtility.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.List;
+import java.util.Map;
+
+import io.envoyproxy.envoymobile.engine.types.EnvoyHTTPCallbacks;
+
+/**
+ * Class to assist with passing types from native code over the JNI. Currently supports
+ * HTTP headers.
+ */
+class JvmBridgeUtility {
```
buildbreaker, line 16: Could we add a couple unit tests for the behaviors expected for this class?

library/java/src/io/envoyproxy/envoymobile/engine/JvmBridgeUtility.java
```diff
@@ -0,0 +1,75 @@
...
+    headerCount++;
+  }
+
+  /**
+   * Retrieves accumulated headers and resets state.
+   *
+   * @return Map, a map of header names to one or more values.
+   */
+  Map<String, List<String>> retrieveHeaders() {
+    Map headers = headerAccumulator;
```
buildbreaker, line 62:
```suggestion
    Map<String, List<String>> headers = headerAccumulator;
```

library/java/src/io/envoyproxy/envoymobile/engine/JvmCallbackContext.java
```diff
@@ -10,118 +10,67 @@
...
    * to be dispatched via the callback.
    *
    * @param length,    the total number of headers included in this header block.
    * @param endStream, whether this header block is the final remote frame.
+   * @return Void,     not used for response callbacks.
    */
-  public void onHeaders(long length, boolean endStream) {
-    startAccumulation(FrameType.HEADERS, length, endStream);
+  public Void onHeaders(long headerCount, boolean endStream) {
+    assert bridgeUtility.validateCount(headerCount);
```
buildbreaker, line 30: This would crash if the validation returns false?

test/common/http/dispatcher_test.cc
```diff
@@ -73,16 +73,18 @@ TEST_F(DispatcherTest, SetDestinationCluster) {
   callbacks_called cc = {0, 0, 0, 0, 0, 0};
   bridge_callbacks.context = &cc;
   bridge_callbacks.on_headers = [](envoy_headers c_headers, bool end_stream,
-                                   void* context) -> void {
-    ASSERT_TRUE(end_stream);
+                                   void* context) -> void* {
+    EXPECT_TRUE(end_stream);
```
buildbreaker, line 77: What's the difference between `ASSERT_TRUE` and `EXPECT_TRUE`?

library/java/src/io/envoyproxy/envoymobile/engine/JvmBridgeUtility.java
junr03, line 16: I might be missing something, but division into this utility class is not predicated on the returned context, right?

library/common/jni/jni_interface.cc
```diff
@@ -143,22 +143,23 @@ static JNIEnv* get_env() {
...
   // Note: be careful of JVM types. Before we casted to jlong we were getting integer problems.
   // TODO: make this cast safer.
-  env->CallVoidMethod(j_context, jmid_onHeaders, (jlong)headers.length,
-                      end_stream ? JNI_TRUE : JNI_FALSE);
+  jobject result = env->CallObjectMethod(j_context, jmid_onHeaders, (jlong)headers.length,
+                                         end_stream ? JNI_TRUE : JNI_FALSE);
 
   env->DeleteLocalRef(jcls_JvmCallbackContext);
-  pass_headers(env, headers, j_context);
+  return result;
```
junr03, line 159: As far as I can tell the returned contexts are not getting used anywhere yet right?

library/java/src/io/envoyproxy/envoymobile/engine/JvmCallbackContext.java
```diff
@@ -10,118 +10,67 @@
...
    * Initializes state for accumulating header pairs via passHeaders, ultimately
    * to be dispatched via the callback.
    *
    * @param length,    the total number of headers included in this header block.
    * @param endStream, whether this header block is the final remote frame.
+   * @return Void,     not used for response callbacks.
    */
-  public void onHeaders(long length, boolean endStream) {
-    startAccumulation(FrameType.HEADERS, length, endStream);
+  public Void onHeaders(long headerCount, boolean endStream) {
```
junr03, line 29: what is capital `V` void?

test/common/http/dispatcher_test.cc
```diff
@@ -73,16 +73,18 @@ TEST_F(DispatcherTest, SetDestinationCluster) {
...
     ResponseHeaderMapPtr response_headers = toResponseHeaders(c_headers);
     EXPECT_EQ(response_headers->Status()->value().getStringView(), "200");
     callbacks_called* cc = static_cast<callbacks_called*>(context);
     cc->on_headers_calls++;
+    return nullptr;
```
junr03, line 82: Do you foresee the context being passed down all the way to the Http::Dispatcher or stopping in the bridge layer?

I ask because it would be great if instead of just updating the return type here we could actually cover the change with testing. But obviously the dispatcher is not seeing the context yet (or ever, hence my question).

test/common/http/dispatcher_test.cc
goaway: `ASSERT_TRUE` aborts execution if the expectation fails. It also requires the enclosing function to have a `void` return type, because it actually calls `return`. `EXPECT_TRUE` does not abort execution, and has no requirements on the return-type of the enclosing function. Both will cause the test to fail if the passed assertion/expectation isn't met.

library/java/src/io/envoyproxy/envoymobile/engine/JvmBridgeUtility.java
goaway: No, but it does encapsulate the complexity a bit better (imo), and it allows re-use (in the case of filters).

library/common/jni/jni_interface.cc
goaway: Correct, they're used in the subsequent PR I have for filters.

library/java/src/io/envoyproxy/envoymobile/engine/JvmCallbackContext.java
goaway: https://docs.oracle.com/javase/7/docs/api/java/lang/Void.html

Note, we could use any other non-instantiable type here, including our own. But `Void` is built in, and there's some precedent/convention for using `Void` for this purpose.

test/common/http/dispatcher_test.cc
goaway: The dispatcher has no need to see it, since it doesn't utilize any signals returned by callbacks. So my inclination would be not to expose it at all there. My line of thinking was that the existing tests actually do cover the change, since they demonstrate that it causes no change in underlying functionality. Additional coverage will be needed for filters usage, of course.

library/java/src/io/envoyproxy/envoymobile/engine/JvmCallbackContext.java
goaway: Yes. It shouldn't ever fail, and the implementation is broken if it fails.

library/java/src/io/envoyproxy/envoymobile/engine/JvmBridgeUtility.java
goaway: Sure, I don't have strong feelings. Is that more common/idiomatic?

library/java/src/io/envoyproxy/envoymobile/engine/JvmBridgeUtility.java
goaway: Yes, though if you're okay with it, I'd prefer to do it in a separate PR to keep things rolling forward on main.

library/java/src/io/envoyproxy/envoymobile/engine/JvmBridgeUtility.java
buildbreaker: Yeah, it's looks a bit bare without the type declarations haha

library/common/jni/jni_interface.cc
```diff
@@ -175,41 +181,47 @@ static void jvm_on_data(envoy_data data, bool end_stream, void* context) {
...
+      env->CallObjectMethod(j_context, jmid_onData, j_data, end_stream ? JNI_TRUE : JNI_FALSE);
 
   data.release(data.context);
   env->DeleteLocalRef(j_data);
   env->DeleteLocalRef(jcls_JvmCallbackContext);
+  return result;
 }
 
-static void jvm_on_metadata(envoy_headers metadata, void* context) {
+static void* jvm_on_metadata(envoy_headers metadata, void* context) {
```
buildbreaker, line 193: This isn't related to this PR but do we want to remove the on metadata callbacks since we are not using them or exposing them?

test/common/http/dispatcher_test.cc
buildbreaker: Just so I understand: we're returning a `nullptr` here since we don't expect the consumer (the dispatcher) to use the return value?

library/java/src/io/envoyproxy/envoymobile/engine/JvmCallbackContext.java
junr03: I see so it is the java way of returning a void*?

test/common/http/dispatcher_test.cc
junr03: > The dispatcher has no need to see it, since it doesn't utilize any signals returned by callbacks. So my inclination would be not to expose it at all there.

ok, that was my read of it.

library/java/src/io/envoyproxy/envoymobile/engine/JvmCallbackContext.java
goaway: Effectively, yes. I switched over to `Object` though as you can see, due to an issue with JNI signature lookup.

goaway:
Obj-c example app works for me locally and it's seemed very inconsistent for me in CI. Going to go ahead and push forward.

goaway:
@junr03 by all means ask questions as you come up with them, and also bear in mind that there's still some cleanup I want to come in and do here afterwards as well.

goaway:
thanks all!

PR #981, opened 2020-07-28
update envoy to f958d39fcd

junr03:

### DIFF
--- a/library/common/BUILD
+++ b/library/common/BUILD
@@ -35,7 +35,7 @@ envoy_cc_library(
     deps = [
         "@envoy//source/common/common:random_generator_lib",
         "@envoy//source/common/runtime:runtime_lib",
-        "@envoy//source/exe:main_common_lib",
+        "@envoy//source/exe:envoy_main_common_lib",
     ],
 )
 
--- a/library/common/BUILD
+++ b/library/common/BUILD
@@ -44,7 +44,7 @@ envoy_cc_library(
     repository = "@envoy",
     deps = [
         ":envoy_mobile_main_common_lib",
-        "@envoy//source/common/version:version_linkstamp",
+        "@envoy//source/exe:envoy_main_common_lib",
     ],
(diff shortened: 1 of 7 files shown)


junr03:
going to merge as is and open up a high priority issue to get a current diff on the size regression

library/common/BUILD
```diff
@@ -35,7 +35,15 @@ envoy_cc_library(
...
-        "@envoy//source/exe:envoy_common_lib",
+        "@envoy//source/exe:main_common_lib",
+    ],
+)
+
+envoy_cc_library(
+    name = "envoy_mobile_main_common_lib_stamped",
+    repository = "@envoy",
+    deps = [
+        ":envoy_mobile_main_common_lib",
```
rebello95, line 46: Why is this now required?

library/common/config_template.cc
```diff
@@ -268,4 +268,11 @@ stats_flush_interval: {{ stats_flush_interval_seconds }}s
     app_id : {{ app_id }}
     app_version : {{ app_version }}
     os: {{ device_os }}
+# Needed due to warning in https://github.com/envoyproxy/envoy/blob/6eb7e642d33f5a55b63c367188f09819925fca34/source/server/server.cc#L546
```
rebello95, line 271: Is there an issue for this or intended to be permanent?

library/common/BUILD
junr03: https://github.com/envoyproxy/envoy/pull/12171 split things for tests, so our targets here were either: a) not having symbols, or having duplicate symbols. This new setup lets us use the new upstream system correctly.

library/common/config_template.cc
junr03: Unfortunately from the TODOs in that code the only way to set that configuration is in runtime, and the only (for now) way we can set runtime in envoy mobile is this way

PR #987, opened 2020-07-30
swift: add `@frozen` annotation to filter statuses

rebello95:
This suppresses the following warning when `switch`ing over filter return values:

> Switch covers known cases, but 'FilterHeaderStatus<RequestHeaders>' may have additional unknown values

Note that while we are not guaranteeing ABI stability, we will use semantic versioning going forward.


rebello95:
Thanks @keith, updated the description

PR #958, opened 2020-07-15
build: update bazel, upstream envoy, & ios rules

rebello95:
- Update bazel to 3.4.1 (depends on https://github.com/envoyproxy/envoy/pull/12123)
- Update envoy upstream to include ^
- Update rules apple and rules swift
- Update our header map iteration to fix breaking changes upstream from https://github.com/envoyproxy/envoy/pull/12103

Signed-off-by: Michael Rebello <me@michaelrebello.com>

PR #988, opened 2020-07-30
envoy: revert recent upstream updates

goaway:
Description: c6e1af6 (#958) broke Android builds on MacOS and 811235d (#981) adopted new Envoy interfaces making it no longer possible to an older SHA. Together this makes Android development extremely difficult. This rolls back those changes on main until the breakage can be resolved.
Risk Level: TBD
Testing: Local and CI


rebello95:
Can you elaborate on the issues you're encountering (CI is green on master)?

I'm concerned about reverting back to a version of Envoy from several weeks ago - each time we upgrade, we typically have several issues that need to be resolved, and getting further out of date makes it harder to upgrade as time goes on. Can we fix the issues instead?

goaway:
Resolved this offline, looks like a local NDK issue we were all experiencing. Thanks for the help @buildbreaker and @keith!

PR #992, opened 2020-07-30
stats: introduce stats client

jingwei99:
Description: Introduces a StatsClient capable of giving out Counters that, when incremented, emit stats from Envoy's internal stats engine. Also refactors the top-level platform Engine interface to make both the StreamClient and the StatsClient available.
Risk Level: Low
Testing: Pending


library/common/engine.cc
```diff
@@ -93,6 +93,18 @@ Engine::~Engine() {
   main_thread_.join();
 }
 
+void Engine::recordCounter(std::string elemenets, uint64_t count) {
```
goaway, line 96: minor typo s/elemenets/elements/

library/common/engine.cc
```diff
@@ -93,6 +93,18 @@ Engine::~Engine() {
...
+  if (server_) {
+    server_->dispatcher().post([this, elemenets, count]() -> void {
+      std::string client_stats = "client_stats";
```
goaway, line 99: let's just make the prefix be "client" I think

goaway:
@rebello95 I like the idea of having Element be a strong type. There are actually invalid character strings that one could pass, and we can prevent that by having Element have a conditional initializer. This would allow error discovery to happen more or less statically.

examples/java/hello_world/MainActivity.java
```diff
@@ -76,7 +76,7 @@ private void makeRequest() {
     RequestHeaders requestHeaders = new RequestHeadersBuilder(RequestMethod.GET, REQUEST_SCHEME,
                                                               REQUEST_AUTHORITY, REQUEST_PATH)
                                         .build();
-    streamClient.newStreamPrototype()
+    engine.getStreamClient().newStreamPrototype()
         .setOnResponseHeaders((responseHeaders, endStream) -> {
```
buildbreaker, line 80:
```suggestion
    engine.getStreamClient()
        .newStreamPrototype()
        .setOnResponseHeaders((responseHeaders, endStream) -> {
```

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
```diff
@@ -36,7 +36,11 @@ kt_android_library(
 envoy_mobile_kt_library(
     name = "envoy_interfaces_lib",
     srcs = glob([
-        "EnvoyClient.kt",
+        "Counter.kt",
+        "Element.kt",
```
rebello95, line 40: I think it'd make sense to group these into a `stats` subdirectory, similarly to `grpc` and `filters`

library/kotlin/src/io/envoyproxy/envoymobile/Counter.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+class Counter internal constructor(
+  private val envoyEngine: WeakReference<EnvoyEngine>,
+  private val elements: List<Element>
+) {
+
+  /**
+   * Increment the counter by the given count.
+   */
+  fun increment(count: Int) {
```
rebello95, line 19: Should we provide a default value of 1?

library/kotlin/src/io/envoyproxy/envoymobile/Counter.kt
```diff
@@ -0,0 +1,24 @@
+package io.envoyproxy.envoymobile
+
+import io.envoyproxy.envoymobile.engine.EnvoyEngine
+import java.lang.ref.WeakReference
+
+/**
+ * A stat.
+ *
+ * Current the supported stat type is counter, and it can increment.
```
rebello95, line 9: I assume we will create different types for non-Counter stats, so maybe this docstring can be updated to simply describe the Counter?

library/kotlin/src/io/envoyproxy/envoymobile/Element.kt
```diff
@@ -0,0 +1,16 @@
+package io.envoyproxy.envoymobile
+
+import java.util.regex.Pattern
+
+/**
+ * Element for stats.
+ */
```
rebello95, line 7: Can you document what the intended value is supposed to be (given the regex below)?

library/kotlin/src/io/envoyproxy/envoymobile/Engine.kt
```diff
@@ -0,0 +1,17 @@
...
+
+  /**
+   *  @return a {@link StreamClient} instance
+   */
+  fun getStreamClient(): StreamClient
+
+  /**
+   *  @return a {@link StatsClient} instance
+   */
+  fun getStatsClient(): StatsClient
```
rebello95, line 16: would it make sense to simply name these `streamClient()` and `statsClient()`?

library/kotlin/src/io/envoyproxy/envoymobile/StatsClient.kt
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Client used to record timeseries metrics.
+ */
+interface StatsClient {
+
+  /**
+   * @return A counter based on the joined elements.
+   */
+  fun getCounter(vararg elements: Element): Counter
```
rebello95, line 11: Similar suggestion here (`counter()` versus `getCounter()`?)

library/swift/src/BUILD
```diff
@@ -5,7 +5,11 @@ licenses(["notice"])  # Apache 2
 swift_static_framework(
     name = "ios_framework",
     srcs = glob([
-        "EnvoyClient.swift",
+        "Counter.swift",
+        "Element.swift",
```
rebello95, line 9: Same comments apply to iOS re: grouping and naming

library/swift/src/Counter.swift
```diff
@@ -0,0 +1,22 @@
+@_implementationOnly import EnvoyEngine
+import Foundation
+
+@objc
+public class Counter: NSObject {
```
rebello95, line 5: Same docstrings as Android?

library/swift/src/Counter.swift
```diff
@@ -0,0 +1,22 @@
...
+  private let series: String
+  private weak var engine: EnvoyEngine?
+
+  internal init(elements: [Element], engine: EnvoyEngine) {
```
rebello95, line 9:
```suggestion
  init(elements: [Element], engine: EnvoyEngine) {
```

library/swift/src/Counter.swift
```diff
@@ -0,0 +1,22 @@
...
+    self.series = elements.map{ $0.description }.joined(separator: ".")
```
rebello95, line 10:
```suggestion
    self.series = elements.map { $0.description }.joined(separator: ".")
```

library/swift/src/Element.swift
```diff
@@ -0,0 +1,19 @@
...
+
+  public init(stringLiteral value: String) {
+    guard value.range(of: kPattern, options: .regularExpression) != nil else {
+      preconditionFailure("Element values must conform to the regex /^[A-Za-z_]+$/.")
+    }
+    self.value = value
+  }
+
+  // CustomStringConvertible
+  public override var description: String { return value }
```
rebello95, line 18: Why not make `value` internal? I think using `description` is fragile since it could break if someone deleted the explicit `description` override (yet would still compile)

library/swift/src/Counter.swift
```diff
@@ -0,0 +1,22 @@
...
+    self.engine = engine
+    super.init()
+  }
+
+  func increment(count: Int) {
```
rebello95, line 15: This needs to be public and is missing docstrings

library/swift/src/Counter.swift
```diff
@@ -0,0 +1,22 @@
...
+    guard let engine = self.engine else {
+      return
+    }
+
+    engine.recordCounter(self.series, count: numericCast(count))
```
rebello95, line 20: Why are we exposing an `Int` publicly but requiring an unsigned integer internally? Shouldn't these be the same (in which case we could avoid a cast)?

library/swift/src/Counter.swift
```diff
@@ -0,0 +1,22 @@
+@_implementationOnly import EnvoyEngine
+import Foundation
+
+@objc
```
rebello95, line 4:
```suggestion
@objcMembers
```

library/swift/src/Element.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+
+private let kPattern = "^[A-Za-z_]+$"
+
+/// Element represents one dot-delimited component of a timeseries name.
+@objc
```
rebello95, line 6:
```suggestion
@objcMembers
```

library/swift/src/Element.swift
```diff
@@ -0,0 +1,19 @@
...
+public final class Element: NSObject, ExpressibleByStringLiteral {
+  private let value: String
+
+  public init(stringLiteral value: String) {
+    guard value.range(of: kPattern, options: .regularExpression) != nil else {
+      preconditionFailure("Element values must conform to the regex /^[A-Za-z_]+$/.")
```
rebello95, line 12:
```suggestion
      preconditionFailure("Element values must conform to the regex '\(kPattern)'")
```

library/kotlin/src/io/envoyproxy/envoymobile/Element.kt
```diff
@@ -0,0 +1,16 @@
...
+class Element(val element: String) {
+  init {
+    if (!Pattern.compile("^[A-Za-z_]+\$").matcher(element).matches()) {
+      throw IllegalArgumentException(
+        "Element values must conform to the regex /^[A-Za-z_]+$/"
```
rebello95, line 12: I'd use a constant here so you don't have to replicate the regex twice (and so they don't deviate over time)

library/swift/src/StatsClient.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+/// Client used to record timeseries metrics.
+@objc
+public protocol StatsClient: AnyObject {
+  /// - returns: A Counter based on the joined elements.
```
rebello95, line 6: missing `- parameter elements`

library/swift/src/StatsClientImpl.swift
```diff
@@ -0,0 +1,19 @@
+@_implementationOnly import EnvoyEngine
+import Foundation
+
+/// Envoy implementation of StatsClient.
+@objcMembers
```
rebello95, line 5: This shouldn't be necessary since internal types don't need to be exposed to public objc consumers

library/swift/src/mocks/MockEnvoyEngine.swift
```diff
@@ -7,6 +7,8 @@ final class MockEnvoyEngine: NSObject {
   static var onRunWithConfig: ((_ config: EnvoyConfiguration, _ logLevel: String?) -> Void)?
   /// Closure called when `run(withConfigYAML:)` is called.
   static var onRunWithYAML: ((_ configYAML: String, _ logLevel: String?) -> Void)?
+  /// Closure called when `recordCounter(_:count:)` is called.
+  static var onRecordCounter: ((_ elements: String, _ count: UInt) -> Void)?
```
rebello95, line 11: Can we add some tests?

library/swift/src/StreamClientImpl.swift
```diff
@@ -0,0 +1,19 @@
+@_implementationOnly import EnvoyEngine
+import Foundation
+
+/// Envoy implementation of StreamClient.
+@objcMembers
```
rebello95, line 5: Same here regarding objc

library/kotlin/src/io/envoyproxy/envoymobile/StatsClientImpl.kt
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Envoy implementation of `StatsClient`.
+ */
+internal class StatsClientImpl constructor(
+  internal val engine: EnvoyEngine
+) : StatsClient {
+
+  override fun getCounter(vararg elements: Element): Counter {
+    return Counter(WeakReference(engine), elements.asList())
```
rebello95, line 14: iOS has a strong ref below, we should keep them consistent (I think a strong ref is what you want here)

library/kotlin/src/io/envoyproxy/envoymobile/Counter.kt
junr03: usually we overload with `increment()` being `increment(1)`. In favor of that.

library/kotlin/src/io/envoyproxy/envoymobile/Engine.kt
jingwei99: @goaway ^

library/kotlin/src/io/envoyproxy/envoymobile/StatsClientImpl.kt
rebello95: Ah I see, this is passing a weak reference to the Counter, whereas on iOS the Counter itself makes the reference weak. Would it make sense to move `WeakReference(engine)` into the counter file's constructor to make it clear that it doesn't retain the engine?

Separately, I'm wondering if it makes sense to allow counters to retain the engine. Do we want the engine to stick around as long as a consumer is holding on to a counter ref?

library/common/engine.cc
```diff
@@ -93,6 +93,19 @@ Engine::~Engine() {
   main_thread_.join();
 }
 
+void Engine::recordCounter(std::string elements, uint64_t count) {
```
junr03, line 96: Can we cover this in tests?

library/kotlin/src/io/envoyproxy/envoymobile/StatsClientImpl.kt
jingwei99: > Would it make sense to move WeakReference(engine) into the counter file's constructor to make it clear that it doesn't retain the engine?

yea

> Do we want the engine to stick around as long as a consumer is holding on to a counter ref?

probably not, counters could be referenced by feature modules of the platform(iOS/android) code, and we probably don't want the feature modules to (indirectly) hold a strong ref to engine

library/common/config_template.cc
```diff
@@ -204,6 +204,9 @@ stats_flush_interval: {{ stats_flush_interval_seconds }}s
         - safe_regex:
             google_re2: {}
             regex: '^http.dispatcher.*'
+        - safe_regex:
+            google_re2: {}
+            regex: '^client.*'
```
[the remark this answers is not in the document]
jingwei99: what rules do you have in mind? 
right now all the elements used for the stats are checked against the regex: `^[A-Za-z_]+\$`

library/common/engine.cc
```diff
@@ -93,6 +93,19 @@ Engine::~Engine() {
...
+  if (server_) {
+    server_->dispatcher().post([this, elements, count]() -> void {
+      std::string client = "client";
+      absl::string_view prefix{client};
```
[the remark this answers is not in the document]
jingwei99: will try it out

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
goaway: Sounds good to me.

library/kotlin/src/io/envoyproxy/envoymobile/Counter.kt
goaway: Sounds good.

library/kotlin/src/io/envoyproxy/envoymobile/Engine.kt
goaway: I'm fine with that.

library/kotlin/src/io/envoyproxy/envoymobile/StatsClient.kt
goaway: Sure, sounds good.

library/swift/src/Element.swift
goaway: Okay, sure.

library/swift/src/mocks/MockEnvoyEngine.swift
goaway: Sure.

library/common/engine.h
```diff
@@ -35,6 +35,11 @@ class Engine {
    */
   Http::Dispatcher& httpDispatcher();
 
+  /**
+   * Increment a counter with a given string of elements and by the given count.
+   */
+  void recordCounter(std::string elements, uint64_t count);
```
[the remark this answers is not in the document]
goaway: discussed offline and decided we'll go with `record` for now

library/common/engine.cc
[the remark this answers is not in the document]
jingwei99: update: tried it out, didn't compile

library/common/engine.cc
```diff
@@ -93,6 +93,19 @@ Engine::~Engine() {
...
+      absl::string_view dynamic_elements{elements};
+      Stats::Utility::counterFromElements(server_->serverFactoryContext().scope(),
```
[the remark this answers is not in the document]
goaway: Based on my reading of the linked doc, the main cost associated with dynamic tokens is that they take up more space and take time to encode - but they don't actually contend locks. It also appears new static tokens can be created via XDS updates, implying this could be an option for us as well as a potential optimization.

At present, all our work is on the main thread anyhow, and with dispatched stat updates, they're going to execute exclusively between I/O events regardless. I'd be fine with prominently documenting that this interface is experimental and may have unknown performance implications, but I feel we shouldn't block on that any more than we have other potential performance sinks in our hot path (extra copies, etc.).

What do you think?

library/swift/src/Counter.swift
goaway: My impression is that Swift convention strongly favors treating base numeric types as signed wherever possible. (And in Java/Kotlin, there's basically no alternative.) On the other hand, Envoy is explicit about counters accepting unsigned values only. `numericCast` imposes strong runtime checks on invalid (in this case, negative) ranges. And this keeps the two platform interfaces aligned.

library/kotlin/src/io/envoyproxy/envoymobile/Element.kt
goaway: clever

library/kotlin/src/io/envoyproxy/envoymobile/StatsClientImpl.kt
goaway: My intuition right now is that it seems natural for engine *clients* to strongly retain the engine, but perhaps less so for the artifacts they produce. This has more to do with users of the library being able to implicitly manage the lifecycle of the engine than anything else. I should say that I'm not _strongly_ convinced this is right, but it seems reasonable for now for experimentation.

library/common/config_template.cc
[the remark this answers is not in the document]
goaway: Talked with @junr03, long-term it might make sense to make people specify this in configuration, so that some thought goes into the stats they emit. For now, as a new, experimental interface, this is probably good enough to move forward.

library/common/engine.cc
goaway: Adding a ticket to cover testing generally at this layer, will link here.

library/common/engine.cc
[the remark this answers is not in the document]
goaway: Chatted with @junr03, we're going to flag the StatsClient interface as experimental, and move forward, while continuing to investigate potential performance ramifications.

library/common/engine.cc
junr03: @goaway don't forget to add this ticket

library/common/engine.cc
```diff
@@ -93,6 +93,19 @@ Engine::~Engine() {
   main_thread_.join();
 }
 
+void Engine::recordCounter(std::string elements, uint64_t count) {
+  if (server_) {
+    server_->dispatcher().post([this, elements, count]() -> void {
+      static const std::string client = "client";
```
jmarantz, line 99: No non-pod statics (static init fiasco). Moreover what you really want here since this string is known at compile-time is to save a StatName for "client" using either a StatNamePool or StatNameManagedStorage. You want to do that in a context object that is created once at process startup and then re-used. Then you have no thread contention issues.

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
```diff
@@ -52,15 +54,18 @@ envoy_mobile_kt_library(
         "ResponseTrailersBuilder.kt",
         "RetryPolicy.kt",
         "RetryPolicyMapper.kt",
+        "StatsClient.kt",
+        "StatsClientImpl.kt",
```
rebello95, line 58: Shouldn't these be in `stats/` which is captured below with globbing?

library/kotlin/src/io/envoyproxy/envoymobile/StatsClientImpl.kt
rebello95: Was there a resolution here? Looks like the weak ref is still being passed from this line

library/kotlin/src/io/envoyproxy/envoymobile/stats/Element.kt
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Element represents one dot-delimited component of a time series name.
+ *
+ * Element values must conform to the regex /^[A-Za-z_]+$/.
+ */
+class Element(val element: String) {
+  init {
+    if (!Pattern.compile("^[A-Za-z_]+\$").matcher(element).matches()) {
+      throw IllegalArgumentException(
+        "Element values must conform to the regex /^[A-Za-z_]+$/"
```
rebello95, line 14: Regex is hardcoded 3 places here. Can we use a constant instead so we don't accidentally fail to change one of the usages in the future?

library/swift/src/BUILD
```diff
@@ -21,15 +23,18 @@ swift_static_framework(
         "ResponseTrailersBuilder.swift",
         "RetryPolicy.swift",
         "RetryPolicyMapper.swift",
+        "StatsClient.swift",
+        "StatsClientImpl.swift",
```
rebello95, line 27: Same here regarding living in `stats/`

library/swift/src/StatsClient.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+/// Client used to record time series metrics.
+///
+/// Note: this an experimental interface and is subject to change. The implementation has not been
+/// optimized, and there may be performance implications in production usage.
+@objc
+public protocol StatsClient: AnyObject {
+  /// - parameter elements: Elements to identify a counter
+  /// - returns: A Counter based on the joined elements.
```
rebello95, line 10:
```suggestion
  /// - parameter elements: Elements to identify a counter
  ///
  /// - returns: A Counter based on the joined elements.
```

library/swift/src/stats/Counter.swift
```diff
@@ -0,0 +1,24 @@
+@_implementationOnly import EnvoyEngine
+import Foundation
+
+/// A time series counter.
+@objcMembers
+public class Counter: NSObject {
```
rebello95, line 6:
```suggestion
public final class Counter: NSObject {
```

library/swift/src/stats/Counter.swift
```diff
@@ -0,0 +1,24 @@
...
+  private let series: String
+  private weak var engine: EnvoyEngine?
+
+  init(elements: [Element], engine: EnvoyEngine) {
+    self.series = elements.map { $0.value }.joined(separator: ".")
+    self.engine = engine
+    super.init()
+  }
+
+  /// Increment the counter by the given count.
```
rebello95, line 16:
```suggestion
  /// Increment the counter by the given count.
  ///
  /// - parameter count: The amount by which to increment the counter.
```

library/swift/src/stats/Counter.swift
```diff
@@ -0,0 +1,24 @@
...
+  public func increment(count: Int = 1) {
+    guard let engine = self.engine else {
+      return
+    }
+
+    engine.recordCounter(self.series, count: numericCast(count))
```
rebello95, line 22:
```suggestion
    self.engine?.recordCounter(self.series, count: numericCast(count))
```

library/swift/src/stats/Element.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+private let kPattern = "^[A-Za-z_]+$"
+
+/// Element represents one dot-delimited component of a time series name.
+/// Element values must conform to the regex /^[A-Za-z_]+$/.
+@objcMembers
+public final class Element: NSObject, ExpressibleByStringLiteral {
+  internal let value: String
```
rebello95, line 9:
```suggestion
  let value: String
```

library/swift/test/StatsClientImplTests.swift
```diff
@@ -0,0 +1,53 @@
...
+import Foundation
+import XCTest
+
+final class StatsClientImplTests: XCTestCase {
+  override func tearDown() {
+    super.tearDown()
+    MockEnvoyEngine.onRecordCounter = nil
+  }
+
+  func testCounterDelegatesToEngine() throws {
```
rebello95, line 12:
```suggestion
  func testCounterDelegatesToEngine() {
```

library/swift/test/StatsClientImplTests.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+    let mockEngine = MockEnvoyEngine()
+    let statsClient = StatsClientImpl(engine: mockEngine)
+    let counter = statsClient.counter(elements: ["test", "stat"])
+    counter.increment()
+    XCTAssertEqual(actualSeries, "test.stat")
+    XCTAssertEqual(actualCount, 1)
+  }
+
+  func testCounterDelegatesToEngineWithCount() throws {
```
rebello95, line 27:
```suggestion
  func testCounterDelegatesToEngineWithCount() {
```

library/swift/test/StatsClientImplTests.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+    let mockEngine = MockEnvoyEngine()
+    let statsClient = StatsClientImpl(engine: mockEngine)
+    let counter = statsClient.counter(elements: ["test", "stat"])
+    counter.increment(count: 5)
+    XCTAssertEqual(actualSeries, "test.stat")
+    XCTAssertEqual(actualCount, 5)
+  }
+
+  func testCounterWeaklyHoldsEngine() throws {
```
rebello95, line 42:
```suggestion
  func testCounterWeaklyHoldsEngine() {
```

library/swift/test/StatsClientImplTests.swift
```diff
@@ -0,0 +1,53 @@
+@testable import Envoy
+import EnvoyEngine
+import Foundation
+import XCTest
+
+final class StatsClientImplTests: XCTestCase {
```
rebello95, line 6: Can we add these tests for Kotlin too?

library/swift/test/BUILD
```diff
@@ -52,9 +52,19 @@ envoy_mobile_swift_test(
...
     srcs = [
-        "StreamClientBuilderTests.swift",
+        "EngineBuilderTests.swift",
+    ],
+    deps = [
+        "//library/objective-c:envoy_engine_objc_lib",
+    ],
+)
+
+envoy_mobile_swift_test(
```
rebello95, line 64: nit: can we alphabetize these test cases?

jingwei99: thinking out loud: prolly worth adding a lint/pre-commit check for this in the long run

library/kotlin/src/io/envoyproxy/envoymobile/BUILD
jingwei99: I think these ideally live here, basically in the same level as `StreamClient`, as the interface `Engine.kt` exposes a `StatsClient` and a `StreamClient`.

library/kotlin/src/io/envoyproxy/envoymobile/stats/Element.kt
jingwei99: fair

library/swift/src/BUILD
jingwei99: same as above, `StatsClient` and its implementation ideally live at the same level as `StreamClient` given the `Engine.swift` interface.

jingwei99:
@rebello95 : for this [comment](https://github.com/lyft/envoy-mobile/pull/992#discussion_r468228297), yea i missed it in my previous commit, and have a fix locally, planned to push along with other suggested changes

library/common/engine.cc
goaway: Created (but it's internal).

library/kotlin/test/io/envoyproxy/envoymobile/StatsClientImplTest.kt
```diff
@@ -0,0 +1,36 @@
...
+    val statsClient = StatsClientImpl(envoyEngine)
+    val counter = statsClient.counter(Element("test"), Element("stat"))
+    counter.increment(5)
+    val elementsCaptor = ArgumentCaptor.forClass(String::class.java)
+    val countCaptor = ArgumentCaptor.forClass(Int::class.java)
+    verify(envoyEngine).recordCounter(elementsCaptor.capture(), countCaptor.capture())
+    assertThat(elementsCaptor.getValue()).isEqualTo("test.stat")
+    assertThat(countCaptor.getValue()).isEqualTo(5)
+  }
+}
```
jingwei99, line 36: tried to emulate `StatsClientImplTests.testCounterWeaklyHoldsEngine`, but I wasn't able to get the `EnvoyEngine` instance garbage collected - as calling System.gc() doesn't guarantee gc.

rebello95:
@jingwei99 @goaway can we update the docs to reflect the interface changes from this PR?

jingwei99:
> @jingwei99 @goaway can we update the docs to reflect the interface changes from this PR?

yup, i'll work on the docs.

rebello95:
Thanks!

PR #995, opened 2020-07-31
assertion filter: initial filter implementation

junr03:
Description: the assertion filter will allow us to programatically test end-to-end bridging code between platform (swift/kotlin) and the native (c++) layers. This PR only lands the filter with subsequent PRs using it to build test suites.
Risk Level: low
Testing: unit tests
Docs Changes: will add additional documentation with the swift and kotlin test suites.


library/common/extensions/filters/http/assertion/filter.h
```diff
@@ -0,0 +1,50 @@
...
+
+private:
+  std::vector<Extensions::Common::Tap::MatcherPtr> matchers_;
+};
+
+typedef std::shared_ptr<AssertionFilterConfig> AssertionFilterConfigSharedPtr;
+
+/**
+ * Harness to bridge Envoy filter invocations up to the platform layer.
+ */
```
rebello95, line 31: I don't think this is accurate since the filter isn't being bubbled up to the platform

library/common/extensions/filters/http/assertion/filter.cc
```diff
@@ -0,0 +1,79 @@
...
+  return Http::FilterHeadersStatus::StopIteration;
+}
+
+Http::FilterDataStatus AssertionFilter::decodeData(Buffer::Instance& data, bool end_stream) {
+  config_->rootMatcher().onRequestBody(data, statuses_);
+  if (config_->rootMatcher().matchStatus(statuses_).matches_) {
+    if (end_stream) {
+      decoder_callbacks_->sendLocalReply(
+          Http::Code::OK, "Request Body match configured expectations", nullptr, absl::nullopt, "");
+      return Http::FilterDataStatus::StopIterationNoBuffer;
```
rebello95, line 51: These are super clever 👏

library/common/extensions/filters/http/assertion/filter.proto
```diff
@@ -0,0 +1,14 @@
...
+import "envoy/config/tap/v3/common.proto";
+
+import "validate/validate.proto";
+
+message Assertion {
+  // The match configuration. If the configuration matches the request frames the filter will send
+  // a local reply with Http::Code::OK on the last frame of the request stream (continuing on
+  // intervening frames). Otherwise, it will send a local reply with Http::Code::BadRequest.
+  envoy.config.tap.v3.MatchPredicate match_config = 1 [(validate.rules).message = {required: true}];
+}
```
rebello95, line 14:
```suggestion
  // The match configuration. If the configuration matches the request frames, the filter will send
  // a local reply with Http::Code::OK on the last frame of the request stream (continuing on
  // intervening frames). Otherwise, it will send a local reply with Http::Code::BadRequest.
  envoy.config.tap.v3.MatchPredicate match_config = 1 [(validate.rules).message = {required: true}];
}
```
It'd be nice if we could validate here that there's at least 1 matcher rule within the config (since it's asserted above), but not a big deal. We should probably document it here though

library/common/extensions/filters/http/assertion/filter.h
junr03: stale copy paste. Will refresh comment

library/common/extensions/filters/http/assertion/filter.proto
junr03: Right, that would need a change to the upstream proto.

library/common/extensions/filters/http/assertion/filter.cc
```diff
@@ -0,0 +1,79 @@
...
+                                         absl::nullopt, "");
+      return Http::FilterHeadersStatus::StopIteration;
+    }
+    return Http::FilterHeadersStatus::Continue;
+  }
+
+  decoder_callbacks_->sendLocalReply(Http::Code::BadRequest,
+                                     "Request Headers do not match configured expectations",
+                                     nullptr, absl::nullopt, "");
+  return Http::FilterHeadersStatus::StopIteration;
```
goaway, line 42: I would probably suggest an inverted control flow, which I think simplifies the reading of the logic, and would also help you avoid some nesting.
In other words:
if not matches -> 400
if end_stream -> 200
continue

library/common/extensions/filters/http/assertion/filter.cc
```diff
@@ -0,0 +1,79 @@
...
+          Http::Code::OK, "Request Body match configured expectations", nullptr, absl::nullopt, "");
+      return Http::FilterDataStatus::StopIterationNoBuffer;
+    }
+    return Http::FilterDataStatus::Continue;
+  }
+
+  decoder_callbacks_->sendLocalReply(Http::Code::BadRequest,
+                                     "Request Body does not match configured expectations", nullptr,
+                                     absl::nullopt, "");
+  return Http::FilterDataStatus::StopIterationNoBuffer;
```
goaway, line 59: Again, I would suggest inverting the control flow here, and removing the nesting.

library/common/extensions/filters/http/assertion/filter.cc
```diff
@@ -0,0 +1,79 @@
...
+  if (config_->rootMatcher().matchStatus(statuses_).matches_) {
+    decoder_callbacks_->sendLocalReply(Http::Code::OK,
+                                       "Request Trailers match configured expectations", nullptr,
+                                       absl::nullopt, "");
+  } else {
+    decoder_callbacks_->sendLocalReply(Http::Code::BadRequest,
+                                       "Request Trailers do not match configured expectations",
+                                       nullptr, absl::nullopt, "");
+  }
+  return Http::FilterTrailersStatus::StopIteration;
```
goaway, line 73: Same as above, also, additional `return` instead of `else`.

test/common/extensions/filters/http/assertion/assertion_filter_test.cc
```diff
@@ -0,0 +1,167 @@
...
+
+TEST_F(AssertionFilterTest, DataNoMatch) {
+  setUpFilter(R"EOF(
+match_config:
+  http_request_generic_body_match:
+    patterns:
+      - string_match: match_me
+)EOF");
+
+  Buffer::InstancePtr body{new Buffer::OwnedImpl("dont_match_ma")};
```
buildbreaker, line 121:
```suggestion
  Buffer::InstancePtr body{new Buffer::OwnedImpl("dont_match_me")};
```

library/common/extensions/filters/http/assertion/filter.cc
```diff
@@ -0,0 +1,79 @@
...
+    const envoymobile::extensions::filters::http::assertion::Assertion& proto_config) {
+  Common::Tap::buildMatcher(proto_config.match_config(), matchers_);
+}
+
+Extensions::Common::Tap::Matcher& AssertionFilterConfig::rootMatcher() const {
+  ASSERT(!matchers_.empty());
+  return *matchers_[0];
+}
+
+AssertionFilter::AssertionFilter(AssertionFilterConfigSharedPtr config) : config_(config) {
```
goaway, line 21: Can we add a docstring to this? It would also be really helpful to link the matcher documentation (whatever there is).

library/common/extensions/filters/http/assertion/filter.h
```diff
@@ -0,0 +1,50 @@
...
+class AssertionFilter final : public Http::PassThroughFilter {
+public:
+  AssertionFilter(AssertionFilterConfigSharedPtr config);
+
+  // StreamDecoderFilter
+  Http::FilterHeadersStatus decodeHeaders(Http::RequestHeaderMap& headers,
+                                          bool end_stream) override;
+  Http::FilterDataStatus decodeData(Buffer::Instance& data, bool end_stream) override;
+  Http::FilterTrailersStatus decodeTrailers(Http::RequestTrailerMap& trailers) override;
+
```
goaway, line 41: Can we validate response as well? Do matchers work on responses?

The current implementation generates 200s on end_stream, but that could easily be handled by a different upstream filter. Meanwhile it would allow testing of, e.g., modification of responses via platform filters.

library/common/extensions/filters/http/assertion/filter.cc
junr03: Already exists in the header.

library/common/extensions/filters/http/assertion/filter.h
junr03: Yeah the matcher can match responses. My thought was that for the initial commit I would do request validation as that is enough expressiveness to fulfill the e2e tests I had in mind for client->bridge->core interactions. The filter validates the request path and the client can validate what is returned on the response path.

So my plan was to land this, and then as you start writing tests we can modify the filter for added complexity.

library/common/extensions/filters/http/assertion/filter.cc
junr03: Sure, can do

library/common/extensions/filters/http/assertion/filter.cc
goaway: Yep, you're right. Can you still add a link to documentation on matchers?

library/common/extensions/filters/http/assertion/filter.h
goaway: Sounds good. 👍

library/common/extensions/filters/http/assertion/filter.cc
junr03: The proto is linked. I think the larger issue is that we need to start generating docs from our protos, cut this issue to track: https://github.com/lyft/envoy-mobile/issues/1001

junr03:
@goaway this is up-to-date, do you mind reviewing?

.github/workflows/coverage.yml
```diff
@@ -13,7 +13,7 @@ jobs:
   coverage:
     name: coverage
     runs-on: ubuntu-18.04
-    timeout-minutes: 60
+    timeout-minutes: 120
```
goaway, line 16: 😢

junr03: Not needed. I just wanted to run for longer because I am actually befuddled about why the coverage run is failing :( it passes locally, and it is failing on the envoy_mobile_main_common_test which I do not touch at all in the PR

junr03:
going to merge and see if the coverage failure remains in main.

PR #996, opened 2020-08-01
android: initial commit for local jar execution

junr03:
Description: in preparation for tests using the native source via #995 
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


junr03:
@buildbreaker @goaway can you take a look at this?

library/common/jni/jni_interface.cc
```diff
@@ -1,19 +1,18 @@
...
-#include <ares.h>
 #include <jni.h>
 
 #include <string>
 
+#include "library/common/jni/jni_utility.h"
 #include "library/common/main_interface.h"
 
-static JavaVM* static_jvm = nullptr;
+JavaVM* global_jvm = nullptr;
```
goaway, line 8: I have some misgivings about making this global. Even a static is a little bit iffy. Can we at least wrap it in a read-only accessor?

goaway: The fact that everyone should be calling `get_env()` is all the more reason to me to hide, rather than expose it.

library/common/jni/BUILD
```diff
@@ -19,3 +22,18 @@ cc_binary(
...
+    name = "java_libenvoy_jni.so",
+    srcs = [
+        "java_utility.cc",
+        "jni_interface.cc",
+        "jni_utility.h",
+    ],
+    includes = [
+        "../../../external/local_jdk/include",
+        "../../../external/local_jdk/include/darwin",
+        "../../../external/local_jdk/include/linux",
```
goaway, line 36: Will this relative path break if things move around? Can it be an absolute path?

library/common/jni/jni_interface.cc
```diff
@@ -60,28 +59,22 @@ extern "C" JNIEXPORT jint JNICALL
...
                                                                    jclass, // class
                                                                    jobject connectivity_manager) {
-  // See note above about c-ares.
-  // c-ares jvm init is necessary in order to let c-ares perform DNS resolution in Envoy.
-  // More information can be found at:
-  // https://c-ares.haxx.se/ares_library_init_android.html
-  ares_library_init_jvm(static_jvm);
-
-  return ares_library_init_android(connectivity_manager);
+  return platform_setup(global_jvm, connectivity_manager);
```
goaway, line 62: Why is this needed? This is already an Android-specific function.

PR #997, opened 2020-08-03
filters: update iOS interface to take explicit filter name

goaway:
Description: This avoids generics and some additional complexity on the Kotlin side. Ultimately, we may move to a solution that removes the need to specify a name entirely.
Risk Level: Low
Testing: Local and CI

### DIFF
--- a/library/swift/src/StreamClientBuilder.swift
+++ b/library/swift/src/StreamClientBuilder.swift
@@ -110,6 +110,7 @@ public final class StreamClientBuilder: NSObject {
 
   /// Add an HTTP filter factory used to construct filters for streams sent by this client.
   ///
+  /// - parameter filterName: Unique name for this filter required for registration.
   /// - parameter factory: Closure returning an instantiated filter. Called once per stream.
   ///
   /// - returns: This builder.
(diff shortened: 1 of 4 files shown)


rebello95:
This looks fine to me, though looking over it with fresh eyes I'm wondering if it'd make sense to simply generate a random string for each registered filter factory and use that, rather than putting that burden on the caller (if we expect each factory to be uniquely identifiable). What do you think?

It's also worth noting that consumers could theoretically return different `Filter` types between invocations of the filter factory's closure, so we should take that into consideration as well

goaway:
Yep, that was actually the suggestion I made in standup Friday. Actually, I think a String with an incremented value would be even better, since it will avoid the chance of a random collision. It won't work here, because we need dynamic configuration to land first though.

goaway:
What I can do, is make that part of the dynamic configuration task.

library/swift/src/StreamClientBuilder.swift
```diff
@@ -114,8 +114,9 @@ public final class StreamClientBuilder: NSObject {
   ///
```
rebello95, line 114: Can we document `filterName` with what we expect the user to specify (i.e., uniqueness)?

goaway: Sure thing.

PR #1000, opened 2020-08-04
swift: add some missing ResponseFilter functions

rebello95:
These are specified in the Kotlin interface, but not in Swift. They'll be wired up separately.


goaway:
I had them for now in Swift, until we can add support. I hadn't gotten to the point of removing them yet in Kotlin, but wouldn't that be preferable?

rebello95:
Given that `RequestFilterCallbacks` and `ResponseFilterCallbacks` are also not wired up but are in the interfaces, I think it's okay to add these with the intent of supporting them in the next week or two, to match what Android has on main and to exemplify the intended functionality

goaway:
Okay, sure, we can add them back in.

PR #1002, opened 2020-08-05
filters: wire Android invocation path

goaway:
Description: Adds invocation path for Android platform-bridged filters, covering request/response headers, data, and trailers. This change does not provide for state modification via the return path.
Risk Level: Moderate
Testing: Local and CI

### DIFF
--- a/examples/kotlin/hello_world/MainActivity.kt
+++ b/examples/kotlin/hello_world/MainActivity.kt
@@ -13,6 +13,7 @@ import io.envoyproxy.envoymobile.Engine
 import io.envoyproxy.envoymobile.RequestHeadersBuilder
 import io.envoyproxy.envoymobile.RequestMethod
 import io.envoyproxy.envoymobile.UpstreamHttpProtocol
+import io.envoyproxy.envoymobile.helloenvoykotlin.DemoFilter
 import io.envoyproxy.envoymobile.shared.Failure
 import io.envoyproxy.envoymobile.shared.ResponseRecyclerViewAdapter
 import io.envoyproxy.envoymobile.shared.Success
@@ -37,6 +38,8 @@ class MainActivity : Activity() {
     setContentView(R.layout.activity_main)
 
     engine = AndroidEngineBuilder(application).build()
+      .addFilter({ DemoFilter() })
+      .build()
 
     recyclerView = findViewById(R.id.recycler_view) as RecyclerView
     recyclerView.layoutManager = LinearLayoutManager(this)
--- a/examples/kotlin/hello_world/MainActivity.kt
+++ b/examples/kotlin/hello_world/MainActivity.kt
@@ -37,7 +37,7 @@ class MainActivity : Activity() {
     super.onCreate(savedInstanceState)
     setContentView(R.layout.activity_main)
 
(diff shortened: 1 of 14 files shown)


examples/kotlin/hello_world/MainActivity.kt
```diff
@@ -36,7 +37,9 @@ class MainActivity : Activity() {
     super.onCreate(savedInstanceState)
     setContentView(R.layout.activity_main)
 
-    streamClient = AndroidStreamClientBuilder(application).build()
+    streamClient = AndroidStreamClientBuilder(application)
+      .addFilter({ DemoFilter() })
```
buildbreaker, line 41:
```suggestion
      .addFilter { DemoFilter() }
```

You don't have to add the parens for this case

library/java/src/io/envoyproxy/envoymobile/engine/JvmFilterContext.java
```diff
@@ -0,0 +1,104 @@
...
+
+import java.nio.ByteBuffer;
+import java.util.Map;
+
+import io.envoyproxy.envoymobile.engine.types.EnvoyHTTPFilter;
+
+/**
+ * Wrapper class for EnvoyHTTPFilter for receiving JNI calls.
+ */
+class JvmFilterContext {
```
buildbreaker, line 11: Is it possible for us to add unit tests to just ensure the the filters are called appropriately?

goaway: https://github.com/lyft/envoy-mobile/pull/1024

buildbreaker: Could we also have a `JvmFilterContext` unit test which helps us with exercising the interaction between the `JvmBridgeUtility` and `EnvoyHTTPFilter`?

goaway: Stepping back for a second, I'm honestly skeptical of the value of **unit** testing functions that purely perform delegation. There's no real black box logic to test; the function makes another call, and you implement that call on one hand and then assert you've implemented it on the other. In my opinion, integration testing is a much better mechanism for testing for correct wiring/coupling of layers.

Translation on the other hand, I think merits unit testing - which is partly why I've extracted logic into the `JvmBridgeUtility` (and also the `JniBridgeUtility` in the upcoming PR).

buildbreaker: That's fair. I would like to capture and exercise the delegation path somewhere.

I assume you have in mind an integration test which would exercise this?

buildbreaker:
Looks good. Let me know once the formatter is green and I can approve

library/java/src/io/envoyproxy/envoymobile/engine/JvmFilterContext.java
goaway: Yes, I'd like to cover the paths through the layer wiring using the AssertionFilter that @junr03 has implemented.

goaway:
Tested example Swift app (which is failing in CI, but should be entirely unrelated anyways) locally.

PR #1006, opened 2020-08-06
filters: add dynamically-generated configuration

goaway:
Description: Adds dynamically-generated configuration for platform filter chains, allowing for registration of custom filters.
Risk Level: Moderate
Testing: Local and CI

### DIFF
--- a/library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
+++ b/library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
@@ -53,6 +53,7 @@ public class EnvoyConfiguration {
   String resolveTemplate(String templateYAML) {
     String resolvedConfiguration =
         templateYAML.replace("{{ stats_domain }}", statsDomain)
+            .replace("{{ platform_filter_chain }}", "")
             .replace("{{ connect_timeout_seconds }}", String.format("%s", connectTimeoutSeconds))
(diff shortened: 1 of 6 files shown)


library/objective-c/EnvoyConfiguration.m
```diff
@@ -33,7 +33,17 @@ - (instancetype)initWithStatsDomain:(NSString *)statsDomain
 }
 
 - (nullable NSString *)resolveTemplate:(NSString *)templateYAML {
+  NSString *filterConfigChain = [[NSString alloc] init];
```
rebello95, line 36: Change LGTM, can we add a test?

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -53,6 +53,7 @@ public EnvoyConfiguration(String statsDomain, int connectTimeoutSeconds, int dns
   String resolveTemplate(String templateYAML) {
     String resolvedConfiguration =
         templateYAML.replace("{{ stats_domain }}", statsDomain)
+            .replace("{{ platform_filter_chain }}", "")
```
rebello95, line 56: Should probably add a TODO here

goaway:
The coverage failure is for `envoy_mobile_main_common_test` which this doesn't touch, so I don't plan to block on it here. Will merge when asan passes.

PR #1010, opened 2020-08-07
coverage: fix main after #995

junr03:
Description: try to further debug why `envoy_mobile_main_common_test` is failing on coverage after #995. As [explained in that PR](https://github.com/lyft/envoy-mobile/pull/995#discussion_r466731628) I am befuddled by why the test is failing, specially since it is not failing in my local linux env.


junr03:
fixed in #1014

PR #1013, opened 2020-08-07
platform: automatically assign filter names

rebello95:
Since we essentially just want to enforce that each filter has a unique name in the registry, we can move this burden into the library and avoid making consumers specify filter names directly.


goaway:
I think this is fine. I do prefer the simplified API. Thanks @rebello95!

Issue #1016, opened 2020-08-10
crash: resetStream post body


junr03:
```
 Lyft:0 - Envoy::Http::ConnectionManagerImpl::doDeferredStreamDestroy(Envoy::Http::ConnectionManagerImpl::ActiveStream&)
    Lyft:0 - Envoy::Http::ConnectionManagerImpl::ActiveStream::onResetStream(Envoy::Http::StreamResetReason, std::__1::basic_string_view<cha...
    Lyft:0 - std::__1::__function::__func<Envoy::Http::Dispatcher::resetStream(long)::$_4, std::__1::allocator<Envoy::Http::Dispatcher::rese...
    Lyft:0 - Envoy::Event::DispatcherImpl::runPostCallbacks()
    Lyft:0 - event_process_active_single_queue
    Lyft:0 - event_base_loop
    Lyft:0 - Envoy::Server::InstanceImpl::run()
    Lyft:0 - Envoy::MainCommonBase::run()
    Lyft:0 - Envoy::Engine::run(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::basic_strin...
    Lyft:0 - void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__...
```

seen with a client release with this diff: https://github.com/lyft/envoy-mobile/compare/7c1e6149035953ca568830fb4b69afd2b105a9f7...da521e4ab4ae631e5884eca0c11b5db22e5f0444

Issue #1017, opened 2020-08-10
crash: doDeferredStreamDestroy


junr03:
```
 Lyft:0 - Envoy::Http::ConnectionManagerImpl::doDeferredStreamDestroy(Envoy::Http::ConnectionManagerImpl::ActiveStream&)
    Lyft:0 - Envoy::Http::ConnectionManagerImpl::ActiveStream::onResetStream(Envoy::Http::StreamResetReason, std::__1::basic_string_view<cha...
    Lyft:0 - std::__1::__function::__func<Envoy::Http::Dispatcher::resetStream(long)::$_4, std::__1::allocator<Envoy::Http::Dispatcher::rese...
    Lyft:0 - Envoy::Event::DispatcherImpl::runPostCallbacks()
    Lyft:0 - event_process_active_single_queue
    Lyft:0 - event_base_loop
    Lyft:0 - Envoy::Server::InstanceImpl::run()
    Lyft:0 - Envoy::MainCommonBase::run()
    Lyft:0 - Envoy::Engine::run(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::basic_strin...
    Lyft:0 - void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__...
```

seen with a client release with this diff: https://github.com/lyft/envoy-mobile/compare/7c1e6149035953ca568830fb4b69afd2b105a9f7...da521e4ab4ae631e5884eca0c11b5db22e5f0444

junr03:
In [`Http::Dispatcher::resetStream`](https://github.com/lyft/envoy-mobile/blob/a9f1a81996ad76321a0411afb8069dfe09215b13/library/common/http/dispatcher.cc#L453) we asserted that the DirectStream needs to live _at least_ as long as the ActiveStream. And we made the assumption that dispatching a post to the Event::Dispatcher which would release the DirectStream __after__ the `deferredDelete` of the ActiveStream happened would __necessarily__ mean that the ActiveStream would be destroyed before the DirectStream. 

[...]

PR #1018, opened 2020-08-10
crash fix: ensure main thread is destructed before Http::Dispatcher

goaway:
Description: Due to inverse destruction order, it was previously theoretically possible that the main thread could still be running and executing the event loop when the Http::Dispatcher had already been destroyed. This ensures that the normal destruction order of the Engine should ensure the main thread has been destructed (and terminated) before the dispatcher. Our presumption that the crash is called by `this` being null in the callback is covered by a new `RELEASE_ASSERT`, so that we can detect if further protection (e.g. capturing the dispatcher via a shared pointer in the callback) is needed.
Risk Level: Low
Testing: CI

### DIFF
--- a/library/common/engine.h
+++ b/library/common/engine.h
@@ -48,12 +48,12 @@ private:
   envoy_engine_callbacks callbacks_;
   Thread::MutexBasicLockable mutex_;
   Thread::CondVar cv_;
-  std::thread main_thread_;
   std::unique_ptr<Http::Dispatcher> http_dispatcher_;
   std::unique_ptr<MobileMainCommon> main_common_ GUARDED_BY(mutex_);
   Server::Instance* server_{};
   Server::ServerLifecycleNotifier::HandlePtr postinit_callback_handler_;
   Event::Dispatcher* event_dispatcher_;
+  std::thread main_thread_;
 };
 
 } // namespace Envoy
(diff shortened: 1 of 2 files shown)


junr03:
Potentially fixes #1016 and #1017

library/common/engine.h
```diff
@@ -48,12 +48,12 @@ class Engine {
   envoy_engine_callbacks callbacks_;
   Thread::MutexBasicLockable mutex_;
   Thread::CondVar cv_;
-  std::thread main_thread_;
   std::unique_ptr<Http::Dispatcher> http_dispatcher_;
   std::unique_ptr<MobileMainCommon> main_common_ GUARDED_BY(mutex_);
   Server::Instance* server_{};
   Server::ServerLifecycleNotifier::HandlePtr postinit_callback_handler_;
   Event::Dispatcher* event_dispatcher_;
+  std::thread main_thread_;
```
rebello95, line 56: Can we document this inline somewhere? We're now relying on implementation details of how destructors are called, which is very fragile. Tests would be ideal, but I understand this is probably quite difficult to test, so we should at least document it

junr03: > We're now relying on implementation details of how destructors are called, which is very fragile

For better or worse this is a common c++ pattern.

> Tests would be ideal

As always agreed. I had the idea for this crash "fix" but I am not entirely sure that the situation described here with instructions on the main_thread being executed after destruction of other objects is the actual problem. I tried writing repro tests but was not able to set it up. I have a related PR with engine tests coming up imminently that repros a different trace. I can try an massage that to repro this as well.

> Can we document this inline somewhere

Absolutely, this should be in the comments. Updated.

rebello95:
Re-running Mac CI

PR #1020, opened 2020-08-11
stats: add scope for external stats

goaway:
Description: Use a dedicated Stats::Scope for external stats.
Risk Level: Low
Testing: Local

### DIFF
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -95,7 +95,7 @@ Engine::~Engine() {
 }
 
 void Engine::recordCounter(std::string elements, uint64_t count) {
-  if (server_ && scope_) {
+  if (server_ && external_scope_) {
     server_->dispatcher().post([this, elements, count]() -> void {
       absl::string_view dynamic_elements{elements};
       Stats::Utility::counterFromElements(*external_scope_, {dynamic_elements}).add(count);
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -50,7 +50,7 @@ envoy_status_t Engine::run(std::string config, std::string log_level) {
     postinit_callback_handler_ = main_common_->server()->lifecycleNotifier().registerCallback(
         Envoy::Server::ServerLifecycleNotifier::Stage::PostInit, [this]() -> void {
           server_ = TS_UNCHECKED_READ(main_common_)->server();
-          external_scope_ = server_->serverFactoryContext().scope().createScope("client.");
+          client_scope_ = server_->serverFactoryContext().scope().createScope("client.");
           auto api_listener = server_->listenerManager().apiListener()->get().http();
           ASSERT(api_listener.has_value());
           http_dispatcher_->ready(server_->dispatcher(), server_->serverFactoryContext().scope(),
@@ -95,10 +95,10 @@ Engine::~Engine() {
 }
 
 void Engine::recordCounter(std::string elements, uint64_t count) {
-  if (server_ && external_scope_) {
+  if (server_ && client_scope_) {
     server_->dispatcher().post([this, elements, count]() -> void {
       absl::string_view dynamic_elements{elements};
-      Stats::Utility::counterFromElements(*external_scope_, {dynamic_elements}).add(count);
+      Stats::Utility::counterFromElements(*client_scope_, {dynamic_elements}).add(count);
     });
   }
 }
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -50,7 +50,7 @@ envoy_status_t Engine::run(std::string config, std::string log_level) {
     postinit_callback_handler_ = main_common_->server()->lifecycleNotifier().registerCallback(
         Envoy::Server::ServerLifecycleNotifier::Stage::PostInit, [this]() -> void {
           server_ = TS_UNCHECKED_READ(main_common_)->server();
-          scope_ = server_->serverFactoryContext().scope().createScope("client.");
+          external_scope_ = server_->serverFactoryContext().scope().createScope("client.");
           auto api_listener = server_->listenerManager().apiListener()->get().http();
           ASSERT(api_listener.has_value());
           http_dispatcher_->ready(server_->dispatcher(), server_->serverFactoryContext().scope(),
@@ -98,7 +98,7 @@ void Engine::recordCounter(std::string elements, uint64_t count) {
   if (server_ && scope_) {
     server_->dispatcher().post([this, elements, count]() -> void {
       absl::string_view dynamic_elements{elements};
-      Stats::Utility::counterFromElements(*scope_, {dynamic_elements}).add(count);
+      Stats::Utility::counterFromElements(*external_scope_, {dynamic_elements}).add(count);
     });
   }
 }
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -97,7 +97,8 @@ Engine::~Engine() {
 void Engine::recordCounter(std::string elements, uint64_t count) {
   if (server_ && client_scope_) {
     server_->dispatcher().post([this, elements, count]() -> void {
-      Stats::Utility::counterFromElements(*client_scope_, {Stats::DynamicName(elements)}).add(count);
+      Stats::Utility::counterFromElements(*client_scope_, {Stats::DynamicName(elements)})
+          .add(count);
     });
   }
 }
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -94,7 +94,7 @@ Engine::~Engine() {
   main_thread_.join();
 }
 
-void Engine::recordCounter(std::string elements, uint64_t count) {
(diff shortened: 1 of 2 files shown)


rebello95:
Can you provide more details in the description as to what the intent of this change is?

goaway:
This avoids the static init fiasco callout @jmarantz made on the previous PR (which we missed), and also adopts the preferred mechanism for applying a fixed prefix to a group of stats.

goaway:
For reference: https://github.com/lyft/envoy-mobile/pull/992#discussion_r468209683

library/common/engine.cc
```diff
@@ -94,14 +95,10 @@ Engine::~Engine() {
 }
 
 void Engine::recordCounter(std::string elements, uint64_t count) {
-  if (server_) {
+  if (server_ && external_scope_) {
     server_->dispatcher().post([this, elements, count]() -> void {
-      static const std::string client = "client";
-      absl::string_view prefix{client};
       absl::string_view dynamic_elements{elements};
```
junr03, line 100: nit: can we check if the string view can be built in place in the call below?

library/common/engine.cc
```diff
@@ -94,14 +95,10 @@ Engine::~Engine() {
 }
 
 void Engine::recordCounter(std::string elements, uint64_t count) {
-  if (server_) {
+  if (server_ && external_scope_) {
```
junr03, line 98: maybe we can asset `external_scope_`? Because if the server is there but the scope is not that seems odd?

library/common/engine.h
```diff
@@ -50,6 +50,7 @@ class Engine {
 private:
   envoy_status_t run(std::string config, std::string log_level);
 
+  Stats::ScopePtr external_scope_;
```
junr03, line 53:
```suggestion
  Stats::ScopePtr stats_client_scope_;
```

library/common/engine.cc
goaway: I chose to assert both because the `server_` is still the sentinel for the dispatcher event loop, but the `Scope` could be constructed independently, and if someone changed it to be, it would silently cause a conditional on `server_` to no longer guard against the null pointer deference.

In other words, the conditional guards exactly the two pointers that are dereferenced, and it's an unenforced detail that one happens to depend on the other right now.

library/common/engine.cc
goaway: I checked it; it doesn't work. In fact, there's another transparent type being created here, which is an `const InlinedVector<variant<Envoy::Stats::StatName, Envoy::Stats::DynamicName>`.

library/common/engine.cc
```diff
@@ -94,14 +95,10 @@ Engine::~Engine() {
 }
 
 void Engine::recordCounter(std::string elements, uint64_t count) {
```
jmarantz, line 97: it's weird to have a scalar string called `elements`, but in any case I suggest passing it by const ref, or better yet, by string_view.

But IMO the higher level issue is that if you define your API to inc a counter based on a string, you are going to have to pay the cost of map lookups and/or locks. Is it not possible for you to have the Engine API offer a wrapper to create a handle that the caller can store, to quickly increment a counter?

library/common/engine.cc
```diff
@@ -94,14 +95,10 @@ Engine::~Engine() {
...
-  if (server_) {
+  if (server_ && client_scope_) {
     server_->dispatcher().post([this, elements, count]() -> void {
-      static const std::string client = "client";
-      absl::string_view prefix{client};
       absl::string_view dynamic_elements{elements};
-      Stats::Utility::counterFromElements(server_->serverFactoryContext().scope(),
-                                          {prefix, dynamic_elements})
-          .add(count);
+      Stats::Utility::counterFromElements(*client_scope_, {dynamic_elements}).add(count);
```
jmarantz, line 101: I would not have expected this to work, but I see the problem. I meant to make DynamicName an `explicit` but allow the implicit conversion from string literal to string_view.

Can you change this to:
```
  Stats::Utility::counterFromElements(*client_scope_, {Stats::DynamicName(elements)}).add(count);
```
I'm going to write a quick PR to add the `explicit` qualifier to the DynamicName constructor which would make the form that you have here invalid.

The reason is that it is an anti-pattern to have a ton of DynamicName objects (due to string bloat) so I want to be able to search for them.

Of course the better solution is really to get callers to create a Counter and keep a handle to it, rather than doing the name lookup all the time, but if you can't do that then please add the explicit qualifier.

Then also you will not need the string_view declaration above.

library/common/engine.cc
```diff
@@ -94,14 +95,10 @@ Engine::~Engine() {
 }
 
 void Engine::recordCounter(std::string elements, uint64_t count) {
-  if (server_) {
+  if (server_ && client_scope_) {
     server_->dispatcher().post([this, elements, count]() -> void {
```
jmarantz, line 99: why the post() here rather than just incrementing the counter directly at the call-site?

counters are thread-safe, at least once they are looked up.

Maybe the issue is that the caller is in some thread that Envoy does not know about, so the TLS lookups underneath fail? In that case can you comment?

library/common/engine.cc
goaway: I think this is ideally what we'd like to do, but we were going for something quick to iterate on here, and lacked the familiarity with the stats API to figure out how best to accomplish this. Would welcome your perspective on a sounder approach.

library/common/engine.cc
goaway: The issue is exactly what you suggest - this call is coming from a random application thread, Envoy has no knowledge of it, and there's no TLS state associated with it. I can definitely add a comment, though that's actually a common theme at this layer of the API.

goaway:
Thank you for the comments @jmarantz! Very much welcome your thoughts on how to improve our approach.

library/common/engine.cc
jmarantz: No worries just add a comment explaining the non-obvious delay.

goaway: On the other hand, if you see a way we could accomplish this without the dispatch in a thread-safe manner, would certainly be interested in hearing more.

goaway:
Synced with @jmarantz who graciously provided his thoughts on how to move to an implementation with reduced overhead. Opening an issue to track further improvements that can be made here: https://github.com/lyft/envoy-mobile/issues/1034

library/common/engine.cc
jmarantz: I don't think you can without changes to ThreadLocalStore...so we'll have to be content with a comment.

Also when you establish an API for creation of a Counter object for callers, the same issue will be there. You will have to populate the underlying Counter objects from a dispatch.

I'll defer to @mattklein123 on what the best TLS strategy should be for accessing stat by name outside the context of worker threads. But in the current code I think it has to be done via post.

library/common/engine.cc
junr03:
```suggestion
void Engine::recordCounter(const std::string& elements, uint64_t count) {
```

per @jmarantz.

library/common/engine.cc
```diff
@@ -93,14 +94,10 @@ Engine::~Engine() {
   main_thread_.join();
 }
 
-void Engine::recordCounter(std::string elements, uint64_t count) {
-  if (server_) {
+void Engine::recordCounter(const std::string& elements, uint64_t count) {
```
jmarantz, line 97: now that I see this in its current form, it's worth a quick check whether 'elements' value gets copied into the lambda provided to post, or whether it holds a ref to an object that might potentially not live long enough for the dispatcher to execute the post.

It might not matter in your unit test, where the lifetime of the string you pass to recordCounter is long enough.

In any case I think the interface to this function should be a const ref, but you may need to copy it explicitly into storage saved in the lambda. Can you check?

goaway: Thanks @jmarantz, good callout.

junr03: Yeah good point! 🤦 the string could definitely be gone as the copy is just in the stack of the main interface function call. What's the least egrigious way to minimize copies:
- pass down the const char* all the way until we capture a copy in the lambda,
- create a unique ptr of a string at the main interface level that we pass down,
- something else?

jmarantz: My recommendation:
 * leave the interface as is (or better yet, witch to a string_view to avoid copying while calling recordValue)
 * make a local copy in the function body (`std::string name = elements;`)
 * capture that copy by value in the lambda passed to `post()`.

jmarantz: My apologies, what you have is fine. I tested it with:
```
TEST(Foo, Bar) {
  std::string foo = "foo";
  std::string& fooref = foo;
  auto test = [fooref]() { EXPECT_EQ("foo", fooref); };
  test();
  foo = "bar";
  test();
}
```

That test passes, which means that the ref's value is captured by the lambda.

goaway:
thanks @jmarantz, @junr03 and @jingwei99!

PR #1021, opened 2020-08-11
shutdown: create avenues for shutdown out of static de-initialization

### LINKED ISSUE #831: crash: trying to use garbage/null pointer

junr03:
Description: this PR attempts to fix two related crashes. By hooking up to iOS's notification system we attempt to terminate the engine earlier, and allow all objects necessary to be available fixing #1035. Potentially fixes #831.
Risk Level: med - changes termination sequence
Testing: added test


library/common/engine.cc
```diff
@@ -21,17 +21,21 @@ Engine::Engine(envoy_engine_callbacks callbacks, const char* config, const char*
...
   {
     Thread::LockGuard lock(mutex_);
     try {
-      char* envoy_argv[] = {strdup("envoy"), strdup("--config-yaml"),   strdup(config.c_str()),
-                            strdup("-l"),    strdup(log_level.c_str()), nullptr};
+      const std::string name = "envoy";
+      const std::string config_flag = "--config-yaml";
+      const std::string log_flag = "-l";
+      const char* envoy_argv[] = {name.c_str(),     config_flag.c_str(), config.c_str(),
+                                  log_flag.c_str(), log_level.c_str(),   nullptr};
```
goaway, line 32: Can you provide some motivation here?

library/common/engine.cc
```diff
@@ -60,13 +64,14 @@ envoy_status_t Engine::run(std::string config, std::string log_level) {
...
 
-  // The above call is blocking; at this point the event loop has exited.
-  callbacks_.on_exit();
-
   // Ensure destructors run on Envoy's main thread.
+  http_dispatcher_->exit();
   postinit_callback_handler_.reset();
   TS_UNCHECKED_READ(main_common_).reset();
 
+  // The above call is blocking; at this point the event loop has exited.
```
goaway, line 72: This comment really applies to the event loop rather than the above calls now.

library/common/engine.cc
junr03: asan correctly complained about the engine test. the old strdup based implementation was leaking all these strings.

junr03:
@goaway any more comments here?

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -193,19 +195,18 @@ - (void)recordCounter:(NSString *)elements count:(NSUInteger)count {
...
 }
 
 - (void)lifecycleDidChangeWithNotification:(NSNotification *)notification {
-  NSLog(@"[Envoy] triggering stats flush (%@)", notification.name);
-  flush_stats();
+  NSLog(@"[Envoy] terminating engine (%@)", notification.name);
+  // TODO: ensure this is called with the correct envoy_engine_t once multiple Engine objects can
+  // be allocated.
+  // https://github.com/lyft/envoy-mobile/issues/332
+  terminate_engine(0);
```
goaway, line 209: Won't this terminate the engine on _any_ lifecycle notification? Or are we only registering for backgrounding/shutdown.

goaway: Also, we should have access to the engine handle here. No need to hardcode `0`.

library/common/engine.cc
```diff
@@ -60,14 +64,14 @@ envoy_status_t Engine::run(std::string config, std::string log_level) {
...
   bool run_success = TS_UNCHECKED_READ(main_common_)->run();
-
   // The above call is blocking; at this point the event loop has exited.
-  callbacks_.on_exit();
 
   // Ensure destructors run on Envoy's main thread.
   postinit_callback_handler_.reset();
   TS_UNCHECKED_READ(main_common_).reset();
 
+  callbacks_.on_exit(callbacks_.context);
```
goaway, line 73: I agree this ordering makes more sense.

library/objective-c/EnvoyEngineImpl.m
junr03: I deleted the observer for `UIApplicationWillResignActiveNotification` so it will only be on termination. ack on the engine handle, will update.

junr03:
@goaway updated!

junr03:
btw https://github.com/lyft/envoy-mobile/runs/1064413662 effectively repros the crash prior to #1067. Merging main to include #1067 gives me a green test 🎉

library/objective-c/EnvoyEngineImpl.m
goaway: I think the notification actually has the lifecycle event on it, so we could be more picky here. My worry is that if someone adds it back in, it won't be obvious they're doing something wrong, and it could cause networking to fail until the app is restarted.

library/common/engine.cc
goaway: We were leaking them intentionally before because we weren't sure how long Envoy needed them to stick around. This is fine though; they're on the stack anyways for the full duration of `run`.

goaway (approved):

goaway (approved):

PR #1024, opened 2020-08-12
coverage: add unit tests for JvmBridgeUtility

goaway:
Description: Adds unit-level coverage to the JvmBridgeUtility.

### DIFF
--- a/library/java/src/io/envoyproxy/envoymobile/engine/BUILD
+++ b/library/java/src/io/envoyproxy/envoymobile/engine/BUILD
@@ -36,3 +36,9 @@ java_library(
         "//library/java/src/io/envoyproxy/envoymobile/engine/types:envoy_c_types_lib",
     ],
 )
+
+filegroup(
+    name = "envoy_base_engine_lib_srcs",
+    srcs = ["JvmBridgeUtility.java"],
+    visibility = ["//visibility:public"],
+)
--- a/library/java/test/io/envoyproxy/envoymobile/engine/JvmBridgeUtilityTest.kt
+++ b/library/java/test/io/envoyproxy/envoymobile/engine/JvmBridgeUtilityTest.kt
@@ -21,7 +21,7 @@ class JvmBridgeUtilityTest {
   }
 
   @Test
-  fun `validateCount returns a boolean that comparison of the  number of header values in the map`() {
+  fun `validateCount checks if the expected number of header values in the map matches the actual`() {
     val utility = JvmBridgeUtility()
     utility.passHeader("test-0".toByteArray(), "value-0".toByteArray(), true)
     utility.passHeader("test-1".toByteArray(), "value-1".toByteArray(), false)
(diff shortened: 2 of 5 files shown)


library/java/src/io/envoyproxy/envoymobile/engine/BUILD
```diff
@@ -36,3 +36,9 @@ java_library(
         "//library/java/src/io/envoyproxy/envoymobile/engine/types:envoy_c_types_lib",
     ],
 )
+
+filegroup(
```
buildbreaker, line 40: This is unfortunate...

https://github.com/lyft/envoy-mobile/issues/1027

library/java/test/io/envoyproxy/envoymobile/engine/JvmBridgeUtilityTest.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    val headers = utility.retrieveHeaders()
+    val expectedHeaders = mapOf(
+      "test-0" to listOf("value-0"),
+      "test-1" to listOf("value-1", "value-2")
+    )
+
+    assertThat(headers)
+      .hasSize(2) // Two keys / header name
+      .usingRecursiveComparison().isEqualTo(expectedHeaders)
```
buildbreaker, line 23:
```suggestion
      .usingRecursiveComparison()
      .isEqualTo(expectedHeaders)
```

library/java/test/io/envoyproxy/envoymobile/engine/JvmBridgeUtilityTest.kt
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @Test
+  fun `validateCount checks if the expected number of header values in the map matches the actual`() {
```
buildbreaker, line 27: I think we can have 2 separate tests (potentially 3 but up to you):
1. `validateCount` returns the appropriate value (this can be split into the positive and negative cases but up to you)
2. `retrieveHeaders()` resets the internal count

Separate question: Wouldn't it be easier if the utility just returned the count value instead of requiring the user to ask it if the utility has N counts?

library/java/test/io/envoyproxy/envoymobile/engine/JvmBridgeUtilityTest.kt
goaway: These two are actually dependently-chained (and you couldn't, for instance, reverse the order of 22 and 23, since 23 changes the type) which was why I put them on one line. I don't feel terribly strongly about it though.

library/java/test/io/envoyproxy/envoymobile/engine/JvmBridgeUtilityTest.kt
goaway: We could do that. I don't know that it would necessarily be easier, but it's definitely a valid approach. I was trying to keep the interface scoped pretty narrowly, and nothing really has any business querying the accumulated count outside of validation purposes.

goaway: Oh also, I'm fine with splitting up the tests.

buildbreaker: Let's split them up :)

I think it'll be clearer

PR #1026, opened 2020-08-12
filters: wire Android return path

goaway:
Description: Wires the return path for Android filter invocations, allowing modification of the current HTTP entity.
Risk Level: Moderate
Testing: Local and CI

### DIFF
--- a/library/common/jni/jni_interface.cc
+++ b/library/common/jni/jni_interface.cc
@@ -275,7 +275,7 @@ jvm_http_filter_on_request_headers(envoy_headers headers, bool end_stream, const
   jobjectArray j_headers = static_cast<jobjectArray>(env->GetObjectArrayElement(result, 1));
 
   return (envoy_filter_headers_status){/*status*/ unbox_integer(env, status),
-                                        /*headers*/ to_native_headers(env, j_headers)};
+                                       /*headers*/ to_native_headers(env, j_headers)};
 }
 
 static envoy_filter_headers_status
@@ -288,7 +288,7 @@ jvm_http_filter_on_response_headers(envoy_headers headers, bool end_stream, cons
   jobjectArray j_headers = static_cast<jobjectArray>(env->GetObjectArrayElement(result, 1));
 
   return (envoy_filter_headers_status){/*status*/ unbox_integer(env, status),
-                                        /*headers*/ to_native_headers(env, j_headers)};
+                                       /*headers*/ to_native_headers(env, j_headers)};
 }
 
 static void* jvm_on_data(const char* method, envoy_data data, bool end_stream, void* context) {
--- a/library/common/jni/jni_interface.cc
+++ b/library/common/jni/jni_interface.cc
@@ -101,6 +101,100 @@ Java_io_envoyproxy_envoymobile_engine_AndroidJniLibrary_flushStats(JNIEnv* env,
   flush_stats();
 }
 
+// Utility functions
+static JNIEnv* get_env() {
+  JNIEnv* env = nullptr;
+  int get_env_res = static_jvm->GetEnv((void**)&env, JNI_VERSION);
+  if (get_env_res == JNI_EDETACHED) {
+    __android_log_write(ANDROID_LOG_VERBOSE, "[Envoy]", "environment is JNI_EDETACHED");
+    // Note: the only thread that should need to be attached is Envoy's engine std::thread.
+    // TODO: harden this piece of code to make sure that we are only needing to attach Envoy
+    // engine's std::thread, and that we detach it successfully.
+    static_jvm->AttachCurrentThread(&env, nullptr);
+    static_jvm->GetEnv((void**)&env, JNI_VERSION);
+  }
+  return env;
+}
+
+static void jni_delete_global_ref(void* context) {
+  JNIEnv* env = get_env();
+  jobject ref = static_cast<jobject>(context);
+  env->DeleteGlobalRef(ref);
+}
+
+static void jni_delete_const_global_ref(const void* context) {
+  jni_delete_global_ref(const_cast<void*>(context));
+}
+
+static int unbox_integer(JNIEnv* env, jobject boxedInteger) {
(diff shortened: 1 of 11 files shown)


library/common/jni/jni_interface.cc
```diff
@@ -101,6 +101,100 @@ Java_io_envoyproxy_envoymobile_engine_AndroidJniLibrary_flushStats(JNIEnv* env,
...
+
+  if (direct_address == nullptr) {
+    jclass jcls_ByteBuffer = env->FindClass("java/nio/ByteBuffer");
+    // We skip checking hasArray() because only direct ByteBuffers or array-backed ByteBuffers
+    // are supported. We will crash here if this is an invalid buffer, but guards may be
+    // implemented in the JVM layer.
+    jmethodID jmid_array = env->GetMethodID(jcls_ByteBuffer, "array", "()[B");
+    jbyteArray array = static_cast<jbyteArray>(env->CallObjectMethod(j_data, jmid_array));
+    return array_to_native_data(env, array);
+  }
```
goaway, line 155: The surrounding code was moved, but this block is new.

library/common/jni/jni_interface.cc
```diff
@@ -236,20 +324,28 @@ static void* jvm_on_response_data(envoy_data data, bool end_stream, void* contex
 
 static envoy_filter_data_status jvm_http_filter_on_request_data(envoy_data data, bool end_stream,
                                                                 const void* context) {
-  envoy_data return_data = copy_envoy_data(data.length, data.bytes);
-  jobject result = static_cast<jobject>(
+  JNIEnv* env = get_env();
+  jobjectArray result = static_cast<jobjectArray>(
```
goaway, line 328: Previously we ignored the `result`; now we extract its elements, map, and return them.

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyHTTPStream.java
```diff
@@ -76,17 +76,4 @@ public void sendTrailers(Map<String, List<String>> trailers) {
...
-    // value.
-    final List<byte[]> convertedHeaders = new ArrayList<byte[]>(2 * headers.size());
-    for (Map.Entry<String, List<String>> entry : headers.entrySet()) {
-      for (String value : entry.getValue()) {
-        convertedHeaders.add(entry.getKey().getBytes(StandardCharsets.UTF_8));
-        convertedHeaders.add(value.getBytes(StandardCharsets.UTF_8));
-      }
-    }
-    return convertedHeaders.toArray(new byte[0][0]);
-  }
```
goaway, line 91: Moved to utility class, which can be tested and reused.

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
```diff
@@ -5,14 +5,16 @@ import java.nio.ByteBuffer
 /*
  * Status to be returned by filters when transmitting or receiving data.
  */
-sealed class FilterDataStatus<T : Headers> {
+sealed class FilterDataStatus<T : Headers>(
+  val status: Int
```
goaway, line 9: Introduce a "static" constant value here that maps to Envoy's enum. It's not possible to directly set these from the native values, but I'd like to add a runtime check and/or test that verifies they're the same.

library/kotlin/src/io/envoyproxy/envoymobile/filters/Filter.kt
```diff
@@ -39,26 +39,60 @@ internal class EnvoyHTTPFilterAdapter(
...
-        is FilterHeadersStatus.StopIteration -> arrayOf(1 /*EnvoyHTTPHeadersStatusStopIteration*/, headers)
+        is FilterHeadersStatus.Continue -> arrayOf(result.status, result.headers.headers)
+        is FilterHeadersStatus.StopIteration -> arrayOf(result.status, headers)
       }
     }
     return arrayOf(0, headers)
   }
 
   override fun onRequestData(data: ByteBuffer, endStream: Boolean): Array<Any> {
+    (filter as? RequestFilter)?.let { requestFilter ->
```
buildbreaker, line 50: I feel like this would be clearer if we just have an if statement
```suggestion
    if (filter is RequestFilter) {
         ...
    }
    return arrayOf(0, data)
```

library/java/src/io/envoyproxy/envoymobile/engine/JniBridgeUtility.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Class to assist with passing types from the JVM to native code. Currently supports
+ * HTTP headers.
+ */
+final class JniBridgeUtility {
+
+  private JniBridgeUtility() {}
+
+  public static byte[][] toJniHeaders(Map<String, List<String>> headers) {
```
buildbreaker, line 16: Could we add a quick test here?

library/kotlin/src/io/envoyproxy/envoymobile/filters/Filter.kt
goaway: `let` parallels the Swift implementation (and convention) more closely, and I was somewhat under the impression that this was idiomatic(?). But I don't have strong feelings. Just wanted to ask if that swayed our preference at all.

library/java/src/io/envoyproxy/envoymobile/engine/JniBridgeUtility.java
goaway: Sure (that was part of why I split it out to begin with). As it's existing logic that's been moved, I'll admit a slight preference for doing that as a follow-up though. wdyt?

buildbreaker: Sounds good. Let's do it as a follow up

goaway:
Re-ran unit tests locally, and they all pass, but letting CI take another crack at it.

PR #1028, opened 2020-08-12
complex filters: add ResumeIteration status

goaway:


goaway:
Note the difference between the signatures with optional types in Swift and Kotlin. I introduced the discrepancy to allow consideration of both options: the approach in Swift allows extra entities to be defaulted to `nil`, while the approach in Kotlin orders the entities in terms of their invocation order or appearance on the stream.

PR #1029, opened 2020-08-12
advanced filters: introduce interface for more complex filter interactions

goaway:
Description: Introduces platform-bridged filter interfaces that allow stopping and stopping filter iteration, and asynchronous resumption of filter chain. The following conventions are adopted with these new interfaces:
- 'Continue' is the status used to indicate HTTP entities should be forwarded when filter iteration is ongoing.
- 'StopIteration' causes forwarding to halt (though invocations of the current filter will continue).
- The 'ResumeIteration' status or an asynchronous resume call must be used to begin iteration and forwarding again. 'Continue' is not valid when iteration is stopped.
- Using an asynchronous call to resume iteration will result in a special 'onResume' invocation of the filter, that will allow interaction with the HTTP stream on the same thread as other invocations, avoiding the need for synchronization.
- Resume mechanisms offer optional parameters attached to the return status that can be used to provide parts of the stream that have not yet been forwarded. e.g., if you resume during an 'onData' invocation, you can attach headers to the 'ResumeIteration' status (and must do so, if headers have not yet been forwarded).
Risk: Low
Testing: Local and CI

### DIFF
--- a/library/kotlin/src/io/envoyproxy/envoymobile/AsycRequestFilter.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/AsycRequestFilter.kt
@@ -5,19 +5,31 @@ package io.envoyproxy.envoymobile
  */
 interface AsyncRequestFilter : RequestFilter {
   /**
-   * Invoked explicitly in response to an asynchronous resumeRequest() callback when filter
-   * iteration has been stopped.
-   *
-   * @return: The resumption status including any previously held entities that remain
-   *          to be forwarded.
-   */
-  fun onResumeRequest(): FilterResumeStatus<RequestHeaders, RequestTrailers>
-
-  /**
    * Called by the filter manager once to initialize the filter callbacks that the filter should
    * use.
    *
    * @param callbacks: The callbacks for this filter to use to interact with the chain.
    */
    fun setRequestFilterCallbacks(callbacks: RequestFilterCallbacks)
+
+  /**
+   * Invoked explicitly in response to an asynchronous resumeRequest() callback when filter
+   * iteration has been stopped. The parameters passed to this invocation will be a snapshot
+   * of any stream state that has not yet been forwarded along the filter chain.
+   *
+   * As with other filter invocations, this will be called on Envoy's main thread, and thus
+   * no additional synchronization is required between this and other invocations.
+   *
+   * @param headers: Headers, if `StopIteration` was returned from `onResponseHeaders`.
+   * @param data: Any data that has been buffered where `StopIterationAndBuffer` was returned.
+   * @param trailers: Trailers, if `StopIteration` was returned from `onReponseTrailers`. 
+   *
+   * @return: The resumption status including any HTTP entities that will be forwarded.
+   */
+  fun onResumeRequest(
+    headers: RequestHeaders?,
+    data: ByteBuffer?
+    trailrs: RequestTrailers?
+    endStream: Boolean
+  ): FilterResumeStatus<RequestHeaders, RequestTrailers>
 }
--- a/library/kotlin/src/io/envoyproxy/envoymobile/AsycRequestFilter.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/AsycRequestFilter.kt
@@ -1,0 +1,23 @@
+package io.envoyproxy.envoymobile
+
+/*
+ * RequestFilter supporting asynchronous resumption.
+ */
+interface AsyncRequestFilter : RequestFilter {
+  /**
+   * Invoked explicitly in response to an asynchronous resumeRequest() callback when filter
+   * iteration has been stopped.
+   *
+   * @return: The resumption status including any previously held entities that remain
+   *          to be forwarded.
+   */
+  fun onResumeRequest(): FilterResumeStatus<RequestHeaders, RequestTrailers>
+
+  /**
+   * Called by the filter manager once to initialize the filter callbacks that the filter should
+   * use.
+   *
+   * @param callbacks: The callbacks for this filter to use to interact with the chain.
+   */
+   fun setRequestFilterCallbacks(callbacks: RequestFilterCallbacks)
+}
--- a/library/kotlin/src/io/envoyproxy/envoymobile/AsycRequestFilter.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/AsycRequestFilter.kt
@@ -20,9 +20,10 @@ interface AsyncRequestFilter : RequestFilter {
    * As with other filter invocations, this will be called on Envoy's main thread, and thus
    * no additional synchronization is required between this and other invocations.
    *
-   * @param headers: Headers, if `StopIteration` was returned from `onResponseHeaders`.
+   * @param headers: Headers, if `StopIteration` was returned from `onRequestHeaders`.
    * @param data: Any data that has been buffered where `StopIterationAndBuffer` was returned.
-   * @param trailers: Trailers, if `StopIteration` was returned from `onReponseTrailers`. 
+   * @param trailers: Trailers, if `StopIteration` was returned from `onRequestTrailers`.
+   * @param endStream: True, if the stream ended with the previous (and thus, last) invocation.
    *
    * @return: The resumption status including any HTTP entities that will be forwarded.
    */
--- a/library/kotlin/src/io/envoyproxy/envoymobile/AsycResponseFilter.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/AsycResponseFilter.kt
@@ -1,0 +1,23 @@
+package io.envoyproxy.envoymobile
+
+/*
+ * ResponseFilter supporting asynchronous resumption.
+ */
+interface AsyncResponseFilter : ResponseFilter {
+  /**
+   * Invoked explicitly in response to an asynchronous resumeResponse() callback when filter
+   * iteration has been stopped.
+   *
+   * @return: The resumption status including any previously held entities that remain
+   *          to be forwarded.
+   */
+  fun onResumeResponse(): FilterResumeStatus<ResponseHeaders, ResponseTrailers>
+
+  /**
+   * Called by the filter manager once to initialize the filter callbacks that the filter should
+   * use.
+   *
+   * @param callbacks: The callbacks for this filter to use to interact with the chain.
+   */
+   fun setResponseFilterCallbacks(callbacks: ResponseFilterCallbacks)
+}
--- a/library/kotlin/src/io/envoyproxy/envoymobile/AsycResponseFilter.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/AsycResponseFilter.kt
@@ -23,6 +23,7 @@ interface AsyncResponseFilter : ResponseFilter {
    * @param headers: Headers, if `StopIteration` was returned from `onResponseHeaders`.
    * @param data: Any data that has been buffered where `StopIterationAndBuffer` was returned.
    * @param trailers: Trailers, if `StopIteration` was returned from `onReponseTrailers`. 
+   * @param endStream: True, if the stream ended with the previous (and thus, last) invocation.
    *
    * @return: The resumption status including any HTTP entities that will be forwarded.
    */
--- a/library/kotlin/src/io/envoyproxy/envoymobile/AsycResponseFilter.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/AsycResponseFilter.kt
@@ -5,19 +5,31 @@ package io.envoyproxy.envoymobile
  */
 interface AsyncResponseFilter : ResponseFilter {
   /**
-   * Invoked explicitly in response to an asynchronous resumeResponse() callback when filter
-   * iteration has been stopped.
-   *
-   * @return: The resumption status including any previously held entities that remain
-   *          to be forwarded.
-   */
-  fun onResumeResponse(): FilterResumeStatus<ResponseHeaders, ResponseTrailers>
-
-  /**
    * Called by the filter manager once to initialize the filter callbacks that the filter should
    * use.
    *
    * @param callbacks: The callbacks for this filter to use to interact with the chain.
    */
    fun setResponseFilterCallbacks(callbacks: ResponseFilterCallbacks)
+
+  /**
+   * Invoked explicitly in response to an asynchronous resumeResponse() callback when filter
+   * iteration has been stopped. The parameters passed to this invocation will be a snapshot
+   * of any stream state that has not yet been forwarded along the filter chain.
+   *
+   * As with other filter invocations, this will be called on Envoy's main thread, and thus
+   * no additional synchronization is required between this and other invocations.
+   *
+   * @param headers: Headers, if `StopIteration` was returned from `onResponseHeaders`.
+   * @param data: Any data that has been buffered where `StopIterationAndBuffer` was returned.
+   * @param trailers: Trailers, if `StopIteration` was returned from `onReponseTrailers`. 
+   *
+   * @return: The resumption status including any HTTP entities that will be forwarded.
+   */
+  fun onResumeResponse(
+    headers: ResponseHeaders?,
+    data: ByteBuffer?
+    trailrs: ResponseTrailers?
+    endStream: Boolean
+  ): FilterResumeStatus<ResponseHeaders, ResponseTrailers>
 }
--- a/library/kotlin/src/io/envoyproxy/envoymobile/AsyncRequestFilter.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/AsyncRequestFilter.kt
@@ -10,7 +10,7 @@ interface AsyncRequestFilter : RequestFilter {
    *
    * @param callbacks: The callbacks for this filter to use to interact with the chain.
    */
-   fun setRequestFilterCallbacks(callbacks: RequestFilterCallbacks)
+  fun setRequestFilterCallbacks(callbacks: RequestFilterCallbacks)
 
   /**
    * Invoked explicitly in response to an asynchronous resumeRequest() callback when filter
@@ -29,8 +29,8 @@ interface AsyncRequestFilter : RequestFilter {
    */
   fun onResumeRequest(
     headers: RequestHeaders?,
-    data: ByteBuffer?
-    trailrs: RequestTrailers?
+    data: ByteBuffer?,
+    trailrs: RequestTrailers?,
     endStream: Boolean
   ): FilterResumeStatus<RequestHeaders, RequestTrailers>
 }
--- a/library/kotlin/src/io/envoyproxy/envoymobile/AsyncResponseFilter.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/AsyncResponseFilter.kt
@@ -10,7 +10,7 @@ interface AsyncResponseFilter : ResponseFilter {
    *
    * @param callbacks: The callbacks for this filter to use to interact with the chain.
    */
-   fun setResponseFilterCallbacks(callbacks: ResponseFilterCallbacks)
+  fun setResponseFilterCallbacks(callbacks: ResponseFilterCallbacks)
 
   /**
    * Invoked explicitly in response to an asynchronous resumeResponse() callback when filter
@@ -22,15 +22,15 @@ interface AsyncResponseFilter : ResponseFilter {
    *
    * @param headers: Headers, if `StopIteration` was returned from `onResponseHeaders`.
    * @param data: Any data that has been buffered where `StopIterationAndBuffer` was returned.
-   * @param trailers: Trailers, if `StopIteration` was returned from `onReponseTrailers`. 
+   * @param trailers: Trailers, if `StopIteration` was returned from `onReponseTrailers`.
    * @param endStream: True, if the stream ended with the previous (and thus, last) invocation.
    *
    * @return: The resumption status including any HTTP entities that will be forwarded.
    */
   fun onResumeResponse(
     headers: ResponseHeaders?,
-    data: ByteBuffer?
-    trailrs: ResponseTrailers?
+    data: ByteBuffer?,
+    trailrs: ResponseTrailers?,
     endStream: Boolean
   ): FilterResumeStatus<ResponseHeaders, ResponseTrailers>
 }
--- a/library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
@@ -29,7 +29,7 @@ sealed class FilterDataStatus {
    * This should be called by filters which must parse a larger block of the incoming data before
    * continuing processing.
    */
-  class StopIterationAndBuffer(val data: ByteBuffer) : FilterDataStatus()
+  class StopIterationAndBuffer : FilterDataStatus()
 
   /**
    * Do not iterate to any of the remaining filters in the chain, and do not internally buffer
@@ -45,4 +45,14 @@ sealed class FilterDataStatus {
    * continuing processing, and will handle their own buffering.
    */
   class StopIterationNoBuffer : FilterDataStatus()
+
+  /**
+   * Resume previously-stopped iteration, possibly forwarding headers, if iteration was previously
+   * stopped during an on*Headers invocation.
+   *
+   * It is an error to return ResumeIteration if iteration is not currently stopped, and it is
+   * an error to include headers if headers have already been forwarded to the next filter
+   * (i.e. iteration was stopped during an on*Data invocation instead of on*Headers).
+   */
+  class ResumeIteration<T : Headers>(val headers: T?, val data: ByteBuffer) : FilterDataStatus()
 }
--- a/library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
@@ -14,7 +14,7 @@ sealed class FilterDataStatus<T : Headers> {
    * before the entirety is sent to the next filter.
    * TODO: add param docs. name param in swift.
    */
-  class Continue(val data: ByteBuffer) : FilterDataStatus<T>()
+  class Continue<T : Headers>(val data: ByteBuffer) : FilterDataStatus<T>()
 
   /**
    * Do not iterate to any of the remaining filters in the chain, and buffer body data for later
@@ -29,7 +29,7 @@ sealed class FilterDataStatus<T : Headers> {
    * This should be called by filters which must parse a larger block of the incoming data before
    * continuing processing.
    */
-  class StopIterationAndBuffer : FilterDataStatus<T>()
+  class StopIterationAndBuffer<T : Headers> : FilterDataStatus<T>()
 
   /**
    * Do not iterate to any of the remaining filters in the chain, and do not internally buffer
@@ -44,7 +44,7 @@ sealed class FilterDataStatus<T : Headers> {
    * This may be called by filters which must parse a larger block of the incoming data before
    * continuing processing, and will handle their own buffering.
    */
-  class StopIterationNoBuffer : FilterDataStatus<T>()
+  class StopIterationNoBuffer<T : Headers> : FilterDataStatus<T>()
 
   /**
    * Resume previously-stopped iteration, possibly forwarding headers, if iteration was previously
--- a/library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
@@ -36,7 +36,7 @@ sealed class FilterDataStatus<T : Headers> {
    *
    * `onData` will continue to be called with new chunks of data.
    *
-   * Returning `Continue` or `ResumeIteration from another filter invocation or calling
+   * Returning `ResumeIteration` from another filter invocation or calling
    * `resumeRequest()`/`resumeResponse()` MUST be called when continued filter iteration is
    * desired.
    *
--- a/library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
@@ -5,7 +5,7 @@ import java.nio.ByteBuffer
 /*
  * Status returned by filters when transmitting or receiving data.
  */
-sealed class FilterDataStatus {
+sealed class FilterDataStatus<T : Headers> {
   /**
    * Continue filter chain iteration. If headers have not yet been sent to the next filter, they
    * will be sent first via `onRequestHeaders()`/`onResponseHeaders()`.
@@ -14,7 +14,7 @@ sealed class FilterDataStatus {
    * before the entirety is sent to the next filter.
    * TODO: add param docs. name param in swift.
    */
-  class Continue(val data: ByteBuffer) : FilterDataStatus()
+  class Continue(val data: ByteBuffer) : FilterDataStatus<T>()
 
   /**
    * Do not iterate to any of the remaining filters in the chain, and buffer body data for later
@@ -29,7 +29,7 @@ sealed class FilterDataStatus {
    * This should be called by filters which must parse a larger block of the incoming data before
    * continuing processing.
    */
-  class StopIterationAndBuffer : FilterDataStatus()
+  class StopIterationAndBuffer : FilterDataStatus<T>()
 
   /**
    * Do not iterate to any of the remaining filters in the chain, and do not internally buffer
@@ -44,7 +44,7 @@ sealed class FilterDataStatus {
    * This may be called by filters which must parse a larger block of the incoming data before
    * continuing processing, and will handle their own buffering.
    */
-  class StopIterationNoBuffer : FilterDataStatus()
+  class StopIterationNoBuffer : FilterDataStatus<T>()
 
   /**
    * Resume previously-stopped iteration, possibly forwarding headers, if iteration was previously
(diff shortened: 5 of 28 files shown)


library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
```diff
@@ -45,4 +45,14 @@ sealed class FilterDataStatus {
...
+
+  /**
+   * Resume previously-stopped iteration, possibly forwarding headers, if iteration was previously
+   * stopped during an on*Headers invocation.
+   *
+   * It is an error to return ResumeIteration if iteration is not currently stopped, and it is
+   * an error to include headers if headers have already been forwarded to the next filter
+   * (i.e. iteration was stopped during an on*Data invocation instead of on*Headers).
+   */
+  class ResumeIteration<T : Headers>(val headers: T?, val data: ByteBuffer) : FilterDataStatus()
```
rebello95, line 57: If the caller omits `headers` here (passes `nil`), that will result in the underlying filter manager passing its internally buffered headers, correct? If so, I think that's worth documenting above

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterTrailersStatus.kt
```diff
@@ -17,4 +17,14 @@ sealed class FilterTrailersStatus<T : Headers> {
...
+
+  /**
+   * Resume previously-stopped iteration, possibly forwarding headers and data, if iteration was
+   * previously stopped during an on*Headers or on*Data invocation.
+   *
+   * It is an error to return ResumeIteration if iteration is not currently stopped, and it is
+   * an error to include headers if headers have already been forwarded to the next filter
+   * (i.e. iteration was stopped during an on*Data invocation instead of on*Headers).
+   */
+  class ResumeIteration<T : Headers>(val headers: Headers?, val data: ByteBuffer?, val trailers: T) : FilterTrailersStatus<T>()
```
rebello95, line 29: Same comment here regarding behavior in the `nil` case

library/swift/src/filters/FilterDataStatus.swift
```diff
@@ -35,4 +35,14 @@ public enum FilterDataStatus: Equatable {
...
+
+  /**
+   * Resume previously-stopped iteration, possibly forwarding headers, if iteration was previously
+   * stopped during an on*Headers invocation.
+   *
+   * It is an error to return `resumeIteration` if iteration is not currently stopped, and it is
+   * an error to include headers if headers have already been forwarded to the next filter
+   * (i.e. iteration was stopped during an on*Data invocation instead of on*Headers).
+   */
+  case resumeIteration(Data, withHeaders: Headers? = nil)
```
rebello95, line 47:
```suggestion
  case resumeIteration(headers: Headers? = nil, data: Data)
```

library/swift/src/filters/FilterTrailersStatus.swift
```diff
@@ -9,4 +9,14 @@ public enum FilterTrailersStatus<T: Headers>: Equatable {
...
+
+  /**
+   * Resume previously-stopped iteration, possibly forwarding headers and data, if iteration was
+   * previously stopped during an on*Headers or on*Data invocation.
+   *
+   * It is an error to return ResumeIteration if iteration is not currently stopped, and it is
+   * an error to include headers if headers have already been forwarded to the next filter
+   * (i.e. iteration was stopped during an on*Data invocation instead of on*Headers).
+   */
+  case resumeIteration(T, withHeaders: Headers? = nil, withData: Data? = nil)
```
rebello95, line 21:
```suggestion
  case resumeIteration(headers: Headers? = nil, data: Data? = nil, trailers: T)
```

rebello95: This ordering matches Kotlin - I also think it's nice to name the values explicitly but don't feel too strongly

library/swift/src/filters/ResponseFilter.swift
```diff
@@ -38,6 +38,14 @@ public protocol ResponseFilter: Filter {
   /// - returns: The trailer status containing body with which to continue or buffer.
   func onResponseTrailers(_ trailers: ResponseTrailers) -> FilterTrailersStatus<ResponseTrailers>
 
+  /// Invoked explicitly in response to an asynchronous resume() callback when filter
+  /// iteration has been stopped.
+  /// 
+  /// @return: The resumption status including any previously held entities that remain
+  ///          to be forwarded.
+  /// 
+  func onResumeIteration() -> FilterResumeStatus<ResponseHeaders, ResponseTrailers>
```
rebello95, line 47: Are we missing the `FilterResumeStatus` type in this PR?

library/kotlin/src/io/envoyproxy/envoymobile/filters/ResponseFilter.kt
```diff
@@ -50,6 +50,15 @@ interface ResponseFilter : Filter {
    */
   fun onResponseTrailers(trailers: ResponseTrailers): FilterTrailersStatus<ResponseTrailers>
 
+  /**
+   * Invoked explicitly in response to an asynchronous resume() callback when filter
+   * iteration has been stopped.
```
rebello95, line 55: - These functions are currently called `continueRequest()` / `continueResponse()`; I think they need to be updated in this PR on the callback types
- Can we add documentation to the callbacks indicating that this function will be called on the filter's execution thread as an immediate result of calling the callback?
- It'd be nice to add an example use case to these descriptions. Personally I think we should be as clear/verbose as possible on these interface descriptions

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
goaway: Another option is for that to be an error (and I can update the documentation accordingly). To simplify reasoning, the contract could _always_ be that all forwarded entities are passed back as part of a return from a filter invocation. If it's not explicitly passed back, it hasn't been forwarded.

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterTrailersStatus.kt
goaway: I intentionally introduced a discrepancy between Kotlin and Swift here to demonstrate two potential approaches (we should ultimately pick one). The Swift approach allows unused entities to be defaulted to `nil` and left off of the return status. The Kotlin approach positions the entities in the order in which they'll appear.

I'm open to either and think there's pros to both.

library/swift/src/filters/ResponseFilter.swift
goaway: They're coming. :)

library/kotlin/src/io/envoyproxy/envoymobile/filters/ResponseFilter.kt
goaway: - Updated.
- Sure.
- Good idea.

goaway:
Thanks for the early comments @rebello95! I pushed a significant update, and I think this is at a good point to drive further discussion/consensus, so please do take another pass (and @buildbreaker and @junr03) when you're able. In parallel, I'm going to finish off getting it to compile cleanly.

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterTrailersStatus.kt
goaway: I was playing around with the Swift playground, and it turns out that with named parameters, defaulted arguments on enum cases can be left off in any position. And with Kotlin I think we can achieve this with multiple constructors. So in short, we should be able to support `.resumeIteration(headers: myHeaders, data: myData)` seamlessly with simple case of `.resumeIteration(data: myData)`, realizing both the benefits of conceptually consistent ordering and defaulting.

goaway:
Sorry, I pushed some in-flight work which makes things a little inconsistent again - I suggest holding off on reviewing until I have a chance to wrap it up tomorrow morning (though by all means, if you're curious, feel free to look at what's changed).

library/kotlin/src/io/envoyproxy/envoymobile/AsycRequestFilter.kt
```diff
@@ -0,0 +1,23 @@
...
+ */
+interface AsyncRequestFilter : RequestFilter {
+  /**
+   * Invoked explicitly in response to an asynchronous resumeRequest() callback when filter
+   * iteration has been stopped.
+   *
+   * @return: The resumption status including any previously held entities that remain
+   *          to be forwarded.
+   */
+  fun onResumeRequest(): FilterResumeStatus<RequestHeaders, RequestTrailers>
```
buildbreaker, line 14: Interface to be updated

library/kotlin/src/io/envoyproxy/envoymobile/AsycRequestFilter.kt
```diff
@@ -0,0 +1,23 @@
...
+
+  /**
+   * Called by the filter manager once to initialize the filter callbacks that the filter should
+   * use.
+   *
+   * @param callbacks: The callbacks for this filter to use to interact with the chain.
+   */
+   fun setRequestFilterCallbacks(callbacks: RequestFilterCallbacks)
```
buildbreaker, line 22: Can we have this first

library/kotlin/src/io/envoyproxy/envoymobile/AsycResponseFilter.kt
```diff
@@ -0,0 +1,23 @@
...
+   */
+  fun onResumeResponse(): FilterResumeStatus<ResponseHeaders, ResponseTrailers>
+
+  /**
+   * Called by the filter manager once to initialize the filter callbacks that the filter should
+   * use.
+   *
+   * @param callbacks: The callbacks for this filter to use to interact with the chain.
+   */
+   fun setResponseFilterCallbacks(callbacks: ResponseFilterCallbacks)
```
buildbreaker, line 22: nit: have this first

(same on swift)

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
```diff
@@ -44,5 +44,15 @@ sealed class FilterDataStatus {
    * This may be called by filters which must parse a larger block of the incoming data before
    * continuing processing, and will handle their own buffering.
    */
-  class StopIterationNoBuffer : FilterDataStatus()
+  class StopIterationNoBuffer<T : Headers> : FilterDataStatus<T>()
+
+  /**
+   * Resume previously-stopped iteration, possibly forwarding headers, if iteration was previously
+   * stopped during an on*Headers invocation.
```
buildbreaker, line 51:
```suggestion
   * Resume previously-stopped iteration, possibly forwarding headers, if iteration was previously
   * stopped during an on*Headers invocation. If headers are already sent, headers should not be passed in the return status.

File: library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterResumeStatus.kt [hunk 11]
```diff
@@ -0,0 +1,18 @@
...
+
+import java.nio.ByteBuffer
+
+/*
+ * Status returned by filters when transmitting or receiving trailers.
+ */
+sealed class FilterResumeStatus<T : Headers, U : Headers> {
+  /**
+   * Resume previously-stopped iteration, possibly forwarding headers and data, if iteration was
+   * previously stopped during an on*Headers or on*Data invocation.
```
buildbreaker, line 11: same as https://github.com/lyft/envoy-mobile/pull/1029/files#r470203669

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterTrailersStatus.kt
```diff
@@ -1,20 +1,33 @@
...
+   * asynchronous filters support resumption after returning StopIteration from on*Trailers.
    * Calling `continueRequest()`/`continueResponse()` MUST occur when continued filter iteration
    * is desired.
    */
-  class StopIteration<T : Headers>(val trailers: T) : FilterTrailersStatus<T>()
+  class StopIteration<T : Headers, U : Headers> : FilterTrailersStatus<T, U>()
+
+  /**
+   * Resume previously-stopped iteration, possibly forwarding headers and data, if iteration was
+   * previously stopped during an on*Headers or on*Data invocation.
```
buildbreaker, line 26: same as https://github.com/lyft/envoy-mobile/pull/1029/files#r470203669

library/swift/src/filters/FilterDataStatus.swift
```diff
@@ -35,4 +35,14 @@ public enum FilterDataStatus: Equatable {
   /// This may be called by filters which must parse a larger block of the incoming data before
   /// continuing processing, and will handle their own buffering.
   case stopIterationNoBuffer
+
+  /**
+   * Resume previously-stopped iteration, possibly forwarding headers, if iteration was previously
+   * stopped during an on*Headers invocation.
```
buildbreaker, line 41: same as https://github.com/lyft/envoy-mobile/pull/1029/files#r470203669

library/swift/src/filters/FilterResumeStatus.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+/// Status returned by filters after resuming iteration asynchronously.
+@frozen
+public enum FilterResumeStatus<T: Headers, U: Headers>: Equatable {
+  /**
+   * Resume previously-stopped iteration, possibly forwarding headers and data, if iteration was
+   * previously stopped during an on*Headers or on*Data invocation.
```
buildbreaker, line 8: same as https://github.com/lyft/envoy-mobile/pull/1029/files#r470203669

library/kotlin/src/io/envoyproxy/envoymobile/AsycRequestFilter.kt
rebello95: What will this be updated to?

library/kotlin/src/io/envoyproxy/envoymobile/AsycRequestFilter.kt
rebello95: +1 to putting this first in the file

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
rebello95: `   * stopped during an on*Headers invocation. If headers have already been sent, headers should not be passed in the return status.`

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterTrailersStatus.kt
```diff
@@ -1,20 +1,33 @@
...
  * Status returned by filters when transmitting or receiving trailers.
  */
-// TODO: create abstract Trailers class.
-sealed class FilterTrailersStatus<T : Headers> {
+sealed class FilterTrailersStatus<T : Headers, U : Headers> {
   /**
    * Continue filter chain iteration, passing the provided trailers through.
    */
-  class Continue<T : Headers>(val trailers: T) : FilterTrailersStatus<T>()
+  class Continue<T : Headers, U : Headers>(val trailers: T) : FilterTrailersStatus<T, U>()
```
rebello95, line 12: Shouldn't this be `val trailers: U`?

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
```diff
@@ -44,5 +44,15 @@ sealed class FilterDataStatus {
...
+   *
+   * It is an error to return ResumeIteration if iteration is not currently stopped, and it is
+   * an error to include headers if headers have already been forwarded to the next filter
+   * (i.e. iteration was stopped during an on*Data invocation instead of on*Headers).
+   */
```
rebello95, line 56: I think it's worth including details as to why trailers are not specified here, similarly to how you have in the `StopIteration` case definition

library/swift/src/filters/AsyncRequestFilter.swift
```diff
@@ -0,0 +1,22 @@
...
+    data: Data?,
+    trailers: RequestTrailers?,
+    endStream: Bool
+  ) -> FilterResumeStatus<RequestHeaders, RequestTrailers>
+
+  /// Called by the filter manager once to initialize the filter callbacks that the filter should
+  /// use.
+  ///
+  /// - parameter callbacks: The callbacks for this filter to use to interact with the chain.
+  func setRequestFilterCallbacks(_ callbacks: RequestFilterCallbacks)
```
rebello95, line 21: Same suggestion as @buildbreaker above regarding listing this first

library/swift/src/filters/AsyncResponseFilter.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+
+/// ResponseFilter supporting asynchronous resumption.
```
rebello95, line 3: Same comments in this file

library/swift/src/filters/AsyncRequestFilter.swift
```diff
@@ -0,0 +1,22 @@
...
+  /// Invoked explicitly in response to an asynchronous resumeRequest() callback when filter
+  /// iteration has been stopped. The parameters passed to this invocation will be a snapshot
+  /// of any stream state that has not yet been forwarded along the filter chain.
+  ///
+  /// - return: The resumption status including any HTTP entities that will be forwarded.
+  func onResumeRequest(
+    headers: RequestHeaders?,
+    data: Data?,
+    trailers: RequestTrailers?,
+    endStream: Bool
```
rebello95, line 14: When will `endStream` be true? Will it be true if *either* there's a data frame with `endStream=true` *or* trailers?

rebello95: I think it's worth adding `- parameter` docs for each of these (same with the response filter) to clarify these points

library/swift/src/filters/AsyncRequestFilter.swift
```diff
@@ -0,0 +1,22 @@
...
+  ) -> FilterResumeStatus<RequestHeaders, RequestTrailers>
```
rebello95, line 15: Thinking about how users will consume this, I think they'll likely end up force-unwrapping the values they called `resume` with, which might be a little unwieldy. However, I suppose we can't really make these non-optional since `resumeRequest()` can be called after headers or data, in which case data will sometimes be nil. Probably nothing actionable here, just thinking out loud (though it might be good to list some examples in docs here)

library/swift/src/filters/AsyncResponseFilter.swift
```diff
@@ -0,0 +1,21 @@
...
+public protocol AsyncResponseFilter: ResponseFilter {
+  /// Invoked explicitly in response to an asynchronous resumeResponse() callback when filter
+  /// iteration has been stopped.
+  ///
+  /// - return: The resumption status including any HTTP entities that will be forwarded.
+  func onResumeResponse(
+    headers: ResponseHeaders?,
+    data: Data?,
+    trailers: ResponseTrailers?,
+    endStream: Bool
```
rebello95, line 13: Same here

library/swift/src/filters/FilterHeadersStatus.swift
```diff
@@ -9,5 +9,5 @@ public enum FilterHeadersStatus<T: Headers>: Equatable {
   /// Returning `continue` from `onRequestData()`/`onResponseData()` or calling
   /// `continueRequest()`/`continueResponse()` MUST occur when continued filter iteration is
```
rebello95, line 10: These docstrings are outdated (can you grep for any references to these functions and update please?)

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterTrailersStatus.kt
goaway: yes, good catch, thank you

library/swift/src/filters/AsyncRequestFilter.swift
goaway: I've been thinking about this two, but short of having multiple different `onResume` invocations with overloaded signatures based on what's present, and their own specific return types, I can't see another way of supporting the different states. (Multiple `onResume`s would probably be annoying to implement also.)

library/kotlin/src/io/envoyproxy/envoymobile/AsycRequestFilter.kt
goaway: Updated to match swift.

goaway: (It has been.)

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
goaway: I'm a little bit concerned that describing an option that's not present might be more confusing than simply describing those that are.

goaway:
Thanks for all the comments so far @buildbreaker and @rebello95! I think I've addressed all of them, so please take another pass when you have the chance.

library/kotlin/src/io/envoyproxy/envoymobile/AsyncRequestFilter.kt
```diff
@@ -0,0 +1,36 @@
...
+   */
+  fun setRequestFilterCallbacks(callbacks: RequestFilterCallbacks)
+
+  /**
+   * Invoked explicitly in response to an asynchronous resumeRequest() callback when filter
+   * iteration has been stopped. The parameters passed to this invocation will be a snapshot
+   * of any stream state that has not yet been forwarded along the filter chain.
+   *
+   * As with other filter invocations, this will be called on Envoy's main thread, and thus
+   * no additional synchronization is required between this and other invocations.
```
rebello95, line 21: 👏

library/kotlin/src/io/envoyproxy/envoymobile/AsyncRequestFilter.kt
```diff
@@ -0,0 +1,36 @@
...
+   *
+   * @param headers: Headers, if `StopIteration` was returned from `onRequestHeaders`.
+   * @param data: Any data that has been buffered where `StopIterationAndBuffer` was returned.
+   * @param trailers: Trailers, if `StopIteration` was returned from `onRequestTrailers`.
+   * @param endStream: True, if the stream ended with the previous (and thus, last) invocation.
```
rebello95, line 26: Got it, so this will be true for trailers or `data, endStream: true` 👍

library/kotlin/src/io/envoyproxy/envoymobile/AsyncRequestFilter.kt
```diff
@@ -0,0 +1,36 @@
...
+   *
+   * @return: The resumption status including any HTTP entities that will be forwarded.
+   */
+  fun onResumeRequest(
+    headers: RequestHeaders?,
+    data: ByteBuffer?,
+    trailrs: RequestTrailers?,
```
rebello95, line 33:
```suggestion
    trailers: RequestTrailers?,
```

library/kotlin/src/io/envoyproxy/envoymobile/AsyncResponseFilter.kt
```diff
@@ -0,0 +1,36 @@
...
+   * @param data: Any data that has been buffered where `StopIterationAndBuffer` was returned.
+   * @param trailers: Trailers, if `StopIteration` was returned from `onReponseTrailers`.
+   * @param endStream: True, if the stream ended with the previous (and thus, last) invocation.
+   *
+   * @return: The resumption status including any HTTP entities that will be forwarded.
+   */
+  fun onResumeResponse(
+    headers: ResponseHeaders?,
+    data: ByteBuffer?,
+    trailrs: ResponseTrailers?,
```
rebello95, line 33:
```suggestion
    trailers: ResponseTrailers?,
```

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
```diff
@@ -14,7 +14,7 @@ sealed class FilterDataStatus {
    * before the entirety is sent to the next filter.
    * TODO: add param docs. name param in swift.
```
rebello95, line 15: Might be worth resolving this while we're here, can also do separately

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
```diff
@@ -29,20 +29,30 @@ sealed class FilterDataStatus {
...
 
   /**
    * Do not iterate to any of the remaining filters in the chain, and do not internally buffer
    * data.
    *
    * This filter will continue to be called with new chunks of data.
    *
-   * Returning `Continue` from `onRequestData()`/`onResponseData()` or calling
-   * `continueRequest()`/`continueResponse()` MUST be called when continued filter iteration is
+   * Returning `Continue` or `ResumeIteration from another filter invocation or calling
```
rebello95, line 40:
```suggestion
   * Returning `Continue` or `ResumeIteration` from another filter invocation or calling
```

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
```diff
@@ -29,20 +29,30 @@ sealed class FilterDataStatus {
...
+   * `resumeRequest()`/`resumeResponse()` MUST be called when continued filter iteration is
```
rebello95, line 41:
```suggestion
   * `resumeRequest()`/`resumeResponse()` MUST occur when continued filter iteration is
```

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterHeadersStatus.kt
```diff
@@ -12,9 +12,9 @@ sealed class FilterHeadersStatus<T : Headers> {
   /**
    * Do not iterate to any of the remaining filters in the chain with headers.
    *
-   * Returning `continue` from `onRequestData()`/`onResponseData()` or calling
-   * `continueRequest()`/`continueResponse()` MUST occur when continued filter iteration is
+   * Returning `Continue` or `ResumeIteration from another filter invocation or calling
```
rebello95, line 15:
```suggestion
   * Returning `Continue` or `ResumeIteration` from another filter invocation or calling
```

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterResumeStatus.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+sealed class FilterResumeStatus<T : Headers, U : Headers> {
+  /**
+   * Resume previously-stopped iteration, potentially forwarding headers, data, and/or trailers
+   * that have not yet been passed along the filter chain.
+   *
+   * It is an error to return ResumeIteration if iteration is not currently stopped, and it is
+   * an error to include headers if headers have already been forwarded to the next filter
+   * (i.e. iteration was stopped during an on*Data invocation instead of on*Headers). It is also
+   * an error to include data or trailers if endStream was previously sent or if trailers have
```
rebello95, line 16:
```suggestion
   * an error to include data or trailers if endStream=true was previously sent or if trailers have
```

library/swift/src/filters/FilterDataStatus.swift
```diff
@@ -15,13 +15,13 @@ public enum FilterDataStatus: Equatable {
   ///
   /// This filter will continue to be called with new chunks of data.
   ///
-  /// Returning `continue` from `onRequestData()`/`onResponseData()` or calling
-  /// `continueRequest()`/`continueResponse()` MUST be called when continued filter iteration is
+  /// Returning `continue` or `resumeIteration from another filter invocation or calling
```
rebello95, line 18:
```suggestion
  /// Returning `continue` or `resumeIteration` from another filter invocation or calling
```

library/swift/src/filters/FilterDataStatus.swift
```diff
@@ -15,13 +15,13 @@ public enum FilterDataStatus: Equatable {
...
+  /// `resumeRequest()`/`resumeResponse()` MUST be called when continued filter iteration is
```
rebello95, line 19:
```suggestion
  /// `resumeRequest()`/`resumeResponse()` MUST occur when continued filter iteration is
```

library/swift/src/filters/AsyncResponseFilter.swift
```diff
@@ -0,0 +1,30 @@
...
+  /// iteration has been stopped. The parameters passed to this invocation will be a snapshot
+  /// of any stream state that has not yet been forwarded along the filter chain.
+  ///
+  /// As with other filter invocations, this will be called on Envoy's main thread, and thus
+  /// no additional synchronization is required between this and other invocations.
+  ///
+  /// - param headers: Headers, if `StopIteration` was returned from `onResponseHeaders`.
+  /// - param data: Any data that has been buffered where `StopIterationAndBuffer` was returned.
+  /// - param trailers: Trailers, if `StopIteration` was returned from `onReponseTrailers`.
+  /// - param endStream: True, if the stream ended with the previous (and thus, last) invocation.
```
rebello95, line 21:
```suggestion
  /// - param headers: Headers, if `stopIteration` was returned from `onResponseHeaders`.
  /// - param data: Any data that has been buffered where `stopIterationAndBuffer` was returned.
  /// - param trailers: Trailers, if `stopIteration` was returned from `onReponseTrailers`.
  /// - param endStream: True, if the stream ended with the previous (and thus, last) invocation.
```

library/swift/src/filters/FilterResumeStatus.swift
```diff
@@ -0,0 +1,15 @@
...
+/// Status returned by filters after resuming iteration asynchronously.
+@frozen
+public enum FilterResumeStatus<T: Headers, U: Headers>: Equatable {
+  /// Resume previously-stopped iteration, potentially forwarding headers, data, and/or trailers
+  /// that have not yet been passed along the filter chain.
+  ///
+  /// It is an error to return resumeIteration if iteration is not currently stopped, and it is
+  /// an error to include headers if headers have already been forwarded to the next filter
+  /// (i.e. iteration was stopped during an on*Data invocation instead of on*Headers). It is also
+  /// an error to include data or trailers if endStream was previously sent or if trailers have
```
rebello95, line 12:
```suggestion
  /// an error to include data or trailers if endStream=true was previously sent or if trailers have
```

library/swift/src/filters/FilterTrailersStatus.swift
```diff
@@ -1,12 +1,24 @@
...
+  /// asynchronous filters support resumption after returning StopIteration from on*Trailers.
+  /// Calling `resumeRequest()`/`resumeResponse()` MUST occur if continued filter iteration
   /// is desired.
-  case stopIteration(T)
+  case stopIteration
+
+  /// Resume previously-stopped iteration, possibly forwarding headers and data if iteration was
+  /// stopped during an on*Headers or on*Data invocation.
+  ///
+  /// It is an error to return ResumeIteration if iteration is not currently stopped, and it is
```
rebello95, line 20:
```suggestion
  /// It is an error to return resumeIteration if iteration is not currently stopped, and it is
```

library/swift/src/filters/FilterDataStatus.swift
```diff
@@ -35,4 +35,12 @@ public enum FilterDataStatus: Equatable {
...
   /// continuing processing, and will handle their own buffering.
   case stopIterationNoBuffer
+
+  /// Resume previously-stopped iteration, possibly forwarding headers if iteration was stopped
+  /// during an on*Headers invocation.
+  ///
+  /// It is an error to return `resumeIteration` if iteration is not currently stopped, and it is
+  /// an error to include headers if headers have already been forwarded to the next filter
+  /// (i.e. iteration was stopped during an on*Data invocation instead of on*Headers).
+  case resumeIteration(headers: T? = nil, data: Data)
```
rebello95, line 45: Something I noticed while testing out these interfaces on a filter migration is that it's not immediately obvious to the consumer that returning `.resumeIteration(...)` **does not call** `onResumeIteration()` - only calling `resumeRequest()` does. Thoughts on how we could clarify that either through naming or documentation?

rebello95: Would it make sense to simply rename `continue` and allow optional headers to be passed there? Since my understanding is the ability to pass headers here is the only difference between the two

library/kotlin/src/io/envoyproxy/envoymobile/AsyncRequestFilter.kt
goaway: Correct, in fact any of:
`onHeaders(... endStream: true)`
`onData(... endStream: true)`
`onTrailers(...)`

library/kotlin/src/io/envoyproxy/envoymobile/AsyncRequestFilter.kt
goaway: nice, thanks

library/swift/src/filters/FilterDataStatus.swift
goaway: I still think "must be called" is accurate and maybe clearer?

library/swift/src/filters/FilterDataStatus.swift
goaway: Based on our discussion elsewhere, I believe we've all settled on having "continue" be for continuing ongoing iteration and "resume" specifically for starting again when iteration has been stopped.

library/kotlin/src/io/envoyproxy/envoymobile/filters/FilterDataStatus.kt
goaway: 👍

PR #1037, opened 2020-08-14
crashes: get more granularity on potential crash

junr03:
Description: introduces temporary state to figure out where the garbage pointer in #1016 is.
Risk Level: low, removing imminently.

### DIFF
--- a/library/common/http/dispatcher.cc
+++ b/library/common/http/dispatcher.cc
@@ -243,6 +243,8 @@ Dispatcher::DirectStream::DirectStream(envoy_stream_t stream_handle, Dispatcher&
 
 Dispatcher::DirectStream::~DirectStream() {
   ENVOY_LOG(debug, "[S{}] destroy stream", stream_handle_);
+  // TODO: delete once https://github.com/lyft/envoy-mobile/issues/1016 is fixed.
+  destroyed_ = true;
 }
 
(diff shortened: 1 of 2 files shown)


library/common/http/dispatcher.cc
```diff
@@ -243,6 +243,8 @@ Dispatcher::DirectStream::DirectStream(envoy_stream_t stream_handle, Dispatcher&
 
 Dispatcher::DirectStream::~DirectStream() {
   ENVOY_LOG(debug, "[S{}] destroy stream", stream_handle_);
+  // TODO: delete once https://github.com/lyft/envoy-mobile/issues/1016 is fixed.
+  destroyed_ = true;
```
goaway, line 247: I'm okay moving forward with this to see what we see, but I still feel a little bit that a more robust approach, at least in the case of the dispatcher, would be to have the state be unowned by the dispatcher and be updated as a side-effect of destruction.

junr03: Yeah, per our offline conversation that would definitely give us access to the state. But I believe that the static function indirection will provide us the stacktrace if the crash is indeed due to that garbage pointer. The state itself is not necessary. My worry was that the compiler might optimize away an empty function.

Lets see what we see with this and we can adapt after. Obviously this is throwaway code anyways.

PR #1050, opened 2020-08-26
jni: make get_env skip JNI calls for the normal case

goaway:
Description: Previously, get_env would make unnecessary JNI calls every time it was called, making it relatively expensive, and encouraging passing the pointer around everywhere. This makes it faster in the normal case by skipping JNI calls except for the first time a thread uses it.
Risk Level: Low
Testing: Local and CI

### DIFF
--- a/library/common/jni/jni_utility.cc
+++ b/library/common/jni/jni_utility.cc
@@ -15,16 +15,18 @@ void set_vm(JavaVM* vm) { static_jvm = vm; }
 JavaVM* get_vm() { return static_jvm; }
 
 JNIEnv* get_env() {
-  if (local_env) { return local_env; }
+  if (local_env) {
+    return local_env;
+  }
 
   jint result = static_jvm->GetEnv(reinterpret_cast<void**>(&local_env), JNI_VERSION);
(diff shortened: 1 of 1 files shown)


library/common/jni/jni_utility.cc
```diff
@@ -8,24 +8,26 @@
...
-    // TODO: harden this piece of code to make sure that we are only needing to attach Envoy
-    // engine's std::thread, and that we detach it successfully.
-    static_jvm->AttachCurrentThread(&env, nullptr);
-    static_jvm->GetEnv(reinterpret_cast<void**>(&env), JNI_VERSION);
+    JavaVMAttachArgs args = {JNI_VERSION, "EnvoyMain", NULL};
+    result = static_jvm->AttachCurrentThread(&local_env, &args);
   }
-  return env;
+  // TODO(goaway): add assertions and uncomment
+  // ASSERT(result == JNI_OK);
```
buildbreaker, line 29: Please open ticket :)

goaway: https://github.com/lyft/envoy-mobile/issues/1054

PR #1052, opened 2020-08-26
Introduce Counter protocol

Augustyniak:
Description: Make `StatsClient` protocol's interface return `Counter` protocol (as opposed to `Counter` class) to make mocking of the `StatsClient` easier.

Use case: in our Lyft application we need a way to mock `StatsClient` (protocol). An instance of the object conforming to this protocol can be created with the use of Envoy engine. In unit tests, we do not want to use Envoy engine and hence we need a way to create a custom object conforming to `StatsClient` protocol.
To implement a `StatsClient` protocol one needs to implements `func counter(_ elements: [Element]) -> Counter` method. In order to implement this method a developer needs to be able to initialize a new instance of `Counter`, it's currently not possible to do from outside of the module in which `Counter` is defined because:
1. `Counter` does  not provide any public initializer.
2. `Counter` cannot be subclassed since public classes don't allow for subclassing from outside of the module in which they are defined.

Testing: No testing performed since it shouldn't influence the way framework works
Docs Changes: No changes
Release Notes: No changes

### DIFF
--- a/library/swift/src/stats/Counter.swift
+++ b/library/swift/src/stats/Counter.swift
@@ -3,7 +3,7 @@ import Foundation
 
 /// A time series counter.
 @objcMembers
-public class Counter: NSObject {
+open public class Counter: NSObject {
   private let series: String
   private weak var engine: EnvoyEngine?
 
--- a/library/swift/src/stats/Counter.swift
+++ b/library/swift/src/stats/Counter.swift
@@ -1,24 +1,15 @@
-@_implementationOnly import EnvoyEngine
 import Foundation
 
 /// A time series counter.
-@objcMembers
-open public class Counter: NSObject {
-  private let series: String
-  private weak var engine: EnvoyEngine?
-
-  init(elements: [Element], engine: EnvoyEngine) {
-    self.series = elements.map { $0.value }.joined(separator: ".")
-    self.engine = engine
-    super.init()
-  }
-
+@objc
+public protocol Counter: AnyObject {
   /// Increment the counter by the given count.
-  public func increment(count: Int = 1) {
-    guard let engine = self.engine else {
-      return
-    }
+  func increment(count: Int)
+}
 
-    engine.recordCounter(self.series, count: numericCast(count))
+extension Counter {
+  /// Increment the counter by 1.
+  public func increment() {
+    self.increment(count: 1)
   }
 }
--- a/library/swift/src/StatsClientImpl.swift
+++ b/library/swift/src/StatsClientImpl.swift
@@ -13,6 +13,6 @@ final class StatsClientImpl: NSObject {
 
 extension StatsClientImpl: StatsClient {
   func counter(elements: [Element]) -> Counter {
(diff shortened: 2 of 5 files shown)


Augustyniak:
PTAL #client-networking-dev

buildbreaker:
@jingwei99 Could you take a first pass and make sure both swift/kotlin interfaces are consistent?

Augustyniak:
One things that is inconsistent is the lack of `Counter` protocol on Android. I opened a GH issue to address this https://github.com/lyft/envoy-mobile/issues/1053. Other than this, this PR doesn't change the signature of any of the public methods - it just changes `Counter` class to be `Counter` protocol instead.

library/swift/src/stats/Counter.swift
```diff
@@ -1,24 +1,15 @@
...
-  }
-
+@objc
+public protocol Counter: AnyObject {
   /// Increment the counter by the given count.
-  public func increment(count: Int = 1) {
-    guard let engine = self.engine else {
-      return
-    }
+  func increment(count: Int)
```
Augustyniak, line 7: no default argument value can be specified in Swift's protocol - that's why I need to have a method defined in an extension.

library/swift/src/stats/Counter.swift
```diff
@@ -1,24 +1,15 @@
-@_implementationOnly import EnvoyEngine
```
jingwei99, line 1: nit: on android we usually use `I` to prefix an interface (similar to ios protocol). if this was an android class, it would be named `ICounter`.

personally i don't have too strong of an opinion on this, but `Counter` definitely doesn't not intuitively suggest an interface.

Augustyniak: looking at other protocols on iOS - `StatsClient` being one example - we don't use `I` prefix for protocols on iOS. Instead, we add `Impl` suffix to objects implementing a given protocol.

jingwei99: > looking at other protocols on iOS 
im assuming you are referencing to other protocols in this repo (instead of swift in general), my original question was more about swift in general.

Augustyniak: In general, `I` is not used as prefix for protocol names in Swift. This is what official Swift guidelines say:

Screen Shot 2020-08-27 at 2 49 58 PM

source: https://swift.org/documentation/api-design-guidelines/

goaway: An `I-` prefix is much more a Java convention. In Swift, if a distinction is necessary, the `-Protocol` suffix would be more typical. Here, we've simply standardized on `-Impl` for the implementation, since it's common enough across languages, and what Envoy itself uses as well.

jingwei99: i see, thank you both.

goaway:
This is great, thanks @Augustyniak!

jingwei99:
looks good to me

PR #1056, opened 2020-08-27
http dispatcher: serialize all dispatch and state management

goaway:
Description: Previously cancel had the contract that by the time a call to `cancel` returned, no further dispatch would occur for a stream. This was guarded by some atomic state on the stream. While this was actually one of the more tested areas of the codebase, we're unwinding that for now, and dispatching cancellation the same as any other stream interaction. This is to potentially simplify reasoning about the code to continue investigation of recent crashes. Some tests have been removed that specifically tested concurrency of cancellation, since ordering tests now cover this.
Risk Level: Moderate
Testing: Local and CI


library/common/http/dispatcher.cc
```diff
@@ -425,63 +412,31 @@ envoy_status_t Dispatcher::sendTrailers(envoy_stream_t stream, envoy_headers tra
...
+      // defer delete the HCM's ActiveStream. That means that the lifetime of the DirectStream
+      // only needs to be as long as that deferred delete. Therefore, we synchronously call
+      // cleanup here which will defer delete the DirectStream, which by definition will be
+      // scheduled **after** the HCM's defer delete as they are scheduled on the same dispatcher
+      // context.
+      //
+      // StreamResetReason::RemoteReset is used as the platform code that issues the
+      // cancellation is considered the remote.
+      if (!direct_stream->hcm_stream_pending_destroy_) {
+        direct_stream->hcm_stream_pending_destroy_ = true;
```
mattklein123, line 433: I still don't understand why this boolean is necessary. Once a stream is pending destroy, it should be gone from the map and the map lookup on stream ID should fail. Can you try to remove this boolean and simplify the logic further?

library/common/http/dispatcher.cc
```diff
@@ -57,16 +57,12 @@ void Dispatcher::DirectStreamCallbacks::encodeHeaders(const ResponseHeaderMap& h
...
-    Thread::BasicLockable* mutex = end_stream ? nullptr : &direct_stream_.dispatch_lock_;
-    Thread::OptionalReleasableLockGuard lock(mutex);
     if (direct_stream_.dispatchable(end_stream)) {
       ENVOY_LOG(debug,
                 "[S{}] dispatching to platform response headers for stream (end_stream={}):\n{}",
                 direct_stream_.stream_handle_, end_stream, headers);
       bridge_callbacks_.on_headers(Utility::toBridgeHeaders(headers), end_stream,
                                    bridge_callbacks_.context);
-      lock.release();
       closeRemote(end_stream);
```
mattklein123, line 66: It is extremely difficult to reason about when the stream is removed from the map in various scenarios and with early returns. Per my other comment, can you remove the pending destroy boolean, and change the logic in all of these functions (with cleanup, etc. whatever) to a) ASSERT the stream is still in the map, and b) remove it if end_stream/reset in a way that is easy to audit without going through every early return path?

library/common/http/dispatcher.cc
goaway: I actually was thinking through whether it was stlil necessary last night. I wasn't 100% convinced it was removable and I erred on the side of changing less, but I'll walk through the paths that use it.

library/common/http/dispatcher.cc
goaway: Per some discussion, we will do additional reworking of the code here to enhance readability in a follow-up PR.

PR #1061, opened 2020-09-01
http dispatcher: use stream map as source of truth

goaway:
Description: Removes most state management from the Http::Dispatcher and updates it with patterns better aligned to Envoy's own HTTP codecs. Existence in the stream map becomes the sole source of truth about whether a stream is open and may continue sending or receiving content.
Risk: Moderate
Testing: Local and CI


library/common/http/dispatcher.cc
```diff
@@ -96,131 +95,109 @@ void Dispatcher::DirectStreamCallbacks::encodeHeaders(const ResponseHeaderMap& h
...
-    // @see Dispatcher::DirectStreamCallbacks::closeRemote.
-    // Otherwise fire the reset from here.
-    if (direct_stream_.local_closed_) {
-      onReset();
-    }
+    // If the stream is not locally closed, the HCM will fire resetStream subsequently. However, we will see
+    // that the stream is already gone from the streams_ map, and direct Envoy to runResetCallbacks() without
+    // redundantly calling on onReset().
+    // @see Dispatcher::DirectStreamCallbacks::resetStream.
+    onReset();
```
mattklein123, line 102: I would somehow move this call back out to the calling function (either via return code or something else) so that all of the reset/closure logic is in one place.

library/common/http/dispatcher.cc
```diff
@@ -96,131 +95,109 @@ void Dispatcher::DirectStreamCallbacks::encodeHeaders(const ResponseHeaderMap& h
...
   }
 }
 
+void Dispatcher::DirectStreamCallbacks::mapLocalDataToError(Buffer::Instance& data, bool end_stream) {
```
mattklein123, line 106: I would probably inline this where it's called. I don't think it's worth moving out, for the same reason I commented on above (all closure logic in one place).

library/common/http/dispatcher.cc
goaway: Sure - in fact, your suggestion led to me to realize I can simplify this even further, by checking success prior to this special handling.

test/common/http/dispatcher_test.cc
```diff
@@ -1273,100 +1273,6 @@ TEST_F(DispatcherTest, StreamResetAfterOnComplete) {
   ASSERT_EQ(cc.on_cancel_calls, 0);
 }
 
-TEST_F(DispatcherTest, ResetStreamLocalHeadersRemoteRaceLocalWins) {
```
junr03, line 1276: What was the failure here? Shouldn't the race between local and remote reset still be covered?

library/common/http/dispatcher.cc
```diff
@@ -34,50 +34,59 @@ void Dispatcher::DirectStreamCallbacks::encodeHeaders(const ResponseHeaderMap& h
...
-    http_dispatcher_.synchronizer_.syncPoint("dispatch_encode_headers");
-
-    if (direct_stream_.dispatchable(end_stream)) {
-      ENVOY_LOG(debug,
-                "[S{}] dispatching to platform response headers for stream (end_stream={}):\n{}",
-                direct_stream_.stream_handle_, end_stream, headers);
-      bridge_callbacks_.on_headers(Utility::toBridgeHeaders(headers), end_stream,
-                                   bridge_callbacks_.context);
-      closeRemote(end_stream);
+  // Error path: missing EnvoyUpstreamServiceTime implies this is a local reply, which we treat as
```
junr03, line 52:
```suggestion
  // Error path: missing EnvoyUpstreamServiceTime implies this is a local reply, which Envoy Mobile treats as
```

nit

library/common/http/dispatcher.cc
```diff
@@ -34,50 +34,59 @@ void Dispatcher::DirectStreamCallbacks::encodeHeaders(const ResponseHeaderMap& h
...
+  // a stream error.
+  if (!success_ && headers.get(Headers::get().EnvoyUpstreamServiceTime) == nullptr) {
+    ENVOY_LOG(debug, "[S{}] intercepted local response", direct_stream_.stream_handle_);
+    mapLocalResponseToError(headers, end_stream);
+    if (end_stream) {
+      onReset();
```
junr03, line 58: looks like in the 200 case we are firing `onComplete()` in `mapLocalResponseToError` and then `onRese()` here?

library/common/http/dispatcher.cc
```diff
@@ -93,134 +102,93 @@ void Dispatcher::DirectStreamCallbacks::encodeHeaders(const ResponseHeaderMap& h
...
-    http_dispatcher_.stats().stream_failure_.inc();
-    bridge_callbacks_.on_error({code, message, attempt_count}, bridge_callbacks_.context);
-
-    // All the terminal callbacks only cleanup if they are dispatchable.
-    // This ensures that cleanup will happen exactly one time.
-    http_dispatcher_.cleanup(direct_stream_.stream_handle_);
-  }
+  ENVOY_LOG(debug, "[S{}] dispatching to platform remote reset stream",
+            direct_stream_.stream_handle_);
+  http_dispatcher_.stats().stream_failure_.inc();
```
junr03, line 187: I wonder if we should move stats charging to closeStream so that it is only done there, and we don't have a chance of double counting?

library/common/http/dispatcher.cc
```diff
@@ -234,38 +202,24 @@ Dispatcher::DirectStream::~DirectStream() {
...
-  // resetStream on the response_encoder_'s Stream. It is up to the response_encoder_ to
-  // runResetCallbacks in order to have the Http::ConnectionManager call doDeferredStreamDestroy in
-  // ConnectionManagerImpl::ActiveStream::onResetStream.
-  //
-  // @see Dispatcher::resetStream's comment on its call to runResetCallbacks for an explanation if
-  // this if guard.
-  if (!hcm_stream_pending_destroy_) {
-    hcm_stream_pending_destroy_ = true;
-    runResetCallbacks(reason);
+  runResetCallbacks(reason);
```
junr03, line 205: drop a comment that this is the expectation in higher layers, like the h2 codec does?

library/common/http/dispatcher.cc
```diff
@@ -234,38 +202,24 @@ Dispatcher::DirectStream::~DirectStream() {
...
+  if (!parent_.getStream(stream_handle_)) {
+    // We don't assert here, because Envoy will issue a stream reset if a stream closes remotely
+    // while still open locally. In this case the stream will already have been removed from
+    // our streams_ map due to the remote closure.
```
junr03, line 209:
```suggestion
    // streams_ map due to the remote closure.
```

nit

library/common/http/dispatcher.cc
```diff
@@ -234,38 +202,24 @@ Dispatcher::DirectStream::~DirectStream() {
...
+
+    // The Http::ConnectionManager does not destroy the stream in doEndStream() when it calls
+    // resetStream on the response_encoder_'s Stream. It is up to the response_encoder_ to
+    // runResetCallbacks in order to have the Http::ConnectionManager call doDeferredStreamDestroy
+    // in ConnectionManagerImpl::ActiveStream::onResetStream.
+    // TODO: explore an upstream fix to get the HCM to clean up ActiveStream itself.
+    // runResetCallbacks(reason);
```
junr03, line 216: this is out of place now?

test/common/http/dispatcher_test.cc
goaway: There is no more race - because of serialization, it's just a matter of ordering, which is covered elsewhere. The specific failure here was that the synchronization point no longer gated what it originally did - cleanup happened before it instead of after. But it's also no longer relevant to the implementation. And the test would no longer be meaningful for what it was covering if I were to simply move the sync point.

library/common/http/dispatcher.cc
goaway: Actually the 200 case will never match because I check success first - I just forgot to delete it.

library/common/http/dispatcher.cc
goaway: Stats are charged on each of the terminal paths of which only one will be called per stream - `onReset`, `onComplete`, and `onCancel`. `closeStream` on the other hand is not guaranteed to be called.

library/common/http/dispatcher.cc
```diff
@@ -34,53 +34,53 @@ void Dispatcher::DirectStreamCallbacks::encodeHeaders(const ResponseHeaderMap& h
...
+            direct_stream_.stream_handle_, end_stream, headers);
+  bridge_callbacks_.on_headers(Utility::toBridgeHeaders(headers), end_stream,
+                               bridge_callbacks_.context);
+  if (end_stream) {
+    onComplete();
+  }
+}
+
+void Dispatcher::DirectStreamCallbacks::mapLocalResponseToError(const ResponseHeaderMap& headers,
+                                                                bool end_stream) {
```
junr03, line 78: unused param ^

goaway:
thanks for the feedback @junr03 and @mattklein123!

PR #1067, opened 2020-09-02
stats: ensure client_scope_ destruction occurs ahead of main_common_ destruction

junr03:
Description: scope destruction should happen ahead of destroying main_common_. Additionally while cross thread destruction is acceptable, this PR also moves destruction to happen in the context of the main thread.
Risk Level: med - resolve crash
Testing: repro in alpha builds of the lyft app.


goaway (approved):
Nice. Seems reasonable for now.

PR #1069, opened 2020-09-03
swift: http acceptance tests with assertion filter

junr03:
Description: this PR add end-to-end tests for (almost) all http streaming functionality in the library from the swift platform layer to the core layer.
Risk Level: low
Testing: adds new testing strategy.


library/swift/test/DemoFilter.swift
```diff
@@ -0,0 +1,30 @@
...
+  func onResponseHeaders(_ headers: ResponseHeaders, endStream: Bool)
+    -> FilterHeadersStatus<ResponseHeaders>
+  {
+    NSLog("Adding new header!")
+    let builder = headers.toResponseHeadersBuilder()
+    builder.add(name: "filter-demo", value: "1")
+    return .continue(headers: builder.build())
+  }
+
+  func setResponseFilterCallbacks(_ callbacks: ResponseFilterCallbacks) {}
```
rebello95, line 14: This shouldn't be required for `ResponseFilter` implementations

library/swift/test/BridgeTests.swift
```diff
@@ -0,0 +1,100 @@
...
+  app_id: {{ app_id }}
+  virtual_clusters: {{ virtual_clusters }}
+"""
+
+private struct TestFilter: Filter {}
+
+final class BridgeTests: XCTestCase {
+  override func tearDown() {
+    super.tearDown()
+  }
```
rebello95, line 28: Can remove this

library/swift/test/BridgeTests.swift
```diff
@@ -0,0 +1,100 @@
...
+                  match_config:
+                    http_request_headers_match:
+                      headers:
+                        - name: ":authority"
+                          exact_match: example.com
+              - name: envoy.router
+                typed_config:
+                  "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
+    """
+    let cond = NSCondition()
```
rebello95, line 73: We should be using XCTestExpectation rather than NSCondition: https://github.com/lyft/envoy-mobile/blob/06fc24a895cb640408450bc220190c01c3f2f2c0/library/swift/test/EngineBuilderTests.swift#L131

library/swift/test/BridgeTests.swift
```diff
@@ -0,0 +1,100 @@
...
+"""
+
+private struct TestFilter: Filter {}
+
+final class BridgeTests: XCTestCase {
+  override func tearDown() {
+    super.tearDown()
+  }
+
+  func testInitial() throws {
```
rebello95, line 30: We should add some more color to what this is actually testing (test names can be quite verbose which is fine)

library/swift/test/BridgeTests.swift
```diff
@@ -0,0 +1,100 @@
...
+
+    let requestHeaders = RequestHeadersBuilder(method: .get, scheme: "https",
+                                               authority: "example.com", path: "/test")
+      .addUpstreamHttpProtocol(.http2)
+      .build()
+    client
+      .newStreamPrototype()
+      .setOnResponseHeaders { responseHeaders, _ in
+        let status = responseHeaders.httpStatus ?? -1
+         XCTAssertEqual(200, status)
```
rebello95, line 89:
```suggestion
         XCTAssertEqual(200, responseHeaders.httpStatus)
```

library/swift/test/BridgeTests.swift
```diff
@@ -0,0 +1,100 @@
...
+      }
+      .setOnResponseData { _, endStream in
+        XCTAssertTrue(endStream)
+        cond.signal()
+      }
+      .start(queue: q)
```
rebello95, line 95: Could probably just use `.main`

junr03: I believe I need to use a different thread for the response to advance the test thread to wait for the expectation. In an earlier draft the test was blocking if I queued on main. I did change some of the test structure, so I can try with main again.

junr03:
@rebello95 finally updated 🎉

library/swift/test/BridgeTests.swift
junr03: I guess expectations work differently under the hood than NSCondition. Changed queue back to main successfully.

library/swift/test/HttpBridgeTests/CancelStreamTest.swift
```diff
@@ -0,0 +1,57 @@
...
+
+    client
+      .newStreamPrototype()
+      .setOnCancel {
+         expectation.fulfill()
+      }
+      .start()
+      .cancel()
+
+    self.waitForExpectations(timeout: 1)
```
rebello95, line 55:
```suggestion
    XCTAssertEqual(XCTWaiter.wait(for: [expectation], timeout: 1), .completed)
```
I'd replace these with assertions that the result of the wait is actually `completed` (throughout this PR)

library/swift/test/HttpBridgeTests/DemoFilter.swift
```diff
@@ -0,0 +1,28 @@
...
+  {
+    NSLog("Adding new header!")
+    let builder = headers.toResponseHeadersBuilder()
+    builder.add(name: "filter-demo", value: "1")
+    return .continue(headers: builder.build())
+  }
+
+  func onResponseData(_ body: Data, endStream: Bool) -> FilterDataStatus<ResponseHeaders> {
+    // TODO(goaway): Can remove this when we have better integration coverage in place.
+    NSLog("Saw data chunk of length \(body.count)")
```
rebello95, line 16: Should we remove these logs?

library/swift/test/HttpBridgeTests/DemoFilter.swift
```diff
@@ -0,0 +1,28 @@
+import Envoy
+import Foundation
+
+struct DemoFilter: ResponseFilter {
+  func onResponseHeaders(_ headers: ResponseHeaders, endStream: Bool)
+    -> FilterHeadersStatus<ResponseHeaders>
+  {
+    NSLog("Adding new header!")
```
rebello95, line 8:
```suggestion
```

library/swift/test/HttpBridgeTests/README.md
```diff
@@ -0,0 +1,11 @@
...
+functionality. It tests both the request side sendHeaders/Data,Close,Cancel; and the response
+side via all the setOnResponseHeaders/Data,setOnError,setOnCancel.
+
+Note: These tests are broke apart into different suites and bazel targets in order to tear down
+app state --and thus static lifetime objects like the Envoy engine -- between tests. When
+multiple engine support (https://github.com/lyft/envoy-mobile/issues/332) lands, all of these
+tests can be collapsed to the same suite/target.
+
+TODO: setOnTrailers is not tested as the neither the direct_response pathway, nor the router
+allow sending trailers programmatically. Add tests once possible.
```
rebello95, line 11:
```suggestion
This test suite tests end-to-end integration of the platform layer and the core layer's HTTP
functionality. It tests both the request side `send{Headers|Data}`, `close`, `cancel`,
as well as the response side via all the` setOnResponse{...}` functions.

TODO: These tests are broken apart into different suites and bazel targets in order to tear down
app state - and thus static lifetime objects like the Envoy engine - between tests. When
multiple engine support (https://github.com/lyft/envoy-mobile/issues/332) lands, all of these
tests can be collapsed to the same suite/target.

TODO: setOnTrailers is not tested as neither the `direct_response` pathway, nor the router
allow sending trailers programmatically. Add tests once possible.
```

library/swift/test/HttpBridgeTests/ReceiveDataTest.swift
```diff
@@ -0,0 +1,78 @@
+import Envoy
+import EnvoyEngine
+import Foundation
+import XCTest
+
+final class ReceiveDataTests: XCTestCase {
+  func testReceiveData() throws {
+    // swiftlint:disable:next line_length
+    let apiListenerType = "type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager"
+    // swiftlint:disable:next line_length
```
rebello95, line 10: If you want you could put one of these at the top of the individual file and do `// swiftlint:disable line_length`

library/swift/test/HttpBridgeTests/ReceiveDataTest.swift
```diff
@@ -0,0 +1,78 @@
...
+      .streamClient()
+
+    let requestHeaders = RequestHeadersBuilder(method: .get, scheme: "https",
+                                               authority: "example.com", path: "/test")
+      .addUpstreamHttpProtocol(.http2)
+      .build()
+    client
+      .newStreamPrototype()
+      .setOnResponseHeaders { responseHeaders, _ in
+         XCTAssertEqual(200, responseHeaders.httpStatus)
```
rebello95, line 66: We should add a second expectation here and wait for it below as well - there's no validation as-is that this assert is actually called

library/swift/test/HttpBridgeTests/ReceiveDataTest.swift
```diff
@@ -0,0 +1,78 @@
...
+      }
+      .setOnResponseData { data, _ in
+        let responseBody = String(decoding: data, as: UTF8.self)
```
rebello95, line 69:
```suggestion
        let responseBody = String(data: data, encoding: .utf8)
```

library/swift/test/HttpBridgeTests/ReceiveErrorTest.swift
```diff
@@ -0,0 +1,69 @@
...
+              name: api_router
+              virtual_hosts:
+                - name: api
+                  domains:
+                    - "*"
+                  routes:
+                    - match:
+                        prefix: "/"
+                      direct_response:
+                        status: 503
```
rebello95, line 36: Would it make sense to document lower in the test that Envoy Mobile maps a 503 to an `EnvoyError` which is what's being tested?

rebello95 (approved):
Looks great overall! Just a few minor comments.

library/swift/test/HttpBridgeTests/ReceiveDataTest.swift
junr03: just want to do these specific lines

library/swift/test/HttpBridgeTests/ReceiveDataTest.swift
junr03: How come there isn't? If the expectation below doesn't fulfill then the test will fail. Meaning that if the test finishes, this code must have ran?

junr03:
@rebello95 updated!

library/swift/test/HttpBridgeTests/ReceiveDataTest.swift
rebello95: My concern is that it relies on the (hopefully always valid) assumption that `onResponseData` will always be called after `onResponseHeaders` (since the expectation being fulfilled is fulfilled upon receiving response data, but nothing is explicitly fulfilled on headers). Adding an expectation here would prevent the theoretical issue of the filter responding with data callbacks but no headers

library/swift/test/HttpBridgeTests/ReceiveDataTest.swift
```diff
@@ -60,19 +59,24 @@ final class ReceiveDataTests: XCTestCase {
...
         let responseBody = String(data: data, encoding: .utf8)
         XCTAssertEqual("response_body", responseBody)
-        expectation.fulfill()
+        dataExpectation.fulfill()
       }
       .start()
       .sendHeaders(requestHeaders, endStream: true)
 
-    XCTAssertEqual(XCTWaiter.wait(for: [expectation], timeout: 1), .completed)
+    XCTAssertEqual(XCTWaiter.wait(for: [headersExpectation, dataExpectation], timeout: 1), .completed)
```
rebello95: you could also pass `enforceOrder: true` here if you want

rebello95 (approved):

rebello95 (approved):

rebello95 (approved):

rebello95:
Merging this - Swift unit tests are green, and the rest are Homebrew flakes.

junr03:
Thanks @rebello95 ❤️

PR #1071, opened 2020-09-03
assertion filter: update to be pass-through

goaway:
Description: Updates theAssertionFilter to always return ::Continue when validation checks pass. Fixes some bugs when requests should have failed, but wouldn't, or should have passed, but wouldn't:
- Processing would be aborted for truncated requests, but potentially still appear to pass: e.g., a filter set up to validate conditions on headers and body could return 200 for headers-only requests, skipping body validation.
- If data was streamed, but the first chunk didn't pass all data assertions on its own, processing would fail, even if later chunks would have passed validation.
- Testing of conditions before and after another filter wasn't possible.

For integration tests, there's not much downside to the pass-through behavior, since the Router can always be configured to return a 200 direct response, or any other sentinel value.
Risk Level: Low
Testing: Local and CI

### DIFF
--- a/library/common/extensions/filters/http/assertion/filter.cc
+++ b/library/common/extensions/filters/http/assertion/filter.cc
@@ -1,6 +1,7 @@
 #include "library/common/extensions/filters/http/assertion/filter.h"
 
 #include "common/http/header_map_impl.h"
+
 #include "envoy/http/codes.h"
 #include "envoy/server/filter_config.h"
 
--- a/library/common/extensions/filters/http/assertion/filter.cc
+++ b/library/common/extensions/filters/http/assertion/filter.cc
@@ -1,9 +1,9 @@
 #include "library/common/extensions/filters/http/assertion/filter.h"
 
-#include "common/http/header_map_impl.h"
-
 #include "envoy/http/codes.h"
 #include "envoy/server/filter_config.h"
+
+#include "common/http/header_map_impl.h"
 
 namespace Envoy {
 namespace Extensions {
--- a/library/common/extensions/filters/http/assertion/filter.cc
+++ b/library/common/extensions/filters/http/assertion/filter.cc
@@ -24,9 +24,11 @@ AssertionFilter::AssertionFilter(AssertionFilterConfigSharedPtr config) : config
   config_->rootMatcher().onNewStream(statuses_);
 }
 
-Http::FilterHeadersStatus AssertionFilter::decodeHeaders(Http::RequestHeaderMap& headers, bool end_stream) {
+Http::FilterHeadersStatus AssertionFilter::decodeHeaders(Http::RequestHeaderMap& headers,
+                                                         bool end_stream) {
   config_->rootMatcher().onHttpRequestHeaders(headers, statuses_);
-  if (!config_->rootMatcher().matchStatus(statuses_).matches_) {
+  auto& matchStatus = config_->rootMatcher().matchStatus(statuses_);
+  if (!matchStatus.matches_ && !matchStatus.might_change_status_) {
     decoder_callbacks_->sendLocalReply(Http::Code::BadRequest,
                                        "Request Headers do not match configured expectations",
                                        nullptr, absl::nullopt, "");
@@ -36,7 +38,8 @@ Http::FilterHeadersStatus AssertionFilter::decodeHeaders(Http::RequestHeaderMap&
   if (end_stream) {
     Buffer::OwnedImpl empty_buffer;
     config_->rootMatcher().onRequestBody(empty_buffer, statuses_);
-    if (!config_->rootMatcher().matchStatus(statuses_).matches_) {
+    auto& matchStatus = config_->rootMatcher().matchStatus(statuses_);
+    if (!matchStatus.matches_ && !matchStatus.might_change_status_) {
       decoder_callbacks_->sendLocalReply(Http::Code::BadRequest,
                                          "Request Body does not match configured expectations",
                                          nullptr, absl::nullopt, "");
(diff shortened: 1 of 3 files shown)


buildbreaker (approved):

buildbreaker:
> Processing would be aborted for truncated requests, but potentially still appear to pass: e.g., a filter set up to validate conditions on headers and body could return 200 for headers-only requests, skipping body validation.

Could we actually exercise the case for this actually?

library/common/extensions/filters/http/assertion/filter.cc
```diff
@@ -26,53 +28,91 @@ AssertionFilter::AssertionFilter(AssertionFilterConfigSharedPtr config) : config
 Http::FilterHeadersStatus AssertionFilter::decodeHeaders(Http::RequestHeaderMap& headers,
                                                          bool end_stream) {
   config_->rootMatcher().onHttpRequestHeaders(headers, statuses_);
-  if (!config_->rootMatcher().matchStatus(statuses_).matches_) {
+  auto& matchStatus = config_->rootMatcher().matchStatus(statuses_);
```
junr03, line 31:
```suggestion
  auto& match_status = config_->rootMatcher().matchStatus(statuses_);
```

library/common/extensions/filters/http/assertion/filter.cc
```diff
@@ -26,53 +28,91 @@ AssertionFilter::AssertionFilter(AssertionFilterConfigSharedPtr config) : config
...
+  if (!matchStatus.matches_ && !matchStatus.might_change_status_) {
```
junr03, line 32: 💡 `might_change_status_` I see.

library/common/extensions/filters/http/assertion/filter.cc
```diff
@@ -26,53 +28,91 @@ AssertionFilter::AssertionFilter(AssertionFilterConfigSharedPtr config) : config
...
                                        nullptr, absl::nullopt, "");
     return Http::FilterHeadersStatus::StopIteration;
   }
 
   if (end_stream) {
-    decoder_callbacks_->sendLocalReply(Http::Code::OK,
-                                       "Request Headers match configured expectations", nullptr,
-                                       absl::nullopt, "");
-    return Http::FilterHeadersStatus::StopIteration;
+    Buffer::OwnedImpl empty_buffer;
```
junr03, line 40: I see, so here we would like to see if there were expectations for body and trailers that because this request ended on a headers frame would not match.

Do you mind adding comments explaining the thought process for posterity?

library/common/extensions/filters/http/assertion/filter.cc
junr03: same elsewhere

test/common/extensions/filters/http/assertion/assertion_filter_test.cc
```diff
@@ -126,6 +119,29 @@ TEST_F(AssertionFilterTest, DataNoMatch) {
   EXPECT_EQ(Http::FilterDataStatus::StopIterationNoBuffer, filter_->decodeData(*body, true));
 }
 
+TEST_F(AssertionFilterTest, DataMissing) {
```
junr03, line 122: We are missing tests for some of  the other combinations e.g, trailer missing in headers, trailer missing in data

library/common/extensions/filters/http/assertion/filter.cc
```diff
@@ -26,53 +28,91 @@ AssertionFilter::AssertionFilter(AssertionFilterConfigSharedPtr config) : config
...
     decoder_callbacks_->sendLocalReply(Http::Code::BadRequest,
                                        "Request Trailers do not match configured expectations",
                                        nullptr, absl::nullopt, "");
     return Http::FilterTrailersStatus::StopIteration;
   }
-
-  decoder_callbacks_->sendLocalReply(
-      Http::Code::OK, "Request Trailers match configured expectations", nullptr, absl::nullopt, "");
-  return Http::FilterTrailersStatus::StopIteration;
+  if (!matchStatus.matches_) {
```
junr03, line 109: I think I am confused about why we would catch this here rather than in the decodeData call, is it because the matcher is not completely done with knowing that data _will not_ match until we see an end_stream == true in decodeData, or a decodeTrailers call (implying there is no more data to match?).

Same as above could we add comments so that the thinking is clear?

junr03 (commented):
Thanks for working on this @goaway!

buildbreaker (approved):
Thanks for doing this, @goaway !

library/common/extensions/filters/http/assertion/filter.cc
```diff
@@ -28,25 +33,27 @@ AssertionFilter::AssertionFilter(AssertionFilterConfigSharedPtr config) : config
...
-    if (!matchStatus.matches_ && !matchStatus.might_change_status_) {
+    auto& match_status = config_->rootMatcher().matchStatus(statuses_);
+    if (!match_status.matches_ && !match_status.might_change_status_) {
       decoder_callbacks_->sendLocalReply(Http::Code::BadRequest,
                                          "Request Body does not match configured expectations",
                                          nullptr, absl::nullopt, "");
       return Http::FilterHeadersStatus::StopIteration;
     }
 
+    // Check of there are unsatisfied assertions about stream trailers.
```
junr03, line 56:
```suggestion
    // Check if there are unsatisfied assertions about stream trailers.
```

library/common/extensions/filters/http/assertion/filter.cc
```diff
@@ -69,25 +80,29 @@ Http::FilterHeadersStatus AssertionFilter::decodeHeaders(Http::RequestHeaderMap&
...
+  auto& match_status = config_->rootMatcher().matchStatus(statuses_);
+  if (!match_status.matches_ && !match_status.might_change_status_) {
     decoder_callbacks_->sendLocalReply(Http::Code::BadRequest,
                                        "Request Body does not match configured expectations",
                                        nullptr, absl::nullopt, "");
     return Http::FilterDataStatus::StopIterationNoBuffer;
   }
 
   if (end_stream) {
+    // Check of there are unsatisfied assertions about stream trailers.
```
junr03, line 92:
```suggestion
    // Check if there are unsatisfied assertions about stream trailers.
```

junr03 (approved):
Awesome, thanks! Just a minor misspell

goaway:
thanks @buildbreaker and @junr03!

PR #1077, opened 2020-09-09
swift: fix build warnings in platform filter

rebello95:
These are unused and triggering warnings when building.


goaway:
I'd actually left these in since the warning seemed benign and they're not going to stay unused. But this is fine. ¯\\\_(ツ)\_/¯

goaway (approved):

rebello95:
@goaway I think we should lean towards not introducing warnings unnecessarily (ideally we could enable warnings as errors, but upstream Envoy has several warnings at the moment).

PR #1084, opened 2020-09-11
platform: add Trailers type for better classification

goaway:
Description: Allows for better scoping of header types; in this case, specifically for advanced filters.
Risk Level: Low
Testing: CI


rebello95 (commented):
This seems reasonable. Can we update the callbacks on the client / streams as well?

goaway:
@rebello95 I think those are all already correct because they're already specific `Request~` or `Response~` variants.

rebello95:
Ah right, good call

rebello95 (approved):

goaway:
thanks @rebello95!

PR #1085, opened 2020-09-11
ci: increase delay before checking iOS simulator

goaway:
Description: Increases the sleep before attempting to cat iOS logs in CI to 5 minutes. We're still occasionally trying too early, resulting in flakiness.
Risk Level: Low
Testing: CI

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -49,7 +49,7 @@ jobs:
       - name: 'Run swift app'
         run: bazelisk run --config=ios //examples/swift/hello_world:app &> /tmp/envoy.log &
       - name: 'Wait for simulator to start & perform requests'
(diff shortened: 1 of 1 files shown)


rebello95:
Are the cases where this fails actually too early or is it possible the tests are flaking for a different reason? Five minutes seems like a very long time to wait for a log to print

goaway:
Whenever I've seen it, it looks like the grep simply happened too early.

goaway:
It will generally pass the next time, for one thing.

rebello95 (approved):
Might be worth considering checking at different intervals and marking the job as successful once one of the greps is successful, to reduce CI time in the "happy path" where it logs quickly

goaway:
I like that idea - definitely seems worth exploring.

PR #1087, opened 2020-09-12
core/platform: add platform-level postinit callback wiring

rebello95:
This PR adds a new set of `onEngineRunning` callbacks to the platform (Swift/Kotlin) interfaces.

The intent behind these callbacks is to allow platform consumers to observe when Envoy Mobile has finished setting itself up internally. When `build()` is called from an `EnvoyEngineBuilder`, internal library setup is not complete when the allocation returns. Instead, Envoy actually does additional setup before being called back with a "post init" step from the core Envoy layer indicating that it's ready to send requests. In practice, this doesn't have any direct effect on consumers since any requests sent into the library during this timeframe are queued for sending automatically once setup is complete.

Based on my localized runs of the library, I noticed 800-1000ms between the start of allocation and the time at which Envoy Mobile has completed its setup. There is room for improvement here which will be looked to in the future.

Changes:
- Add the new interface to platform layers
- Implement the core wiring for this callback and pipe that all the way through JNI / Objective-C up to Kotlin/Swift
- Update sample apps to utilize this functionality
- Add core unit test to validate the core callback completes

Co-authored-by: Alan Chiu <achiu@lyft.com>


library/common/engine.cc
```diff
@@ -51,14 +53,23 @@ envoy_status_t Engine::run(const std::string config, const std::string log_level
     // When we improve synchronous failure handling and/or move to dynamic forwarding, we only need
     // to wait until the dispatcher is running (and can drain by enqueueing a drain callback on it,
     // as we did previously).
+    int64_t start_time = std::chrono::duration_cast<std::chrono::milliseconds>(
+                             std::chrono::system_clock::now().time_since_epoch())
+                             .count();
```
rebello95, line 58: Envoy lints against using system time - I think I'll actually revisit this and avoid doing the timing within the library and simply make the closure `void`

library/proguard.txt
```diff
@@ -11,6 +11,10 @@
     native <methods>;
 }
 
+-keep, includedescriptorclasses class io.envoyproxy.envoymobile.engine.types.EnvoyEngineOnSetupComplete {
```
buildbreaker, line 14: 💯

docs/root/api/starting_envoy.rst
```diff
@@ -184,6 +184,25 @@ This functionality is used for stat segmentation.
...
+ready to start dispatching requests. Any requests sent through Envoy before this setup completes
+will be queued automatically, and this function is typically used purely for observability.
+
+**Example**::
+
+  // Kotlin
+  builder.setOnSetupComplete { /*do something*/ }
+
+  // Swift
+  builder.setOnSetupComplete { /*do something*/ }
```
goaway, line 204: What about `onEngineRunning`?

There's a separate, potentially interesting callback point where objects have been initialized, but the main run loop hasn't started and configuration hasn't yet been loaded. `onEngineRunning` is maybe a little bit more explicit about the juncture to which it refers.

rebello95: Okay, updated.

buildbreaker (approved):
🎉

library/common/jni/jni_interface.cc
```diff
@@ -30,6 +30,20 @@ extern "C" JNIEXPORT jlong JNICALL Java_io_envoyproxy_envoymobile_engine_JniLibr
...
+
+  JNIEnv* env = get_env();
+  jobject j_context = static_cast<jobject>(context);
+  jclass jcls_JvmonEngineRunningContext = env->GetObjectClass(j_context);
+  jmethodID jmid_onEngineRunning = env->GetMethodID(
+      jcls_JvmonEngineRunningContext, "invokeOnEngineRunning", "()Ljava/lang/Object;");
+  env->CallObjectMethod(j_context, jmid_onEngineRunning);
+
+  env->DeleteLocalRef(jcls_JvmonEngineRunningContext);
+  env->DeleteGlobalRef(j_context);
```
goaway, line 44: Can you put a comment here? This works only because the other current engine callback doesn't use the context, but if others get added and/or when we move to per-instance runs, this will need to be moved.

goaway (approved):
Looks good, nice work. One minor comment.

library/common/jni/jni_interface.cc
rebello95: Sure, will add in a separate PR since this one is green

rebello95: https://github.com/lyft/envoy-mobile/pull/1101

PR #1091, opened 2020-09-14
objc: block properties should be copy, not strong

goaway:


rebello95 (approved):
Not strictly required with ARC, but good for semantics (further reading: https://stackoverflow.com/a/27156860/1145804)

goaway:
Yep, you're correct it's not a requirement, @rebello95. It's misleading to put anything else though, since ARC _will_ copy them.

PR #1092, opened 2020-09-14
ci: continuously check iOS example app log output

goaway:
Description: This should reduce the flakiness of the iOS example app CI jobs, and allow them to terminate quicker on success. Notes:
- `touch` log file so that `tail` doesn't terminate early if it starts before output.
- `tail -F` so it notices when the log file is replaced.
- `sed` to also stream all log output to `stdout`.

Risk Level: Low
Testing: CI

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -48,15 +48,8 @@ jobs:
       # Run the app in the background and redirect logs.
       - name: 'Run swift app'
         run: bazelisk run --config=ios //examples/swift/hello_world:app &> /tmp/envoy.log &
-      - name: 'Wait for simulator to start & perform requests'
-        run: sleep 300
-      # Print out the log in case the liveliness check fails.
-      # This allows the author to decide if the failure is a flake or a real problem.
-      - name: 'Cat log'
-        run: 'cat /tmp/envoy.log'
-      # Check for the sentinel value that shows the app is alive and well.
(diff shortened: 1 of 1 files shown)


keith:
what ever makes this stop?

goaway:
@keith the timeout on the job or `sed` hitting a match. I was able to come up with a working version that used `read -t` to use a shorter timeout, but not one that also wrote output to `stdout`.

goaway:
@keith fwiw, this looks like it's equivalent to what Android does right now. It should exit early if there's a match, so the happy path won't take a long time. And if it is taking a long time, you should be getting streamed output that you can view and determine there's a problem before it times out. If `timeout` was available I could use that, but I didn't play around with it, since this is running on MacOS. I guess we could install it as part of the job.

keith (approved):

goaway:
Tested failure here: https://github.com/lyft/envoy-mobile/pull/1093

goaway:
This passed (https://github.com/lyft/envoy-mobile/pull/1092/checks?check_run_id=1113958872), but it looks like GitHub Actions and pulled in a check from the intentionally failling branch. Doing a quick re-run to see if it can figure itself out, but otherwise will merge since output on both branches is as expected.

goaway:
This works, I think GitHub Actions just got messed up when I pushed two branches with the same sha.

PR #1098, opened 2020-09-14
stats: remove ability to force flush until required

junr03:
Description: this functionality is not currently being used. Instead of having non-used code in the library (which is also not covered by tests), we should delete and bring back when it is actively being worked on
Risk Level: low - unused functionality
Related to: https://github.com/lyft/envoy-mobile/issues/754

### DIFF
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -110,16 +110,6 @@ void Engine::recordCounter(const std::string& elements, uint64_t count) {
   }
 }
 
-void Engine::flushStats() {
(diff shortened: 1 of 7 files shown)


rebello95:
I'm not necessarily opposed to this but it is a bit unfortunate to be removing this rather than fixing the lifecycle flushing issue you linked (which would result in this code being used). That said, we can revert this when that fix goes in.

rebello95 (approved):

junr03:
Agreed it would be best to get the functionality in place. But I'd rather bring this code back when we decide to prioritize that.

PR #1099, opened 2020-09-15
filters: add coverage for onDestroy and release_filter

goaway:


junr03 (approved):
Thanks! talked offline about failing CI, @goaway is fixing locally and will push imminently

PR #1100, opened 2020-09-16
filters: support new platform ResumeIteration status

goaway:
Description: Adds support for resuming iteration to platform filters via a new *ResumeIteration filter status. The API for platform filters requires all HTTP entities that should be forwarded to be attached to a result status of a filter invocation. When iteration is stopped, previous entities that were not forwarded may now be attached to the special ResumeIteration status, e.g., pending headers might be attached to ResumeIteration returned during an on-data invocation. See code documentation for additional detail. Full API documentation will be included in an upcoming PR.
Risk Level: Moderate
Testing: Unit tests

### DIFF
--- a/library/common/extensions/filters/http/platform_bridge/BUILD
+++ b/library/common/extensions/filters/http/platform_bridge/BUILD
@@ -10,7 +10,7 @@ api_proto_package()
 envoy_cc_library(
     name = "platform_bridge_filter_lib",
     srcs = [
-        "c_types.cc",
+        "c_type_definitions.h",
         "filter.cc",
     ],
     hdrs = [
--- a/library/common/extensions/filters/http/platform_bridge/c_type_definitions.h
+++ b/library/common/extensions/filters/http/platform_bridge/c_type_definitions.h
@@ -1,7 +1,7 @@
 // NOLINT(namespace-envoy)
-#include "library/common/extensions/filters/http/platform_bridge/c_types.h"
-
 #include "envoy/http/filter.h"
+
+#include "library/common/extensions/filters/http/platform_bridge/c_types.h"
 
 const envoy_filter_headers_status_t kEnvoyFilterHeadersStatusContinue =
     static_cast<envoy_filter_headers_status_t>(Envoy::Http::FilterHeadersStatus::Continue);
@@ -42,5 +42,5 @@ const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusContinue =
 const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusStopIteration =
     static_cast<envoy_filter_trailers_status_t>(Envoy::Http::FilterTrailersStatus::StopIteration);
 // See comment above.
-extern const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusResumeIteration =
+const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusResumeIteration =
     kEnvoyFilterTrailersStatusContinue - 1;
--- a/library/common/extensions/filters/http/platform_bridge/c_types.h
+++ b/library/common/extensions/filters/http/platform_bridge/c_types.h
@@ -47,7 +47,7 @@ extern const envoy_filter_data_status_t kEnvoyFilterDataStatusResumeIteration;
 typedef struct {
   envoy_filter_data_status_t status;
   envoy_data data;
-  envoy_headers* extra_headers;
+  envoy_headers* pending_headers;
 } envoy_filter_data_status;
 
 /**
@@ -64,8 +64,8 @@ extern const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusResumeIter
 typedef struct {
   envoy_filter_trailers_status_t status;
   envoy_headers trailers;
-  envoy_headers* extra_headers;
-  envoy_data* extra_data;
+  envoy_headers* pending_headers;
+  envoy_data* pending_data;
 } envoy_filter_trailers_status;
 
 #ifdef __cplusplus
--- a/library/common/extensions/filters/http/platform_bridge/c_types.h
+++ b/library/common/extensions/filters/http/platform_bridge/c_types.h
@@ -5,6 +5,16 @@
 // NOLINT(namespace-envoy)
 
 /**
+ * Convenience constant indicating no changes to data.
+ */
+extern const envoy_data envoy_unaltered_data;
+
+/**
+ * Convenience constant indicating no changes to headers.
+ */
+extern const envoy_headers envoy_unaltered_headers;
+
+/**
  * Return codes for on-headers filter invocations. @see envoy/http/filter.h
  */
 typedef int envoy_filter_headers_status_t;
@@ -37,7 +47,7 @@ extern const envoy_filter_data_status_t kEnvoyFilterDataStatusResumeIteration;
 typedef struct {
   envoy_filter_data_status_t status;
   envoy_data data;
-  envoy_headers extra_headers;
+  envoy_headers* extra_headers;
 } envoy_filter_data_status;
 
 /**
@@ -54,8 +64,8 @@ extern const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusResumeIter
 typedef struct {
   envoy_filter_trailers_status_t status;
   envoy_headers trailers;
-  envoy_headers extra_headers;
-  envoy_data extra_data;
+  envoy_headers* extra_headers;
+  envoy_data* extra_data;
 } envoy_filter_trailers_status;
 
 #ifdef __cplusplus
--- a/library/common/extensions/filters/http/platform_bridge/c_types.h
+++ b/library/common/extensions/filters/http/platform_bridge/c_types.h
@@ -12,6 +12,7 @@ extern const envoy_filter_headers_status_t kEnvoyFilterHeadersStatusContinue;
 extern const envoy_filter_headers_status_t kEnvoyFilterHeadersStatusStopIteration;
 extern const envoy_filter_headers_status_t kEnvoyFilterHeadersStatusContinueAndEndStream;
 extern const envoy_filter_headers_status_t kEnvoyFilterHeadersStatusStopAllIterationAndBuffer;
+extern const envoy_filter_headers_status_t kEnvoyFilterHeadersStatusResumeIteration;
 
 /**
  * Compound return type for on-headers filter invocations.
@@ -28,6 +29,7 @@ typedef int envoy_filter_data_status_t;
 extern const envoy_filter_data_status_t kEnvoyFilterDataStatusContinue;
 extern const envoy_filter_data_status_t kEnvoyFilterDataStatusStopIterationAndBuffer;
 extern const envoy_filter_data_status_t kEnvoyFilterDataStatusStopIterationNoBuffer;
+extern const envoy_filter_data_status_t kEnvoyFilterDataStatusResumeIteration;
 
 /**
  * Compound return type for on-data filter invocations.
@@ -35,6 +37,7 @@ extern const envoy_filter_data_status_t kEnvoyFilterDataStatusStopIterationNoBuf
 typedef struct {
   envoy_filter_data_status_t status;
   envoy_data data;
+  envoy_headers extra_headers;
 } envoy_filter_data_status;
 
 /**
@@ -43,6 +46,7 @@ typedef struct {
 typedef int envoy_filter_trailers_status_t;
 extern const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusContinue;
 extern const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusStopIteration;
+extern const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusResumeIteration;
 
 /**
  * Compound return type for on-trailers filter invocations.
@@ -50,6 +54,8 @@ extern const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusStopIterat
 typedef struct {
   envoy_filter_trailers_status_t status;
   envoy_headers trailers;
+  envoy_headers extra_headers;
+  envoy_data extra_data;
 } envoy_filter_trailers_status;
 
 #ifdef __cplusplus
--- a/library/common/extensions/filters/http/platform_bridge/c_types.h
+++ b/library/common/extensions/filters/http/platform_bridge/c_types.h
@@ -22,6 +22,8 @@ extern const envoy_filter_headers_status_t kEnvoyFilterHeadersStatusContinue;
 extern const envoy_filter_headers_status_t kEnvoyFilterHeadersStatusStopIteration;
 extern const envoy_filter_headers_status_t kEnvoyFilterHeadersStatusContinueAndEndStream;
 extern const envoy_filter_headers_status_t kEnvoyFilterHeadersStatusStopAllIterationAndBuffer;
+// Note this return status is unique to platform filters and used only to resume iteration after
+// it has been previously stopped.
 extern const envoy_filter_headers_status_t kEnvoyFilterHeadersStatusResumeIteration;
 
 /**
@@ -39,6 +41,8 @@ typedef int envoy_filter_data_status_t;
 extern const envoy_filter_data_status_t kEnvoyFilterDataStatusContinue;
 extern const envoy_filter_data_status_t kEnvoyFilterDataStatusStopIterationAndBuffer;
 extern const envoy_filter_data_status_t kEnvoyFilterDataStatusStopIterationNoBuffer;
+// Note this return status is unique to platform filters and used only to resume iteration after
+// it has been previously stopped.
 extern const envoy_filter_data_status_t kEnvoyFilterDataStatusResumeIteration;
 
 /**
@@ -56,6 +60,8 @@ typedef struct {
 typedef int envoy_filter_trailers_status_t;
 extern const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusContinue;
 extern const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusStopIteration;
+// Note this return status is unique to platform filters and used only to resume iteration after
+// it has been previously stopped.
 extern const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusResumeIteration;
 
 /**
--- a/library/common/extensions/filters/http/platform_bridge/filter.cc
+++ b/library/common/extensions/filters/http/platform_bridge/filter.cc
@@ -104,8 +104,10 @@ Http::FilterDataStatus PlatformBridgeFilter::onData(Buffer::Instance& data, bool
   }
 
   envoy_data in_data;
-  if (iteration_state_ == IterationState::Stopped && internal_buffer &&
-      internal_buffer->length() > 0) {
+  bool already_buffering = iteration_state_ == IterationState::Stopped && internal_buffer &&
+    internal_buffer->length() > 0;
+
+  if (already_buffering) {
     // Pre-emptively buffer data to present aggregate to platform.
     internal_buffer->move(data);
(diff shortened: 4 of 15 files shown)


goaway:
Will be adding additional coverage, but implementation is ready for review.

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -66,31 +77,33 @@ Http::FilterHeadersStatus PlatformBridgeFilter::onHeaders(Http::HeaderMap& heade
...
+  case kEnvoyFilterHeadersStatusContinue:
+    PlatformBridgeFilter::replaceHeaders(headers, result.headers);
+    return Http::FilterHeadersStatus::Continue;
+
+  case kEnvoyFilterHeadersStatusStopIteration:
+    iteration_state_ = IterationState::Stopped;
+    return Http::FilterHeadersStatus::StopIteration;
+
+  default:
+    PANIC("invalid filter state: unsupported status for platform filters");
```
junr03, line 91: Can we change this to NOT_REACHED_GCOVR_EXCL_LINE? We can add a new macro for NOT_REACHED that takes a string if we want to modify the message. Same with everywhere else.

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -101,27 +114,26 @@ Http::FilterDataStatus PlatformBridgeFilter::onData(Buffer::Instance& data, bool
   }
 
   envoy_filter_data_status result = on_data(in_data, end_stream, platform_filter_.instance_context);
-  Http::FilterDataStatus status = static_cast<Http::FilterDataStatus>(result.status);
-  switch (status) {
-  case Http::FilterDataStatus::Continue:
+
+  switch (result.status) {
+  case kEnvoyFilterDataStatusContinue:
     if (iteration_state_ == IterationState::Stopped) {
```
junr03, line 120: Instead of having the if and the panic we could just `RELEASE_ASSERT(iteration_state_ != IterationState::Stopped, msg)`

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -133,24 +145,32 @@ Http::FilterDataStatus PlatformBridgeFilter::onData(Buffer::Instance& data, bool
...
+    return Http::FilterDataStatus::StopIterationNoBuffer;
 
-  // TODO(goaway): Current platform implementations expose immutable data, thus any modification
-  // necessitates a full copy. Add 'modified' bit to determine when we can elide the copy. See also
-  // https://github.com/lyft/envoy-mobile/issues/949 for potential future optimization.
-  if (iteration_state_ == IterationState::Ongoing) {
-    data.drain(data.length());
-    data.addBufferFragment(*Buffer::BridgeFragment::createBridgeFragment(result.data));
+  case kEnvoyFilterDataStatusResumeIteration:
+    if (iteration_state_ != IterationState::Stopped) {
```
junr03, line 151: same as above with the release assert

library/common/extensions/filters/http/platform_bridge/c_type_definitions.h
```diff
@@ -13,15 +13,34 @@ const envoy_filter_headers_status_t kEnvoyFilterHeadersStatusContinueAndEndStrea
...
+//
+// Regarding enum values, the C++11 standard (7.2/2) states:
+//   If the first enumerator has no initializer, the value of the corresponding constant is zero.
+//   An enumerator-definition without an initializer gives the enumerator the value obtained by
+//   increasing the value of the previous enumerator by one.
+//
+// Creating a new return status like this is brittle, but at least somewhat more resilient to
+// a new status being added in Envoy, since it won't overlap as long as the new status is added
+// rather than prepended.
+const envoy_filter_headers_status_t kEnvoyFilterHeadersStatusResumeIteration =
```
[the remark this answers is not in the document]
junr03: After looking at onData, I am still confused why we need two different statuses rather than driving difference via the presence of absence of headers to replace (or headers/data in onTrailers). I might just be being dense, but I wonder if we can eliminate the need of another status to support the use case?

If we _do_ need the new status, can we write more documentation about it in the filter.cc file?

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -133,24 +145,32 @@ Http::FilterDataStatus PlatformBridgeFilter::onData(Buffer::Instance& data, bool
...
+      PANIC("invalid filter state: ResumeIteration  may only be used when filter iteration is "
+            "stopped");
+    }
+    if (result.extra_headers) {
```
junr03, line 155: Why are extra headers only allowed here?

library/common/extensions/filters/http/platform_bridge/c_type_definitions.h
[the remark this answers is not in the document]
junr03: I _think_ I see? In continue iteration the buffer copied is only the current frame, but in resume iteration it is the entire internal buffer?

junr03 (commented):
some initial questions about the core. 

I actually would like @rebello95 and @buildbreaker to review this, as we are getting to more and more complicated implementations and I'd like all of us to understand.

library/common/extensions/filters/http/platform_bridge/c_type_definitions.h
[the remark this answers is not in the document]
goaway: @junr03 I think it will be helpful for you to reference the API doc we put together back when we discussed our implementation for advanced filters. It lays out differences between Envoy's native filters and the platform filters. Also the docstrings on the platform implementations of these APIs contain some detail as well, e.g.
https://github.com/lyft/envoy-mobile/blob/main/library/swift/src/filters/AsyncRequestFilter.swift
https://github.com/lyft/envoy-mobile/blob/main/library/swift/src/filters/FilterResumeStatus.swift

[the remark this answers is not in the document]
goaway: Sorry actually that's the other side of the coin - I mixed up the PRs because I'm working on the asynchronous side right now. For the synchronous side there's this (e.g.):
https://github.com/lyft/envoy-mobile/blob/main/library/swift/src/filters/FilterDataStatus.swift
Because iteration may have been stopped on headers, we need a mechanism for the platform filter to still inform Envoy of what modifications, if any, it wants to perform on headers when iteration is resumed. Same for headers _and_ data, when resumption occurs on trailers.

library/common/extensions/filters/http/platform_bridge/BUILD
```diff
@@ -10,7 +10,7 @@ api_proto_package()
 envoy_cc_library(
     name = "platform_bridge_filter_lib",
     srcs = [
-        "c_types.cc",
+        "c_type_definitions.h",
```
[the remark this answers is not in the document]
goaway: It's not possible to switch on externed constants without compile-time visibility into their defined values. Making this a header allows me to import their actual definitions as part of the only file that needs them at compile time - the filter implementation.

library/common/extensions/filters/http/platform_bridge/c_type_definitions.h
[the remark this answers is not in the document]
goaway: Part of the reason to have two statuses is that it keeps things simple for the main/golden path. When all you're doing is continuing, you never need to worry about sending previous entities.

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -56,6 +57,16 @@ void PlatformBridgeFilter::onDestroy() {
   platform_filter_.instance_context = nullptr;
 }
 
+void PlatformBridgeFilter::replaceHeaders(Http::HeaderMap& headers, envoy_headers c_headers) {
```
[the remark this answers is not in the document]
goaway: The standard request/response path produces new headers; this is the only place we modify them, and moreover, need to modify them in place. I guess I'm not opposed to it, but I don't see an obvious use for it elsewhere at the moment.

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -189,36 +248,56 @@ Http::FilterDataStatus PlatformBridgeFilter::decodeData(Buffer::Instance& data,
...
+  if (decoder_callbacks_->decodingBuffer()) {
+    decoder_callbacks_->modifyDecodingBuffer([&internal_buffer](Buffer::Instance& mutable_buffer) {
       internal_buffer = &mutable_buffer;
     });
   }
 
-  return onData(data, end_stream, internal_buffer, platform_filter_.on_response_data);
+  auto status = onTrailers(trailers, internal_buffer, &pending_request_headers_,
+                           platform_filter_.on_request_trailers);
+  if (status == Http::FilterTrailersStatus::StopIteration) {
```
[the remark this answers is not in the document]
goaway: Filters can stop iteration on trailers and still asynchronously resume iteration later. Consider for instance a filter that buffers an entire request, then kicks off some async side request, then resumes iteration when that side request completes. (We actually internally have some examples of exactly this.)

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: Is that strictly correct though? I mean this _could_ be reached - we receive the value from a public API.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: Yes, you're absolutely right.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: They're only allowed where pending headers might still exist. That means onData or onTrailers (or async onResume) when iteration has been previously stopped.

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -133,24 +145,32 @@ Http::FilterDataStatus PlatformBridgeFilter::onData(Buffer::Instance& data, bool
...
-  if (iteration_state_ == IterationState::Ongoing) {
-    data.drain(data.length());
-    data.addBufferFragment(*Buffer::BridgeFragment::createBridgeFragment(result.data));
+  case kEnvoyFilterDataStatusResumeIteration:
+    if (iteration_state_ != IterationState::Stopped) {
+      PANIC("invalid filter state: ResumeIteration  may only be used when filter iteration is "
+            "stopped");
+    }
+    if (result.extra_headers) {
+      PlatformBridgeFilter::replaceHeaders(**pending_headers, *result.extra_headers);
```
rebello95, line 156: `extra_headers` is a bit of a misnomer IMO because it's actually replacing the headers rather than adding to them, right?

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -133,24 +145,32 @@ Http::FilterDataStatus PlatformBridgeFilter::onData(Buffer::Instance& data, bool
...
+      *pending_headers = nullptr;
+      free(result.extra_headers);
+    }
+    internal_buffer->drain(internal_buffer->length());
+    internal_buffer->addBufferFragment(*Buffer::BridgeFragment::createBridgeFragment(result.data));
+    return Http::FilterDataStatus::Continue;
+
+  default:
```
rebello95, line 164: I think it's worth having this be exhaustive so that if upstream Envoy adds a new iteration state Envoy Mobile won't compile

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -159,25 +179,64 @@ PlatformBridgeFilter::onTrailers(Http::HeaderMap& trailers,
...
-        Http::Utility::convertToString(result.trailers.headers[i].value));
-  }
-  // The C envoy_trailers struct can be released now because the trailers have been copied.
-  release_envoy_headers(result.trailers);
-  return status;
+
+  switch (result.status) {
+  case kEnvoyFilterTrailersStatusContinue:
+    if (iteration_state_ == IterationState::Stopped) {
+      PANIC("invalid filter state: ResumeIteration  may only be used when filter iteration is "
```
rebello95, line 186:
```suggestion
      PANIC("invalid filter state: ResumeIteration may only be used when filter iteration is "
```

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -159,25 +179,64 @@ PlatformBridgeFilter::onTrailers(Http::HeaderMap& trailers,
...
+    PlatformBridgeFilter::replaceHeaders(trailers, result.trailers);
+    return Http::FilterTrailersStatus::Continue;
+
+  case kEnvoyFilterTrailersStatusStopIteration:
+    iteration_state_ = IterationState::Stopped;
+    return Http::FilterTrailersStatus::StopIteration;
+
+  case kEnvoyFilterTrailersStatusResumeIteration:
+    if (iteration_state_ != IterationState::Stopped) {
+      PANIC("invalid filter state: ResumeIteration  may only be used when filter iteration is "
```
rebello95, line 198:
```suggestion
      PANIC("invalid filter state: ResumeIteration may only be used when filter iteration is "
```

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -159,25 +179,64 @@ PlatformBridgeFilter::onTrailers(Http::HeaderMap& trailers,
...
+            "stopped");
+    }
+    if (result.extra_headers) {
+      PlatformBridgeFilter::replaceHeaders(**pending_headers, *result.extra_headers);
+      *pending_headers = nullptr;
+      free(result.extra_headers);
+    }
+    if (result.extra_data) {
```
rebello95, line 206: `extra_data` is also a bit deceiving since it's a replacement set of data

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -159,25 +179,64 @@ PlatformBridgeFilter::onTrailers(Http::HeaderMap& trailers,
...
+      internal_buffer->drain(internal_buffer->length());
+      internal_buffer->addBufferFragment(
+          *Buffer::BridgeFragment::createBridgeFragment(*result.extra_data));
+      free(result.extra_data);
+    }
+    PlatformBridgeFilter::replaceHeaders(trailers, result.trailers);
+    return Http::FilterTrailersStatus::Continue;
+
+  default:
```
rebello95, line 215: Same here regarding default switch statements

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -189,36 +248,56 @@ Http::FilterDataStatus PlatformBridgeFilter::decodeData(Buffer::Instance& data,
...
+    encoder_callbacks_->modifyEncodingBuffer([&internal_buffer](Buffer::Instance& mutable_buffer) {
+      internal_buffer = &mutable_buffer;
+    });
+  }
 
-Http::FilterHeadersStatus PlatformBridgeFilter::encodeHeaders(Http::ResponseHeaderMap& headers,
-                                                              bool end_stream) {
-  // Delegate to shared implementation for request and response path.
-  return onHeaders(headers, end_stream, platform_filter_.on_response_headers);
+  return onData(data, end_stream, internal_buffer, &pending_response_headers_,
```
rebello95, line 264: Is passing a null buffer here acceptable? (I assume yes)

library/common/extensions/filters/http/platform_bridge/filter.h
```diff
@@ -55,15 +55,23 @@ class PlatformBridgeFilter final : public Http::PassThroughFilter,
...
+                                        Buffer::Instance* internal_buffer,
+                                        Http::HeaderMap** pending_headers,
                                         envoy_filter_on_trailers_f on_trailers);
   const std::string filter_name_;
   IterationState iteration_state_;
   envoy_http_filter platform_filter_;
+  Http::HeaderMap* pending_request_headers_{};
+  Http::HeaderMap* pending_response_headers_{};
+  Http::HeaderMap* pending_request_trailers_{};
+  Http::HeaderMap* pending_response_trailers_{};
```
rebello95, line 74: I actually don't see that these references are being cleared - are we assuming they'll be automatically released when the filter is deallocated and that's sufficient?

library/objective-c/EnvoyBridgeUtility.h
```diff
@@ -3,12 +3,26 @@
...
+
   uint8_t *native_bytes = (uint8_t *)safe_malloc(sizeof(uint8_t) * data.length);
   memcpy(native_bytes, data.bytes, data.length);
   envoy_data ret = {data.length, native_bytes, free, native_bytes};
   return ret;
 }
 
+static inline envoy_data *toNativeDataPtr(NSData *data) {
+  if (data == nil) {
+    return nil;
```
rebello95, line 18: Below you're using `NULL`, can we keep these consistent?

rebello95 (dismissed):
Looks good to me after adding inline documentation about behavior (similar to what we have on the platform layer) and addressing Jose's other comments.

library/common/extensions/filters/http/platform_bridge/filter.h
goaway: These member pointers are all being initialized to `nullptr`. This syntax is idiomatic (though it's also newer and perhaps less obvious to an unfamiliar observer than ` = nullptr;`). When they're set to something else in the implementation, they're being set to point to an object they don't manage - something else owns and will clean up that object. (This may be inferred by the type signatures that provide those objects.) The actual space the pointer itself occupies will be automatically reclaimed when the object is destructed.

(Don't worry - these details and their syntax are not all that intuitive.)

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: Yes, but good eye for catching the possibility. Here, `nullptr` represents "none".

library/objective-c/EnvoyBridgeUtility.h
goaway: Yeah, this should be `NULL` because it's not a valid Objective-C object.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: I'll rename it to `pending_data`.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: Yeah, I'll use `pending_`. I used `extra` to differentiate between the internal entity that hadn't been sent yet and those attached to the status as "extra", but I agree it's probably a confusing choice.

library/common/extensions/filters/http/platform_bridge/c_types.h
```diff
@@ -43,13 +56,16 @@ typedef struct {
...
+extern const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusResumeIteration;
 
 /**
  * Compound return type for on-trailers filter invocations.
  */
 typedef struct {
   envoy_filter_trailers_status_t status;
   envoy_headers trailers;
+  envoy_headers* extra_headers;
+  envoy_data* extra_data;
```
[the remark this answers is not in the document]
goaway: Inline documentation coming.

goaway:
Thanks for the review @rebello95!

buildbreaker (dismissed):

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: Unfortunately we won't be able to get these compile-time guarantees here regardless. C/C++ don't enforce this, and static analysis would miss it too, since we're iterating over a constant set only partially defined in terms of the internal enum (ResumeIteration doesn't map to anything internal).

PR #1102, opened 2020-09-16
test: add tests around main interface

junr03:
Description: adds explicit tests for main interface functions.
Risk Level: low
Testing: add new integration style tests for the main interface static functions.

### DIFF
--- a/library/common/http/dispatcher.h
+++ b/library/common/http/dispatcher.h
@@ -100,6 +100,10 @@ public:
   envoy_status_t cancelStream(envoy_stream_t stream);
 
   const DispatcherStats& stats() const;
+  // Used to fill response code details for streams that are cancelled via cancelStream.
+  const std::string& getCancelDetails() {
+    CONSTRUCT_ON_FIRST_USE(std::string, "client cancelled stream");
+  }
 
   // Used for testing.
   Thread::ThreadSynchronizer& synchronizer() { return synchronizer_; }
@@ -168,11 +172,14 @@ private:
     const Network::Address::InstanceConstSharedPtr& connectionLocalAddress() override {
       return parent_.address_;
     }
+    absl::string_view responseDetails() override { return details_; }
     // TODO: https://github.com/lyft/envoy-mobile/issues/825
     void readDisable(bool /*disable*/) override {}
     uint32_t bufferLimit() override { return 65000; }
     // Not applicable
     void setFlushTimeout(std::chrono::milliseconds) override {}
+
+    void setDetails(absl::string_view details) { details_ = details; }
 
     const envoy_stream_t stream_handle_;
 
@@ -181,6 +188,8 @@ private:
     // Used to receive incoming HTTP stream operations.
     DirectStreamCallbacksPtr callbacks_;
     Dispatcher& parent_;
+    // Response details used by the connection manager.
+    absl::string_view details_;
   };
 
   using DirectStreamSharedPtr = std::shared_ptr<DirectStream>;
--- a/library/common/http/dispatcher.h
+++ b/library/common/http/dispatcher.h
@@ -172,14 +172,16 @@ private:
     const Network::Address::InstanceConstSharedPtr& connectionLocalAddress() override {
       return parent_.address_;
     }
-    absl::string_view responseDetails() override { return details_; }
+    absl::string_view responseDetails() override { return response_details_; }
     // TODO: https://github.com/lyft/envoy-mobile/issues/825
     void readDisable(bool /*disable*/) override {}
     uint32_t bufferLimit() override { return 65000; }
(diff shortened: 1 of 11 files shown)


junr03:
had to pull in Envoy update for buffer filter, and also preemptively fixed behavior that would have caused crashes due to stream cancellation. Funny enough writing the test caught it :)

test/common/main_interface_test.cc
```diff
@@ -0,0 +1,128 @@
...
+      "envoy.extensions.filters.http.router.v3.Router\"}}]}}}]},\"layered_runtime\":{\"layers\":[{"
+      "\"name\":\"static_layer_0\",\"static_layer\":{\"overload\":{\"global_downstream_max_"
+      "connections\":50000}}}]}}";
+  const std::string level = "debug";
+  absl::Notification exit;
+  envoy_engine_callbacks engine_cbs{[](void* context) -> void {
+                                      auto* exit = static_cast<absl::Notification*>(context);
+                                      exit->Notify();
+                                    },
+                                    &exit};
```
rebello95, line 35: You'll need to merge main for this; there's an `on_engine_running` callback now

test/common/main_interface_test.cc
```diff
@@ -0,0 +1,128 @@
...
+  run_engine(0, engine_cbs, config.c_str(), level.c_str());
+
+  absl::Notification on_complete_notification;
+  envoy_http_callbacks stream_cbs{[](envoy_headers, bool, void*) -> void* { return nullptr; },
+                                  nullptr,
+                                  nullptr,
+                                  nullptr,
+                                  nullptr,
```
rebello95, line 43: Can you `/*document*/` these?

test/common/main_interface_test.cc
```diff
@@ -0,0 +1,128 @@
...
+
+  start_stream(stream, stream_cbs);
+
+  reset_stream(stream);
+
+  ASSERT_TRUE(on_cancel_notification.WaitForNotificationWithTimeout(absl::Seconds(10)));
+
+  terminate_engine(0);
+
+  ASSERT_TRUE(exit.WaitForNotificationWithTimeout(absl::Seconds(10)));
```
rebello95, line 126: Same comments for this test case as above

test/common/main_interface_test.cc
```diff
@@ -0,0 +1,128 @@
...
+
+  envoy_stream_t stream = init_stream(0);
+
+  start_stream(stream, stream_cbs);
+
+  send_headers(stream, c_headers, false);
+  send_data(stream, c_data, false);
+  send_trailers(stream, c_trailers);
+
+  ASSERT_TRUE(on_complete_notification.WaitForNotificationWithTimeout(absl::Seconds(10)));
```
rebello95, line 70: I think it's worth a docstring somewhere here explaining that the config above triggers a direct response which is why the stream completes. Also, should we validate the response itself?

library/common/http/dispatcher.h
```diff
@@ -168,19 +172,24 @@ class Dispatcher : public Logger::Loggable<Logger::Id::http> {
...
+
     const envoy_stream_t stream_handle_;
 
     // Used to issue outgoing HTTP stream operations.
     RequestDecoder* request_decoder_;
     // Used to receive incoming HTTP stream operations.
     DirectStreamCallbacksPtr callbacks_;
     Dispatcher& parent_;
+    // Response details used by the connection manager.
+    absl::string_view details_;
```
rebello95, line 192:
```suggestion
    absl::string_view response_details_;
```

library/common/http/dispatcher.h
```diff
@@ -168,19 +172,24 @@ class Dispatcher : public Logger::Loggable<Logger::Id::http> {
...
       return parent_.address_;
     }
+    absl::string_view responseDetails() override { return details_; }
     // TODO: https://github.com/lyft/envoy-mobile/issues/825
     void readDisable(bool /*disable*/) override {}
     uint32_t bufferLimit() override { return 65000; }
     // Not applicable
     void setFlushTimeout(std::chrono::milliseconds) override {}
 
+    void setDetails(absl::string_view details) { details_ = details; }
```
rebello95, line 182: `setResponseDetails`? I'm also curious as to what the benefit is of having an explicit setter when `details_` has the same access level below and can be modified directly

rebello95 (approved):
LGTM modulo some comments/questions

library/common/http/dispatcher.h
junr03: yep, good eye. I actually want to turn all the members to private to prevent a conn manager <-> async stream situation here. So wanted to prepare for that.

rebello95 (dismissed):

PR #1103, opened 2020-09-17
assertion filter: add config coverage

junr03:
Description: assertion filter - add config coverage
Risk Level: low
Testing: added unit tests

### DIFF
--- a/test/common/extensions/filters/http/assertion/assertion_filter_test.cc
+++ b/test/common/extensions/filters/http/assertion/assertion_filter_test.cc
@@ -1,9 +1,11 @@
 #include "test/mocks/http/mocks.h"
 #include "test/test_common/utility.h"
+#include "test/mocks/server/factory_context.h"
 
 #include "gtest/gtest.h"
 #include "library/common/extensions/filters/http/assertion/filter.h"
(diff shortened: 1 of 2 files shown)


test/common/extensions/filters/http/assertion/assertion_filter_test.cc
```diff
@@ -495,6 +497,30 @@ TEST_F(AssertionFilterTest, ResponseTrailersNoMatch) {
   EXPECT_EQ(Http::FilterTrailersStatus::StopIteration, filter_->encodeTrailers(response_trailers));
 }
 
+TEST(AssertionFilterFactoryTest, Config) {
```
rebello95, line 500: Maybe this is an upstream envoy convention to name the test minimally (`Config`), but I think it might be nice to make the name more clearly demonstrate what specifically the test is validating. In our platform tests, names are usually closer to sentences describing the behavior

rebello95 (approved):

test/common/extensions/filters/http/assertion/assertion_filter_test.cc
junr03: 😅 yeah convention for this type of very very basic acceptance test. Let me merge and change in subsequent PR.

PR #1104, opened 2020-09-17
record counter: return envoy_status_t and add interface test

junr03:
Description: all main interface functions return envoy_status_t. This PR fixes record_counter to follow suit and adds a test.
Risk Level: med - API breaking change
Testing: added test


library/common/engine.h
```diff
@@ -39,8 +39,9 @@ class Engine {
    * Increment a counter with a given string of elements and by the given count.
    * @param elements, joined elements of the timeseries.
    * @param count, amount to add to the counter.
+   * @param envoy_status_t, the resulting status of the operation.
    */
-  void recordCounter(const std::string& elements, uint64_t count);
+  envoy_status_t recordCounter(const std::string& elements, uint64_t count);
```
goaway, line 44: Since we're here let's add the `envoy_engine_t` argument as well.

library/common/main_interface.h
```diff
@@ -98,8 +98,11 @@ envoy_status_t set_preferred_network(envoy_network_t network);
 
 /**
  * Increment a counter with the given elements and by the given count.
+ * @param elements, the string that identifies the counter to increment.
+ * @param count, the count to increment by.
+ * @param envoy_status_t, the resulting status of the operation.
  */
-void record_counter(const char* elements, uint64_t count);
+envoy_status_t record_counter(const char* elements, uint64_t count);
```
goaway, line 105: `envoy_engine_t` arg here too.

library/common/engine.h
junr03: I wanted to be more selective in this PR and then have that in a subsequent one. wdyt @goaway?

library/swift/test/StatsClientImplTests.swift
```diff
@@ -19,7 +19,7 @@ final class StatsClientImplTests: XCTestCase {
     let mockEngine = MockEnvoyEngine()
     let statsClient = StatsClientImpl(engine: mockEngine)
     let counter = statsClient.counter(elements: ["test", "stat"])
-    counter.increment()
+    XCTAssertEqual(counter.increment(), 0)
```
junr03, line 22: @rebello95 I was not able how to import kEnvoySuccess here

rebello95: If you `@testable import EnvoyEngine` it should work I believe

junr03: worked, thanks!

library/swift/src/stats/Counter.swift
```diff
@@ -4,12 +4,14 @@ import Foundation
...
-  func increment(count: Int)
+  /// - returns: A status indicating if the action was successful.
+  func increment(count: Int) -> Int32
 }
 
 extension Counter {
   /// Increment the counter by 1.
-  public func increment() {
+  /// - returns: A status indicating if the action was successful.
+  public func increment() -> Int32{
```
rebello95, line 14:
```suggestion
  ///
  /// - returns: A status indicating if the action was successful.
  public func increment() -> Int32 {
```

library/swift/src/stats/Counter.swift
```diff
@@ -4,12 +4,14 @@ import Foundation
 @objc
 public protocol Counter: AnyObject {
   /// Increment the counter by the given count.
-  func increment(count: Int)
+  /// - returns: A status indicating if the action was successful.
+  func increment(count: Int) -> Int32
```
rebello95, line 8:
```suggestion
  ///
  /// - returns: A status indicating if the action was successful.
  func increment(count: Int) -> Int32
```

library/swift/src/stats/CounterImpl.swift
```diff
@@ -14,11 +14,11 @@ class CounterImpl: NSObject, Counter {
   }
 
   /// Increment the counter by the given count.
```
rebello95, line 16:
```suggestion
  /// Increment the counter by the given count.
  ///
  /// - returns: A status indicating if the action was successful.
```

library/swift/test/CounterTests.swift
```diff
@@ -5,13 +5,14 @@ final class CounterTests: XCTestCase {
     func testConvenientMethodDelegatesToTheMainMethod() {
         class MockCounterImpl: Counter {
             var count: Int?
-            func increment(count: Int) {
+            func increment(count: Int) -> Int32 {
                 self.count = count
+                return 0
```
rebello95, line 10: Should we use `kEnvoySuccess` in this file too?

test/common/engine_test.cc
```diff
@@ -49,4 +49,42 @@ TEST_F(EngineTest, EarlyExit) {
 
   start_stream(0, {});
 }
+
+// Move to main_interface_test
```
rebello95, line 53: `TODO`?

library/kotlin/src/io/envoyproxy/envoymobile/stats/Counter.kt
```diff
@@ -7,5 +7,6 @@ interface Counter {
   /**
    * Increment the counter by the given count.
    */
+  // TODO: return A status indicating if the action was successful.
```
rebello95, line 10: Why is this implemented in Swift but not Kotlin?

goaway, line 10: nit: I'd drop this TODO for now - I'm not certain we _do_ want to expose this publicly via a return code. (We can also throw an exception. Or log. Or do nothing.)

library/swift/src/stats/Counter.swift
goaway, line 8: At this layer I'd have it return void. We don't currently expose Envoy return codes beyond the bridge layer elsewhere, and they're only return codes there because of a lack of a platform-agnostic way to signal success, if relevant.

goaway (commented):
I think `Counter#increment` should generally return `void`. I'm open to having a discussion about whether and how to signal failures, but a return code wouldn't be my preference when better mechanisms exist at these layers.

junr03:
reverted the platform return and created https://github.com/lyft/envoy-mobile/issues/1113

library/kotlin/src/io/envoyproxy/envoymobile/stats/CounterImpl.kt
```diff
@@ -15,6 +15,7 @@ internal class CounterImpl : Counter {
     this.elements = elements
   }
 
+  /// TODO: raise error to platform if the operation is not successful.
```
goaway, line 18:
```suggestion
  /// TODO: Potentially signal error to platform if the operation is not successful.
```

It's not altogether certain to me that we should raise, though it is an option.

library/swift/src/stats/CounterImpl.swift
```diff
@@ -14,6 +14,7 @@ class CounterImpl: NSObject, Counter {
...
+  /// TODO: raise error to platform if the operation is not successful.
```
goaway, line 17: Same as above.

goaway (approved):
Please soften language in TODOs, otherwise LGTM.

PR #1105, opened 2020-09-17
jni: minor clarification

goaway:


goaway:
@rebello95 sorry I should have been more clear in my PR comment. Deleting the local ref here makes total sense. It's deleting the global ref that's a little bit odd in terms of timing and merits some description.

rebello95 (dismissed):

PR #1108, opened 2020-09-18
build: add asan config runnable on MacOS

goaway:
Description: Bazel config `clang-asan` currently is unable to be run on MacOS due to a macro redefinition and missing dependencies. This alternative config works around these issues.
Risk Level: Low
Testing: Local

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -34,17 +34,18 @@ build:ios --define=manual_stamp=manual_stamp
 # Default flags for builds targeting Android
 build:android --define=logger=android
 
-# ASAN config runnable on MacOS
+# Locally-runnable ASAN config for MacOS & Linux with standard dev environment
+# See also:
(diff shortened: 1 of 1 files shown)


rebello95:
Should we add a CI job to run these tests on macOS CI?

goaway:
Upstream Envoy actually has this issue right now that would prevent us from enabling in CI: https://github.com/envoyproxy/envoy/issues/12629

It's actually a real issue, but doesn't affect us, because we don't bind local IPv6 ports. So one could argue in favor of it - but I think our current CI asan test is probably still good enough for now.

I'd still suggest merging this, since it's actually the only config I was able to get running on either OS locally.

goaway:
@junr03 updated per our discussion.

junr03 (approved):

PR #1110, opened 2020-09-18
core: add missing coverage

junr03:
Description: add areas of missing coverage across the core codebase.
Risk Level: low
Testing: adds more testing!

### DIFF
--- a/test/common/http/dispatcher_test.cc
+++ b/test/common/http/dispatcher_test.cc
@@ -1324,5 +1324,127 @@ TEST_F(DispatcherTest, ResetWhenRemoteClosesBeforeLocal) {
   ASSERT_EQ(cc.on_error_calls, 0);
 }
 
+TEST_F(DispatcherTest, Encode100Continue) {
+  ready();
+
+  envoy_stream_t stream = 1;
+  envoy_http_callbacks bridge_callbacks;
+
+  // Build a set of request headers.
+  TestRequestHeaderMapImpl headers;
+  HttpTestUtility::addDefaultHeaders(headers);
+  envoy_headers c_headers = Utility::toBridgeHeaders(headers);
+
+  // Create a stream.
+  Event::PostCb start_stream_post_cb;
+  EXPECT_CALL(event_dispatcher_, post(_)).WillOnce(SaveArg<0>(&start_stream_post_cb));
+  EXPECT_EQ(http_dispatcher_.startStream(stream, bridge_callbacks), ENVOY_SUCCESS);
+
+  // Grab the response encoder in order to dispatch responses on the stream.
+  // Return the request decoder to make sure calls are dispatched to the decoder via the dispatcher
+  // API.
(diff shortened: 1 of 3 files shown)


test/common/http/dispatcher_test.cc
```diff
@@ -1324,5 +1324,127 @@ TEST_F(DispatcherTest, ResetWhenRemoteClosesBeforeLocal) {
...
+  Event::PostCb send_headers_post_cb;
+  EXPECT_CALL(event_dispatcher_, post(_)).WillOnce(SaveArg<0>(&send_headers_post_cb));
+  http_dispatcher_.sendHeaders(stream, c_headers, true);
+
+  EXPECT_CALL(request_decoder_, decodeHeaders_(_, true));
+  send_headers_post_cb();
+
+  // Encode 100 continue should blow up.
+  TestResponseHeaderMapImpl response_headers{{":status", "200"}};
+  EXPECT_DEATH(response_encoder_->encode100ContinueHeaders(response_headers), "panic: not implemented");
```
rebello95, line 1363: Cool that this works

rebello95 (approved):

test/common/network/synthetic_address_impl_test.cc
```diff
@@ -22,8 +22,10 @@ TEST(SyntheticAddressImplTest, Accessors) {
   SyntheticAddressImpl address;
   ASSERT_EQ(address.ip(), nullptr);
   ASSERT_EQ(address.pipe(), nullptr);
+  ASSERT_EQ(address.envoyInternalAddress(), nullptr);
   ASSERT_EQ(address.sockAddr(), nullptr);
   ASSERT_EQ(address.sockAddrLen(), 0);
+  address.socketInterface();
```
goaway, line 28: ?

test/common/main_interface_test.cc
```diff
@@ -218,6 +218,75 @@ TEST(MainInterfaceTest, ResetStream) {
...
+  terminate_engine(0);
+
+  ASSERT_TRUE(engine_cbs_context.on_exit.WaitForNotificationWithTimeout(absl::Seconds(10)));
+}
+
+TEST(MainInterfaceTest, InitEngineReturns1) {
+  // TODO(goaway): return new handle once multiple engine support is in place.
+  // https://github.com/lyft/envoy-mobile/issues/332
+  ASSERT_EQ(1, init_engine());
+}
```
goaway, line 288: This doesn't really feel like a useful test to me. I get that we're just trying to increase coverage, but I don't know that we actually want to be asserting or implying that this returns a specific value. If we were to change this to return 2 tomorrow, everything should still work, and the only thing that would be different is that now we'd have a broken test.

junr03: I see this as a sentinel that we have to have coverage for `init_engine()`. So in my mind having a test that breaks is, in and of itself, useful.

goaway (approved):

PR #1111, opened 2020-09-18
coverage: improve AssertionFilter coverage

goaway:
Description: Removes two branches that aren't possible to hit with Envoy's matcher implementation today. Adds two tests that cover an alternate path through body assertion failure.
Risk Level: Low
Testing: Local and CI


junr03 (approved):
Makes sense to me. I think this should cover the missing branches. Lets merge and I will report back. Thanks so much for writing all this!

PR #1114, opened 2020-09-18
Implement isEqual method

Augustyniak:
Description: Providing the "correct" implementation of `isEqual` method for `Element` object. This should help us for cases when we want have unit tests that want to verify whether a metrics with right `Element`'s is emitted or not.
Risk Level: Small
Testing: Unit Tests
Docs Changes: None
Release Notes: None


library/swift/test/BUILD
```diff
@@ -77,3 +77,10 @@ envoy_mobile_swift_test(
         "CounterTests.swift",
     ],
 )
+
+envoy_mobile_swift_test(
+    name = "element_tests",
+    srcs = [
+        "ElementTests.swift",
+    ],
+)
```
rebello95, line 86: Trailing space 😃

library/swift/src/stats/Element.swift
```diff
@@ -14,4 +14,12 @@ public final class Element: NSObject, ExpressibleByStringLiteral {
     }
     self.value = value
   }
+
+  public override func isEqual(_ object: Any?) -> Bool {
+    if let object = object as? Element {
+      return self.value == object.value
+    } else {
+      return false
+    }
```
rebello95, line 23:
```suggestion
    return (object as? Element)?.value == self.value
```

library/swift/test/ElementTests.swift
```diff
@@ -0,0 +1,14 @@
...
+
+final class ElementTests: XCTestCase {
+    func testElementNotEqualToObjectOfUnrelatedType() {
+        let element = Element("foo")
+        XCTAssertFalse(element.isEqual("bar"))
+    }
+
+    func testElementsEquality() {
+      XCTAssertEqual(Element("foo"), Element("foo"))
+      XCTAssertNotEqual(Element("foo"), Element("bar"))
```
rebello95, line 12: Should we explicitly do an `isEqual` check here too?

library/swift/test/ElementTests.swift
```diff
@@ -0,0 +1,14 @@
+import Envoy
+import XCTest
+
+final class ElementTests: XCTestCase {
+    func testElementNotEqualToObjectOfUnrelatedType() {
+        let element = Element("foo")
+        XCTAssertFalse(element.isEqual("bar"))
+    }
+
+    func testElementsEquality() {
```
rebello95, line 10:
```suggestion
    func testElementsAreEqualIfStringValuesAreEqual() {
```

library/swift/test/ElementTests.swift
Augustyniak: maybe lets add 2 additional assertions - they would use `isEqual` method. WDYT?

library/swift/test/BUILD
Augustyniak: it wasn't here previously so I though that it was not needed 🤔

rebello95: Screen Shot 2020-09-18 at 3 42 46 PM

I was referring to this ^ (not showing up on the main GitHub screen for some reason, only when you view files)

library/swift/test/ElementTests.swift
rebello95: sure

rebello95 (dismissed):

rebello95 (dismissed):
Thanks!

rebello95 (approved):

buildbreaker (approved):

PR #1115, opened 2020-09-18
config: rename decompressor name from "basic" to "gzip"

rebello95:
This filter is configured exclusively with gzip decompression. If we were to add more in the future (i.e., brotli), this would be a good way to differentiate.

This change is especially useful since it'll transform `x-envoy-decompressor-basic-uncompressed-bytes` to use "gzip" instead of "basic" in the trailers it adds to responses.


junr03 (approved):
Seems reasonable to me! @rebello95 I would take a look at our internal stats macros. I believe some of them will stop matching because this will also change the stats name emitted.

rebello95:
Yep, will do @junr03!

PR #1117, opened 2020-09-22
stats: add gauge methods to engine and interface layers

jingwei99:
Description: Adds methods to support stats type Gauge to the engine and interface layers.
Risk Level: Medium
Testing: Unit tests


test/common/engine_test.cc
```diff
@@ -49,42 +49,4 @@ TEST_F(EngineTest, EarlyExit) {
 
   start_stream(0, {});
 }
-
-// TODO: move to main_interface_test.
```
jingwei99, line 53: moved to `main_interface_test`

junr03: thanks!

test/common/main_interface_test.cc
```diff
@@ -13,6 +13,26 @@ typedef struct {
   absl::Notification on_exit;
 } engine_test_context;
 
+// This config is the minimal envoy mobile config that allows for running the engine.
+// There is nothing functional about the config, as the created stream is only used for
+// send_metadata.
+const std::string CONFIG =
```
junr03, line 19:
```suggestion
const std::string MINIMAL_NOOP_CONFIG =
```

test/common/main_interface_test.cc
```diff
@@ -13,6 +13,26 @@ typedef struct {
   absl::Notification on_exit;
 } engine_test_context;
 
+// This config is the minimal envoy mobile config that allows for running the engine.
+// There is nothing functional about the config, as the created stream is only used for
+// send_metadata.
```
junr03, line 18: do you mind updating the comment, as this is now used in more places?

test/common/main_interface_test.cc
```diff
@@ -13,6 +13,26 @@ typedef struct {
   absl::Notification on_exit;
 } engine_test_context;
 
+// This config is the minimal envoy mobile config that allows for running the engine.
```
junr03, line 16: Thanks! Great clean up

test/common/main_interface_test.cc
junr03: `There is nothing functional about the config, as the created stream is only used for send_metadata.` This part of the comment (and similar parts in other deleted comments) could be left in the test itself explaining why the minimal config is being used.

junr03 (commented):
Awesome, lgtm. Just a few comments

library/common/engine.h
```diff
@@ -39,10 +39,30 @@ class Engine {
...
-   * @param envoy_status_t, the resulting status of the operation.
    */
   envoy_status_t recordCounter(const std::string& elements, uint64_t count);
 
+  /**
+   * Set a gauge of a given string of elements with the given value.
+   * @param elements, joined elements of the timeseries.
+   * @param value, value to set to the gauge.
+   */
+  envoy_status_t setGauge(const std::string& elements, uint64_t value);
```
goaway, line 50: I had originally been thinking we could group all stats-recording functions by prefixing them with `record*`. These could be called `recordGuageSet`, `recordGuageAdd`, `recordGuageSub`. `recordCounter` could be change to `recordCounterInc` or stay the same. What do you think?

goaway:
Nice work @jingwei99! This looks good. Just one comment on naming at the main_interface/engine layer.

library/common/engine.h
jingwei99: Grouping all the stats-recording functions by the same prefix sounds a good idea to me. 

I'll update the gauge related methods to the suggested names in this pr. 

As for the renaming of `recordCounter`, I assume we want the renaming to go all the way to the platform apis. It will be a breaking change, and I will split that change into a separate pr.

How does this plan sound? @goaway

goaway (approved):

PR #1118, opened 2020-09-23
filters: add onAsyncResume invocation

goaway:


library/common/extensions/filters/http/platform_bridge/c_type_definitions.h
```diff
@@ -44,3 +44,7 @@ const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusStopIteration =
 // See comment above.
 const envoy_filter_trailers_status_t kEnvoyFilterTrailersStatusResumeIteration =
     kEnvoyFilterTrailersStatusContinue - 1;
+
+// These values don't exist in Envoy and are essentially arbitrary.
+const envoy_filter_resume_status_t kEnvoyFilterResumeStatusStopIteration = 1;
+const envoy_filter_resume_status_t kEnvoyFilterResumeStatusResumeIteration = -1;
```
rebello95, line 50: If these are truly arbitrary, why not start zero-indexed? I assume the `-1` is coming from the same reasoning above, in which case we should probably make that clearer here

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -313,9 +321,111 @@ PlatformBridgeFilter::encodeTrailers(Http::ResponseTrailerMap& trailers) {
...
+  Buffer::Instance* internal_buffer = nullptr;
+  if (decoder_callbacks_->decodingBuffer()) {
+    decoder_callbacks_->modifyDecodingBuffer([&internal_buffer](Buffer::Instance& mutable_buffer) {
+      internal_buffer = &mutable_buffer;
+    });
+  }
+
+  envoy_headers* pending_headers = nullptr;
+  envoy_data* pending_data = nullptr;
+  envoy_headers* pending_trailers = nullptr;
```
rebello95, line 338: nits on naming:
- `pending_*` here might be easily misread between `pending_headers`/`pending_request_header`. Might be worth changing to `headers_to_send` or something like that
- why `internal_buffer` instead of `pending_request_data` to align with the others?

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -313,9 +321,111 @@ PlatformBridgeFilter::encodeTrailers(Http::ResponseTrailerMap& trailers) {
...
+      platform_filter_.on_resume_request(pending_headers, pending_data, pending_trailers,
+                                         request_complete_, platform_filter_.instance_context);
+  if (result.status == kEnvoyFilterResumeStatusStopIteration) {
+    return;
+  }
+  if (pending_request_headers_) {
+    PlatformBridgeFilter::replaceHeaders(*pending_request_headers_, *result.pending_headers);
+    pending_request_headers_ = nullptr;
+    free(result.pending_headers);
+  }
```
rebello95, line 359: Should we assert in the `else` case that `result.pending_headers` is not passed?

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -313,9 +321,111 @@ PlatformBridgeFilter::encodeTrailers(Http::ResponseTrailerMap& trailers) {
...
+    inject_data.addBufferFragment(
+        *Buffer::BridgeFragment::createBridgeFragment(*result.pending_data));
+    decoder_callbacks_->addDecodedData(inject_data, false);
+    free(result.pending_data);
+  }
+  if (pending_request_trailers_) {
+    PlatformBridgeFilter::replaceHeaders(*pending_request_trailers_, *result.pending_trailers);
+    pending_request_trailers_ = nullptr;
+    free(result.pending_trailers);
+  }
```
rebello95, line 376: Same here regarding assert

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -313,9 +321,111 @@ PlatformBridgeFilter::encodeTrailers(Http::ResponseTrailerMap& trailers) {
...
+  if (internal_buffer) {
+    internal_buffer->drain(internal_buffer->length());
+    internal_buffer->addBufferFragment(
+        *Buffer::BridgeFragment::createBridgeFragment(*result.pending_data));
+    free(result.pending_data);
+  } else if (result.pending_data) {
+    Buffer::OwnedImpl inject_data;
+    inject_data.addBufferFragment(
+        *Buffer::BridgeFragment::createBridgeFragment(*result.pending_data));
+    decoder_callbacks_->addDecodedData(inject_data, false);
```
rebello95, line 369: is `false` referring to `end_stream` here? If so, is it safe to always assume `false` (how does the consumer close the stream?)?

library/common/extensions/filters/http/platform_bridge/filter.cc
rebello95: I now see below that `internal_buffer` is reused between request/response, which I assume is the reason for that name. Are there any concerns around that state not being properly cleared before starting to receive response data?

library/common/extensions/filters/http/platform_bridge/filter.h
```diff
@@ -50,6 +50,18 @@ class PlatformBridgeFilter final : public Http::PassThroughFilter,
 public:
   PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr config);
 
+  // Scheduled on the dispatcher when resumeRequest is called from platform
+  // filter callbacks. Provides a snapshot of pending request state to the
+  // platform filter, and consumes invocation results to modify pending HTTP
+  // entities before resuming decoding.
+  void onResumeDecoding();
```
rebello95, line 57: I know decoding/encoding is an Envoy construct, but I think `onResumeRequest`/`onResumeResponse` would be clearer here since these are executed as a result of functions with similar names in Envoy Mobile

library/objective-c/EnvoyEngine.h
```diff
@@ -74,6 +74,17 @@ extern const int kEnvoyFilterTrailersStatusContinue;
...
+/// Return codes for on-resume filter invocations. These are unique to platform filters,
+/// and used exclusively after an asynchronous request to resume iteration via callbacks.
+extern const int kEnvoyFilterResumeStatusStopIteration;
+extern const int kEnvoyFilterResumeStatusResumeIteration;
+
+@protocol EnvoyHTTPFilterCallbacks
+
+- (void)resumeIteration;
+
+@end
```
rebello95, line 86: Mind documenting this?

library/swift/src/filters/ResponseFilterCallbacksImpl.swift
```diff
@@ -0,0 +1,18 @@
...
+
+  init(callbacks: EnvoyHTTPFilterCallbacks) {
+    self.callbacks = callbacks
+    super.init()
+  }
+}
+
+extension ResponseFilterCallbacksImpl: ResponseFilterCallbacks {
+  func resumeResponse() {
+    callbacks.resumeIteration()
```
rebello95, line 16:
```suggestion
    self.callbacks.resumeIteration()
```

library/swift/src/filters/RequestFilterCallbacksImpl.swift
```diff
@@ -0,0 +1,18 @@
...
+
+  init(callbacks: EnvoyHTTPFilterCallbacks) {
+    self.callbacks = callbacks
+    super.init()
+  }
+}
+
+extension RequestFilterCallbacksImpl: RequestFilterCallbacks {
+  func resumeRequest() {
+    callbacks.resumeIteration()
```
rebello95, line 16:
```suggestion
    self.callbacks.resumeIteration()
```

library/objective-c/EnvoyEngine.h
```diff
@@ -88,6 +99,20 @@ extern const int kEnvoyFilterTrailersStatusResumeIteration;
 
 @property (nonatomic, copy) NSArray * (^onResponseTrailers)(EnvoyHeaders *trailers);
 
+@property (nonatomic, copy) void (^setRequestFilterCallbacks)
+    (id<EnvoyHTTPFilterCallbacks> callbacks);
+
+@property (nonatomic, copy) NSArray * (^onResumeRequest)
```
rebello95, line 105: I know adding type annotations here isn't possible since they're varied types, but can we at least document the expected contents and lengths of these `NSArray`s?

library/objective-c/EnvoyFilterCallbacksImpl.m
```diff
@@ -0,0 +1,20 @@
...
+@implementation EnvoyHTTPFilterCallbacks {
+}
+
+- (instancetype)init {
+  self = [super init];
+  if (!self) {
+    return nil
+  }
+  return self;
+}
```
rebello95, line 14: Why do we need this?

library/objective-c/EnvoyFilterCallbacksImpl.m
```diff
@@ -0,0 +1,20 @@
+#import "library/objective-c/EnvoyEngine.h"
+
+#pragma mark - EnvoyHTTPFilterCallbacksImpl
+
+@implementation EnvoyHTTPFilterCallbacks {
+}
```
rebello95, line 6:
```suggestion
@implementation EnvoyHTTPFilterCallbacks
```

library/objective-c/EnvoyFilterCallbacksImpl.m
```diff
@@ -0,0 +1,20 @@
...
+- (instancetype)init {
+  self = [super init];
+  if (!self) {
+    return nil
+  }
+  return self;
+}
+
+- (void)resumeIteration {
+  NSLog(@"async resume");
```
rebello95, line 17: Should there be a TODO here?

library/swift/src/filters/Filter.swift
```diff
@@ -105,5 +105,55 @@ extension EnvoyHTTPFilter {
...
+    if let asyncRequestFilter = filter as? AsyncRequestFilter {
+      self.setRequestFilterCallbacks = { envoyCallbacks in
+        asyncRequestFilter.setRequestFilterCallbacks(
+          RequestFilterCallbacksImpl(callbacks: envoyCallbacks)
+        )
+      }
+
+      self.onResumeRequest = { envoyHeaders, data, envoyTrailers, endStream in
+        let result = asyncRequestFilter.onResumeRequest(
+          headers: envoyHeaders.flatMap(RequestHeaders.init),
```
rebello95, line 118:
```suggestion
          headers: envoyHeaders.map(RequestHeaders.init),
```

library/swift/src/filters/Filter.swift
```diff
@@ -105,5 +105,55 @@ extension EnvoyHTTPFilter {
...
+          data: data,
+          trailers: envoyTrailers.flatMap(RequestTrailers.init),
```
rebello95, line 120:
```suggestion
          trailers: envoyTrailers.map(RequestTrailers.init),
```

library/swift/src/filters/Filter.swift
```diff
@@ -105,5 +105,55 @@ extension EnvoyHTTPFilter {
...
+    if let asyncResponseFilter = filter as? AsyncResponseFilter {
+      self.setResponseFilterCallbacks = { envoyCallbacks in
+        asyncResponseFilter.setResponseFilterCallbacks(
+          ResponseFilterCallbacksImpl(callbacks: envoyCallbacks)
+        )
+      }
+
+      self.onResumeResponse = { envoyHeaders, data, envoyTrailers, endStream in
+        let result = asyncResponseFilter.onResumeResponse(
+          headers: envoyHeaders.flatMap(ResponseHeaders.init),
```
rebello95, line 143:
```suggestion
          headers: envoyHeaders.map(ResponseHeaders.init),
```

library/swift/src/filters/Filter.swift
```diff
@@ -105,5 +105,55 @@ extension EnvoyHTTPFilter {
...
+          data: data,
+          trailers: envoyTrailers.flatMap(ResponseTrailers.init),
```
rebello95, line 145:
```suggestion
          trailers: envoyTrailers.map(ResponseTrailers.init),
```

library/swift/src/filters/Filter.swift
```diff
@@ -105,5 +105,55 @@ extension EnvoyHTTPFilter {
         }
       }
     }
+
+    if let asyncRequestFilter = filter as? AsyncRequestFilter {
+      self.setRequestFilterCallbacks = { envoyCallbacks in
+        asyncRequestFilter.setRequestFilterCallbacks(
```
rebello95, line 111: `asyncRequestFilter` is being retained here - I assume that's intentional, but want to flag

rebello95 (dismissed):

buildbreaker (commented):
Looks fine for the most part (assuming @rebello95's comments are addressed).

I assume we're going to be using the assertion filter to exercise the behaviors we are introducing here?

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: `internal_buffer` is a temporary pointer to the buffer needed for that invocation. The naming here is consistent with the other invocations, which do share more code. (The actual underlying memory is never shared between the request and response path.) I elected not to share code in this case as I felt it introduced more complexity rather than simplified things, due to the number of pending references in this singular case.

But honestly no strong feelings. Since the code isn't shared, we could be more particular about the naming.

library/common/extensions/filters/http/platform_bridge/c_type_definitions.h
goaway: I simply set them to be consistent with other invocations, but the comment is meant to suggest that one shouldn't rely on them being anything in particular (other than constant). They could also be zero-indexed (and I actually don't have a preference).

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: Sure. It will explode either way, but an `ASSERT` lets us give a nice reason for it. I'll update.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: The `false` actually is part of an unrelated Envoy API that has to do with whether buffer watermarking will occur - which has to do with backpressure, and isn't applicable for a library. (The documentation on the API is actually a little bit ambiguous to me, but `false` seemed correct based on my reading of it and other usage I've encountered.)

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: Sounds good, I'll update.

library/common/extensions/filters/http/platform_bridge/filter.h
goaway: The other functions in this class all use `decode` and `encode` (see L69+ below), due to the names being dictated by Envoy's own interfaces. I use request/response everywhere above this, because I do think it's clearer, especially for our use case, but my feeling was that it was worthwhile being consistent at least within a single class.

library/objective-c/EnvoyEngine.h
goaway: Sounds good.

library/objective-c/EnvoyFilterCallbacksImpl.m
goaway: Yeah, I'll add one.

library/objective-c/EnvoyFilterCallbacksImpl.m
goaway: The next PR has stuff that needs to go here, so I went ahead and dropped in the boilerplate when I created the class.

library/swift/src/filters/Filter.swift
goaway: Oh is `map` preferred/idiomatic for optional over `flatMap` now? (and did that change or did I misremember?)

library/swift/src/filters/RequestFilterCallbacksImpl.swift
goaway: Cool; for same reason every language has a different convention around this. :p

library/common/extensions/filters/http/platform_bridge/filter.cc
rebello95: Got it. Let's do an inline `/* watermark */` to annotate

library/swift/src/filters/RequestFilterCallbacksImpl.swift
rebello95: yea the bias I've kept in this repo (same at Lyft) is to explicitly use `self.` to make the source clearer (it's required within closures as well)

library/swift/src/filters/Filter.swift
rebello95: `flatMap` will unwrap the optional _and the optional returned as the result of the closure_. Since the closure returns a non-optional, there's no reason to use `flatMap` over `map`

goaway: Thanks, that makes sense.

library/swift/src/filters/Filter.swift
goaway: Yep, good callout, but should be fine since both the internal filter and downcasted reference are both tied to the lifecycle of the parent Filter.

library/common/extensions/filters/http/platform_bridge/c_types.h
```diff
@@ -103,11 +124,30 @@ typedef envoy_filter_data_status (*envoy_filter_on_data_f)(envoy_data data, bool
...
+                                                               envoy_headers* trailers,
+                                                               bool end_stream,
+                                                               const void* context);
+
 /**
  * Function signature to release a filter instance once the filter chain is finished with it.
  */
 typedef void (*envoy_filter_release_f)(const void* context);
 
+typedef void (*envoy_filter_resume_f)(const void* context);
```
rebello95, line 142: Let's document these typedefs

rebello95 (approved):

rebello95 (approved):

PR #1119, opened 2020-09-29
stats: sanitize incoming stat names

goaway:
Description: Ensure stat names passed into the public interface pass internal validity checks.
Risk Level: Low
Testing: CI


rebello95 (approved):
Can you please update the description with some context as to what this does in practice?

PR #1120, opened 2020-09-29
coverage: continue to upload on error

junr03:
Description: let the coverage report be uploaded even on failure. This allows contributors to see the report when the failure happens due to coverage below threshold errors.
Risk Level: low
Testing: testing by setting the coverage bar too high in an earlier commit.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -28,6 +28,8 @@ jobs:
           export CXX=clang++
           export COVERAGE_THRESHOLD=97
           ./envoy/test/run_envoy_bazel_coverage.sh //test/...
+      - name: 'package coverage'
+        run: |
           tar -czvf coverage.tar.gz generated/coverage
       - uses: actions/upload-artifact@v1
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage.yml
```diff
@@ -21,11 +21,12 @@ jobs:
       - name: 'Install dependencies'
         run: ./ci/linux_ci_setup.sh
       - name: 'Run coverage'
+        continue-on-error: true
         run: |
           export PATH=/usr/lib/llvm-10/bin:$PATH
           export CC=clang
           export CXX=clang++
-          export COVERAGE_THRESHOLD=95
+          export COVERAGE_THRESHOLD=97
```
goaway, line 29: Let's not fold this into this change. I'd also like to discuss before we continue to push this upwards.

junr03: > Testing: testing by setting the coverage bar too high in an earlier commit.

Using to test that the report does in fact upload when caused by failing to meet the threshold. I'll revert once the job runs.

goaway: Gotcha - missed the comment above. 👍

goaway (dismissed):

goaway (approved):

PR #1122, opened 2020-09-30
bugfix: fix stack-use-after-scope with stat name

goaway:


rebello95 (approved):
(For reference, this is caught by asan, which was not running on `main` when the breaking commit landed)

PR #1123, opened 2020-09-30
upstream: update with apple dns resolver

rebello95:
Updating to pull in https://github.com/envoyproxy/envoy/pull/13074 which will use Apple's DNS resolver for iOS in order to allow Envoy Mobile to ship on iOS 14 without triggering the "local area network" modal.


buildbreaker (dismissed):

rebello95:
Seems to require https://github.com/envoyproxy/envoy/pull/13338

rebello95:
new breakage upstream, fixing in https://github.com/envoyproxy/envoy/pull/13342

buildbreaker (approved):

PR #1127, opened 2020-10-01
filters: fix out of bound access to result status

goaway:
Description: The high-level representation of filter statuses with attached HTTP entities to be forwarded can't be directly represented across the bridge layer and must be mapped. Uniquely, resume iteration statuses may contain pending entities that have not yet been forwarded. This quick fix checks bounds explicitly for these cases in Objective-C.
Risk Level: Low

### DIFF
--- a/library/objective-c/EnvoyEngineImpl.m
+++ b/library/objective-c/EnvoyEngineImpl.m
@@ -67,9 +67,12 @@ static envoy_filter_data_status ios_http_filter_on_request_data(envoy_data data,
 
   NSData *platformData = to_ios_data(data);
   NSArray *result = filter.onRequestData(platformData, end_stream);
+  // Result is typically a pair of status and entity, but uniquely in the case of
+  // ResumeIteration it will (optionally) contain additional pending elements.
+  envoy_headers *pending_headers = toNativeHeadersPtr(result.count == 3 ? result[2] : nil);
   return (envoy_filter_data_status){/*status*/ [result[0] intValue],
                                     /*data*/ toNativeData(result[1]),
-                                    /*pending_headers*/ toNativeHeadersPtr(result[2])};
+                                    /*pending_headers*/ pending_headers};
 }
 
 static envoy_filter_data_status ios_http_filter_on_response_data(envoy_data data, bool end_stream,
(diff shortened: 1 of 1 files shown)


library/objective-c/EnvoyEngineImpl.m
```diff
@@ -83,9 +86,12 @@ static envoy_filter_data_status ios_http_filter_on_response_data(envoy_data data
 
   NSData *platformData = to_ios_data(data);
   NSArray *result = filter.onResponseData(platformData, end_stream);
+  // Result is typically a pair of status and entity, but uniquely in the case of
+  // ResumeIteration it will (optionally) contain additional pending elements.
+  envoy_headers *pending_headers = toNativeHeadersPtr(result.count == 3 ? result[2] : nil);
```
rebello95, line 91:
```suggestion
  envoy_headers *pending_headers = toNativeHeadersPtr(result.count == 3 ? result[2] : NULL);
```

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -67,9 +67,12 @@ static envoy_filter_data_status ios_http_filter_on_request_data(envoy_data data,
 
   NSData *platformData = to_ios_data(data);
   NSArray *result = filter.onRequestData(platformData, end_stream);
+  // Result is typically a pair of status and entity, but uniquely in the case of
+  // ResumeIteration it will (optionally) contain additional pending elements.
+  envoy_headers *pending_headers = toNativeHeadersPtr(result.count == 3 ? result[2] : nil);
```
rebello95, line 72:
```suggestion
  envoy_headers *pending_headers = toNativeHeadersPtr(result.count == 3 ? result[2] : NULL);
```

rebello95 (approved):

library/objective-c/EnvoyEngineImpl.m
goaway: This one should be `nil` - it's being passed as a missing objc object, not as a raw pointer.

library/objective-c/EnvoyEngineImpl.m
goaway: This one should be `nil` - it's being passed as a missing objc object, not as a raw pointer.

PR #1129, opened 2020-10-13
filters: add core and iOS async filter callback support

goaway:


library/objective-c/BUILD
```diff
@@ -32,6 +31,15 @@ objc_library(
     ],
 )
 
+objc_library(
```
rebello95, line 34: Should this be added as a dependency somewhere?

goaway: (was wip) :)

goaway:
Currently addressing the test failures, but the implementation is reviewable.

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -325,7 +369,20 @@ PlatformBridgeFilter::encodeTrailers(Http::ResponseTrailerMap& trailers) {
   return status;
 }
 
+void PlatformBridgeFilter::resumeDecoding() {
+  auto weak_self = weak_from_this();
+  dispatcher_.post([weak_self]() -> void {
```
goaway, line 374: I want to call out that there is a potential shutdown race here, but it all depends on timing. We don't retain the filter and so when Envoy is being torn down, it's possible the filter is torn down before the dispatcher, in which case everything is fine. If the inverse happens, however, then there's narrow window here where we could successfully acquire the filter and attempt to access the dispatcher, which is already gone.

As we discussed in person, there are a few solutions to this. We could allow the dispatcher itself to be shared and locked (this is seems ideal to me, since we would never have a garbage reference here or anywhere else we needed to call in from platform code on an arbitrary thread). We could lock the top-level engine object, but the downside is that all incoming calls we need access to it (and indeed, there's not an obvious reason a filter should have access to it, let alone a mechanism to make it available short of ugly hacks). And finally, we could take ownership of the dispatcher and have it be something passed into Envoy to ensure it was always valid. (There are other advantages to this last approach, and it might be desirable eventually regardless, but it would initially require non-trivial changes to Envoy.)

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -17,14 +17,38 @@ namespace Extensions {
 namespace HttpFilters {
 namespace PlatformBridge {
 
+static void envoy_filter_release_callbacks(const void* context) {
```
junr03, line 20: nit: I wonder if we should move the static helpers to their own utility file.

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -44,6 +68,24 @@ PlatformBridgeFilter::PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr c
   ASSERT(platform_filter_.instance_context,
          fmt::format("init_filter unsuccessful for {}", filter_name_));
   iteration_state_ = IterationState::Ongoing;
+
+  if (platform_filter_.set_request_callbacks) {
+    platform_request_callbacks_.resume_iteration = envoy_filter_callback_resume_decoding;
+    platform_request_callbacks_.release_callbacks = envoy_filter_release_callbacks;
+    platform_request_callbacks_.callback_context =
+        new PlatformBridgeFilterWeakPtr{weak_from_this()};
```
junr03, line 76: we should have an inline comment about the ownership model

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -172,6 +214,7 @@ Http::FilterDataStatus PlatformBridgeFilter::onData(Buffer::Instance& data, bool
       data.drain(data.length());
       data.addBufferFragment(*Buffer::BridgeFragment::createBridgeFragment(result.data));
     }
+    iteration_state_ = IterationState::Ongoing;
```
junr03, line 217: The absence of this assignment was a bug, right?

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -325,7 +369,20 @@ PlatformBridgeFilter::encodeTrailers(Http::ResponseTrailerMap& trailers) {
   return status;
 }
 
+void PlatformBridgeFilter::resumeDecoding() {
+  auto weak_self = weak_from_this();
```
junr03, line 373: nit: usually `weak_this` in the Envoy codebase

library/common/extensions/filters/http/platform_bridge/filter.cc
junr03: can we leave a todo in the code linking to an issue that we can track for any of the solutions you list?

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -325,7 +369,20 @@ PlatformBridgeFilter::encodeTrailers(Http::ResponseTrailerMap& trailers) {
...
+  dispatcher_.post([weak_self]() -> void {
+    if (auto self = weak_self.lock()) {
+      self->onResumeDecoding();
+    }
+  });
+}
+
 void PlatformBridgeFilter::onResumeDecoding() {
+  if (iteration_state_ == IterationState::Ongoing) {
+    return;
```
junr03, line 383: In other parts of the filter we have assertions about invalid combination of actions vs filter state. I want to hear your thoughts about why here we have a no-op.

test/common/extensions/filters/http/platform_bridge/platform_bridge_filter_test.cc
```diff
@@ -1,3 +1,4 @@
+#include "test/mocks/event/mocks.h"
```
junr03, line 1: Missing tests, right?

junr03 (commented):
First pass!

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: In my view, It's expected that this could happen, and it shouldn't be treated as an error. Say you have a filter that halts iteration for the next data chunk only for up to a 30s timeout and then resumes regardless. You'd need to resume asynchronously after the 30 seconds are up, but you might get the next data chunk right around then too. The ordering of the dispatch of those two events is unpredictable. You would know if the async call won during your on-data call, because you'd have received an on-resume call (on the same thread - the stream/filter thread). But if the inverse was true, and you wanted to proceed, you'd have no way to abort the already-dispatched asynchronous call (at least not without making the implementation a whole lot more complicated). This allows you to use async resume in conjunction with sync resume without a whole lot of extra bookkeeping in the filter implementation and said complexity in the implementation.

test/common/extensions/filters/http/platform_bridge/platform_bridge_filter_test.cc
goaway: But coverage isn't failing! ;)

In all seriousness, there's not a lot extra to cover here, but I am going to add coverage to cause the on-resume tests to be triggered by an "async callback", and basic coverage of callback state management. Also that bug you noted that's fixed in this PR.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: yes

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: I referenced the cache filter for this, and it used "self". (shrug emoji)

I have no preference.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: It was, though not one we would have encountered yet. I am adding coverage for it.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: Sure.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: Well, they're static, so they're not accessible ouside this file. And we don't need them to be, since they're just what we use to set the function pointers on the struct.

library/objective-c/EnvoyHTTPFilterCallbacksImpl.h
```diff
@@ -0,0 +1,12 @@
+#import <Foundation/Foundation.h>
```
rebello95, line 1: So multiple objc header files work now? If so, maybe we can (in a separate PR) look into splitting up the EnvoyEngine.h file?

rebello95 (approved):
Objective-C LGTM, will let @junr03 give a final pass on C++

library/objective-c/EnvoyHTTPFilterCallbacksImpl.h
goaway: Not quite. Private headers can be used either by listing them as `srcs` or by moving them to a separate library target. But anything exposed to Swift still needs to be in the single header (because that becomes the bridge header and we hit a clang error with transitive imports).

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: We could include them in a bare header, but I suppose my preference is to put them here, to limit their visibility strictly to this file.

library/common/extensions/filters/http/platform_bridge/filter.h
```diff
@@ -46,9 +46,18 @@ enum class IterationState { Ongoing, Stopped };
...
-  PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr config);
+  PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr config, Event::Dispatcher& dispatcher);
+
+  // Asynchronously trigger resumption of filter iteration, if applicable.
+  // This is a no-op if filter iteration is already ongoing.
+  void resumeDecoding();
+
+  // Asynchronously trigger resumption of filter iteration, if applicable.
+  // This is a no-op if filter iteration is already ongoing.
+  void resumeEncoding();
```
junr03, line 60: should the onResume callbacks below be made private now that we have the async triggers exposed?

junr03 (commented):
lgtm, just one visibility question

library/common/extensions/filters/http/platform_bridge/filter.h
goaway: It could be, though right now its visibility aligns with the other `on*` functions in the class (which perhaps could arguably all be made private).

goaway: Actually, I'm mistaken, they're already private. Let me move it.

junr03 (approved):

PR #1132, opened 2020-10-15
swiftlint: bump target version

rebello95:


goaway:
Do we have a documented min target SDK somewhere?

rebello95:
@goaway I don't believe we have it in any of our docs (we can add it), but this is currently what we build with in all the bazel configs.

goaway (approved):

PR #1143, opened 2020-10-21
java: add Android and Java async filter callback support

goaway:
Description: Adds support for asynchronous filter callbacks to be set and called on Android and Java filters.
Risk Level: Moderate
Testing: Local and CI

### DIFF
--- a/library/common/jni/jni_interface.cc
+++ b/library/common/jni/jni_interface.cc
@@ -363,7 +363,8 @@ static envoy_filter_trailers_status jvm_http_filter_on_response_trailers(envoy_h
                                         /*trailers*/ native_headers};
 }
 
-static void jvm_http_filter_set_request_callbacks(envoy_http_filter_callbacks callbacks, const void* context) {
+static void jvm_http_filter_set_request_callbacks(envoy_http_filter_callbacks callbacks,
+                                                  const void* context) {
 
   __android_log_write(ANDROID_LOG_VERBOSE, "[Envoy]", "jvm_http_filter_set_request_callbacks");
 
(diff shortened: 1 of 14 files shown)


goaway:
Test plan: In addition to the unit testing present on the PlatformBridgeFilter, I plan to update the example apps to leverage the asynchronous callback pathway to get some end-to-end coverage. After that, the plan is still to provide more in depth integration coverage via the AssertionFilter.

library/common/jni/jni_interface.cc
```diff
@@ -362,6 +363,122 @@ static envoy_filter_trailers_status jvm_http_filter_on_response_trailers(envoy_h
                                         /*trailers*/ native_headers};
 }
 
+static void jvm_http_filter_set_request_callbacks(envoy_http_filter_callbacks callbacks,
+                                                  const void* context) {
+
+  __android_log_write(ANDROID_LOG_VERBOSE, "[Envoy]", "jvm_http_filter_set_request_callbacks");
+
+  JNIEnv* env = get_env();
+  jobject j_context = static_cast<jobject>(const_cast<void*>(context));
```
buildbreaker, line 372: I might have my rules mixed up but do we need to delete this local ref here?

goaway: Yep, great catch. In a separate branch I started on last week, I'm adding caching to a bunch of the JNI refs, which in addition to being a noteworthy performance improvement, also reduces the number of locations we need to track refs like this and risk leaking.

buildbreaker (approved):
🎉

PR #1144, opened 2020-10-21
filters: tag set callbacks methods as mutating in swift protocols

goaway:
Description: This allows structs to implement to protocol and store callbacks in an instance variable (as well as classes, which ignore it).
Risk Level: Low - doesn't break class-based implementations.
Testing: Local and CI

### DIFF
--- a/library/swift/src/filters/AsyncRequestFilter.swift
+++ b/library/swift/src/filters/AsyncRequestFilter.swift
@@ -6,7 +6,7 @@ public protocol AsyncRequestFilter: RequestFilter {
   /// use.
   ///
   /// - parameter callbacks: The callbacks for this filter to use to interact with the chain.
-  func setRequestFilterCallbacks(_ callbacks: RequestFilterCallbacks)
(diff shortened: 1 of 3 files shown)


rebello95:
I'm curious as to the reasoning around allowing for this? In this case we'd have structs (value types) storing callback classes (reference types) which seems a little odd to me and could potentially lead to users introducing leaks

goaway:
We don't have to go down this route, certainly. Ultimately the ownership of the native memory is tied to the callback object, so as long as its ref count hits 0, it's not particularly relevant if the struct was value-copied here or there. But I see your point. The alternative is that in order to implement the async protocol you _must_ use a class and can't use a struct. But maybe that's desirable?

Honestly, I'm fine either way. I made the change simply because I had been using structs for testing.

goaway:
I'll close this and we can switch to classes. The more I think about it, the more using a struct seems likely to lead to bugs.

rebello95:
> The alternative is that in order to implement the async protocol you must use a class and can't use a struct. But maybe that's desirable?

Right, that was what I was thinking. This is also a bit more objective-c friendly and will make people think about how their filter's memory management is working

PR #1146, opened 2020-10-27
filters: fix weak pointer capture for filter callbacks

goaway:
Description: This change moves platform filter callback setup from the PlatformBridgeFilter constructor into its own callback setup path. It's not possible to use shared/weak_from_this in the constructor because the shared pointer state doesn't exist yet.
Risk Level: Low
Testing: Local


rebello95 (approved):

goaway:
Checking on the asan failure.

goaway:
Looks like it was a flake - moving forward.

PR #1148, opened 2020-10-27
filters: fix objc null checks to account for collections

goaway:
Description: Objective-C collections that don't allow nil use NSNull as a sentinel for missing values. Because these mappers may be used on values that come from collections, this fixes their null checks to be sensitive to NSNull as well as nil.
Risk Level: Low
Testing: Local

### DIFF
--- a/library/objective-c/EnvoyBridgeUtility.h
+++ b/library/objective-c/EnvoyBridgeUtility.h
@@ -3,7 +3,7 @@
 #import "library/common/types/c_types.h"
 
 static inline envoy_data toNativeData(NSData *data) {
-  if (data == nil || data == (id)[NSNull null]) {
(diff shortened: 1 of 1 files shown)


library/objective-c/EnvoyBridgeUtility.h
```diff
@@ -3,7 +3,7 @@
 #import "library/common/types/c_types.h"
 
 static inline envoy_data toNativeData(NSData *data) {
-  if (data == nil) {
+  if (data == nil || data == (id)[NSNull null]) {
```
rebello95, line 6: Might make more sense from a readability standpoint to do `(id)data == [NSNull null]` but not a huge deal

rebello95 (dismissed):

rebello95 (approved):

PR #1151, opened 2020-10-28
filters: fix some missing JNI wiring

goaway:
Description: A miss at the JNI layer was causing optional pending entities attached to resume iteration return statuses to be dropped. Upcoming additional coverage for this will include new example app filters and integration tests.
Risk Level: Low
Testing: Local

### DIFF
--- a/library/common/jni/jni_interface.cc
+++ b/library/common/jni/jni_interface.cc
@@ -265,8 +265,9 @@ static envoy_filter_data_status jvm_http_filter_on_request_data(envoy_data data,
   jobject j_data = static_cast<jobjectArray>(env->GetObjectArrayElement(result, 1));
 
   envoy_headers* pending_headers = nullptr;
+  // Avoid out-of-bounds access to array when checking for optional pending entities.
   if (size == 3) {
-    jobjectArray j_headers = static_cast<jobjectArray>(env->GetObjectArrayElement(result, 1));
+    jobjectArray j_headers = static_cast<jobjectArray>(env->GetObjectArrayElement(result, 2));
     pending_headers = to_native_headers_ptr(env, j_headers);
     env->DeleteLocalRef(j_headers);
   }
@@ -294,8 +295,9 @@ static envoy_filter_data_status jvm_http_filter_on_response_data(envoy_data data
   jobject j_data = static_cast<jobjectArray>(env->GetObjectArrayElement(result, 1));
 
   envoy_headers* pending_headers = nullptr;
+  // Avoid out-of-bounds access to array when checking for optional pending entities.
   if (size == 3) {
-    jobjectArray j_headers = static_cast<jobjectArray>(env->GetObjectArrayElement(result, 1));
+    jobjectArray j_headers = static_cast<jobjectArray>(env->GetObjectArrayElement(result, 2));
     pending_headers = to_native_headers_ptr(env, j_headers);
     env->DeleteLocalRef(j_headers);
   }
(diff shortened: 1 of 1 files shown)


library/common/jni/jni_interface.cc
```diff
@@ -259,10 +259,18 @@ static envoy_filter_data_status jvm_http_filter_on_request_data(envoy_data data,
   JNIEnv* env = get_env();
   jobjectArray result = static_cast<jobjectArray>(
       jvm_on_data("onRequestData", data, end_stream, const_cast<void*>(context)));
+  jsize size = env->GetArrayLength(result);
 
   jobject status = env->GetObjectArrayElement(result, 0);
   jobject j_data = static_cast<jobjectArray>(env->GetObjectArrayElement(result, 1));
 
+  envoy_headers* pending_headers = nullptr;
+  if (size == 3) {
```
junr03, line 268: sorry, I am a bit decontextualized here. Why 3? Can we have a constant?

library/common/jni/jni_interface.cc
```diff
@@ -271,18 +279,27 @@ static envoy_filter_data_status jvm_http_filter_on_request_data(envoy_data data,
...
                                                                  const void* context) {
   JNIEnv* env = get_env();
   jobjectArray result = static_cast<jobjectArray>(
       jvm_on_data("onResponseData", data, end_stream, const_cast<void*>(context)));
+  jsize size = env->GetArrayLength(result);
 
   jobject status = env->GetObjectArrayElement(result, 0);
   jobject j_data = static_cast<jobjectArray>(env->GetObjectArrayElement(result, 1));
 
+  envoy_headers* pending_headers = nullptr;
```
junr03, line 296: Can we create static funcs for the repeated code in these functions?

library/common/jni/jni_interface.cc
```diff
@@ -328,39 +346,69 @@ static envoy_filter_trailers_status jvm_http_filter_on_request_trailers(envoy_he
...
   jobject status = env->GetObjectArrayElement(result, 0);
   jobjectArray j_trailers = static_cast<jobjectArray>(env->GetObjectArrayElement(result, 1));
 
   int unboxed_status = unbox_integer(env, status);
-  envoy_headers native_headers = to_native_headers(env, j_trailers);
+  envoy_headers native_trailers = to_native_headers(env, j_trailers);
+
+  envoy_headers* pending_headers = nullptr;
+  envoy_data* pending_data = nullptr;
+  if (size == 4) {
```
junr03, line 359: Same two comments as above

library/common/jni/jni_interface.cc
goaway: I think the common factoring that's already here complicates following the code in this file somewhat, and would hesitate to push it further.

library/common/jni/jni_interface.cc
goaway: Yeah, this isn't a great check. I'd like to move to fully-bridged types for filter statuses. But this ensures we don't attempt out of bounds access on the array.

junr03: Sure the comment helps. Can we state why 3 here and 4 below? Maybe make them constants?

library/common/jni/jni_interface.cc
junr03: Not sure I completely agree. But feel free to TIOLI!

library/common/jni/jni_interface.cc
goaway: Sure, I'll do that.

library/common/jni/jni_interface.cc
goaway: Thanks, I do appreciate the perspective. I'll think further about how this could be simplified, but for now I'm going to stick with my inclination to leave it. Note there's a great deal of dispatch to common-factored paths already occurring. e.g. the jvm_http_filter_on_response_headers delegates to jvm_on_response_headers, which delegates pass_headers, all of which make use of further extracted common utility functions. While this common factoring can serve to encapsulate logic, it also means you need to jump between more locations to follow what's actually happening.

library/common/jni/jni_interface.cc
goaway: Updated with a different and clearer check.

junr03 (commented):
Thanks for the update. I'll let @buildbreaker take a final pass.

buildbreaker (approved):
Looks good to me @goaway

PR #1154, opened 2020-10-28
filters: fix ordering of optional entities in swift statuses

goaway:
Description: The public and private representations of platform filters' compound statuses differ in ordering. In the public case, the ordering aligns with our chosen API, whereas in the private case, it's dictated by constraints in the type representation. This change fixes a bug which arose out of that misalignment. Note that fully-bridged types would avoid the potential for this sort of mapping bug.
Risk Level: Low
Testing: Pending integration


rebello95 (approved):
We should definitely prioritize having this covered in upcoming integration tests.

Is there any way we can align the internal types and the public bridge interfaces?

PR #1157, opened 2020-10-31
filters: add buffer demo filter to swift

goaway:

### DIFF
--- a/examples/swift/hello_world/BufferDemoFilter.swift
+++ b/examples/swift/hello_world/BufferDemoFilter.swift
@@ -1,0 +1,41 @@
+import Envoy
+import Foundation
+
+/// Example of a more complex HTTP filter that pauses processing on the response filter chain,
+/// buffers until the response is complete, then resumes filter iteration while setting a new
+/// header.
+final class BufferDemoFilter: ResponseFilter {
+  private var headers: ResponseHeaders!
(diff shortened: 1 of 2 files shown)


examples/swift/hello_world/BufferDemoFilter.swift
```diff
@@ -0,0 +1,41 @@
...
+      return .resumeIteration(headers: builder.build(), data: self.body)
+    }
+    return .stopIterationAndBuffer
+  }
+
+  func onResponseTrailers(
+    _ trailers: ResponseTrailers
+  ) -> FilterTrailersStatus<ResponseHeaders, ResponseTrailers> {
+    // Trailers imply end of stream; resume processing of the (now fully-buffered) response.
+    return .resumeIteration(data: self.body, trailers: trailers)
```
rebello95, line 35: I think for requests that are headers and trailers (no body), this would crash because `self.body` is force unwrapped. Should we make `body` an optional in the declaration as `var body: Data?`?

goaway: Very good point. Updated.

rebello95: I think you forgot to push 🙃

rebello95 (approved):

PR #1158, opened 2020-10-31
filters: fix edge case where on-data receives Envoy's internal buffer

goaway:
Description: In certain cases - for example, when a previous filter in the chain stopped and buffered the request, and a subsequent filter stops iteration as well - the on-data filter invocation receives Envoy's internal buffer directly. This makes sense, since that's where Envoy has buffered all data, but it means that an attempt to preemptively aggregate new data into the buffer should be skipped: it's already there.
Risk Level: Low
Testing: Local and CI (#1157)


rebello95:
Shouldn't https://github.com/lyft/envoy-mobile/pull/1157 be failing without this change?

goaway:
Yep, it did: https://github.com/lyft/envoy-mobile/pull/1157/checks?check_run_id=1335120176

rebello95:
Ah cool, hadn't yet failed when I was reviewing

rebello95 (approved):
Great that this is covered by the new test!

PR #1161, opened 2020-11-01
java: show which key is unresolved in config errors

goaway:
Description: Added to assist while debugging filters.
Risk Level: Low
Testing: Local and CI

### DIFF
--- a/library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
+++ b/library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
@@ -19,7 +19,7 @@ public class EnvoyConfiguration {
   public final String appId;
   public final String virtualClusters;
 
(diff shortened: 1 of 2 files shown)


library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -17,6 +19,8 @@
   public final String appId;
   public final String virtualClusters;
 
+  private static final Pattern KEY_PATTERN = Pattern.compile("\\{\\{ (.+) \\}\\}");
```
buildbreaker, line 22:
```suggestion
  private static final Pattern UNRESOLVED_KEY_PATTERN = Pattern.compile("\\{\\{ (.+) \\}\\}");
```

buildbreaker (dismissed):
One suggestion. I think this is fine for now since it's to provide a quick improvement to the debugging message.

buildbreaker (approved):

PR #1165, opened 2020-11-02
platform: add option for specifying custom filter names

rebello95:
Being able to specify a custom name can be helpful for reading `trace` logs which include named filters.


goaway (approved):

buildbreaker (approved):
Cool, didn't know this interaction with a lambda and default parameter works!

PR #1166, opened 2020-11-05
stats: move regex pattern to companion object

jingwei99:
Description: Re-use regex pattern across all `Element` instances, since `Pattern.compile()` comes with performance cost.
Risk Level: Low
Testing: Unit test


library/kotlin/test/io/envoyproxy/envoymobile/ElementTest.kt
```diff
@@ -0,0 +1,18 @@
...
+
+  @Test
+  fun `creates new element as expected`() {
+    // Should just work
+    Element("foo")
+  }
+
+  @Test(expected = IllegalArgumentException::class)
+  fun `throw exception when element name is rejected`() {
+    // Should just work
```
buildbreaker, line 15:
```suggestion
    // Should throw
```?

junr03 (commented):
lgtm, @buildbreaker on the kotlin pattern.

Do we do a similar check in swift?

buildbreaker:
@jingwei99 Could we actually get some more consistency between kotlin and swift: https://github.com/lyft/envoy-mobile/blob/46503043b9d1d539bdbe45b647f17f18902f46e2/library/swift/src/stats/Element.swift#L3?

jingwei99:
> @jingwei99 Could we actually get some more consistency between kotlin and swift:
> 
> https://github.com/lyft/envoy-mobile/blob/46503043b9d1d539bdbe45b647f17f18902f46e2/library/swift/src/stats/Element.swift#L3
> 
> ?

it is the same check on the pattern, right? or you mean something different?

buildbreaker:
I meant let's try to get the structure and variable names the same too

buildbreaker:
Sorry to clarify what I meant was to mirror the `value` field name and internal visibility from swift

buildbreaker (approved):
Thanks!

PR #1172, opened 2020-11-10
core: create LocalErrorFilter to map responses to errors

goaway:
Description: Creates a filter to handle mapping local responses to errors, instead of baking the logic into the dispatcher. This additionally allows platform filters to receive onError callbacks.
Risk Level: Moderate
Testing: Dispatcher integration, Swift integration, unit and manual

### DIFF
--- a/library/common/extensions/filters/http/local_error/filter.cc
+++ b/library/common/extensions/filters/http/local_error/filter.cc
@@ -1,0 +1,88 @@
+#include "library/common/extensions/filters/http/local_error/filter.h"
+
+#include "envoy/http/codes.h"
+#include "envoy/server/filter_config.h"
+
+#include "common/http/codes.h"
+#include "common/http/header_map_impl.h"
+#include "common/http/utility.h"
+
+#include "library/common/http/headers.h"
+#include "library/common/types/c_types.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace LocalError {
+
+LocalErrorFilterConfig::LocalErrorFilterConfig(
+    const envoymobile::extensions::filters::http::local_error::LocalError& proto_config)
+    : enabled_(proto_config.enabled()) {}
+
+LocalErrorFilter::LocalErrorFilter(LocalErrorFilterConfigSharedPtr config) : config_(config) {}
+
+Http::FilterHeadersStatus LocalErrorFilter::encodeHeaders(Http::ResponseHeaderMap& headers,
+                                                          bool end_stream) {
+  uint64_t response_status = Http::Utility::getResponseStatus(headers);
+  bool success = Http::CodeUtility::is2xx(response_status);
+
+  // TODO: ***HACK*** currently Envoy sends local replies in cases where an error ought to be
+  // surfaced via the error path. There are ways we can clean up Envoy's local reply path to
+  // make this possible, but nothing expedient. For the immediate term this is our only real
+  // option. See https://github.com/lyft/envoy-mobile/issues/460
+
+  // Absence of EnvoyUpstreamServiceTime header implies this is a local reply, which we treat as
+  // a stream error.
+  if (!success && headers.get(Http::Headers::get().EnvoyUpstreamServiceTime).empty()) {
+    ENVOY_LOG(debug, "intercepted local response");
+    processingError_ = true;
+    headers_ = &headers;
+    mapLocalResponseToError(headers);
+    return end_stream ? Http::FilterHeadersStatus::Continue
+                      : Http::FilterHeadersStatus::StopIteration;
+  }
+
+  return Http::FilterHeadersStatus::Continue;
+}
+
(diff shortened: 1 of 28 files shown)


library/common/extensions/filters/http/local_error/filter.cc
```diff
@@ -0,0 +1,88 @@
...
+    break;
+  case 503:
+    headers.addCopy(Http::InternalHeaders::get().ErrorCode,
+                    std::to_string(ENVOY_CONNECTION_FAILURE));
+    break;
+  default:
+    headers.addCopy(Http::InternalHeaders::get().ErrorCode, std::to_string(ENVOY_UNDEFINED_ERROR));
+  }
+
+  headers.setStatus(218);
```
junr03, line 82: I know this won't surface to the application because we end up mapping to an on error response. But I wonder if we should use a non 2xx code here?

library/common/http/dispatcher.cc
```diff
@@ -35,32 +36,41 @@ void Dispatcher::DirectStreamCallbacks::encodeHeaders(const ResponseHeaderMap& h
...
-    mapLocalResponseToError(headers);
     if (end_stream) {
       onError();
     }
     return;
   }
 
-  // Normal response path.
+  // Track success for later bookkeeping (stream could still be reset).
+  success_ = CodeUtility::is2xx(response_status);
```
junr03, line 73: related to my comment above re: 218. Should the local error path be counted as success?

test/common/http/dispatcher_test.cc
```diff
@@ -773,17 +773,29 @@ TEST_F(DispatcherTest, MultipleStreams) {
   ASSERT_EQ(cc.on_complete_calls, 1);
 }
 
-TEST_F(DispatcherTest, EnvoyLocalReply) {
+TEST_F(DispatcherTest, EnvoyLocalReplyNotAnError) {
```
junr03, line 776: This is the suite where you are going to add more test? Or are you creating a test suite for the filter? Probably both, right?

library/common/extensions/filters/http/local_error/filter.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+
+package envoymobile.extensions.filters.http.local_error;
+
+message LocalError {
+  bool enabled = 1;
```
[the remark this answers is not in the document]
goaway: It is, but I wanted to have at least one configuration value, so I wasn't sure what would happen with an empty config. There are other examples of filters with an enabled boolean.

library/common/extensions/filters/http/local_error/filter.cc
goaway: I went with it because its unique, and currently used by Apache for a similar purpose. I wanted to reduce the odds of it triggering any status code based behavior in other filters. (Long-term I'd prefer not to use a mechanism like this at all, but rather implement a real `onError` pathway in Envoy's filter chain.)

library/common/http/dispatcher.cc
goaway: It won't be. The function returns a couple lines above.

test/common/http/dispatcher_test.cc
goaway: Will add both.

library/common/extensions/filters/http/local_error/filter.proto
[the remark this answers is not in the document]
junr03: Oh if that's the case, then you can just delete. It is fine to have an empty message.

library/common/extensions/filters/http/local_error/filter.cc
junr03: Sure. Do we have an open issue to get together thoughts on an error pathway in the HCM? I think it is worth opening at this point.

test/common/http/dispatcher_test.cc
junr03: Per offline discussion it sounds like you will add in subsequent PRs. Do you mind opening issue for that?

goaway: Sure - https://github.com/envoyproxy/envoy-mobile/issues/1207

library/common/extensions/filters/http/local_error/filter.cc
goaway: https://github.com/envoyproxy/envoy-mobile/issues/1208

library/common/extensions/filters/http/local_error/filter.proto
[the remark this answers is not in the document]
junr03: Pending this.

junr03 (approved):

junr03 (dismissed):
Need to fix format, otherwise lgtm.

Also, do you mind merging instead of rebasing a force-pushing? Force pushing breaks the github review diff and makes it harder to see what changed since last review.

goaway:
Will merge when this goes green (it should, since #1200 did with the same fix).

PR #1179, opened 2020-11-12
native resource registry: fix construction

goaway:
Description: Cleans up and and fixes an earlier rebase issue with the NativeResourceRegistry and adds test coverage.
Risk Level: Low
Testing: Local, CI, and in app.


library/java/test/io/envoyproxy/envoymobile/engine/EnvoyNativeResourceRegistryTest.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.junit.Test
+
+class EnvoyNativeResourceRegistryTest {
+
+  @Test
+  fun `release callbacks are invoked when EnvoyNativeResourceWrappers are flagged as unreachable`() {
+    val latch = CountDownLatch(1)
+    val testHandle: Long = 77
+    // Force Wrapper out of scope so the GC picks it up. ¯\_(ツ)_/¯
+    for (i in 0..1 step 2) {
```
buildbreaker, line 16: Would this still work if we create a private method which does this instead of inlining it into each test (this is so we can more easily add a new test or change the wrapper)?

goaway: It... might? I wouldn't be able to simply move the wrapper in and out of scope though like I'm doing in these two tests though. This was the simplest approach I could come up with to get the GC to run consistently. Various lambda-based approaches I tried did *not* work. (Perhaps the GC is somewhat more aggressive with loops due to their potential to create lots of garbage in short order.)

goaway: TBQH I'm somewhat disinclined to put even more time in this, given the time it took to get something that was both simple and seemed to work consistently.

buildbreaker (approved):

PR #1188, opened 2020-11-21
bridge: add register string accessor functionality plus expose on Android

junr03:
Description: this PR adds the ability to register string accessors from the platform to the engine. This allows the core layer to access strings from the platform at runtime. Two highlights:

- Eventually this mechanism should be made generic to be able to access and set arbitrary types from the platform. This is very type-specific as a POC. https://github.com/envoyproxy/envoy-mobile/issues/1192
- This PR only wires up Android's platform layer all the way. A subsequent PR will do the same for iOS.

Risk Level: low - new API
Testing: using the example app.


junr03:
@goaway I wanted to get this out early to get your opinion.

library/java/src/io/envoyproxy/envoymobile/engine/JvmStringAccessorContext.java
```diff
@@ -0,0 +1,18 @@
+package io.envoyproxy.envoymobile.engine;
+
+import java.nio.ByteBuffer;
+
+import io.envoyproxy.envoymobile.engine.types.EnvoyStringAccessor;
+
+class JvmStringAccessorContext {
```
junr03, line 7: Given that this accessor runs on the threading context that calls it I am not sure if we need it versus just passing the EnvoyStringAccessor to the jni and retaining that.

library/java/src/io/envoyproxy/envoymobile/engine/types/EnvoyStringAccessor.java
```diff
@@ -0,0 +1,11 @@
+package io.envoyproxy.envoymobile.engine.types;
+
+import java.nio.ByteBuffer;
+
+public interface EnvoyStringAccessor {
+
+  /**
+   * Called to retrieve a string from the Application
+   */
+  ByteBuffer getString();
```
junr03, line 10: decided to return a ByteBuffer here rather than a String to take advantage of the utilities we already have that bridge it into an envoy_data. 

Although it wouldn't be too complicated to create a bridging utility from String to envoy_data if we'd rather use that type.

library/common/jni/jni_interface.cc
```diff
@@ -610,6 +611,19 @@ static const void* jvm_http_filter_init(const void* context) {
   return retained_filter;
 }
 
+// EnvoyStringAccessor
+
+static envoy_data jvm_get_string(void* context) {
+  JNIEnv* env = get_env();
+  jobject j_context = static_cast<jobject>(context);
+  jclass jcls_JvmStringAccessorContext = env->GetObjectClass(j_context);
```
goaway, line 619: This is a leak - the local ref needs to be deleted at the end of this function.

library/common/jni/jni_interface.cc
```diff
@@ -610,6 +611,19 @@ static const void* jvm_http_filter_init(const void* context) {
...
+  jmethodID jmid_getString =
+      env->GetMethodID(jcls_JvmStringAccessorContext, "getString", "()Ljava/lang/Object;");
```
goaway, line 621: (This is not a leak - jmethodID is an int.)

library/common/jni/jni_interface.cc
```diff
@@ -610,6 +611,19 @@ static const void* jvm_http_filter_init(const void* context) {
...
+  // Passed as a java.nio.ByteBuffer.
+  jobject data = env->CallObjectMethod(j_context, jmid_getString);
```
goaway, line 623: Also leaking.

library/java/src/io/envoyproxy/envoymobile/engine/JvmStringAccessorContext.java
goaway: The concrete class allows methods to be cached.

library/java/src/io/envoyproxy/envoymobile/engine/types/EnvoyStringAccessor.java
goaway: Seems fine. Don't we already have a utility for String though?

library/common/jni/jni_interface.cc
junr03: @goaway you mean the local ref, right?

library/java/src/io/envoyproxy/envoymobile/engine/types/EnvoyStringAccessor.java
junr03: Not afaict? We have a byte buffer and a byte array utility. I guess the byte array one can be used with a jstring? But you have to copy.

junr03:
@goaway updated, lmk if you have other comments :)

library/common/api/c_types.h
```diff
@@ -0,0 +1,20 @@
...
+extern "C" { // function pointers
+#endif
+
+typedef envoy_data (*envoy_get_string_f)(void* context);
+
+#ifdef __cplusplus
+} // function pointers
+#endif
+
+typedef struct {
```
goaway, line 17: Please add a docstring here and clarify that this is ready only.

goaway (approved):
Looks good. One request.

goaway (approved):

PR #1193, opened 2020-12-01
config: add ability to register native filters

junr03:
Description: adding templating to register native filters.
Risk Level: low - optional feature
Testing: used in example apps
Docs Changes: pending


junr03:
iOS code will be pushed ASAP but wanted to get the ball rolling with Android code. This can be reviewed as is for now.

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -60,15 +63,23 @@ public EnvoyConfiguration(String statsDomain, int connectTimeoutSeconds, int dns
    * @throws ConfigurationException, when the template provided is not fully
    *                                 resolved.
    */
-  String resolveTemplate(final String templateYAML, final String filterTemplateYAML) {
+  String resolveTemplate(final String templateYAML, final String platformFilterTemplateYAML, final String nativeplatformFilterTemplateYAML) {
```
rebello95, line 66: `nativeplatformFilterTemplateYAML` shouldn't this be `nativeFilterTemplateYAML`? Also should be documented above because I don't think this is immediately obvious as to what should be passed here (it's a bit of a "magic" string)

rebello95: Can you actually elaborate on how this will be used directly? I.e., what will a consumer need to specify to make this work?

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -38,7 +40,7 @@
   public EnvoyConfiguration(String statsDomain, int connectTimeoutSeconds, int dnsRefreshSeconds,
                             int dnsFailureRefreshSecondsBase, int dnsFailureRefreshSecondsMax,
                             List<EnvoyHTTPFilterFactory> httpFilterFactories, int statsFlushSeconds,
-                            String appVersion, String appId, String virtualClusters) {
+                            String appVersion, String appId, String virtualClusters, Map<String, String> nativeFilters) {
```
goaway, line 43: Doesn't this need an order? Two lists or a list of pairs might be a better serialization.

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
junr03: bad find and replace. will fix.

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
junr03: whoops, used to ordered map in c++. I agree, will fix.

junr03:
@goaway @rebello95 updated + added iOS piping. Unit tests tomorrow morning.

library/objective-c/EnvoyConfiguration.m
```diff
@@ -9,11 +9,12 @@ - (instancetype)initWithStatsDomain:(NSString *)statsDomain
...
        dnsFailureRefreshSecondsBase:(UInt32)dnsFailureRefreshSecondsBase
         dnsFailureRefreshSecondsMax:(UInt32)dnsFailureRefreshSecondsMax
-                        filterChain:(NSArray<EnvoyHTTPFilterFactory *> *)httpFilterFactories
+                platformFilterChain:(NSArray<EnvoyHTTPFilterFactory *> *)httpFilterFactories
                   statsFlushSeconds:(UInt32)statsFlushSeconds
                          appVersion:(NSString *)appVersion
                               appId:(NSString *)appId
-                    virtualClusters:(NSString *)virtualClusters {
+                    virtualClusters:(NSString *)virtualClusters
+                  nativeFilterChain:(NSArray<EnvoyNativeFilterConfig *> *)nativeFilterChain {
```
rebello95, line 17: nit: I think it'd be nice to list native/platform filters together in the parameter order

library/objective-c/EnvoyEngine.h
```diff
@@ -218,6 +218,15 @@ extern const int kEnvoyFilterResumeStatusResumeIteration;
 
 @end
 
+#pragma mark - EnvoyNativeFilterConfig
+
+@interface EnvoyNativeFilterConfig : NSObject
+
+@property (nonatomic, strong) NSString *name;
+@property (nonatomic, strong) NSString *typedConfig;
```
rebello95, line 226: docs for these, especially since they're just strings?

library/swift/src/EngineBuilder.swift
```diff
@@ -1,6 +1,14 @@
 @_implementationOnly import EnvoyEngine
 import Foundation
 
+extension EnvoyNativeFilterConfig {
```
rebello95, line 4: Why not put this in objective-c since that's where the type is defined and these are primitives? If you want to keep it here, it should be `private extension`

rebello95 (commented):
LGTM but I think it'd be great if we could add more documentation. Will review after tests are added

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -51,31 +52,35 @@ public EnvoyConfiguration(String statsDomain, int connectTimeoutSeconds, int dns
...
+    this.nativeFilterChain = nativeFilterChain;
   }
 
   /**
    * Resolves the provided configuration template using properties on this
    * configuration.
    *
    * @param templateYAML the template configuration to resolve.
+   * @param platformFilterTemplateYAML helper template to build platform http filters.
+   * @param nativeFilterTemplateYAML helper template to build native http filters.
```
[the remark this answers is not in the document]
junr03: these are not intended to be publicly replaced. I am guessing that @goaway added the platform string (and I copied him here) because the jni needs to be called to get the string. Vs. in objc you'll notice these are not params because objc can use the c string.

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyNativeFilterConfig.java
```diff
@@ -0,0 +1,18 @@
...
+/* Datatype used by the EnvoyConfiguration to create a native http filter chain. */
+public class EnvoyNativeFilterConfig {
+  public final String name;
+  public final String typedConfig;
+
+  /**
+   * Create a new instance of the configuration
+   *
+   * @param name        the name of the filter.
+   * @param typedConfig the filter configuration.
```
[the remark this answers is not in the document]
junr03: using the envoy field name for consistency

library/java/src/io/envoyproxy/envoymobile/engine/JniLibrary.java
```diff
@@ -210,3 +210,4 @@ protected static native int runEngine(long engine, String config, String logLeve
    *         native filter configuration.
    */
   public static native String nativeFilterTemplateString();
+}
```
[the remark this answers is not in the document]
junr03: accidental

library/objective-c/EnvoyEngine.h
junr03: doesn't seem like we traditionally add docstrings here? I added them in the EngineBuilder method.

rebello95: ok if it's in the builder I guess that's sufficient - the reason i called it out here is because it's quite unclear to the reader what a `typedConfig` should actually contain

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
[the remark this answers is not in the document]
goaway: Yep, that's right.

rebello95 (approved):

PR #1194, opened 2020-12-02
ios: allow engine to be run with custom yaml template

goaway:
Description: Custom yaml will be treated as a configuration template, rather than a fixed set of configuration. It's still possible to pass fixed configuration, simply by not including any tokens for interpolation. This usage was already present and appeared supported by the API (e.g. see HttpBridgeTests), but simply wouldn't work, without any indication that it wouldn't.
Risk Level: Low
Testing: Local & CI

### DIFF
--- a/library/objective-c/EnvoyEngine.h
+++ b/library/objective-c/EnvoyEngine.h
@@ -302,15 +302,17 @@ extern const int kEnvoyFailure;
 /**
  Run the Envoy engine with the provided yaml string and log level.
 
- @param configYAML The configuration yaml with which to start Envoy.
+ @param template The configuration template with which to start Envoy.
+ @param config The EnvoyConfiguration used to start Envoy.
  @param logLevel The log level to use when starting Envoy.
  @param onEngineRunning Closure called when the engine finishes its async startup and begins
  running.
  @return A status indicating if the action was successful.
  */
-- (int)runWithConfigYAML:(NSString *)configYAML
-                logLevel:(NSString *)logLevel
-         onEngineRunning:(nullable void (^)())onEngineRunning;
+- (int)runWithTemplate:(NSString *)yaml
+                config:(EnvoyConfiguration *)config
+              logLevel:(NSString *)logLevel
+       onEngineRunning:(nullable void (^)())onEngineRunning;
 
 /**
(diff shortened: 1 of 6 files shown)


goaway:
Android change to follow.

library/objective-c/EnvoyEngine.h
```diff
@@ -302,15 +302,17 @@ extern const int kEnvoyFailure;
 /**
  Run the Envoy engine with the provided yaml string and log level.
 
- @param configYAML The configuration yaml with which to start Envoy.
+ @param template The configuration template with which to start Envoy.
```
rebello95, line 305:
```suggestion
 @param yaml The configuration template with which to start Envoy.
```

library/swift/src/EngineImpl.swift
```diff
@@ -52,10 +52,10 @@ final class EngineImpl: NSObject {
   /// - parameter engine:          The underlying engine to use for starting Envoy.
   /// - parameter onEngineRunning: Closure called when the engine finishes its async
   ///                              initialization/startup.
-  convenience init(configYAML: String, logLevel: LogLevel = .info, engine: EnvoyEngine,
-                   onEngineRunning: (() -> Void)?)
+  convenience init(yaml: String, config: EnvoyConfiguration, logLevel: LogLevel = .info,
```
rebello95, line 55: the docs here are outdated now

library/swift/src/EngineImpl.swift
```diff
@@ -9,8 +9,8 @@ final class EngineImpl: NSObject {
   private let streamClientImpl: StreamClientImpl
 
   private enum ConfigurationType {
-    case yaml(String)
-    case typed(EnvoyConfiguration)
+    case custom(String, EnvoyConfiguration)
+    case standard(EnvoyConfiguration)
```
rebello95, line 13:
```suggestion
    case custom(yaml: String, config: EnvoyConfiguration)
    case standard(config: EnvoyConfiguration)
```

It might be nice to name these now since `String` is ambiguous, but not a big deal since this is private

rebello95 (dismissed):

rebello95 (approved):

PR #1197, opened 2020-12-04
python: create python bindings stub

crockeo:
Description: I'm going to be working with @goaway on setting up Python bindings to envoy-mobile to be used for a requests-like library that uses envoy-mobile, rather than urllib3, for its underlying networking. This PR creates the stubs for the Python platform implementation, which will be translated into a Python module using [pybind11](https://pybind11.readthedocs.io/en/stable/).

This PR contains only the stubs for the core engine implementation (i.e. no gRPC, no stats, no filter). Stubs and implementations for those features will come later.

Risk Level: Low

Testing: N/A for now. We'll put together tests in Python and/or C++ for the types exposed from this library when they have associated functionality.

Docs Changes: N/A

Release Notes: N/A (yet?)


crockeo:
I accidentally made this PR before I meant to (accidental Cmd+Enter 🙂) so bear with me as I read the documentation I meant to beforehand.

library/python/envoy_mobile/BUILD
```diff
@@ -0,0 +1,49 @@
+load("@pybind11_bazel//:build_defs.bzl", "pybind_extension")
```
goaway, line 1: Having `envoy_mobile` in the path seems a little bit funky. Could this be `pybind` or `cc_platform`?

Also maybe the `cc_library` belongs in `//library/cc` and just the `pybind_extension` should go here.

library/python/envoy_mobile/executor.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <functional>
+
+
+class Executor {
```
goaway, line 6: I don't know that we should necessarily have this, since it's basically where we'd want to plug in native python dispatch. I mean, it _is_ a flexible and pretty agnostic interface, but does it really align with python based dispatch mechanisms?

crockeo: it's necessary to have a type defined in C++ for the C++ side of the lib to interact with it, e.g. to call into it from `on_XXX`. in practice this is going to work like:

[executor_base.h](https://github.com/crockeo/py-envoy-mobile/blob/main/py_envoy_mobile/wrapper/executor_base.h) being implemented by something akin to [asyncio_executor.py](https://github.com/crockeo/py-envoy-mobile/blob/main/py_envoy_mobile/asyncio_executor.py) or [gevent_executor.py](https://github.com/crockeo/py-envoy-mobile/blob/main/py_envoy_mobile/gevent_executor.py) inside of Python

library/python/envoy_mobile/BUILD
crockeo: original naming scheme was just trying to match up folder name w/ how it would be imported in python. right now the `.so` built from `//library/python/envoy_mobile:envoy_mobile.so` can just be imported from python as `import envoy_mobile`, which would be canonical naming scheme 

there's nothing keeping the `.so` name and the dir name coupled, though, so i'm happy to change it to whatever

goaway:
I'm mostly on board with this approach, though it raises a couple questions for me that I think are worth mulling over. For instance, in the Swift and Kotlin cases (with their Objective-C and Java underpinnings), we have our own types, because they can't (easily) directly leverage Envoy's types anyways. But here, for instance, does it really make sense to have _another_ C++ headers type, when Envoy's native type is available? (Or at least its interface, which could be implemented?)

I don't know that I have the answer at this point, but wanted to raise it for consideration.

It's also possible that we may be able to thin this layer down further by refining the interfaces exposed by Envoy Mobile's existing C++ API.

crockeo:
Yeah, that's a really good point. I think this is what I tried, and failed, to bring up when we were chatting about supporting a library of C++ types for the Python implementation. I'm not familiar with Envoy internals, but I'd have to imagine some serious part of this PR is just reconstructing Envoy internals but from the C API provided by envoy-mobile.

There's value to be gained for the uniformity between implementations by keeping this as "just another platform" with no privilege to peek into envoy internals. If we base this structured C++ implementation on only the `c_types` and `main_interface` then any work to preserve that interface across internal changes and/or refactors will mean the Python implementation continues to Just Work™ like the others. On the flip side, if we did base this off of the internal object model then there would be less plumbing to do when breaking changes did occur.

I'm not sure where I fall either

library/python/envoy_mobile/executor.h
goaway: Okay, cool - if it's directly supported with gevent/asyncio (and the "right" way to integrate) then that's totally fine.

goaway:
I think it's the alignment between all three platforms that this approach achieves that really wins me over. It will just make it a lot easier to maintain.

As I mentioned offline, long term we can move all three platforms onto a more fully-bridged solution, reducing the footprint of the bridge layer in favor of bridged types.

crockeo:
squashed down all the earlier commits into a single commit with a sign-off so that the DCO check succeeds. also includes a formatting fix for the `//library/python/envoy_mobile` `BUILD` file

library/cc/headers_builder.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include "headers.h"
+
+class HeadersBuilder {
+public:
+  HeadersBuilder& add(const std::string& name, std::string& value);
```
crockeo, line 7: make this `const std::string&`

library/cc/log_level.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include <string>
+
+/**
+ * See: https://pybind11.readthedocs.io/en/stable/classes.html#enumerations-and-internal-types
+ * this set up will let Python see:
+ *
+ * LogLevel.Trace = the enum value
+ * LogLevel.to_string(LogLeve.Trace) = the string value
```
crockeo, line 10: `LogLeve` -> `LogLevel`

library/cc/log_level.h
```diff
@@ -0,0 +1,32 @@
...
+    Trace,
+    Debug,
+    Info,
+    Warn,
+    Error,
+    Critical,
+    Off,
+  };
+
+  _LogLevel from_string(std::string log_level);
```
crockeo, line 27: `static`

library/cc/request_headers.h
```diff
@@ -0,0 +1,26 @@
...
+#include "headers.h"
+#include "request_headers_builder.h"
+#include "request_method.h"
+#include "retry_policy.h"
+#include "upstream_http_protocol.h"
+
+class RequestHeaders : public Headers {
+public:
+  RequestMethod request_method() const;
+  std::string scheme() const;
```
crockeo, line 14: this can probably just be `const std::string&` and avoid copying strings here, same for the rest of them

library/cc/request_headers_builder.h
```diff
@@ -0,0 +1,30 @@
...
+#include "upstream_http_protocol.h"
+
+class RequestHeaders;
+
+class RequestHeadersBuilder : public HeadersBuilder {
+public:
+  RequestHeadersBuilder(RequestMethod::_RequestMethod request_method, const std::string& scheme,
+                        const std::string& authority, const std::string& path);
+
+  using HeadersBuilder::add;
```
crockeo, line 18: C++ question:

1. do I even need this, or will inheritance automatically allow subclasses to have the superclass's implementation?

2. i have a feeling that this will return a `HeadersBuilder&` instead of a `RequestHeadersBuilder&`, is that right?

library/cc/retry_policy.h
```diff
@@ -0,0 +1,36 @@
...
+};
+
+struct RetryPolicy {
+  int max_retry_count;
+  int retry_on;
+  std::vector<int> retry_status_codes;
+  int per_try_timeout_ms;
+  int total_upstream_timeout_ms;
+
+  RawHeaders output_headers() const;
```
crockeo, line 34: it could be more performant to put this directly into an already-existing `RawHeaders` instance, or otherwise just add headers to a `HeadersBuilder` reference, but the API wouldn't be nearly as nice

library/cc/stream_callbacks.h
```diff
@@ -0,0 +1,34 @@
...
+  std::optional<OnErrorCallback> on_error;
+};
+
+class EnvoyHttpCallbacksAdapter {
+public:
+  EnvoyHttpCallbacksAdapter(Executor executor, StreamCallbacks callbacks);
+
+private:
+  // TODO(crockeo): figure out lifecycle management here
+  Executor& executor_;
```
crockeo, line 32: i've been futzing around in rust-land where i get to feel comfortable about reference lifetimes. this isn't really _safe_ per se, but its lifetime will be articulated in a [pybind11::keep_alive](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#keep-alive) on the python obj definition

library/cc/stream_prototype.h
```diff
@@ -0,0 +1,34 @@
...
+  Stream start(Executor executor);
+
+  StreamPrototype& set_on_response_headers(
+      std::function<void(const ResponseHeaders& headers, bool end_stream)> closure);
+  StreamPrototype& set_on_response_data(
+      std::function<void(const std::vector<std::byte>& bytes, bool end_stream)> closure);
+  StreamPrototype&
+  set_on_response_trailers(std::function<void(const ResponseTrailers& trailers)> closure);
+  StreamPrototype& set_on_error(std::function<void(const EnvoyError& error)> closure);
+  StreamPrototype& set_on_cancel(std::function<void()> closure);
```
crockeo, line 29: use callback types here to make sure they stay the same

library/python/envoy_mobile/BUILD
crockeo, line 1: rm this build file

crockeo (commented):
leaving these comments as a bit of a self-review / reminder to go back and do these.

@goaway obviously the relationship between these classes + the Python library is a little obfuscated, but I'd like to hammer out the work required to get from where we are now -> having these stubs merged

library/cc/stream_callbacks.h
goaway: You could use std::weak_ptr for a stronger guarantee. Or even just a shared_ptr to ensure its lifetime.

library/cc/log_level.h
```diff
@@ -0,0 +1,11 @@
...
+
+enum LogLevel {
+  Trace,
+  Debug,
+  Info,
+  Warn,
+  Error,
+  Critical,
+  Off,
+};
```
goaway, line 11: Is there an Envoy enum we can leverage for this?

library/cc/request_method.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+enum RequestMethod {
```
goaway, line 3: Here, too, I'd look to just use an Envoy-defined enum, I think.

library/cc/stream.h
```diff
@@ -0,0 +1,22 @@
...
+#include "library/common/types/c_types.h"
+#include "request_headers.h"
+#include "request_trailers.h"
+
+class Stream {
+public:
+  Stream(envoy_stream_t handle);
+
+  Stream& send_headers(const RequestHeaders& headers, bool end_stream);
+  Stream& send_data(const std::vector<std::byte>& data);
```
goaway, line 15: Is there a pybind reason to use this buffer representation? If not, I'd propose we use uint8_t (which Envoy currently uses internally) and probably just simple arrays.

goaway:
Overall I think this looks good. Have you thought through how object lifecycles will be managed? Skimming the docs, it looks like smart pointers work with pybind. This is the approach Envoy uses generally to manage memory (also preferring unique_ptrs to shared_ptrs wherever possible, though shared_ptr might be required for types that end up exposed in Python).

library/cc/stream.h
crockeo: there is not, i'll use uint8_t.

library/cc/stream_callbacks.h
crockeo: the best case scenario for this is that someone in python land creates and then immediately moves ownership of the Executor into the C++ codebase. i'm not sure that we can enforce move semantics in python, though, so that may be a moot attempt.

i'd say let's use `std::unique_ptr` for now to represent the ideal world and then we can backtrack into whatever ends up working (probably a `std::weak_ptr` or `std::shared_ptr`

crockeo: >In other words, there is nothing special that needs to be done. While returning unique pointers in this way is allowed, it is illegal to use them as function arguments. For instance, the following function signature cannot be processed by pybind11.

[from pybind11 docs](https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html#std-unique-ptr)

looks like we're going to have to use `shared_ptr`s

crockeo:
good q about object lifecycles. i don't believe you can mix arg types for a particular type when exposed to Python. e.g. having one function take `const RequestHeaders&` and another take `std::shared_ptr<RequestHeaders>` isn't possible. 

i think the most straightforward approach is to make most everything a `std::shared_ptr` off the bat and optimize from there. luckily the choice between shared_ptr vs. raw references doesn't actually influence the python interface at all, so we can make breaking changes internally that are exposed the same to python

library/cc/request_headers_builder.h
goaway: Subclasses will indeed inherit their superclasses' interface/implementations.

goaway:
Yeah I think standardizing on shared_ptr makes sense. Note that Envoy usually defines a type for the smart pointer reference in headers, e.g.
`using MyTypeSharedPtr = std::shared_ptr<MyType>;`

library/cc/retry_policy.h
crockeo: resolving this for now bc it's a perf issue that we can figure out later

crockeo:
...signing all of my commits again because I forgot

fool me once, shame on you, fool me twice, shame on me :grin:

goaway:
@crockeo I'll do another full pass on this during the work day today.

crockeo:
btw i noticed some header changes after starting in on [the implementation](https://github.com/crockeo/envoy-mobile/tree/ch/python-bindings-impl); i imagine that's going to continue to happen as i hammer down exactly what each component should do here.

if this is Good Enough™ we can merge and revise this with feedback as we go along if its easier to review chunked PRs.

goaway (approved):
Thanks @crockeo! This looks good to me as a point to iterate from. Let's go ahead and merge and proceed from here.

library/cc/stream.h
```diff
@@ -0,0 +1,25 @@
...
+#include "library/common/types/c_types.h"
+#include "request_headers.h"
+#include "request_trailers.h"
+
+class Stream {
+public:
+  Stream(envoy_stream_t handle);
+
+  Stream& send_headers(RequestHeadersSharedPtr headers, bool end_stream);
+  Stream& send_data(const std::vector<uint8_t>& data);
```
goaway, line 16: After thinking about this a bit, I think we should switch to using `envoy_data` directly in this implementation to represent buffers. Python's buffer protocol doesn't work with vectors, nor do any buffer representations in Envoy. We've already have a bridge implementation of Envoy's internal buffer type defined in terms of `envoy_data`. It seems like using vectors will just result in extra indirection, and potentially, copies.

library/cc/engine.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+// NOLINT(namespace-envoy)
```
goaway, line 3: Let's namespace everything in these files. `Envoy::Mobile`, `Envoy::Platform`, `Envoy::Lib/rary` all seem like reasonable candidates to me.

PR #1199, opened 2020-12-07
filters: wire JVM/Android on-cancel and on-error pathways

goaway:
Description: Adds JVM/Android wiring for on-cancel and on-error filter invocations.
Risk Level: Moderate
Testing: Pending

### DIFF
--- a/library/common/jni/jni_interface.cc
+++ b/library/common/jni/jni_interface.cc
@@ -593,6 +593,14 @@ static void* jvm_on_cancel(void* context) {
   return result;
 }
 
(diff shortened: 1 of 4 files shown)


buildbreaker (dismissed):
Do you happen to have test cases in the top of mind for this change?

goaway:
@buildbreaker I'd like to cover it the same way as I've covered the Swift side, with AssertionFilter-level integration tests. That will require the JVM-only build though. The Swift tests at least serve to assert that the core logic and wiring is correct, the piece that will remain uncovered for the time being here is just the Android wiring.

PR #1200, opened 2020-12-07
filters: dispatch on-error invocations from PlatformBridgeFilter

goaway:
Description: Leverages the LocalErrorFilter to provide on-error invocations for platform filters.
Risk Level: Moderate
Testing: Integration test & manual

### DIFF
--- a/library/common/extensions/filters/http/platform_bridge/filter.cc
+++ b/library/common/extensions/filters/http/platform_bridge/filter.cc
@@ -351,7 +351,8 @@ Http::FilterHeadersStatus PlatformBridgeFilter::encodeHeaders(Http::ResponseHead
     }
 
     if (platform_filter_.on_error) {
-      platform_filter_.on_error({error_code, error_message, attempt_count}, platform_filter_.instance_context);
+      platform_filter_.on_error({error_code, error_message, attempt_count},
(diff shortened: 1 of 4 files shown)


library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -321,6 +322,36 @@ Http::FilterHeadersStatus PlatformBridgeFilter::encodeHeaders(Http::ResponseHead
                                                               bool end_stream) {
   ENVOY_LOG(trace, "PlatformBridgeFilter({})::encodeHeaders", filter_name_);
 
+  // Presence of internal error header indicates an error that should be surfaced as an
+  // error callback (rather than an HTTP response).
+  const auto error_code_header = headers.get(Http::InternalHeaders::get().ErrorCode);
```
buildbreaker, line 327: Are we able to have a test case for verifying this behavior?

goaway: Added passing test.

buildbreaker (approved):
Just making sure I'm not missing something: would a Kotlin test also apply here?

Going to approve for now since this looks fine. Thank you @goaway !

PR #1201, opened 2020-12-07
documentation: add missing docstrings in JvmFilterContext

goaway:

### DIFF
--- a/library/java/src/io/envoyproxy/envoymobile/engine/JvmFilterContext.java
+++ b/library/java/src/io/envoyproxy/envoymobile/engine/JvmFilterContext.java
@@ -117,7 +117,13 @@ class JvmFilterContext {
   }
 
   /**
+   * Invokes onResumeRequest callback with pending HTTP entities.
    *
+   * @param headerCount,  total pending headers included in the header block.
+   * @param data,         buffered body data.
+   * @param trailerCount, total pending trailers included in the trailer block.
+   * @param endStream,    whether the stream is closed at this point.
+   * @return Object[],    tuple of status with updated entities to be forwarded.
    */
   public Object onResumeRequest(long headerCount, byte[] data, long trailerCount,
                                 boolean endStream) {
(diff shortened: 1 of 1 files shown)


library/java/src/io/envoyproxy/envoymobile/engine/JvmFilterContext.java
```diff
@@ -117,7 +117,13 @@ public Object onResponseTrailers(long trailerCount) {
   }
 
   /**
+   * Invokes onResumeRequest callback with pending HTTP entities.
    *
+   * @param headerCount,  total pending headers included in the header block.
+   * @param data,         buffered body data.
+   * @param trailerCount, total pending trailers included in the trailer block.
+   * @param endStream,    whether the stream is closed at this point.
+   * @return Object[],    tuple of status with updated entities to be forwarded.
```
rebello95, line 126: Isn't this only an `Object`?

library/java/src/io/envoyproxy/envoymobile/engine/JvmFilterContext.java
```diff
@@ -139,7 +145,13 @@ public Object onResumeRequest(long headerCount, byte[] data, long trailerCount,
   }
 
   /**
+   * Invokes onResumeResponse callback with pending HTTP entities.
    *
+   * @param headerCount,  total pending headers included in the header block.
+   * @param data,         buffered body data.
+   * @param trailerCount, total pending trailers included in the trailer block.
+   * @param endStream,    whether the stream is closed at this point.
+   * @return Object[],    tuple of status with updated entities to be forwarded.
```
rebello95, line 154: Same here

library/java/src/io/envoyproxy/envoymobile/engine/JvmFilterContext.java
goaway: technically it is an `Object[]` but it's declared as `Object` to make the JNI lookup consistent.

rebello95 (approved):

Issue #1192, opened 2020-12-01
bridge: allow accessing arbitrary types from the platform


junr03:
The platform bridge filter introduced the concept of runtime platform APIs. Furthermore, in #1188 a very specific string accessor was introduced. 

The ability to pass types across the bridge layer is useful and powerful. The string accessor mechanism should be extended generically to be able to pass arbitrary types through the bridge layer via runtime-registered platform APIs.

PR #1202, opened 2020-12-08
ios: add string accessor API

junr03:
Description: iOS parallel to #1188 
Risk Level: low
Testing: example app. Pending test with assertion filter.
Docs Changes: pending full API docs


examples/swift/hello_world/DemoStringAccessor.swift
```diff
@@ -0,0 +1,9 @@
+import Envoy
+import Foundation
+
+/// Example of a simple HTTP filter that adds a response header.
+struct DemoStringAccessor: EnvoyStringAccessor {
```
junr03, line 5: @rebello95 do I need a swift protocol to be able to use`EnvoyStringAccessor` objc protocol on user defined accessors?

rebello95: Sorry can you elaborate? I'm not sure what you're asking

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -261,6 +268,17 @@ - (int)registerFilterFactory:(EnvoyHTTPFilterFactory *)filterFactory {
   return kEnvoySuccess;
 }
 
+- (int)registerStringAccessor:(NSString *)name
+               stringAccessor:(EnvoyStringAccessor *)stringAccessor {
```
rebello95, line 272: Have we considered registering a lambda/closure that returns a string instead? That'll probably be easier to interact with on the platform layer, since `getString()` is a little overly generic IMO

examples/swift/hello_world/DemoStringAccessor.swift
junr03: The high level question is: I want DemoStringAccessor to implement the EnvoyStringAccessor interface, but I am not sure how to import EnvoyStringAccessor to this file.

rebello95: Ah, you won't be able to do that because we only expose Swift types externally from `Envoy.framework`. You'd need to declare a `public protocol` in Swift instead. However, per my comments below I still think we should be registering lambdas/closures instead of a type with a `getString()` method

junr03: @rebello95 finally having the space to come back to this. The idea here is that the `EnvoyStringAccessor` protocol/interface will eventually be made into a generic getter/setter of arbitraty types across the bridge layer (https://github.com/envoyproxy/envoy-mobile/issues/1192). So I wanted to have a concrete type rather than free-floating lambdas. 

However, I can see us exposing the engine builder API as registering a lambda, and then the objc code builds an object around it. The only friction is that I would need to redo the kotlin code to maintain cross-platform homogeneity.

rebello95: Yea personally I'd opt towards the lambda option since that's more consistent with the other APIs Envoy Mobile exposes today (request/response/stream callbacks, for example), and it solves the ambiguity concern around something called `getString()`. @buildbreaker do you have thoughts?

buildbreaker: @rebello95 Yeah, I remember chatting with @junr03 on this. He wanted to use an object/class here because it allows us to extend/add more accessors for data. Would you be open to a rename of said object instead?

One alternative that I just thought about is if we have a lambda like `()->AccessorObject` and have the `AccessorObject` just be the extensible metadata object.

rebello95: Talked with @junr03 about this a bit more offline (thank you both for responding). I'm ok with this general approach of allowing the consumer to register instances of different types that conform to this relatively generic protocol, assuming the intent is that those types are registered within specific contexts.

Maybe the naming of this would help improve a bit?

```swift
protocol EnvoyStringAccessor: NSObject {
    func envoyStringValue() -> String
}
```

Also it sounds like Java is currently exposing this interface up to the platform; we should switch that over to Kotlin

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -297,6 +314,10 @@ - (int)runWithConfig:(EnvoyConfiguration *)config
     [self registerFilterFactory:filterFactory];
   }
 
+  [config.stringAccessors enumerateKeysAndObjectsUsingBlock:^(id name, id accessor, BOOL *stop) {
```
rebello95, line 317: You could use a for loop here like on line 313 above 🤷🏽‍♂️

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -313,6 +334,10 @@ - (int)runWithTemplate:(NSString *)yaml
...
```
rebello95, line 337: Same here

library/swift/test/EngineBuilderTests.swift
```diff
@@ -199,6 +199,20 @@ final class EngineBuilderTests: XCTestCase {
     self.waitForExpectations(timeout: 0.01)
   }
 
+  func testAddingStringAccessorToConfigurationWhenRunningEnvoy() throws {
+    let expectation = self.expectation(description: "Run called with expected data")
+    MockEnvoyEngine.onRunWithConfig = { config, _ in
+      XCTAssertEqual(1, config.stringAccessors.count)
```
rebello95, line 205: A better test might be doing `XCTAssertEqual("hello", config.stringAccessors. getEnvoyString())`

library/swift/test/EngineBuilderTests.swift
```diff
@@ -199,6 +199,20 @@ final class EngineBuilderTests: XCTestCase {
...
+      expectation.fulfill()
+    }
+
+    _ = try EngineBuilder()
+      .addEngineType(MockEnvoyEngine.self)
+      .addStringAccessor(name: "name", accessor: { return "hello"})
```
rebello95, line 211:
```suggestion
      .addStringAccessor(name: "name", accessor: { "hello" })
```
This is equivalent if you want

rebello95 (dismissed):

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -284,6 +291,16 @@ - (int)registerFilterFactory:(EnvoyHTTPFilterFactory *)filterFactory {
   return kEnvoySuccess;
 }
 
+- (int)registerStringAccessor:(NSString *)name accessor:(EnvoyStringAccessor *)accessor {
+  // TODO(goaway): Everything here leaks, but it's all be tied to the life of the engine.
+  // This will need to be updated for https://github.com/lyft/envoy-mobile/issues/332
```
rebello95, line 296: Is this true in the context of a string accessor? Seems like it should be deallocated after usage since this is creating new strings each time it's accessed?

rebello95 (dismissed):
LGTM, but one question. Also we should update Android after this to match

library/objective-c/EnvoyEngineImpl.m
junr03: yeah, still true. The accessor is the object being leaked, and it should be cleaned up.

The strings themselves get managed via the existing bridging utilities; the strings get boxed in envoy_data and that gets managed as an envoy buffer, which in turn cleans everything (all the way to the platform layer) when it is destroyed.

rebello95: got it, sounds good

PR #1209, opened 2020-12-15
python: first-pass platform implementation

crockeo:
Description: First-pass python platform implementation. This is _a lot_ of code to review, so I'm happy to split this out into a bunch of different PRs. I'll reply in-line with known work items.

Risk Level: Low

Testing: N/A, will follow up on this PR or another subsequent PR with tests _in Python_ for each of these components

Docs Changes: N/A

Release Notes: N/A


library/cc/bridge_utility.cc
```diff
@@ -0,0 +1,111 @@
+#include "bridge_utility.h"
+
+// NOLINT(namespace-envoy)
```
crockeo, line 3: move from nolint to `Envoy::Platform` namespace

library/cc/bridge_utility.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+// NOLINT(namespace-envoy)
+
+#include <string>
+#include <vector>
+
+#include "absl/container/flat_hash_map.h"
```
crockeo, line 8: don't need this after including `headers.h`

library/cc/engine_builder.cc
```diff
@@ -0,0 +1,92 @@
...
+       std::to_string(this->dns_failure_refresh_seconds_base_)},
+      {"{{ dns_failure_refresh_rate_seconds_max }}",
+       std::to_string(this->dns_failure_refresh_seconds_max_)},
+      {"{{ dns_refresh_rate_seconds }}", std::to_string(this->dns_refresh_seconds_)},
+      {"{{ native_filter_chain }}", ""},
+      {"{{ platform_filter_chain }}", ""},
+      {"{{ stats_domain }}", this->stats_domain_},
+      {"{{ stats_flush_interval_seconds }}", std::to_string(this->stats_flush_seconds_)},
+      {"{{ virtual_clusters }}", this->virtual_clusters_},
+  };
```
crockeo, line 76: [fmt](https://fmt.dev/) doesn't seem well-suited for this, since it requires single-braced formatting (and will just throw errors for this template string). thoughts on:

- alternatives in C++
- otherwise moving it to Python?

library/cc/log_level.cc
```diff
@@ -0,0 +1,31 @@
...
+
+// NOLINT(namespace-envoy)
+
+#include <stdexcept>
+
+static const std::pair<LogLevel, std::string> LOG_LEVEL_LOOKUP[]{
+    {LogLevel::trace, "trace"}, {LogLevel::debug, "debug"}, {LogLevel::info, "info"},
+    {LogLevel::warn, "warn"},   {LogLevel::error, "error"}, {LogLevel::critical, "critical"},
+    {LogLevel::off, "off"},
+};
```
crockeo, line 11: is there a canonical solution to mapping between `enum` and `std::string` in C++? this kind of lookup table (which i use in several places) is what i came up with

library/cc/bridge_utility.cc
```diff
@@ -0,0 +1,111 @@
...
+
+// TODO(crockeo): we always copy memory across boundaries; consider allowing for moves and/or
+// shared ownership w/ reference counting via envoy-mobile's release callbacks
+
+#include <sstream>
+
+envoy_data buffer_as_envoy_data(const std::vector<uint8_t>& data) {
```
crockeo, line 10: this whole file is going to be a little messy since i hacked it together quickly

i imagine there's already code that translates between the C types and managed C++ types. i'd prefer to reuse that rather than reconstructing complexity here

library/cc/request_headers.cc
```diff
@@ -0,0 +1,38 @@
...
+absl::optional<RetryPolicy> RequestHeaders::retry_policy() const {
+  try {
+    return absl::optional<RetryPolicy>(RetryPolicy::from_raw_headers(this->all_headers()));
+  } catch (std::exception) {
+    return absl::optional<RetryPolicy>();
+  }
+}
+
+absl::optional<UpstreamHttpProtocol> RequestHeaders::upstream_http_protocol() const {
+  const auto header_name = ":x-envoy-mobile-upstream-protocol";
```
crockeo, line 24: nix `:`

library/cc/response_headers.cc
```diff
@@ -0,0 +1,21 @@
+#include "response_headers.h"
+
+// NOLINT(namespace-envoy)
+
+int ResponseHeaders::http_status() const {
+  if (!this->contains(":status")) {
+    throw std::out_of_range("ResponseHeaders does not contain :status");
```
crockeo, line 7: `out_of_range` doesn't feel right when the function doesn't take an arg. maybe change to something that explicates this is an internal issue

library/cc/response_headers.h
```diff
@@ -5,16 +5,15 @@
 #include "headers.h"
 #include "response_headers_builder.h"
 
+class ResponseHeadersBuilder;
+
 class ResponseHeaders : public Headers {
 public:
+  ResponseHeaders(RawHeaders headers) : Headers(std::move(headers)) {}
```
crockeo, line 12: try to remember why i changed this from private + friend to public. choose one for all of the headers types

library/cc/stream_callbacks.cc
```diff
@@ -0,0 +1,85 @@
...
+
+EnvoyHttpCallbacksAdapter::EnvoyHttpCallbacksAdapter(ExecutorSharedPtr executor,
+                                                     StreamCallbacksSharedPtr callbacks)
+    : executor_(executor), stream_callbacks_(callbacks) {}
+
+envoy_http_callbacks EnvoyHttpCallbacksAdapter::as_envoy_http_callbacks() {
+  envoy_http_callbacks callbacks{
+      .on_headers = &EnvoyHttpCallbacksAdapter::dispatch_on_headers,
+      .on_data = &EnvoyHttpCallbacksAdapter::dispatch_on_data,
+      // on_metadata is not used
```
crockeo, line 15: is this intentional? i notice that `on_metadata` isn't used in the kotlin callbacks

library/python/e2e_test.py
```diff
@@ -0,0 +1,136 @@
+import threading
```
crockeo, line 1: this depends on `gevent` and `python3`. need to figure out how to do reproducible python builds in Bazel

library/cc/stream_prototype.h
```diff
@@ -6,31 +6,30 @@
...
 #include "response_trailers.h"
 #include "stream.h"
 #include "stream_callbacks.h"
 
-class Engine;
-
 class StreamPrototype {
 public:
-  StreamPrototype(Engine engine);
+  StreamPrototype(envoy_engine_t engine);
```
crockeo, line 19: here and in `StreamClient` i ended up passing down an `envoy_engine_t` because it fixes a whole bunch of `#include` cycles. thoughts?

library/cc/stream_callbacks.cc
crockeo: originally proposed in spdy, disallowed in HTTP/2, proposal for HTTP/2. going to keep it piped up to here but nothing more

library/cc/log_level.cc
crockeo: looking around on the internet, seems like there isn't a great approach to mapping an enum to an arbitrary string without depending on compiler-specific internals. going to just resolve this.

library/cc/bridge_utility.cc
crockeo: there's also a bunch of overhead from copying data across the platform/bridge layers. i'm going to come back and fix it later, but i thought i'd get it correct first, and make it right later.

PR #1212, opened 2020-12-18
ios: split comma-delimited header values when creating platform representation

goaway:
Description: Splits list-formatted header value on creation of header maps.  Per RFC 7230, does not perform this parsing on "set-cookie" headers, and additionally skips it on "www-authenticate" and "proxy-authenticate", due to their potential to contain commas in contravention of the spec.
Risk Level: Low
Testing: Pending

### DIFF
--- a/library/objective-c/EnvoyBridgeUtility.h
+++ b/library/objective-c/EnvoyBridgeUtility.h
@@ -91,8 +91,7 @@ static inline EnvoyHeaders *to_ios_headers(envoy_headers headers) {
     }
 
     // By convention, these headers disregard the RFC and contain commas single values.
-    if (headerKey == @"set-cookie" ||
-        headerKey == @"www-authenticate" ||
+    if (headerKey == @"set-cookie" || headerKey == @"www-authenticate" ||
         headerKey == @"proxy-authenticate") {
       [headerValueList addObject:headerValue];
     } else {
@@ -100,7 +99,7 @@ static inline EnvoyHeaders *to_ios_headers(envoy_headers headers) {
       NSArray *newValueList = [headerValue componentsSeparatedByString:@","];
       for (NSString *value in newValueList) {
         NSString *trimmedValue =
-          [value stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
+            [value stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
         [headerValueList addObject:trimmedValue];
       }
     }
--- a/library/objective-c/EnvoyBridgeUtility.h
+++ b/library/objective-c/EnvoyBridgeUtility.h
@@ -90,9 +90,11 @@ static inline EnvoyHeaders *to_ios_headers(envoy_headers headers) {
       headerDict[headerKey] = headerValueList;
     }
 
-    // By convention, these headers disregard the RFC and contain commas single values.
-    if (headerKey == @"set-cookie" || headerKey == @"www-authenticate" ||
-        headerKey == @"proxy-authenticate") {
+    // These headers may contain commas in single values, in contravention of the RFC.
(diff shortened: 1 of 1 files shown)


goaway:
Update to this coming early tomorrow.

library/objective-c/EnvoyBridgeUtility.h
```diff
@@ -83,12 +83,27 @@ static inline EnvoyHeaders *to_ios_headers(envoy_headers headers) {
...
+    if (headerKey == @"set-cookie" ||
+        headerKey == @"www-authenticate" ||
+        headerKey == @"proxy-authenticate") {
+      [headerValueList addObject:headerValue];
+    } else {
+      // Add trimmed, comma-separated values as individual members of the list.
+      NSArray *newValueList = [headerValue componentsSeparatedByString:@","];
+      for (NSString *value in newValueList) {
+        NSString *trimmedValue =
+          [value stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
```
rebello95, line 103: Is this necessary? I assume that envoy doesn't add white spaces, in which case we may more closely preserve the "original" content by not trimming

rebello95 (commented):
Looks good overall, but I think we should add some tests for this?

library/objective-c/EnvoyBridgeUtility.h
goaway: It's desirable because comma delimiting may take the form of `","` or `", "`, and both are common. Users will likely find themselves inadvertently failing to match simple string comparisons if leading or trailing spaces are left on that are an artifact of how the header was coalesced and encoded.

Moreover, the RFCs state that all header values are encoded with optional leading and trailing whitespace, implying one should not treat it as part of the value.

rebello95: Sounds good. This implementation seems to have warnings when building:

```
INFO: From Compiling library/objective-c/EnvoyEngineImpl.m:
In file included from library/objective-c/EnvoyEngineImpl.m:2:
./library/objective-c/EnvoyBridgeUtility.h:94:19: warning: direct comparison of a string literal has undefined behavior [-Wobjc-string-compare]
    if (headerKey == @"set-cookie" || headerKey == @"www-authenticate" ||
                  ^  ~~~~~~~~~~~~~
./library/objective-c/EnvoyBridgeUtility.h:94:19: note: use 'isEqual:' instead
    if (headerKey == @"set-cookie" || headerKey == @"www-authenticate" ||
                  ^~
        [          isEqual:       ]
./library/objective-c/EnvoyBridgeUtility.h:94:49: warning: direct comparison of a string literal has undefined behavior [-Wobjc-string-compare]
    if (headerKey == @"set-cookie" || headerKey == @"www-authenticate" ||
                                                ^  ~~~~~~~~~~~~~~~~~~~
./library/objective-c/EnvoyBridgeUtility.h:94:49: note: use 'isEqual:' instead
    if (headerKey == @"set-cookie" || headerKey == @"www-authenticate" ||
                                                ^~
                                      [          isEqual:             ]
./library/objective-c/EnvoyBridgeUtility.h:95:19: warning: direct comparison of a string literal has undefined behavior [-Wobjc-string-compare]
        headerKey == @"proxy-authenticate") {
                  ^  ~~~~~~~~~~~~~~~~~~~~~
./library/objective-c/EnvoyBridgeUtility.h:95:19: note: use 'isEqual:' instead
        headerKey == @"proxy-authenticate") {
                  ^~
        [          isEqual:               ]
```

rebello95 (approved):
+1, let's merge and add a test in a follow-up PR per discussion around objective-C testing bazel issues (can you open an issue?)

library/objective-c/EnvoyBridgeUtility.h
goaway: Switched to `caseInsensitiveCompare`

goaway:
https://github.com/envoyproxy/envoy-mobile/issues/1215

PR #1213, opened 2020-12-18
android: split comma-delimited header values when creating platform representation

goaway:
Description: Splits list-formatted header value on creation of header maps.  Per RFC 7230, does not perform this parsing on "set-cookie" headers, and additionally skips it on "www-authenticate" and "proxy-authenticate", due to their potential to contain commas in contravention of the spec.
Risk Level: Low
Testing: Pending

### DIFF
--- a/library/java/src/io/envoyproxy/envoymobile/engine/JvmBridgeUtility.java
+++ b/library/java/src/io/envoyproxy/envoymobile/engine/JvmBridgeUtility.java
@@ -53,8 +53,7 @@ class JvmBridgeUtility {
     }
 
     // By convention, these headers disregard the RFC and contain commas single values.
-    if (headerKey.equals("set-cookie") ||
-        headerKey.equals("www-authenticate") ||
+    if (headerKey.equals("set-cookie") || headerKey.equals("www-authenticate") ||
         headerKey.equals("proxy-authenticate")) {
       values.add(headerValue);
     } else {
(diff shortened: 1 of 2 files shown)


library/java/src/io/envoyproxy/envoymobile/engine/JvmBridgeUtility.java
```diff
@@ -45,12 +45,26 @@ void passHeader(byte[] key, byte[] value, boolean start) {
...
+    // By convention, these headers disregard the RFC and contain commas single values.
+    if (headerKey.equals("set-cookie") ||
+        headerKey.equals("www-authenticate") ||
+        headerKey.equals("proxy-authenticate")) {
+      values.add(headerValue);
+    } else {
+      // Add trimmed, comma-separated values as individual members of the list.
+      String[] newValues = headerValue.split(",");
+      for (int i = 0; i < newValues.length; i++) {
+        values.add(newValues[i].trim());
```
rebello95, line 64: Similar comment as on the iOS PR as to whether we want to trim here

rebello95 (commented):
Thanks! We should add some tests before merging

library/java/src/io/envoyproxy/envoymobile/engine/JvmBridgeUtility.java
goaway: Replied on the other PR. :)

rebello95 (approved):

PR #1214, opened 2020-12-18
Expose histogram envoy stats methods to interface & engines

donyu:
**Description:** Expose the histogram stat in `envoy/include/envoy/stats/` to envoy-mobile engine. This will be leveraged in subsequent PRs to support a new set of stats collection in the iOS and Android layers.
**Risk Level:** Medium (due to changes in the core engine impl, but should be Low given no one depends on these APIs yet)
**Testing:** Unit tests, CI, Build Envoy Mobile Library and compile into mobile apps to ensure that changes do not break the build.
**Docs Changes:** Done in https://github.com/envoyproxy/envoy-mobile/blob/main/docs/root/api/stats.rst
**Release Notes:**
Expose histogram as a stat exposed via Pulse


library/common/engine.cc
```diff
@@ -154,6 +154,19 @@ envoy_status_t Engine::recordGaugeSub(const std::string& elements, uint64_t amou
   return ENVOY_FAILURE;
 }
 
+envoy_status_t Engine::recordHistogramDurationMs(const std::string& elements, uint64_t amount) {
+  if (server_ && client_scope_) {
+    std::string name = Stats::Utility::sanitizeStatsName(elements);
+    server_->dispatcher().post([this, name, amount]() -> void {
+      Stats::Utility::histogramFromElements(*client_scope_, {Stats::DynamicName(name)},
```
donyu, line 161: Why don't we use the *FromStatNames methods in utility.h? I see that we are passing a dynamic name, but I'm not sure why it needs to be a dynamic name.

library/common/engine.h
```diff
@@ -63,6 +63,13 @@ class Engine {
    */
   envoy_status_t recordGaugeSub(const std::string& elements, uint64_t amount);
 
+  /**
+   * Record a value for the histogram with the given string of elements
+   * @param elements, joined elements of the timeseries.
+   * @param amount, amount to add to the aggregated pool of events for quantile calculations
+   */
+  envoy_status_t recordHistogramDurationMs(const std::string& elements, uint64_t amount);
```
donyu, line 71: Put this down as a DurationMs since it's the most common use case I see on the mobile side, but I'm going to move this to an enum so clients can record any unit values e.g. bytes/ms/us

library/common/main_interface.h
```diff
@@ -133,6 +133,15 @@ envoy_status_t record_gauge_add(envoy_engine_t engine, const char* elements, uin
...
  */
 envoy_status_t record_gauge_sub(envoy_engine_t engine, const char* elements, uint64_t amount);
+
+/**
+ * Add another recorded amount to the histogram with the given string of elements.
+ * @param engine, the engine that owns the histogram.
+ * @param elements, the string that identifies the histogram to subtract from.
+ * @param amount, amount to record as a new value for the histogram.
+ */
+envoy_status_t record_histogram_duration_ms(envoy_engine_t engine, const char* elements, uint64_t amount);
```
donyu, line 143: This is then called in jni_interface.cc for Android clients.

library/java/src/io/envoyproxy/envoymobile/engine/AndroidEngineImpl.java
```diff
@@ -60,6 +60,16 @@ public int recordGaugeSub(String elements, int amount) {
     return envoyEngine.recordGaugeSub(elements, amount);
   }
 
+  @Override
+  public int recordGaugeSub(String elements, int amount) {
+    return envoyEngine.recordGaugeSub(elements, amount);
+  }
+
+  @Override
+  public int recordHistogramDurationMs(String elements, int duration) {
```
donyu, line 69: This will be used in a follow-up PR for the actual interface that clients will use to record histogram stats.

donyu:
@jingwei99 Mind taking a quick look on my WIP for histogram? Had a few questions too on certain functions.

library/common/engine.cc
```diff
@@ -154,6 +154,19 @@ envoy_status_t Engine::recordGaugeSub(const std::string& elements, uint64_t amou
...
+                                        Histogram::Unit::Milliseconds)
+          .recordValue(amount);
```
donyu, line 163: This is the only method exposed in `histogram.h` and all we really need to assign more values to the collection of data points in the histogram.

library/common/engine.cc
jingwei99: if i remember it correctly, it's bc the other way(w/o dynamic name) it requires changes on the c++ layer, which we might/could port to the platform layer, it adds complexity tho, there was a discussion on this when counter was introduced

library/common/engine.h
jingwei99: probably just do `recordHistogram` here, and client can decide the unit for the histogram

jingwei99: and `uint64_t amount` -> `uint64_t value`

library/common/engine.cc
donyu: Ah got it. Thanks for the context - I'll stick to dynamic names then.

library/kotlin/src/io/envoyproxy/envoymobile/PulseClient.kt
```diff
@@ -15,4 +15,10 @@ interface PulseClient {
 
   /** @return A gauge based on the joined elements. */
   fun gauge(vararg elements: Element): Gauge
+
+  /** @return A histogram based on the joined elements that can be used as a timer. */
+  fun histogramTimer(vararg elements: Element): Histogram
+
+  /** @return A histogram based on the joined elements that tracks an unspecified unit measure */
+  fun histogramGeneric(vararg elements: Element): Histogram
```
jingwei99, line 23: +1 to APIs

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyEngine.java
```diff
@@ -72,5 +73,15 @@ int runWithConfig(EnvoyConfiguration envoyConfiguration, String logLevel,
...
 
+  /**
+   * Add another recorded amount to the histogram with the given string of elements.
+   *
+   * @param elements Elements of the histogram stat.
+   * @param value Amount to record as a new value for the histogram distribution.
+   * @param unitMeasure the unit of measurement (e.g. milliseconds, bytes, etc.)
+   * @return A status indicating if the action was successful.
+   */
+  int recordHistogramValue(String elements, int value, HistogramUnit unitMeasure);
```
jingwei99, line 84: it probably worth documenting (as a comment in this pr) the reason of pushing `HistogramUnit` enum to this layer (instead of keeping in below jni)

jingwei99:
lgtm

probably need to change the pr description, will ping you over slack

library/kotlin/src/io/envoyproxy/envoymobile/PulseClient.kt
```diff
@@ -15,4 +15,10 @@ interface PulseClient {
 
   /** @return A gauge based on the joined elements. */
   fun gauge(vararg elements: Element): Gauge
+
+  /** @return A histogram based on the joined elements that can be used as a timer. */
+  fun histogramTimer(vararg elements: Element): Histogram
+
+  /** @return A histogram based on the joined elements that tracks an unspecified unit measure */
```
jingwei99, line 22:
```suggestion
/** @return A histogram based on the joined elements that tracks generic value */
```

donyu: 👍

library/java/src/io/envoyproxy/envoymobile/engine/EnvoyEngine.java
donyu: I updated to hide the enum at the JNI layer. I think it's cleaner to have 2 separate implementations here.

buildbreaker (dismissed):
Looks good to me. Great work @donyu !

cc: @junr03 @goaway @rebello95 in case they have objections or comments

docs/root/api/stats.rst
```diff
@@ -82,3 +82,33 @@ The value of a ``Gauge`` can be incremented, decremented, or reassigned.
...
+
+  // Add a new duration to the underlying timer distribution
+  // Kotlin
+  timer.recordDuration(5)
+
+  // Swift
+  timer.recordDuration(durationMs: 5)
+
+-------------
+``Histogram``
```
goaway, line 102: nit: I'd called this `Distribution`.

goaway:
I'm fine with this split, but IMO the strongest case to be made for having a timer type is the facility to pass a closure to a timer and have it emit the duration automatically when it exits. I'd suggest adding that interface here.

library/common/engine.h
```diff
@@ -63,6 +63,15 @@ class Engine {
    */
   envoy_status_t recordGaugeSub(const std::string& elements, uint64_t amount);
 
+  /**
+   * Record a value for the histogram with the given string of elements and unit measurement
+   * @param elements, joined elements of the timeseries.
+   * @param value, value to add to the aggregated distribution of values for quantile calculations
+   * @param unit_measure, the unit of measurement (e.g. milliseconds, bytes, etc.)
+   */
+  envoy_status_t recordHistogramValue(const std::string& elements, uint64_t value,
```
goaway, line 72:
```suggestion
  envoy_status_t recordDistributionValue(const std::string& elements, uint64_t value,
```

goaway: etc.

docs/root/api/stats.rst
donyu: Sure, I like that.

library/common/engine.h
donyu: 👍

donyu:
> the strongest case to be made for having a timer type is the facility to pass a closure to a timer and have it emit the duration automatically when it exits. I'd suggest adding that interface here.

I agree. In a follow-up PR (since this one is already fairly large), I plan to expose a `complete` call that track the elapsed duration similar to https://github.com/envoyproxy/envoy/blob/main/source/common/stats/timespan_impl.cc

docs/root/api/stats.rst
donyu: On 2nd thought, I'd like to follow the histogram naming since it's a more common terminology throughout metrics. Do you feel strongly about having this called `Distribution`?

goaway (approved):

PR #1217, opened 2020-12-18
python: update headers from impl experience

crockeo:
**Description:** This PR contains only the header changes from https://github.com/envoyproxy/envoy-mobile/pull/1209 which were made while developing. The plan is:

- Get this PR merged
- Split the concrete implementations of https://github.com/envoyproxy/envoy-mobile/pull/1209 up by category (e.g. engine vs. stream vs. headers types)
- Be able to parallelize review process

**Risk Level:** Low
**Testing:** N/A, no behavior changes
**Docs Changes:** N/A
**Release Notes:** N/A


library/cc/bridge_utility.h
```diff
@@ -0,0 +1,21 @@
...
+#include <string>
+#include <vector>
+
+#include "headers.h"
+#include "library/common/types/c_types.h"
+
+namespace Envoy {
+namespace Platform {
+
+envoy_data buffer_as_envoy_data(const std::vector<uint8_t>& data);
```
goaway, line 12: My inclination is not to introduce std::vector<uint8_t> as a buffer representation unless there's a good pythonic reason for doing so. Can we simply directly use envoy_data?

crockeo: yeah--this translates into `List[int]` in Python, but Eventually™ you can tell pybind11 to reinterpret a `std::vector<uint8_t>` as a `bytes` object.

a better approach (if it's supported by pybind11) is to just provide a `bytes` view onto an `envoy_data` instance

library/cc/stream_callbacks.h
```diff
@@ -32,9 +36,21 @@ class EnvoyHttpCallbacksAdapter {
 public:
   EnvoyHttpCallbacksAdapter(ExecutorSharedPtr executor, StreamCallbacksSharedPtr callbacks);
 
+  envoy_http_callbacks as_envoy_http_callbacks();
+
 private:
+  static void* dispatch_on_headers(envoy_headers headers, bool end_stream, void* context);
```
goaway, line 42: Maybe `c_on_headers`?

library/cc/headers.h
```diff
@@ -1,12 +1,13 @@
...
-
 #include <string>
 #include <vector>
 
 #include "absl/container/flat_hash_map.h"
 
+namespace Envoy {
+namespace Platform {
+
 using RawHeaders = absl::flat_hash_map<std::string, std::vector<std::string>>;
```
goaway, line 11: Maybe `RawHeaderMap`?

library/cc/bridge_utility.h
goaway: Okay, cool - I'd just prefer not to introduce yet another buffer representation if at all possible.

library/cc/executor.h
```diff
@@ -1,15 +1,19 @@
...
 
+namespace Envoy {
+namespace Platform {
+
 class Executor {
 public:
-  virtual ~Executor();
+  virtual ~Executor() {}
 
   virtual void execute(std::function<void()> closure) = 0;
```
goaway, line 13: Envoy has a PURE macro we should use for this.

library/cc/pulse_client.h
```diff
@@ -1,20 +1,20 @@
...
 // objects defined under stats. this will not be fully stubbed until stats is stubbed
 
-// #include "stats/counter.h"
-// #include "stats/element.h"
-// #include "stats/gauge.h"
-
 class PulseClient {
 public:
   // virtual Counter counter(Element element) = 0;
   // virtual Gauge gauge(Element element) = 0;
```
goaway, line 14: Use PURE macro (though it's commented out).

library/cc/stream_callbacks.h
goaway: Or maybe not - I was thinking these would be the raw function pointers to go on the struct, but that doesn't look to be the case from the implementation.

library/cc/executor.h
```diff
@@ -1,15 +1,19 @@
...
 
-// NOLINT(namespace-envoy)
-
 #include <functional>
 #include <memory>
 
+namespace Envoy {
+namespace Platform {
+
 class Executor {
```
goaway, line 9: I'm still not certain that we need this. If callbacks in this context are `std::function` then they can have all the state embedded they need for dispatch. I'm not sure there's really a need for the new interface, and I'm a little bit hesitant to introduce one that bindings will come to rely on.

library/cc/stream_callbacks.h
crockeo: changed these all to `c_...`, hope that's ok!

crockeo: although i noticed that maybe they're still the same in engine.h, let me go back and fix that.

goaway (approved):
This is great, thank you @crockeo!

PR #1219, opened 2020-12-18
python: http header impl

crockeo:
**Description:**

This PR implements all of the header-related classes--all of the request/response headers/trailers, their builders, and their base classes.

**Risk Level:** Low
**Testing:** N/A, will follow up when everything is finished with e2e testing + unit testing in Python and/orC++
**Docs Changes:** N/A
**Release Notes:** N/A


library/cc/BUILD
```diff
@@ -6,6 +6,18 @@ envoy_package()
 
 envoy_cc_library(
     name = "envoy_engine_cc_lib",
+    srcs = [
+        "cceaders.cc",
```
goaway, line 10: cceaders?

crockeo: oops

PR #1220, opened 2020-12-18
python: implement enum-related code

crockeo:
**Description:**

This PR implements all of the enum-related code, like translating between string- and enum-representations. It also implements `RetryPolicy`, since it's related to `RetryRule` (and defined in the same file).

**Risk Level:** Low
**Testing:** N/A, will follow up when everything is finished with e2e testing + unit testing in Python and/orC++
**Docs Changes:** N/A
**Release Notes:** N/A

### DIFF
--- a/library/cc/retry_policy.cc
+++ b/library/cc/retry_policy.cc
@@ -1,0 +1,105 @@
+#include "retry_policy.h"
+
+namespace Envoy {
+namespace Platform {
+
+static const std::pair<RetryRule, std::string> RETRY_RULE_LOOKUP[]{
+    {RetryRule::Status5xx, "5xx"},
+    {RetryRule::GatewayError, "gateway-error"},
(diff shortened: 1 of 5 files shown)


library/cc/retry_policy.cc
```diff
@@ -0,0 +1,105 @@
+#include "retry_policy.h"
+
+namespace Envoy {
+namespace Platform {
+
+static const std::pair<RetryRule, std::string> RETRY_RULE_LOOKUP[]{
+    {RetryRule::Status5xx, "5xx"},
+    {RetryRule::GatewayError, "gateway-error"},
+    {RetryRule::ConnectFailure, "conect-failure"},
```
goaway, line 9: "connect-failure"

goaway (approved):
Thank you!

PR #1222, opened 2020-12-18
python: implement bridge utilities

crockeo:
**Description:**

This implements all of the util functions used to convert between envoy-mobile types and cc-lib types. Note that right now it copies _all_ data at the border between envoy-mobile and the cc-lib, which is inefficient. As noted in the TODO, this is going to be a good place for optimization.

**Risk Level:** Low
**Testing:** N/A, will follow up with testing in only C++, since this won't be exposed into the Python layer.
**Docs Changes:** N/A
**Release Notes:** N/A

### DIFF
--- a/library/cc/bridge_utility.cc
+++ b/library/cc/bridge_utility.cc
@@ -1,0 +1,70 @@
+#include "bridge_utility.h"
+
+#include <sstream>
+
+namespace Envoy {
+namespace Platform {
+
+// TODO(crockeo): we always copy memory across boundaries; consider allowing for moves and/or
+// shared ownership w/ reference counting via envoy-mobile's release callbacks
+
+envoy_data string_as_envoy_data(const std::string& str) {
+  uint8_t* bytes = static_cast<uint8_t*>(safe_malloc(str.size()));
+  memcpy(bytes, &str[0], str.size());
+  return envoy_data{
+      .length = str.size(),
+      .bytes = bytes,
+      .release = envoy_noop_release,
+      .context = nullptr,
+  };
+}
+
+std::string envoy_data_as_string(envoy_data data) {
+  std::string str;
+  str.resize(data.length);
+  memcpy(&str[0], data.bytes, data.length);
+  return str;
+}
+
+envoy_headers raw_header_map_as_envoy_headers(const RawHeaderMap& headers) {
+  size_t header_count = 0;
+  for (const auto& pair : headers) {
+    header_count += pair.second.size();
+  }
+
+  envoy_header* headers_list =
+      static_cast<envoy_header*>(safe_malloc(sizeof(envoy_header) * header_count));
+
+  size_t i = 0;
+  for (const auto& pair : headers) {
+    const auto& key = pair.first;
+    for (const auto& value : pair.second) {
+      envoy_header& header = headers_list[i++];
+      header.key = string_as_envoy_data(key);
+      header.value = string_as_envoy_data(value);
+    }
+  }
+
+  envoy_headers raw_headers{
+      .length = static_cast<envoy_header_size_t>(header_count),
+      .headers = headers_list,
+  };
+  return raw_headers;
+}
+
+RawHeaderMap envoy_headers_as_raw_headers(envoy_headers raw_headers) {
+  RawHeaderMap headers;
+  for (auto i = 0; i < raw_headers.length; i++) {
+    auto key = envoy_data_as_string(raw_headers.headers[i].key);
+    auto value = envoy_data_as_string(raw_headers.headers[i].value);
+
+    if (!headers.contains(key)) {
+      headers.emplace(key, std::vector<std::string>());
+    }
+    headers[key].push_back(value);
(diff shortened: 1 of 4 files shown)


library/cc/bridge_utility.cc
```diff
@@ -0,0 +1,115 @@
...
+
+#include <sstream>
+
+namespace Envoy {
+namespace Platform {
+
+// TODO(crockeo): we always copy memory across boundaries; consider allowing for moves and/or
+// shared ownership w/ reference counting via envoy-mobile's release callbacks
+
+envoy_data buffer_as_envoy_data(const std::vector<uint8_t>& data) {
```
goaway, line 11: Presumably we don't need this anymore?

library/cc/bridge_utility.cc
```diff
@@ -0,0 +1,115 @@
...
+      .release = envoy_noop_release,
+      .context = nullptr,
+  };
+
+  return raw_data;
+}
+
+envoy_headers raw_headers_as_envoy_headers(const RawHeaders& headers) {
+  envoy_header* headers_list =
+      static_cast<envoy_header*>(safe_malloc(sizeof(envoy_header) * headers.size()));
```
goaway, line 43: I think this is maybe the wrong size - since the number of entries in the map won't account for header names that have multiple values.

library/cc/bridge_utility.cc
```diff
@@ -0,0 +1,115 @@
...
+}
+
+RawHeaders envoy_headers_as_raw_headers(envoy_headers raw_headers) {
+  RawHeaders headers;
+  for (auto i = 0; i < raw_headers.length; i++) {
+    auto key = envoy_data_as_string(raw_headers.headers[i].key);
+    auto value = envoy_data_as_string(raw_headers.headers[i].value);
+
+    // split the header value by ","
+    std::vector<std::string> values;
```
goaway, line 100: Need to check if the map already has a `values` entry for this header name, I believe.

library/cc/bridge_utility.cc
```diff
@@ -0,0 +1,70 @@
...
+// TODO(crockeo): we always copy memory across boundaries; consider allowing for moves and/or
+// shared ownership w/ reference counting via envoy-mobile's release callbacks
+
+envoy_data string_as_envoy_data(const std::string& str) {
+  uint8_t* bytes = static_cast<uint8_t*>(safe_malloc(str.size()));
+  memcpy(bytes, &str[0], str.size());
+  return envoy_data{
+      .length = str.size(),
+      .bytes = bytes,
+      .release = envoy_noop_release,
```
goaway, line 17: `release` should probably be `free` here (since you're using `malloc`).

library/cc/bridge_utility.cc
```diff
@@ -0,0 +1,70 @@
...
+      .context = nullptr,
+  };
+}
+
+std::string envoy_data_as_string(envoy_data data) {
+  std::string str;
+  str.resize(data.length);
+  memcpy(&str[0], data.bytes, data.length);
```
goaway, line 25: By library convention envoy_data should be released here.

library/cc/bridge_utility.cc
crockeo: ohhh my bad. i should have looked at the implementation of `envoy_noop_release`; i assumed that it was a `free` call w/o further action (noop). in retrospect it's obviously `noop` = no free :grin:

library/cc/bridge_utility.cc
```diff
@@ -0,0 +1,71 @@
...
+  RawHeaderMap headers;
+  for (auto i = 0; i < raw_headers.length; i++) {
+    auto key = envoy_data_as_string(raw_headers.headers[i].key);
+    auto value = envoy_data_as_string(raw_headers.headers[i].value);
+
+    if (!headers.contains(key)) {
+      headers.emplace(key, std::vector<std::string>());
+    }
+    headers[key].push_back(value);
+  }
```
goaway, line 66: Call `release_envoy_headers` before returning.

library/cc/bridge_utility.cc
```diff
@@ -0,0 +1,72 @@
...
+// shared ownership w/ reference counting via envoy-mobile's release callbacks
+
+envoy_data string_as_envoy_data(const std::string& str) {
+  uint8_t* bytes = static_cast<uint8_t*>(safe_malloc(str.size()));
+  memcpy(bytes, &str[0], str.size());
+  return envoy_data{
+      .length = str.size(),
+      .bytes = bytes,
+      .release = free,
+      .context = nullptr,
```
goaway, line 18: `context` needs to be the `bytes` pointer in order for `release` to work.

goaway:
Overall looks good; one minor fix needed.

library/cc/bridge_utility.cc
crockeo: doh

goaway:
Please resolve BUILD conflict, and I'll approve and merge.

PR #1218, opened 2020-12-18
python: implement stream-related classes

crockeo:
**Description:**

All header changes in this PR come from https://github.com/envoyproxy/envoy-mobile/pull/1217, that should be reviewed and merged first.

This PR implements all of the stream-related classes: `Stream`, `StreamCallbacks `EnvoyHttpCallbacksAdapter`, and `StreamPrototype`.

**Risk Level:** Low
**Testing:** N/A, will follow up when everything is finished with e2e testing + unit testing in Python and/orC++
**Docs Changes:** N/A
**Release Notes:** N/A

PR #1223, opened 2020-12-18
python: engine implementation

crockeo:
**Description:**

This PR implements all of the engine-related classes, like `Engine` and `EngineBuilder`.

**Risk Level:** Low
**Testing:** N/A, will follow up when everything is finished with e2e testing + unit testing in Python and/or C++
**Docs Changes:** N/A
**Release Notes:** N/A

### DIFF
--- a/library/cc/engine.cc
+++ b/library/cc/engine.cc
@@ -1,0 +1,42 @@
+#include "engine.h"
+
+#include "library/common/main_interface.h"
+#include "library/common/types/c_types.h"
+
+namespace Envoy {
+namespace Platform {
+
+Engine::Engine(envoy_engine_t engine, const std::string& configuration, LogLevel log_level,
+               std::function<void()> on_engine_running)
+    : engine_(engine), on_engine_running_(on_engine_running) {
+  envoy_engine_callbacks callbacks{
+      .on_engine_running = &Engine::c_on_engine_running,
+      .on_exit = &Engine::c_on_exit,
+      .context = this,
+  };
+
+  run_engine(this->engine_, callbacks, configuration.c_str(),
+             log_level_to_string(log_level).c_str());
+
+  this->stream_client_ = std::make_shared<StreamClient>(this->engine_);
+  this->pulse_client_ = std::make_shared<PulseClient>();
+}
+
+Engine::~Engine() { terminate_engine(this->engine_); }
+
+StreamClientSharedPtr Engine::stream_client() { return this->stream_client_; }
+PulseClientSharedPtr Engine::pulse_client() { return this->pulse_client_; }
+
+void Engine::c_on_engine_running(void* context) {
+  Engine* self = static_cast<Engine*>(context);
+  self->on_engine_running_();
+}
+
+void Engine::c_on_exit(void* context) {
+  // NOTE: this function is intentionally empty
+  // as we don't actually do any post-processing on exit.
+  (void)context;
+}
+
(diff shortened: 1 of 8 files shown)


goaway:
Waiting on this to be updated with executor change.

crockeo:
Should be ready for review now. I changed it to no longer depend on the executor, but because there's no trampoline class any longer we have to explicitly take the GIL when we execute the callback since it's coming from another thread. That's just represented in the same `..._shim` function pattern I had in the headers defn.

library/cc/engine.cc
```diff
@@ -0,0 +1,42 @@
...
+  this->stream_client_ = std::make_shared<StreamClient>(this->engine_);
+  this->pulse_client_ = std::make_shared<PulseClient>();
+}
+
+Engine::~Engine() { terminate_engine(this->engine_); }
+
+StreamClientSharedPtr Engine::stream_client() { return this->stream_client_; }
+PulseClientSharedPtr Engine::pulse_client() { return this->pulse_client_; }
+
+void Engine::c_on_engine_running(void* context) {
```
goaway, line 30: I don't think it needs to be a blocker or necessarily changed here, but I might consider putting these in an anonymous namespace instead of using private static functions.

library/cc/engine.cc
```diff
@@ -0,0 +1,42 @@
...
+  Engine* self = static_cast<Engine*>(context);
```
goaway, line 31: nit: since this is static, let's maybe say `engine` instead of `self` for clarity.

library/cc/engine_builder.cc
```diff
@@ -0,0 +1,96 @@
...
+
+  std::string config_str(config_template);
+  for (const auto& pair : replacements) {
+    const auto& key = pair.first;
+    const auto& value = pair.second;
+
+    size_t idx = 0;
+    while ((idx = config_str.find(key, idx)) != std::string::npos) {
+      config_str.replace(idx, key.size(), value);
+    }
```
goaway, line 87: I'd love to improve the performance of this, but this is no worse than what the other platforms are doing right now.

library/cc/engine.cc
crockeo: sounds good with me. to keep this consistent it's either got to be an update across this and #1218 or a fast follow. i'd prefer the latter

crockeo: ohhh actually i remember why i did this in the first place. i wanted to scope the closures for `on_engine_running_` and `on_exit_` (if ever used) privately under `Engine`. putting the static methods under `Engine` allowed them to access those closures.

is there another pattern that would let me do that w/ an anonymous namespace?

crockeo:
rebased onto main bc of BUILD problem

library/cc/engine.cc
goaway: Anonymous namespaces are only accessible from the file in which they're declared. The header won't even need to expose that they exist. Is that what you mean?

crockeo: synced offline--going to keep as-is right now because i want to be able to access `Engine::on_engine_running_` (the `std::function`) from inside `Engine::c_on_engine_running`.

long-term vision = move the closure definition into an `EngineCallbacks`

goaway (approved):

goaway:
Once BUILD conflicts are resolved, I can merge this.

goaway (approved):

Issue #1233, opened 2021-01-12
bazel 3.7.0 iOS issues


junr03:
Updating bazel to 3.7.0 or later creates the following errors in the iOS builds:

```
ERROR: /Users/jnino/src/envoy-mobile/library/swift/src/BUILD:5:23: every rule of type _swift_static_framework implicitly depends upon the target '@bazel_tools//tools/objc:libtool', but this target could not be found because of: no such target '@bazel_tools//tools/objc:libtool': target 'libtool' not declared in package 'tools/objc' (did you mean 'libtool.sh'?) defined by /private/var/tmp/_bazel_jnino/e2e6858b12cc5d95c5e2dcf9dce57ac4/external/bazel_tools/tools/objc/BUILD
ERROR: Analysis of target '//:ios_dist' failed; build aborted: Analysis failed
INFO: Elapsed time: 17.452s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (12 packages loaded, 11 targets configured)
```

I have tested with the 0.18.0 release of rules_swift and the 0.21.2 release of rules_apple.

junr03:
cc @keith

keith:
Is envoy using the toolchain from rules_cc? if so that needs to be updated

junr03:
Upstream Envoy? If I understand the bazel definitions correctly, yes. All the cc_* defs in envoy use cc_* defs loaded from @rules_cc under the hood.

keith:
not just the rules but the toolchain itself, rules_cc has a crosstool checked in that's kinda sorta a copy of the one from bazel

keith:
it required an update at the time https://github.com/bazelbuild/rules_cc/pull/82

keith:
Ah nevermind, this is invalid now https://github.com/envoyproxy/envoy-mobile/blob/9543bd6a84297df02652b84c16f19a4c9b1633c7/bazel/swift_static_framework.bzl#L135

keith:
I removed it from bazel here https://github.com/bazelbuild/bazel/commit/f8b51ffc2a8a953f64657ae4a21c9e1f1793dae2

keith:
Can you test your upgrade branch with this? https://github.com/envoyproxy/envoy-mobile/pull/1234

PR #1234, opened 2021-01-12
build: fix libtool label for bazel 3.7+

keith:
This tool was meant to be a bazel implementation detail so its public
accessor was removed in https://github.com/bazelbuild/bazel/commit/f8b51ffc2a8a953f64657ae4a21c9e1f1793dae2

Using the cc API from bazel we can dynamically get this. This is still a
bit of a hack because we're writing our own args for this. We might be
able to fetch the args, based on bazel's crosstool, dynamically as well,
but ideally we remove this rule instead.

Fixes https://github.com/envoyproxy/envoy-mobile/issues/1233

### DIFF
--- a/bazel/swift_static_framework.bzl
+++ b/bazel/swift_static_framework.bzl
@@ -3,6 +3,7 @@ This rules creates a fat static framework that can be included later with
 static_framework_import
 """
 
+load("@bazel_tools//tools/build_defs/cc:action_names.bzl", "CPP_LINK_STATIC_LIBRARY_ACTION_NAME")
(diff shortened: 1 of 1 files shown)


rebello95 (approved):

keith:
we should verify it fixes that issue with 3.7 before merging (although I think it'll be compatible before too

goaway:
Confirmed Android and iOS distros both build on bazel 3.7.2 with this change.

goaway (approved):

goaway:
Thank you @keith!

PR #1238, opened 2021-01-16
ci: set ndk version to 21

jingwei99:
Description: Set ndk version to 21 for the Mac image. Github's Mac image [installs latest version of ndk](https://github.com/actions/virtual-environments/blob/58a67845820857c641552a4d25075fe2d2fc466c/images/macos/provision/core/android-toolsets.sh#L69), which had been updated to 22 recently.
Risk Level: low
Testing: ci


ci/mac_ci_setup.sh
```diff
@@ -9,6 +9,15 @@ set -e
 # https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md for
 # a list of pre-installed tools in the macOS image.
 
+export ANDROID_HOME=$HOME/Library/Android/sdk
```
junr03, line 12: Do you mind moving to the bottom of the file and leaving a comment stating why this is done?

Also do you mind cleaning CI workflows that set any of these variables. Now that they are exported here they don't need to be set elsewhere.

junr03 (commented):
Thank you so much for taking this! Looks like the android job ran green 🎉. A couple comments.

ci/mac_ci_setup.sh
jingwei99: >Do you mind moving to the bottom of the file and leaving a comment stating why this is done?
yup, it's planned, it's at the top for debugging - to get executed first

>Also do you mind cleaning CI workflows that set any of these variables. Now that they are exported here they don't need to be set elsewhere.
`ANDROID_HOME` and `ANDROID_NDK_HOME`? sure

junr03 (approved):
Awesome, thanks!

PR #1240, opened 2021-01-20
python: normalize callback structs

crockeo:
**Description:** Because there was no separate callbacks struct, `c_on_engine_running` and `c_on_exit` were namespaced under `Engine`. This PR moves the `std::function` closures to a dedicated `EngineCallbacks` and moves `c_on_engine_running` and `c_on_exit` into an anonymous namespace.

This PR also normalizes the callback structure for both `EngineCallbacks` and `StreamCallbacks`, so they're constructed & passed around the same.

**Risk Level:** Low, none of this code is built into a production target.

**Testing:** N/A

**Docs Changes:** N/A

**Release Notes:** N/A


library/cc/engine.cc
```diff
@@ -6,16 +6,31 @@
 namespace Envoy {
 namespace Platform {
 
+namespace {
+
+static void c_on_engine_running(void* context) {
```
goaway, line 11: You can drop the `static` keyword - the anonymous namespace makes it redundant.

goaway:
Looks good after minor fix above.

goaway (approved):
This is great, thank you @crockeo!

PR #1241, opened 2021-01-20
python: e2e test and fixes

crockeo:
**Description:** implements an end-to-end test with the new python bindings interface. also makes some changes to the cc lib implementation, as it had some remaining bugs.

note that this does _not_ implement the gevent wrapper. i'm going to build on this PR with a gevent wrapper either attached to this PR, or in another PR.

**Risk Level:** Low

**Testing:** This is the test!

**Docs Changes:** N/A

**Release Notes:** N/A


library/cc/stream_callbacks.cc
```diff
@@ -79,15 +79,15 @@ void* EnvoyHttpCallbacksAdapter::c_on_error(envoy_error raw_error, void* context
 void* EnvoyHttpCallbacksAdapter::c_on_complete(void* context) {
   auto self = static_cast<EnvoyHttpCallbacksAdapter*>(context);
   if (self->stream_callbacks_->on_complete.has_value()) {
-    self->stream_callbacks_->on_complete.value();
+    self->stream_callbacks_->on_complete.value()();
```
goaway, line 82: nit: this is a little bit wacky. If nothing else, maybe pull into a local variable first.

library/python/e2e_test.py
```diff
@@ -0,0 +1,57 @@
...
+        .set_on_error(lambda error: stream_complete.set())
+        .set_on_cancel(lambda cancel: stream_complete.set())
+        .start()
+    )
+    stream.send_headers(
+        envoy_engine.RequestHeadersBuilder(
+            envoy_engine.RequestMethod.GET,
+            "https",
+            "www.google.com",
+            "/",
```
goaway, line 49: Can we run this test without using a public endpoint requiring network access?

For instance I set up this test to use a custom config to open a port with the library and issue a direct response:
https://github.com/envoyproxy/envoy-mobile/blob/main/library/swift/test/HttpBridgeTests/CancelStreamTest.swift
You don't necessarily have to use Envoy for that either (which is maybe a little bit brittle).

If we can't, I'd at least suggest using the /ping endpoint the example apps use.

goaway:
This is great as a start. It would be nice to structure it to have some parity with the Swift e2e integration tests. If that's a lot of work, it could be done subsequently.

library/python/e2e_test.py
crockeo: yeah, honestly i _meant_ to make it without calling google.com...and then forgot to get around to it :grin:

without further changes to the `EngineBuilder` interface, we can't do that same swift example because i don't expose a way to directly declare the config yaml. that may be worth doing, though.

in the meantime, it sounds easy _and_ good to point it at the lyft api endpoint instead.

goaway:
Looks like this needs a minor update to resolve conflicts.

crockeo:
3 merges with main later, I think I'm on the right place :grin:

@goaway I'd like to modify this to spin up a local server At Some Point™, but I don't think that point has to be now. let me know if you disagree and I can spin up a PR to allow the `Engine` to take a totally custom .yaml config

crockeo:
wow hard to stay tracking `main`

goaway (approved):
Nice work, thank you!

PR #1244, opened 2021-01-22
python: interface file for envoy_engine.so

crockeo:
**Description**: I believe mypy and other static analysis tools can't introspect a `.so` file for a Python library's types. This PR introduces a [mypy stub file](https://mypy.readthedocs.io/en/stable/stubs.html) to provide type information for both raw usage, and for typechecking the gevent & asyncio wrappers.

**Risk Level:** Low

**Testing:** N/A

**Docs Changes:** N/A

**Release Notes:** N/A


crockeo:
I want to migrate this to be automatically updated via [stubgen](https://mypy.readthedocs.io/en/stable/stubgen.html#automatic-stub-generation-stubgen) at some point, so that we can make sure the interface file is in sync with the actual `.so` we're building.

I think this is a good first step to make development easier while I'm working on the requests lib part of thigns.

goaway (approved):
Nice, this is great to have.

PR #1245, opened 2021-01-22
filters: refactor PlatformBridgeFilter and fix iteration state

goaway:
Description: Better encapsulates directional state in the filter chain, and allows for iteration state to be maintained independently in each direction, fixing a previous bug.
Risk Level: Low
Testing: Local & CI


library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -18,6 +18,18 @@ namespace Extensions {
 namespace HttpFilters {
 namespace PlatformBridge {
 
+namespace {
+void replaceHeaders(Http::HeaderMap& headers, envoy_headers c_headers) {
+  headers.clear();
+  for (envoy_header_size_t i = 0; i < c_headers.length; i++) {
```
rebello95, line 24: Does `length` get recomputed on each iteration? If so, I wonder if we should set that to a variable above instead

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -613,21 +550,22 @@ void PlatformBridgeFilter::onResumeEncoding() {
...
     free(result.pending_data);
   } else if (result.pending_data) {
-    Buffer::OwnedImpl inject_data;
-    inject_data.addBufferFragment(
-        *Buffer::BridgeFragment::createBridgeFragment(*result.pending_data));
-    encoder_callbacks_->addEncodedData(inject_data, /* watermark */ false);
-    free(result.pending_data);
-  }
-  if (pending_response_trailers_) {
+    PANIC("invalid filter state: data injection is unsupported at present");
```
rebello95, line 553: Should this be an assert instead? Not sure how often we use `PANIC`

library/common/extensions/filters/http/platform_bridge/filter.h
```diff
@@ -77,45 +79,54 @@ class PlatformBridgeFilter final : public Http::PassThroughFilter,
...
-  void onResumeDecoding();
-
-  // Scheduled on the dispatcher when resumeResponse is called from platform
-  // filter callbacks. Provides a snapshot of pending response state to the
-  // platform filter, and consumes invocation results to modify pending HTTP
-  // entities before resuming encoding.
-  void onResumeEncoding();
+  /**
+   * Internal delegate for managing logic and state that exists for both the request (decoding)
+   * and response (decoding) paths.
```
rebello95, line 84: shouldn't one of these be `encoding` rather than `decoding`?

library/common/extensions/filters/http/platform_bridge/filter.h
```diff
@@ -77,45 +79,54 @@ class PlatformBridgeFilter final : public Http::PassThroughFilter,
...
+    Http::FilterHeadersStatus onHeaders(Http::HeaderMap& headers, bool end_stream);
+
+    Http::FilterDataStatus onData(Buffer::Instance& data, bool end_stream,
+                                  Buffer::Instance* internal_buffer);
+
+    Http::FilterTrailersStatus onTrailers(Http::HeaderMap& trailers,
+                                          Buffer::Instance* internal_buffer);
+
+    // Scheduled on the dispatcher when resume* is called from platform
+    // filter callbacks. Provides a snapshot of pending request state to the
```
rebello95, line 106:
```suggestion
    // filter callbacks. Provides a snapshot of pending stream state to the
```

library/common/extensions/filters/http/platform_bridge/filter.h
```diff
@@ -77,45 +79,54 @@ class PlatformBridgeFilter final : public Http::PassThroughFilter,
...
+    // platform filter, and consumes invocation results to modify pending HTTP
+    // entities before resuming decoding.
```
rebello95, line 108: Is this only decoding?

rebello95 (approved):

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: In this case `length` translates to an offset in a struct, so it's equivalent to a variable. In either case the value is probably already optimized into a register.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: `PANIC` is effectively equivalent to `RELEASE_ASSERT(false,...)` and it is used, but in this case that's more an artifact of the code I commented out, and it could be re-written as a `RELEASE_ASSERT`. I'll go ahead and do so.

library/common/extensions/filters/http/platform_bridge/filter.h
goaway: Yep, good catch.

library/common/extensions/filters/http/platform_bridge/filter.h
goaway: Nope, will fix these comments up, thanks.

rebello95 (approved):

rebello95 (approved):

PR #1246, opened 2021-01-27
filters: add support for data and trailers injection

goaway:
Description: Adds support for injecting data and/or trailers specifically during onResume() callbacks when resuming filter iteration asynchronously. This is accomplished by passing back data and/or trailers where none were originally present at the point of resumption.
Risk Level: Moderate
Testing: Pending

### DIFF
--- a/library/common/extensions/filters/http/platform_bridge/filter.cc
+++ b/library/common/extensions/filters/http/platform_bridge/filter.cc
@@ -534,11 +534,11 @@ void PlatformBridgeFilter::ResponseFilterBase::addTrailers(envoy_headers trailer
 }
 
 void PlatformBridgeFilter::RequestFilterBase::resumeIteration() {
-  parent_.encoder_callbacks_->continueEncoding();
+  parent_.decoder_callbacks_->continueDecoding();
 }
 
 void PlatformBridgeFilter::ResponseFilterBase::resumeIteration() {
-  parent_.decoder_callbacks_->continueDecoding();
+  parent_.encoder_callbacks_->continueEncoding();
 }
 
 // Technically-speaking to align with Envoy's internal API this method should take
--- a/library/common/extensions/filters/http/platform_bridge/filter.cc
+++ b/library/common/extensions/filters/http/platform_bridge/filter.cc
@@ -58,26 +58,11 @@ PlatformBridgeFilterConfig::PlatformBridgeFilterConfig(
       platform_filter_(static_cast<envoy_http_filter*>(
           Api::External::retrieveApi(proto_config.platform_filter_name()))) {}
 
-PlatformBridgeFilter::FilterBase
-PlatformBridgeFilter::FilterBase::createRequestBase(PlatformBridgeFilter& parent) {
-  return FilterBase{
-      parent, parent.platform_filter_.on_request_headers, parent.platform_filter_.on_request_data,
-      parent.platform_filter_.on_request_trailers, parent.platform_filter_.on_resume_request};
-}
-
-PlatformBridgeFilter::FilterBase
-PlatformBridgeFilter::FilterBase::createResponseBase(PlatformBridgeFilter& parent) {
-  return FilterBase{
-      parent, parent.platform_filter_.on_response_headers, parent.platform_filter_.on_response_data,
-      parent.platform_filter_.on_response_trailers, parent.platform_filter_.on_resume_response};
-}
-
 PlatformBridgeFilter::PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr config,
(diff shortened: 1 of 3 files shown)


library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -378,23 +364,16 @@ Http::FilterHeadersStatus PlatformBridgeFilter::encodeHeaders(Http::ResponseHead
...
-  Buffer::Instance* internal_buffer = nullptr;
-  if (request_filter_base_.iteration_state_ == IterationState::Stopped &&
-      decoder_callbacks_->decodingBuffer()) {
-    decoder_callbacks_->modifyDecodingBuffer([&internal_buffer](Buffer::Instance& mutable_buffer) {
-      internal_buffer = &mutable_buffer;
-    });
-  }
-
-  return request_filter_base_.onData(data, end_stream, internal_buffer);
+  // Delegate to base implementation for request and response path.
```
rebello95, line 375: This is only the request path looks like

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -406,31 +385,15 @@ Http::FilterDataStatus PlatformBridgeFilter::encodeData(Buffer::Instance& data,
...
-  Buffer::Instance* internal_buffer = nullptr;
-  if (response_filter_base_.iteration_state_ == IterationState::Stopped &&
-      encoder_callbacks_->encodingBuffer()) {
-    encoder_callbacks_->modifyEncodingBuffer([&internal_buffer](Buffer::Instance& mutable_buffer) {
-      internal_buffer = &mutable_buffer;
-    });
-  }
-
-  return response_filter_base_.onData(data, end_stream, internal_buffer);
+  // Delegate to base implementation for request and response path.
```
rebello95, line 388: Only the response path (there are a few more below - can fix these later)

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -551,9 +492,9 @@ void PlatformBridgeFilter::FilterBase::onResume(Buffer::Instance* internal_buffe
     internal_buffer->addBufferFragment(
         *Buffer::BridgeFragment::createBridgeFragment(*result.pending_data));
     free(result.pending_data);
-  } else {
-    RELEASE_ASSERT(!result.pending_data,
-                   "invalid filter state: data injection is unsupported at present");
+  } else if (result.pending_data) {
```
rebello95, line 495: 🎉

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -562,9 +503,70 @@ void PlatformBridgeFilter::FilterBase::onResume(Buffer::Instance* internal_buffe
...
+  parent_.encoder_callbacks_->continueEncoding();
+}
+
+void PlatformBridgeFilter::ResponseFilterBase::resumeIteration() {
+  parent_.decoder_callbacks_->continueDecoding();
+}
+
+// Technically-speaking to align with Envoy's internal API this method should take
+// a closure to execute with the available buffer, but since we control all usage,
+// this shortcut works for now.
```
rebello95, line 546: Doesn't seem like there'd be a reason to ever change this unless we _wanted_ to support the platform-layer modifying the data with a closure (which seems like extra complexity), no?

rebello95 (approved):

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: I put this in more as a note to a future reader that we're actually doing something that might seem a little unexpected/wrong from the perspective of Envoy's API. The form of Envoy's API kind of implies (but doesn't state) that you should perform all modifications in the context of a closure rather than cheating and extracting a pointer to the buffer like we're doing. But since we only use it internally and we only use it in the scope of this call stack, it seems unlikely to ever be a problem, unless Envoy's filter pipeline and buffer management are dramatically overhauled. Tests would also reveal right away that this was broken if it ever changes.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: It's both - i.e., `onData` is a common method implemented in the `FilterBase` superclass of both `RequestFilterBase` and `ResponseFilterBase`. The comment is just to explain that we're delegating to a common execution path. If there's a clearer way to state it, I'm also open to suggestions.

rebello95 (approved):

goaway:
Going to proceed with merge - swift tests passed on previous (identical) CI run.

PR #1249, opened 2021-02-02
platform filters: fix memory leaks

junr03:
Description: this PR fixes leaks in the platform filter stack:
1. iOS was leaking a weak ptr allocated on the heap when the filter was synchronous. Android was not because Android unconditionally wraps the heap allocated pointer with a GC managed object.
2. headers and trailer structs were leaking under some conditions on filter.cc
3. error messages were leaking if the platform filter did not have a on error callback set.

Risk Level: med - previous versions of Envoy Mobile that use platform filters were leaking.
Testing: ran the Lyft app under the leak checker.


junr03:
@goaway when you are back in office I'd like to go over filter.cc with you.

1. I wonder if I have missed other conditions under which native structs are leaking.
2. I would like to see if we can come up with a way of conditionally allocating the weak ptr on the heap. The previous if guard was always true as the struct fields (`set_request/response_callbacks) were always set with static functions.

buildbreaker (dismissed):

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -161,6 +171,8 @@ Http::FilterHeadersStatus PlatformBridgeFilter::FilterBase::onHeaders(Http::Head
   case kEnvoyFilterHeadersStatusStopIteration:
     pending_headers_ = &headers;
     iteration_state_ = IterationState::Stopped;
+    // returned headers from the platform filter are not used. Release them here.
+    release_envoy_headers(result.headers);
```
goaway, line 175: These will (should) always be nullptr. Can we replace this with an `ASSERT`?

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -286,6 +298,8 @@ Http::FilterTrailersStatus PlatformBridgeFilter::FilterBase::onTrailers(Http::He
   case kEnvoyFilterTrailersStatusStopIteration:
     pending_trailers_ = &trailers;
     iteration_state_ = IterationState::Stopped;
+    // returned trailers from the platform filter are not used. Release them here.
+    release_envoy_headers(result.trailers);
```
goaway, line 302: These will (should) always be nullptr. Can we replace this with an `ASSERT`?

goaway:
@junr03 thanks for investigating this while I was out! I'd like to suggest a slightly different solution: let's clear the function pointers up front on iOS when it's not an AsyncRequest/ResponseFilter, rather than unconditionally allocating and "knowing" to deallocate later.

junr03:
alright @goaway tests updated.

junr03:
Would you like all nil checks on EnvoyEngimeImpl.m to be asserts?

https://github.com/envoyproxy/envoy-mobile/blob/c44e7a17da4303efa33b1031540ea90954fc1597/library/objective-c/EnvoyEngineImpl.m#L77

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -74,26 +73,41 @@ PlatformBridgeFilter::PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr c
   if (platform_filter_.init_filter == nullptr) {
     ENVOY_LOG(debug, "PlatformBridgeFilter({}): missing initializer", filter_name_);
     platform_filter_ = {};
+
+    // Does this make sense, or should we rather protect the rest of the code against nullptr for
+    // both request_filter_base_ and response_filter_base_.
```
goaway, line 78: This makes sense, but I'd switch to an `if`/`else` and invert the check so the "happy" path comes first, that way the constructor can terminate with setting the two bases in either case.

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -74,26 +73,41 @@ PlatformBridgeFilter::PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr c
...
+    response_filter_base_ = std::make_unique<ResponseFilterBase>(*this);
+
     return;
   }
 
   // Set the instance_context to the result of the initialization call. Cleanup will ultimately
   // occur within the onDestroy() invocation below.
   ENVOY_LOG(trace, "PlatformBridgeFilter({})->init_filter", filter_name_);
-  platform_filter_.instance_context = platform_filter_.init_filter(platform_filter_.static_context);
+  platform_filter_.instance_context = platform_filter_.init_filter(&platform_filter_);
```
goaway, line 88: This looks like it should be reverted, I think?

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -74,26 +73,41 @@ PlatformBridgeFilter::PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr c
...
 
+  // TODO(goaway): currently both platform APIs unconditionally set this field, meaning that the
+  // heap allocation below occurs when it could be avoided.
   if (platform_filter_.set_request_callbacks) {
     platform_request_callbacks_.resume_iteration = envoy_filter_callback_resume_decoding;
     platform_request_callbacks_.release_callbacks = envoy_filter_release_callbacks;
     // We use a weak_ptr wrapper for the filter to ensure presence before dispatching callbacks.
+    // Note that the weak_ptr here is heap allocated because it needs to be managed and held by
+    // platform specific runtimes. Therefore, it is NECESSARY that the platform deallocates the
+    // memory. Otherwise the weak ptr will leak.
```
goaway, line 110: Wait, aren't we going to address this in this PR?

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: Nevermind, I see.

library/swift/src/filters/Filter.swift
```diff
@@ -26,7 +26,7 @@ extension EnvoyHTTPFilter {
         case .continue(let headers):
           return [kEnvoyFilterHeadersStatusContinue, headers.headers]
         case .stopIteration:
-          return [kEnvoyFilterHeadersStatusStopIteration, envoyHeaders]
+          return [kEnvoyFilterHeadersStatusStopIteration, NSNull()]
```
goaway, line 29: I think you can use `nil` and it will become `NSNull()` automagically when mapped to Objective-C.

goaway: Alternatively it might still just be clearer to handle a different tuple length for `StopIteration`.

library/common/extensions/filters/http/platform_bridge/filter.cc
junr03: Yeah, it is addressed in the leaking case. But I wanted to leave the comment, because it is the case that if the pointer gets allocated it is up to the platform code to release.

library/swift/src/filters/Filter.swift
junr03: the compiler complained about nil, because it is a 0 value rather than an object that conform to Any.

The rest of the code is pretty nice in its absence of branching so I felt it was good to leave this way?

goaway: Does making the type `[Any?]` resolve this? Using `NSNull` explicitly feels a little bit against the grain.

library/common/extensions/filters/http/platform_bridge/filter.cc
goaway: Okay, sure, we can leave the comment for clarity. Maybe simplify to something like "The weak_ptr is heap-allocated, because it must be managed (and eventually released) by platform code."

library/swift/src/filters/Filter.swift
junr03: talked offline. Decided to leave NSNull to not leak branching into the static bridge functions.

goaway (dismissed):

goaway (approved):

PR #1250, opened 2021-02-04
stats: avoid string join during stats reporting

jingwei99:
Description: construct the stats name in constructor, avoid string join during stats reporting (inc, set, etc.)
Risk Level: low
Testing: unit, ci


donyu (approved):

buildbreaker (approved):
Hah. This looks like a good thing to have. Thank you @jingwei99 !

goaway (approved):

PR #1251, opened 2021-02-05
errors: map locally-generated gRPC errors

goaway:
Description: Because Envoy may generate local grpc replies for grpc requests, this is required (for now) to maintain existing error mapping behavior.
Risk Level: Low
Testing: Pending


goaway:
Planning to add integration coverage to this as a follow-up, similar to the existing HTTP coverage.

library/common/extensions/filters/http/local_error/filter.cc
```diff
@@ -15,25 +17,75 @@ namespace Extensions {
...
-  // Absence of EnvoyUpstreamServiceTime header implies this is a local reply, which we treat as
-  // a stream error.
-  if (!success && headers.get(Http::Headers::get().EnvoyUpstreamServiceTime).empty()) {
+  const auto grpc_status = Grpc::Common::getGrpcStatus(headers);
+  if (grpc_status && grpc_status.value() != Grpc::Status::WellKnownGrpcStatus::Ok) {
+    ASSERT(end_stream,
+           "Local gRPC responses must consist of a single headers frame. If Envoy changes "
+           "this expectation, this code needs to be updated.");
+    ENVOY_LOG(debug, "intercepted local GRPC response");
+    grpcError_ = true;
```
rebello95, line 76: It looks like this is never read, is it needed?

rebello95 (approved):
Looks good, pending tests in a future PR

library/common/extensions/filters/http/local_error/filter.cc
goaway: Actually you're right, it's not; good catch. I was using it previously, but was able to simplify things. I'll go ahead and remove.

rebello95 (approved):

PR #1258, opened 2021-02-12
stats: unregister custom stats sink

jingwei99:
Description: unregister custom stats sink until the backend service is ready
Risk Level: low
Testing: ci/unit


junr03 (commented):
Thanks!

You also need to update: https://github.com/envoyproxy/envoy-mobile/blob/main/library/common/config_template.cc#L210 and https://github.com/envoyproxy/envoy-mobile/blob/main/envoy_build_config/extensions_build_config.bzl

junr03 (approved):

PR #1260, opened 2021-02-16
 filters: only access mutable buffer when iteration is stopped on current 

goaway:
Description: A previous change caused the pointer to the mutable buffer to be retrieved regardless of iteration state which trips an internal assertion in Envoy.
Risk Level: Low
Testing: New integration test

### DIFF
--- a/library/common/extensions/filters/http/platform_bridge/filter.cc
+++ b/library/common/extensions/filters/http/platform_bridge/filter.cc
@@ -559,6 +559,9 @@ void PlatformBridgeFilter::ResponseFilterBase::resumeIteration() {
 // this shortcut works for now.
 Buffer::Instance* PlatformBridgeFilter::RequestFilterBase::buffer() {
   Buffer::Instance* internal_buffer = nullptr;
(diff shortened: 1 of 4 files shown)


rebello95 (approved):

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -559,7 +559,7 @@ void PlatformBridgeFilter::ResponseFilterBase::resumeIteration() {
 // this shortcut works for now.
 Buffer::Instance* PlatformBridgeFilter::RequestFilterBase::buffer() {
   Buffer::Instance* internal_buffer = nullptr;
-  if (parent_.decoder_callbacks_->decodingBuffer()) {
+  if (iteration_state_ == IterationState::Stopped && parent_.decoder_callbacks_->decodingBuffer()) {
```
junr03, line 562: do you mind leaving a comment that refreshes the reader on why a iteration state needs to be stopped?

PR #1262, opened 2021-02-22
coverage: integration test for LocalErrorFilter when using gRPC

goaway:
Description: This test covers the case where the LocalErrorFilter transforms a locally generated response for a gRPC request to an error.

Note: Kotlin tests can be added once we have #1263.
Risk Level: Low
Testing: This commit adds coverage.

### DIFF
--- a/library/swift/test/HttpBridgeTests/BUILD
+++ b/library/swift/test/HttpBridgeTests/BUILD
@@ -36,6 +36,17 @@ envoy_mobile_swift_test(
 )
 
 envoy_mobile_swift_test(
+    name = "grpc_receive_error_test",
+    srcs = [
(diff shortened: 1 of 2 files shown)


junr03 (commented):
Should we also have the same test case in Kotlin given that we have two separate (albeit small) gRPC codecs?

goaway:
We don't currently have support for this level of integration testing on Kotlin. It is something that I'd like for us to add, but there are a few missing pieces we need first--most notably a non-Android build target.

junr03 (approved):
Awesome, thanks for opening the JVM issue @goaway

goaway:
thanks @junr03!

PR #1265, opened 2021-02-24
swift/kotlin: enforce unsigned :status code values

rebello95:
We can prevent an [assert from being hit within Envoy](https://github.com/envoyproxy/envoy/blob/17e815122ff53d0ac6cb2d64cdbf1bfc547bb7e8/source/common/http/utility.cc#L441-L448) if the end consumer attempts to set a negative status code via these platform interfaces.

### DIFF
--- a/library/kotlin/src/io/envoyproxy/envoymobile/ResponseHeaders.kt
+++ b/library/kotlin/src/io/envoyproxy/envoymobile/ResponseHeaders.kt
@@ -15,7 +15,7 @@ class ResponseHeaders : Headers {
    * HTTP status code received with the response.
    */
   val httpStatus: Int? by lazy {
-    value(":status")?.first()?.toIntOrNull().takeIf { it >= 0 }
+    value(":status")?.first()?.toIntOrNull()?.takeIf { it >= 0 }
   }
 
   /**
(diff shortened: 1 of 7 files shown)


buildbreaker (approved):

buildbreaker (approved):

library/kotlin/src/io/envoyproxy/envoymobile/ResponseHeaders.kt
```diff
@@ -14,7 +14,9 @@ class ResponseHeaders : Headers {
   /**
    * HTTP status code received with the response.
    */
-  val httpStatus: Int? by lazy { value(":status")?.first()?.toIntOrNull() }
+  val httpStatus: Int? by lazy {
```
goaway, line 17: Should this be a UInt?

goaway: Ah, sorry this is the Kotlin one. Got it.

rebello95: Kotlin doesn't have the concept of unsigned integers available to us yet: https://kotlinlang.org/docs/basic-types.html#unsigned-integers, which is why we have a unit test for this now. Meant to mention that in the PR but forgot

buildbreaker (approved):

PR #1266, opened 2021-02-26
stats: add apis for custom tags

jingwei99:
Description: 

This PR adds APIs for custom tags for both iOS and Android, and for all four types of stats.

Two APIs are supported for each type of stats: 1) stats creation with tags; 2) creating a new stats from an existing stats with a list of tags attached.

The first API is designed for stats creation during compile time, calling the first API to create stats during runtime is not restricted. The second API is designed for stats creation during runtime, similar to the first API, calling this API to create stats during compile time is not restricted. 

`Tag` classes do not enforce checks like `Element` classes, the idea is to leave the implementations calling this set of APIs to enforce their own system/platform specific restrictions on tag keys and tag values.

Risk Level: Medium

Testing: unit tests, deployed to Lyft iOS and Android apps for smoke tests

Signed-off-by: Jingwei Hao jingwei.hao@gmail.com


library/kotlin/src/io/envoyproxy/envoymobile/stats/Counter.kt
```diff
@@ -8,4 +8,6 @@ interface Counter {
    * Increment the counter by the given count.
    */
   fun increment(count: Int = 1)
+
+  fun attach(tags: List<Tag>): Counter
```
jingwei99, line 12: This method allows dynamically adding tags during runtime for clients.

library/kotlin/src/io/envoyproxy/envoymobile/PulseClient.kt
```diff
@@ -13,6 +13,8 @@ interface PulseClient {
    */
   fun counter(vararg elements: Element): Counter
 
+  fun counter(vararg elements: Element, tags: List<Tag>): Counter
```
jingwei99, line 16: i imagine this api will be used to add tags during(pre) compile time.

caveat on kotlin: kotlin does not allow two vararg parameters for methods, hence `List`.

for swift the [current api](https://github.com/envoyproxy/envoy-mobile/blob/fe12371436065d52da1a3807956d6dbd17d3fd63/library/swift/src/PulseClient.swift#L12) uses an array for elements

library/kotlin/src/io/envoyproxy/envoymobile/stats/Counter.kt
goaway: I would suggest having tags be an optional parameter to `increment` (and other "record" stats methods).

junr03: I was discussing this with @jingwei99 during our 1:1. What I like about the current "attach" API is that the caller can hold the returned Counter and use it for subsequent increments rather than always having to pass down the tag list.

I can also see how it might be easier on the caller to call increment with tag lists and let the stats client deal with incrementing the correct counter rather than manually maintaining counter objects to select based on runtime tags. 

I could see how one might use either based on the situation.

jingwei99: had a discussion over slack, will go with the suggestions and keep both apis, ty both :)

library/kotlin/io/envoyproxy/envoymobile/stats/Stats.kt
```diff
@@ -0,0 +1,13 @@
...
+
+/** Interface for stats */
+interface Stats {
+
+  fun convert(tags: List<Tag>): Map<String, String> {
+    val tagMap = mutableMapOf<String, String>()
+    tags.forEach { it ->
+      tagMap[it.key] = it.value
+    }
+    return tagMap
```
jingwei99, line 11: This interface is created to host the shared logic to convert list of `Tag` to `Map<String, String>`

library/kotlin/io/envoyproxy/envoymobile/stats/Tag.kt
```diff
@@ -0,0 +1,5 @@
+package io.envoyproxy.envoymobile
+
+/** Tag represents the [key,value] pair component of a time series name. */
+
+class Tag(internal val key: String, internal val value: String)
```
jingwei99, line 5: The choice was between having a dedicated `Tag` class and pass around tags as `List<Tag>` or directly use a `Map<String, String>`. 

Decided to have a `Tag` class: it makes code easier to read for the public APIs. It will also make it easier to add enforcements to tags if needed down the line.

library/swift/stats/Stats.swift
```diff
@@ -0,0 +1,12 @@
...
+
+struct Stats {
+  /// Convert tags.
+ static func convert(tags: [Tag]) -> [String: String] {
+    var tagMap = [String: String]()
+    for tag in tags {
+        tagMap[tag.key] = tag.value
+    }
+    return tagMap
+  }
```
jingwei99, line 11: For swift, the more common way is to make a static function (as oppose to having a method in an interface like kotlin), and for the parity, this function is namespaced as `Stats` - same to the kotlin `Stats` interface

library/kotlin/io/envoyproxy/envoymobile/PulseClient.kt
```diff
@@ -13,12 +13,32 @@ interface PulseClient {
    */
   fun counter(vararg elements: Element): Counter
 
+  /**
+   * @return A counter based on the joined elements with list of tags.
+   */
+  fun counter(vararg elements: Element, tags: List<Tag>): Counter
```
murki, line 19: Do Kotlin interfaces support optional params? If so we could simplify and just have one method like
```kotlin
fun counter(vararg elements: Element, tags: List<Tag> = emptyList()): Counter
```

you might need to add `JvmOverloads` for java interop

library/kotlin/io/envoyproxy/envoymobile/stats/Stats.kt
murki: My Kotlin is a little rusty but I wonder if we could simplify this by using an extension method instead, something like
```kotlin
internal fun List<Tag>.toMap(): Map<String, String> { ... }
```

murki:
Overall lgtm but all defer to the rest of the reviewers for approval

library/kotlin/io/envoyproxy/envoymobile/stats/Stats.kt
jingwei99: I thought about it, and wasn't sure about the opinion for having kotlin extension here - my reading is different codebase tends to have different preferences on kotlin extension.

Similarly for iOS, the static convert function can be facilitated with an extension too.

Would like to hear @junr03 and @goaway 's thoughts on this.

library/kotlin/io/envoyproxy/envoymobile/stats/Counter.kt
```diff
@@ -3,9 +3,12 @@ package io.envoyproxy.envoymobile
...
  */
-interface Counter {
+interface Counter : Stats {
   /**
-   * Increment the counter by the given count.
+   * Increments the counter by the given count.
    */
   fun increment(count: Int = 1)
+
+  /** Gets a counter with tags attached. */
```
junr03, line 12: format

library/kotlin/io/envoyproxy/envoymobile/stats/Tag.kt
junr03: I had written a comment about this before I got to read this. I appreciate the use of a clear class here, but I would like to see if we can do it in a way that pushes its use all the way to the bridge layer. i.e, we are converting from one thing to another before we _need_ to, and the conversion is not free. 

I'd like to hear @rebello95 and @buildbreaker opinion on the balance of the exposed API vs adding more conversion overhead.

buildbreaker: What if we just have a `Tags` class and have this class just be the `List<Tag>` internally and it can expose a nicer API? This way, you can also unify the `Stats.convert` method into this Tags class.

jingwei99: Having a `Tags` class sounds a good option to me. Would like to hear more input too.

jingwei99: @junr03 about the layers for the conversion:

there are two conversions: 1) from `Tag` or `Tags` to `Map<String, String>`, and 2) from `Map<String, String>` to native. The second conversion: From Map<String, String>` to native is done currently in `EnvoyEngineImpl` - same as headers. 

For the first conversion, I was thinking about pushing it down to the `EnvoyEngineImpl` and have it done along with the second conversion, it will require `EnvoyEngineImpl` to import `Tag` or `Tags` class, and I thought at the time it probably was not a good idea.

library/kotlin/io/envoyproxy/envoymobile/stats/Stats.kt
goaway: Why do we convert it to a map at all? Isn't it simpler to keep it as a list, given we'll be passing it to the JNI as an array anyways?

library/kotlin/io/envoyproxy/envoymobile/stats/Tag.kt
goaway: Commented above; I don't see the need for having a Map intermediary.

library/kotlin/io/envoyproxy/envoymobile/stats/TimerImpl.kt
```diff
@@ -7,15 +7,27 @@ import java.lang.ref.WeakReference
...
+    this.tags = convert(tags)
+  }
+
+  private constructor(engine: WeakReference<EnvoyEngine>, series: String, tags: List<Tag>) {
+    this.envoyEngine = engine
+    this.series = series
+    this.tags = convert(tags)
   }
 
   override fun completeWithDuration(durationMs: Int) {
```
goaway, line 26: unrelated nit: I _really_ would like to see this method renamed to `recordDuration`; I think the current name is quite misleading.

goaway:
What's the purpose of the `attach` API? Why not just allow someone to pass tags in when making a record-* call (since it's supported internally throughout the rest of the pipeline, anyways)?

Attaching tags on creation I can see - some things that are tags might be static (e.g., OS version).

If it's not static, the implication is that in can change, at which point attaching it makes less sense to me.

library/kotlin/io/envoyproxy/envoymobile/stats/Tag.kt
junr03: Similar to Headers, you can have a Tags class that allows for additional functionality and cleaner API. But it stores tags internally in a map<string, string>. That way there's no conversation or copy when passing it down.

goaway: If it starts as a map and stays as a map, that's fine with me.

jingwei99: sounds good to me

library/kotlin/io/envoyproxy/envoymobile/stats/TimerImpl.kt
jingwei99: i created a [ticket](https://jira.lyft.net/browse/RES-2173) for this one

jingwei99:
> What's the purpose of the attach API? Why not just allow someone to pass tags in when making a record-* call (since it's supported internally throughout the rest of the pipeline, anyways)?
> 
> Attaching tags on creation I can see - some things that are tags might be static (e.g., OS version).
> 
> If it's not static, the implication is that in can change, at which point attaching it makes less sense to me.

chatted over slack, there was a misunderstanding way back when the apis were first under discussion.

the conclusion is to have two places (apis) to add tags:

library/kotlin/io/envoyproxy/envoymobile/stats/Stats.kt
snowp: Reading this conversation makes me realize that Envoy won't treat the list of tags like a set, so if we can ensure consistent ordering (e.g. via an ordered map) we would avoid duplicate stat entries that really has the same set of tags. 

Maybe this should be fixed in Envoy, I'll take a look

jingwei99: 👍
fwiw, currently tags are passed into envoy as [Stats::StatNameTagVector](https://github.com/envoyproxy/envoy/blob/1d1b708c7bf6efa02c41d9ce22cbf1e4a1aeec2c/include/envoy/stats/tag.h#L27)

buildbreaker (approved):
Good work @jingwei99

Issue #949, opened 2020-07-10
performance: create fully-bridged types for headers and data


goaway:
Instead of converting from a platform type to a bridge (c) type to an Envoy-native type when crossing library layer boundaries, instead we could create platform and Envoy-native implementations backed by the bridge type, such that conversion would be largely unnecessary. This could substantially reduce call overhead.

This optimization could be especially beneficial for platform-bridged filters, where frequent additional traversals of library boundaries would occur with header/data payloads. Note that at present the interfaces for platform filters utilize immutable types for payloads; these would need to change to mutable types (which Envoy utilizes for native filters) to realize this benefit.

PR #1271, opened 2021-02-27
stats: engine layer change for cardinality

jingwei99:
Description: Engine layer change for cardinality
Risk Level: Medium
Testing: Unit test

### DIFF
--- a/library/common/data/utility.h
+++ b/library/common/data/utility.h
@@ -43,6 +43,13 @@ envoy_data copyToBridgeData(const Buffer::Instance&);
  */
 std::string copyToString(envoy_data data);
 
+/**
+ * Transform from an std::string to an envoy_data.
+ * @param s, the std::string to transform envoy_data from.
+ * @return envoy_data the envoy_data transformed from s.
+ */
+envoy_data toEnvoyData(const std::string& s);
+
 } // namespace Utility
 } // namespace Data
 } // namespace Envoy
--- a/library/common/data/utility.h
+++ b/library/common/data/utility.h
@@ -43,13 +43,6 @@ envoy_data copyToBridgeData(const Buffer::Instance&);
  */
 std::string copyToString(envoy_data data);
 
-/**
- * Transform from an std::string to an envoy_data.
- * @param s, the std::string to transform envoy_data from.
- * @return envoy_data the envoy_data transformed from s.
- */
-envoy_data toEnvoyData(const std::string& s);
-
 } // namespace Utility
 } // namespace Data
 } // namespace Envoy
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -4,6 +4,8 @@
 
 #include "common/common/lock_guard.h"
 
+#include "library/common/stats/utility.h"
+
 namespace Envoy {
 
 Engine::Engine(envoy_engine_callbacks callbacks, const char* config, const char* log_level,
@@ -57,6 +59,9 @@ envoy_status_t Engine::run(const std::string config, const std::string log_level
         Envoy::Server::ServerLifecycleNotifier::Stage::PostInit, [this]() -> void {
           server_ = TS_UNCHECKED_READ(main_common_)->server();
           client_scope_ = server_->serverFactoryContext().scope().createScope("pulse.");
+          // StatNameSet is lock-free, the benefit of using it is being able to create StatsName
+          // on-the-fly without risking contention on system with lots of threads.
+          // It also comes with ease of programming.
           stat_name_set_ = client_scope_->symbolTable().makeSet("pulse");
           auto api_listener = server_->listenerManager().apiListener()->get().http();
           ASSERT(api_listener.has_value());
@@ -75,6 +80,7 @@ envoy_status_t Engine::run(const std::string config, const std::string log_level
   // Ensure destructors run on Envoy's main thread.
   postinit_callback_handler_.reset(nullptr);
   client_scope_.reset(nullptr);
+  stat_name_set_.reset();
   TS_UNCHECKED_READ(main_common_).reset(nullptr);
 
   callbacks_.on_exit(callbacks_.context);
@@ -107,26 +113,30 @@ Engine::~Engine() {
   main_thread_.join();
 }
 
-envoy_status_t Engine::recordCounterInc(
-  const std::string& elements, const char * tag_arr[][2], uint64_t tag_arr_len, uint64_t count) {
+envoy_status_t Engine::recordCounterInc(const std::string& elements, envoy_stats_tags tags,
+                                        uint64_t count) {
   if (server_ && client_scope_ && stat_name_set_) {
-    Stats::StatNameTagVector tags = transform(tag_arr, tag_arr_len);
+    Stats::StatNameTagVector tags_vctr =
+        Stats::Utility::transformToStatNameTagVector(tags, stat_name_set_);
     std::string name = Stats::Utility::sanitizeStatsName(elements);
-    server_->dispatcher().post([this, name, tags, count]() -> void {
-      Stats::Utility::counterFromElements(
-        *client_scope_, {Stats::DynamicName(name)}, tags).add(count);
+    server_->dispatcher().post([this, name, tags_vctr, count]() -> void {
+      Stats::Utility::counterFromElements(*client_scope_, {Stats::DynamicName(name)}, tags_vctr)
+          .add(count);
     });
     return ENVOY_SUCCESS;
   }
   return ENVOY_FAILURE;
 }
 
-envoy_status_t Engine::recordGaugeSet(const std::string& elements, uint64_t value) {
-  if (server_ && client_scope_) {
+envoy_status_t Engine::recordGaugeSet(const std::string& elements, envoy_stats_tags tags,
+                                      uint64_t value) {
+  if (server_ && client_scope_ && stat_name_set_) {
+    Stats::StatNameTagVector tags_vctr =
+        Stats::Utility::transformToStatNameTagVector(tags, stat_name_set_);
     std::string name = Stats::Utility::sanitizeStatsName(elements);
-    server_->dispatcher().post([this, name, value]() -> void {
+    server_->dispatcher().post([this, name, tags_vctr, value]() -> void {
       Stats::Utility::gaugeFromElements(*client_scope_, {Stats::DynamicName(name)},
-                                        Stats::Gauge::ImportMode::NeverImport)
+                                        Stats::Gauge::ImportMode::NeverImport, tags_vctr)
           .set(value);
     });
     return ENVOY_SUCCESS;
@@ -134,12 +144,15 @@ envoy_status_t Engine::recordGaugeSet(const std::string& elements, uint64_t valu
   return ENVOY_FAILURE;
 }
 
-envoy_status_t Engine::recordGaugeAdd(const std::string& elements, uint64_t amount) {
-  if (server_ && client_scope_) {
+envoy_status_t Engine::recordGaugeAdd(const std::string& elements, envoy_stats_tags tags,
+                                      uint64_t amount) {
+  if (server_ && client_scope_ && stat_name_set_) {
+    Stats::StatNameTagVector tags_vctr =
+        Stats::Utility::transformToStatNameTagVector(tags, stat_name_set_);
     std::string name = Stats::Utility::sanitizeStatsName(elements);
-    server_->dispatcher().post([this, name, amount]() -> void {
+    server_->dispatcher().post([this, name, tags_vctr, amount]() -> void {
       Stats::Utility::gaugeFromElements(*client_scope_, {Stats::DynamicName(name)},
-                                        Stats::Gauge::ImportMode::NeverImport)
+                                        Stats::Gauge::ImportMode::NeverImport, tags_vctr)
           .add(amount);
     });
     return ENVOY_SUCCESS;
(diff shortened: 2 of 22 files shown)


library/common/engine.cc
```diff
@@ -106,11 +107,14 @@ Engine::~Engine() {
   main_thread_.join();
 }
 
-envoy_status_t Engine::recordCounterInc(const std::string& elements, uint64_t count) {
-  if (server_ && client_scope_) {
+envoy_status_t Engine::recordCounterInc(
```
jingwei99, line 110: Currently in this pr the change is only applied to counter, after a green light of this pr by everyone, I'll expand the implementation to gauge and histogram.

library/common/engine.h
```diff
@@ -38,9 +38,12 @@ class Engine {
   /**
    * Increment a counter with a given string of elements and by the given count.
    * @param elements, joined elements of the timeseries.
+   * @param tag_arr, the array of {key, value} pairs of tags.
+   * @param tag_arr_len, length of tag_arr
    * @param count, amount to add to the counter.
    */
-  envoy_status_t recordCounterInc(const std::string& elements, uint64_t count);
+  envoy_status_t recordCounterInc(const std::string& elements, const char* tag_arr[][2],
```
jingwei99, line 45: use a 3d array to pass the tag key value pairs, it's the only structure compilable for c, cpp and objective-c

library/common/engine.cc
```diff
@@ -57,6 +57,7 @@ envoy_status_t Engine::run(const std::string config, const std::string log_level
         Envoy::Server::ServerLifecycleNotifier::Stage::PostInit, [this]() -> void {
           server_ = TS_UNCHECKED_READ(main_common_)->server();
           client_scope_ = server_->serverFactoryContext().scope().createScope("pulse.");
+          stat_name_set_ = client_scope_->symbolTable().makeSet("pulse");
```
jingwei99, line 60: for context: https://docs.google.com/document/d/1PHvqLEFwiV062RRIMJ2JUd492ZJABOwH-vdK0H-oIRw/edit?disco=AAAALqBrykI

library/common/main_interface.h
```diff
@@ -106,9 +106,12 @@ envoy_status_t set_preferred_network(envoy_network_t network);
  * Increment a counter with the given elements and by the given count.
  * @param engine, the engine that owns the counter.
  * @param elements, the string that identifies the counter to increment.
+ * @param tag_arr, the array of {key, value} pairs of tags.
```
junr03, line 109: I'd create an envoy_pulse_tags (or better name) struct in order to hold both the tags and the length.

library/common/engine.h
```diff
@@ -74,8 +77,10 @@ class Engine {
 
 private:
   envoy_status_t run(std::string config, std::string log_level);
+  Stats::StatNameTagVector transform(const char* tag_arr[][2], uint64_t tag_arr_len);
```
junr03, line 80: Move this to a static utility function similar to what we do with functions that transform `envoy_headers` to `HeaderMap`

junr03: That also allows for unit testing this functionality.

library/common/engine.h
junr03: Can we somehow leverage the bridging we do for headers, which is also a set of key, value pairs?

jingwei99:
>Can we somehow leverage the bridging we do for headers, which is also a set of key, value pairs?

library/common/engine.h
jingwei99: that sounds a good idea

library/common/engine.h
jingwei99: i read through code on headers, it brings extra complexity to leverage the bridge. 

there is [another comment](https://github.com/envoyproxy/envoy-mobile/pull/1271#discussion_r586909893) on creating a struct, that sounds a good idea to me.

junr03: The other comment was my initial thought before I realized the tags were exactly the same data structure as `envoy_headers`. 

I don't think reusing that data structure brings additional complexity. In fact it might reduce some given that we can re-use utility functions. 

My thought is that we can rename `envoy_headers` to `envoy_repeated_string_pair` (or something better); changing the field names to approriate names. Then `typedef envoy_repeated_string_pair envoy_headers` and same with `envoy_pulse_tags`.

The only difference I see here is in memory management, where with `envoy_headers` we have allocated memory that envoy mobile's c/c++ manages, but with your PR it seems that `char *` (with memory owned by the platform) is passed down.

As a reminder, I am out the next two days, but @goaway can advise.

jingwei99: i get the inspiration for re-use, in general it's a good idea.

i looked more closely about the renaming, `envoy_header` carries the `context` and `envoy_release_f` fields ([link](https://github.com/envoyproxy/envoy-mobile/blob/43105fd9b89a7e9d4f6e723a8d47e7ccbaee0251/library/common/types/c_types.h#L85-L86)) from `envoy_data`, they are essential for headers, they are not necessary for tags. I'm aware that one option is split out a standalone `envoy_repeated_string_pair` struct and make another level of embedding for `envoy_header`, i'm not seeing the gain for the split/refactor at this point.

i'd like to hear more about the differences you brought in on memory management, what are the risks you see of having the memory allocated on the envoy mobile c/c++ vs platform? i guess the most obvious answer is memory leak, and i'd like to hear more about the intricacy of it.

junr03: Let me give more context. At a high level, the header data structure and the tags data structure is the same. It is an array of string pairs. 

Envoy Mobile has comprehensive infrastructure (types, utility functions, tests) for bridging an array of string pairs across all supported languages. Of course the more specific application is for headers. My high level point is that we should leverage the same infrastructure for another specific application of an array of string pairs; tags.

It is true that the memory management between tags and headers is different based on the code you have in this PR, but that is part of the reason I am advocating for leveraging the already existing infrastructure: if we have all arrays of string pairs (headers, tags) using the same infrastructure, any improvements we make will be used by both. To give a concrete example: https://github.com/envoyproxy/envoy-mobile/issues/949 will fully bridge the types we are discussing here allowing less copy overhead. If we use the same underlying types for both headers and tags, that improvement will be immediately available to both.

Lastly, thanks to `typedef` we will be able to alias all necessary types such that readability is not hindered, and in fact, improved over passing individual pointers.

library/common/engine_utility.h
```diff
@@ -0,0 +1,13 @@
...
+
+#include "common/stats/utility.h"
+
+#include "library/common/types/c_types.h"
+#include "library/common/types/c_types_utility.h"
+
+namespace Envoy {
+namespace Utility {
+
+Stats::StatNameTagVector transform(envoy_stats_tags tags, Stats::StatNameSetPtr& stat_name_set);
```
junr03, line 11: lets create library/common/stats/utility.h and put this functionality there. under

```c++ 
namespace Envoy {
namespace Stats {
namespace Utility {
```

library/common/engine.cc
```diff
@@ -57,6 +60,7 @@ envoy_status_t Engine::run(const std::string config, const std::string log_level
...
```
junr03, line 63: Can we add an inline comment here that documents the decision making.

library/common/types/c_types_utility.h
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Envoy {
+namespace Utility {
+
+/**
+ * Copy envoy_data into an std::string.
+ * @param s the envoy_data to copy.
+ * @return std::string the string constructed from s.
+ */
+std::string convertToString(envoy_data s);
```
junr03, line 13: Please merge main. This functionality has moved to Buffer::Utility. All in concert to help with #1302

test/common/test_utility.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include "library/common/types/c_types.h"
+
+namespace Envoy {
+namespace TestUtil {
+
+envoy_data make_envoy_data(const std::string& s);
```
junr03, line 8: This is also cleaned up in main.

junr03 (commented):
Just a few comments around moving the stats utility and inline comment for historical context.

library/common/stats/engine_utility.h
```diff
@@ -0,0 +1,17 @@
...
+
+// From upstream Envoy
+#include "common/stats/utility.h"
+
+#include "library/common/types/c_types.h"
+
+namespace Envoy {
+namespace Stats {
+namespace Engine {
+namespace Utility {
```
jingwei99, line 11:
```
namespace Envoy {
namespace Stats {
namespace Utility {
```
yields a conflict from the `@envoy/source/common/stats/utility.h`, therefore added the `Engine` namespace here

junr03: https://github.com/envoyproxy/envoy/pull/15603 will allow us to add the function without adding the Engine namespace. Will update as soon as that lands.

library/common/stats/engine_utility.cc
```diff
@@ -0,0 +1,25 @@
...
+
+#include "library/common/http/header_utility.h"
+#include "library/common/types/c_types.h"
+
+namespace Envoy {
+namespace Stats {
+namespace Engine {
+namespace Utility {
+
+Stats::StatNameTagVector transform(envoy_stats_tags tags, Stats::StatNameSetPtr& stat_name_set) {
```
junr03, line 12: nit: should we have a more specific name? we generally use a more specific name for utilities like this e.g., `copyToString`

test/common/buffer/utility_test.cc
```diff
@@ -47,26 +47,5 @@ TEST(DataConstructorTest, FromCppToC) {
   c_data.release(c_data.context);
 }
 
-TEST(DataConstructorTest, CopyFromCppToC) {
```
junr03, line 50: merge conflict

test/common/test_utility.cc
```diff
@@ -0,0 +1,13 @@
+#include <string>
+
+#include "library/common/types/c_types.h"
+
+namespace Envoy {
+namespace TestUtil {
+
+envoy_data make_envoy_data(const std::string& s) {
```
junr03, line 8: Can we move this to the data utility file?

test/common/stats/engine_utility_test.cc
```diff
@@ -0,0 +1,46 @@
...
+  stat_name_set_ = symbol_table_->makeSet("pulse");
+  envoy_stats_tags tags = make_envoy_stats_tags({{"os", "android"}, {"flavor", "dev"}});
+  Stats::StatNameTagVector tagsVctr = transform(tags, stat_name_set_);
+  ASSERT_EQ(tagsVctr.size(), 2);
+  ASSERT_EQ(symbol_table_->toString(tagsVctr[0].first), "os");
+  ASSERT_EQ(symbol_table_->toString(tagsVctr[0].second), "android");
+  ASSERT_EQ(symbol_table_->toString(tagsVctr[1].first), "flavor");
+  ASSERT_EQ(symbol_table_->toString(tagsVctr[1].second), "dev");
+}
+
```
junr03, line 42: We should have a test for envoy_stats_notags

library/common/stats/engine_utility.cc
goaway: (Also, please say "copy" only if we're creating a copy.)

test/common/test_utility.cc
goaway: And name it in line with other mapping utilities, please. :)

library/common/stats/engine_utility.cc
jingwei99: my assumption was that the parameter(s), the return type and the function name put together makes the intention clear, but let me update it to `transformToStatNameTagVector` - or lmk what would be the name you prefer :)

test/common/test_utility.cc
jingwei99: ya

library/common/data/utility.h
```diff
@@ -43,6 +43,13 @@ envoy_data copyToBridgeData(const Buffer::Instance&);
  */
 std::string copyToString(envoy_data data);
 
+/**
+ * Transform from an std::string to an envoy_data.
+ * @param s, the std::string to transform envoy_data from.
+ * @return envoy_data the envoy_data transformed from s.
+ */
+envoy_data toEnvoyData(const std::string& s);
```
junr03, line 51: I think we can just delete this function and replace uses to `copyToBridgeData`. `const std::string&` are implicitly converted to `absl::string_view`

test/common/stats/utility_test.cc
```diff
@@ -0,0 +1,53 @@
...
+  envoy_stats_tags tags = make_envoy_stats_tags({{"os", "android"}, {"flavor", "dev"}});
+  Stats::StatNameTagVector tags_vctr = transformToStatNameTagVector(tags, stat_name_set_);
+  ASSERT_EQ(tags_vctr.size(), 2);
+  ASSERT_EQ(symbol_table_->toString(tags_vctr[0].first), "os");
+  ASSERT_EQ(symbol_table_->toString(tags_vctr[0].second), "android");
+  ASSERT_EQ(symbol_table_->toString(tags_vctr[1].first), "flavor");
+  ASSERT_EQ(symbol_table_->toString(tags_vctr[1].second), "dev");
+}
+
+TEST(TransformTestNoTags, FromEnvoyStatsTagsToStatNameTagVector) {
```
junr03, line 42: nit: the first string is the test suite name and the second string is the specific test case. So the first string should be the same between tests and the second different.

junr03 (commented):
awesome. two small comments, otherwise lgtm

junr03 (approved):

PR #1272, opened 2021-03-01
core: separate internal dispatch from http path

goaway:
Description: Previously the Http::Dispatcher served the dual purpose of dispatching incoming library calls onto Envoy's main thread event loop and also functioning as the HTTP client. This conflation led to a bunch of incidental complexity, including that the Http::Dispatcher needed to exist as soon as the Engine did, but before Envoy was actually up and running and could provide the interfaces the Http::Dispatcher would need to call to actually accomplish anything.

This change splits the logic into an Http::Client that is largely single-threaded code and contains no synchronization/locking and an Event::ProvisionalDispatcher that wraps Envoy's internal Event::Dispatcher but can persist beyond (before and after) the underlying dispatcher's lifecycle.

The change simplifies the dispatch and threading model of incoming calls, opens the door to resolving some types of races during Engine shutdown, and supports non-singleton integrations.

Risk Level: High
Testing: Local, Unit, Integration, E2E, CI


library/common/main_interface.cc
```diff
@@ -10,30 +10,39 @@
...
 
+static Envoy::Engine* engine(envoy_engine_callbacks callbacks = {0}) {
+  static Envoy::Engine* engine = new Envoy::Engine{callbacks, preferred_network_};
+  return engine; 
+}
+
 envoy_stream_t init_stream(envoy_engine_t) { return current_stream_handle_++; }
 
 envoy_status_t start_stream(envoy_stream_t stream, envoy_http_callbacks callbacks) {
-  if (auto e = engine_.lock()) {
```
rebello95, line 21: Really nice that we can remove these locks now

library/common/main_interface.cc
```diff
@@ -10,30 +10,39 @@
...
-    return e->httpDispatcher().startStream(stream, callbacks);
+  if (auto e = engine()) {
```
rebello95, line 24: When do we expect that this could be nil - is that only during the app shutting down? Wondering because it is a static instance above

goaway: We actually never expect that it will be nullptr with the way the interfaces are used right now by the bridge/platform layers right now. But it's a safety check that mirrors how `lock()` was used previously, and it becomes relevant if someone is calling this interface directly (e.g. with a new platform).

rebello95: got it, makes sense

library/common/main_interface.cc
goaway: :)

library/common/http/client.h
```diff
@@ -99,7 +97,19 @@ class Dispatcher : public Logger::Loggable<Logger::Id::http> {
    */
   envoy_status_t cancelStream(envoy_stream_t stream);
 
-  const DispatcherStats& stats() const;
+  /**
+   * Set listener with thread safety check.
+   * @param api_listener, the listener.
+   */
+  void setListener(ApiListener& api_listener);
```
junr03: I don't think I see where this is getting set?

junr03 (commented):
Thanks for these changes I think this: a) separates concerns more cleanly, b) allocating the engine on the heap and not stopping the event loop/destructing will obviously solve the pesky shutdown crashes that remain.

First pass of comments. I know you have more commits coming, so some of this might have changed already. 

Also can you refresh me on the thread safety/locking conversation we had a couple weeks ago and the decisions we came to? I can't quite remember, but to me it looks like all the locking has been preserved here? Sorry I am a little foggy, the last two weeks where a whirlwind as you know 😅

library/common/http/client.h
```diff
@@ -200,40 +210,35 @@ class Dispatcher : public Logger::Loggable<Logger::Id::http> {
...
-  // stats_ is not currently const because the Http::Dispatcher is constructed before there is
+  Event::ProvisionalDispatcher& dispatcher_;
+  ApiListener* api_listener_ ;
+  // stats_ is not currently const because the Http::Client is constructed before there is
   // access to MainCommon's stats scope.
-  // This can be fixed when queueing logic is moved out of the Http::Dispatcher, as at that point
-  // the Http::Dispatcher could be constructed when access to all objects set in
-  // Http::Dispatcher::ready() is done; obviously this would require "ready()" to not me a member
-  // function of the dispatcher, but rather have a static factory method.
+  // This can be fixed when queueing logic is moved out of the Http::Client, as at that point
```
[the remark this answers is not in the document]
goaway: It's already done.

library/common/http/client.cc
```diff
@@ -0,0 +1,413 @@
...
+
+  streams_.emplace(new_stream_handle, std::move(direct_stream));
+  ENVOY_LOG(debug, "[S{}] start stream", new_stream_handle);
+
+  return ENVOY_SUCCESS;
+}
+
+envoy_status_t Client::sendHeaders(envoy_stream_t stream, envoy_headers headers,
+                                       bool end_stream) {
+  ASSERT(dispatcher_.isThreadSafe());
```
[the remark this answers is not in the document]
goaway: I deliberately added these 1:1 everywhere I removed a call to `post` to validate the context is unchanged.

library/common/event/provisional_dispatcher.h
```diff
@@ -0,0 +1,58 @@
...
+#include "library/common/types/c_types.h"
+
+namespace Envoy {
+namespace Event {
+
+/**
+ * Wrapper around Envoy's Event::Dispatcher that queues functors until the underlying
+ * Event::Dispatcher has started.
+ */
+class ProvisionalDispatcher : public Logger::Loggable<Logger::Id::http> {
```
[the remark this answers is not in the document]
goaway: already updated.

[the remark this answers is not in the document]
goaway: Honestly, I spent forever thinking of a name, and this was the best I came up with. Open to suggestions though.

Provisional governments can become real governments once they ratify a constitution, etc.

But here's some other possibilities:

TransitionalDispatcher ("transition" is problematic because it's overloaded with other programming concepts)
InterimDispatcher (I like this better than "transitional", but it doesn't necessarily carry the connotation of something that can be permanent)
ProxyDispatcher (doesn't really capture any notion of serving a purpose between states; "proxy" is an overloaded term for networking)
QueueingDispatcher (but the real dispatcher queues, too)
DispatcherWrapper (sure, but why?)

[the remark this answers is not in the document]
goaway: And I tried on like a hundred other names for size. ¯\\\_(ツ)\_/¯

library/common/event/provisional_dispatcher.h
```diff
@@ -0,0 +1,58 @@
...
+public:
+  Dispatcher(std::atomic<envoy_network_t>& preferred_network);
+
+  //TODO(goaway): return ENVOY_FAILURE after the underlying dispatcher has exited.
+  /**
+   * Drains all queued callbacks to the real dispatcher once it becomes available. Further posts
```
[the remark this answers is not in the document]
goaway: Actually this comment isn't even accurate anymore.

library/common/event/provisional_dispatcher.h
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Wrapper around Envoy's Event::Dispatcher that queues functors until the underlying
+ * Event::Dispatcher has started.
+ */
+class ProvisionalDispatcher : public Logger::Loggable<Logger::Id::http> {
+public:
+  Dispatcher(std::atomic<envoy_network_t>& preferred_network);
+
+  //TODO(goaway): return ENVOY_FAILURE after the underlying dispatcher has exited.
```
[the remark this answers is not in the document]
goaway: Sure.

goaway:
@junr03 please take a look and comment on approach (still needs fixed tests and cleanup).

library/common/event/provisional_dispatcher.h
```diff
@@ -0,0 +1,67 @@
...
+   * Submits an item for deferred delete. Must be called from context where
+   * ProvisionalDispatcher::isThreadSafe() is true.
+   */
+  void deferredDelete(DeferredDeletablePtr&& to_delete);
+
+  // Used for testing.
+  Thread::ThreadSynchronizer& synchronizer() { return synchronizer_; }
+
+private:
+  // TODO(goaway): This class supports a straightforward case-specific lock-free implementation, but
```
junr03, line 57: Can we document this in an issue. 

Is this referring to what we discussed a couple weeks ago? I thought we were going to go for the lock-free approach in this PR.

library/common/http/client.cc
```diff
@@ -0,0 +1,388 @@
...
+const HttpClientStats& Client::stats() const { return stats_; }
+
+Client::DirectStreamSharedPtr Client::getStream(envoy_stream_t stream) {
+  auto direct_stream_pair_it = streams_.find(stream);
+  // Returning will copy the shared_ptr and increase the ref count.
+  return (direct_stream_pair_it != streams_.end()) ? direct_stream_pair_it->second : nullptr;
+}
+
+void Client::removeStream(envoy_stream_t stream_handle) {
+  RELEASE_ASSERT(dispatcher_.isThreadSafe(),
```
junr03, line 322: Should this be an ASSERT like above, or should those calls be RELEASE_ASSERTS like this?

junr03 (commented):
Awesome clean up. 

re: approach -- it matches the synchronization model we had before just more cleanly thanks to logic separation in two (more) appropriate classes.

 Per my previous comment, I thought we had discussed potentially going lock-free here... but I cannot remember the solution we arrived to. Not necessarily saying we should go there in this PR. I would just like it documented in an issue. In fact, it is probably best to land this, and get it out as it doesn't change the model and has less change for unforeseen edge cases, specially since tests should largely be the same as before just rejiggered for the class separation.

library/common/http/client.h
```diff
@@ -200,40 +204,23 @@ class Dispatcher : public Logger::Loggable<Logger::Id::http> {
...
-  // stats_ is not currently const because the Http::Dispatcher is constructed before there is
-  // access to MainCommon's stats scope.
-  // This can be fixed when queueing logic is moved out of the Http::Dispatcher, as at that point
-  // the Http::Dispatcher could be constructed when access to all objects set in
-  // Http::Dispatcher::ready() is done; obviously this would require "ready()" to not me a member
-  // function of the dispatcher, but rather have a static factory method.
-  // Related issue: https://github.com/lyft/envoy-mobile/issues/720
-  const std::string stats_prefix_;
-  absl::optional<DispatcherStats> stats_ GUARDED_BY(ready_lock_){};
+  ApiListener& api_listener_;
```
junr03, line 221: we could make all these members const

library/common/event/provisional_dispatcher.h
goaway: Sorry for the confusion.

This PR isn't meant to remove all locks. It removes all locks from the HTTP client and simplifies reasoning about concurrency. Somewhat less locking will occur on the golden path.

This comment refers to a specialized, lock-free data structure that could be implemented here, based on the specific usage of the queue (many writers, one reader, one full non-atomic drain pass). It's an optimization that is straightforward enough to realize, but not a priority over the overall restructuring of the code.

library/common/http/client.cc
goaway: Well... pick your poison I guess. I made the ones above `ASSERT` because they're mostly to validate that this large change doesn't have unexpected side effects (and of course we run with regular `ASSERT`s enabled). I don't _expect_ it to have context-changing side effects, so I didn't make them `RELEASE_ASSERTS`. I think you could probably make a reasonable argument either way, and I don't have particularly strong feelings.

Like I mentioned above, I added these `ASSERT`s everywhere I removed a previously-inline `post`.

I guess my slight longer-term bias would be to `ASSERT`, since it has the effect of documenting how the class is expected to be used, but doesn't necessarily incur the overhead of checking all the time for something that really shouldn't be happening.

library/common/engine.h
```diff
@@ -85,6 +85,7 @@ class Engine {
 private:
   envoy_status_t main(std::string config, std::string log_level);
 
+  Event::Dispatcher* event_dispatcher_;
```
junr03, line 88: I don't think this is ever getting set?

library/common/event/provisional_dispatcher.cc
```diff
@@ -7,29 +7,45 @@
 namespace Envoy {
 namespace Event {
 
-void ProvisionalDispatcher::drain() {
+void ProvisionalDispatcher::drain(Event::Dispatcher& event_dispatcher) {
+  // TODO(goaway): Must be called from the real dispatcher's thread, but we can't assert here because the main dispatch loop has not yet started: event_dispatcher_->isThreadSafe() will crash.
```
junr03, line 11: Now I am a little confused. Is the original reason why we delay getting the Event::Dispatcher ref from the server because the object itself is not valid until the post init callback, not because the event loop is not running?

library/common/event/provisional_dispatcher.cc
```diff
@@ -7,29 +7,45 @@
...
 }
 
+bool ProvisionalDispatcher::isThreadSafe() {
+  // Doesn't require locking because if a thread has a stale view of drained_, then by definition
+  // this wasn't a threadsafe call.
+  ENVOY_LOG(trace, "ProvisionalDispatcher::isThreadSafe");
+  return TS_UNCHECKED_READ(drained_) && event_dispatcher_->isThreadSafe();
+}
+
+void ProvisionalDispatcher::deferredDelete(DeferredDeletablePtr&& to_delete) {
```
junr03, line 44: Where is this used?

library/common/engine.h
junr03, line 88:
```suggestion
  Event::Dispatcher* event_dispatcher_{};
```

library/common/event/provisional_dispatcher.h
```diff
@@ -19,14 +19,13 @@ namespace Event {
  */
 class ProvisionalDispatcher : public Logger::Loggable<Logger::Id::main> {
 public:
-  ProvisionalDispatcher(Event::Dispatcher& event_dispatcher)
-      : event_dispatcher_(event_dispatcher) {}
+  ProvisionalDispatcher() {}
```
junr03, line 22:
```suggestion
  ProvisionalDispatcher() = default;
```

junr03 (commented):
Just a few comments. 

I think a format run is missing like you said.

library/common/event/provisional_dispatcher.cc
goaway: in Http::Client

library/common/event/provisional_dispatcher.cc
goaway: Just what I observed - `post` works here, but `isThreadSafe` consistently crashed and the logs indicated the event loop didn't start running until after. That said, provided we fix stats, I think we probably don't need the postinit callback anymore.

library/common/engine.h
goaway: It's returned by the dispatcher() accessor, which is used in main_interface.cc

library/common/event/provisional_dispatcher.h
goaway: I mean technically I can also leave it off altogether for the same effect, right?

library/common/engine.cc
```diff
@@ -112,18 +121,18 @@ Engine::~Engine() {
...
-    if (!main_common_) {
+    if (!event_dispatcher_) {
       cv_.wait(mutex_);
     }
 
-    ASSERT(main_common_);
+    ASSERT(event_dispatcher_);
 
     // Exit the event loop and finish up in Engine::run(...)
     event_dispatcher_->exit();
```
junr03, line 131: the joinable call above should be updated, as it doesn't mean what we thought. And do we want to ultimately allow for exit from the main_thread's context? Per my experiments exiting from that context results in the event loop never actually exiting.

library/common/http/dispatcher.cc
```diff
@@ -0,0 +1,439 @@
+#include "library/common/http/dispatcher.h"
```
junr03: This file should be gone, right?

junr03 (commented):
some comments. I can't quite see from reviewing what is causing the ASAN and TSAN failures. I will take a look at the CI failures in more detail.

library/common/http/dispatcher.cc
goaway: Yes it looks like something got screwed up in the merge.

library/common/engine.cc
```diff
@@ -70,123 +75,130 @@ envoy_status_t Engine::run(const std::string config, const std::string log_level
...
       cv_.wait(mutex_);
     }
 
-    ASSERT(main_common_);
+    ASSERT(event_dispatcher_);
 
     // Exit the event loop and finish up in Engine::run(...)
-    event_dispatcher_->exit();
-  } // _mutex
+    if (std::this_thread::get_id() == main_thread_.get_id()) {
```
junr03, line 131: Didn't we want to clean this to a RELEASE_ASSERT?

library/common/event/BUILD
```diff
@@ -0,0 +1,38 @@
...
+envoy_package()
+
+envoy_cc_library(
+    name = "provisional_dispatcher_lib",
+    srcs = ["provisional_dispatcher.cc"],
+    hdrs = ["provisional_dispatcher.h"],
+    external_deps = ["abseil_optional"],
+    repository = "@envoy",
+    deps = [
+        #        "//library/common/buffer:bridge_fragment_lib",
```
junr03, line 14: clean up

library/common/main_interface.cc
```diff
@@ -121,23 +151,28 @@ envoy_status_t register_platform_api(const char* name, void* api) {
...
-  // still exists and acquire a shared pointer to it - ensuring the engine persists at least for
-  // the duration of the call.
-  engine_ = strong_engine_;
-  return ENVOY_SUCCESS;
+  return ENVOY_FAILURE;
 }
 
-void terminate_engine(envoy_engine_t) { strong_engine_.reset(); }
+void terminate_engine(envoy_engine_t) {
+  auto e = strong_engine_;
```
junr03, line 175: leave comment explaining the local var? Or maybe it is self-explanatory?

test/common/http/client_test.cc
```diff
@@ -5,6 +5,7 @@
 #include "common/stats/isolated_store_impl.h"
```
junr03, line 5: Are unit tests for the provisional dispatcher coming in a subsequent PR?

test/common/mocks/event/mocks.cc
```diff
@@ -0,0 +1,23 @@
...
+using testing::NiceMock;
+using testing::Return;
+using testing::ReturnNew;
+using testing::ReturnPointee;
+using testing::SaveArg;
+
+namespace Envoy {
+namespace Event {
+
+// MockFileEvent::MockFileEvent() = default;
```
junr03, line 19: ?

test/python/integration/test_send_headers.py
```diff
@@ -48,54 +48,55 @@
...
-            True,
-        )
-        stream_complete.wait()
-
-        assert response == {
-            "status": 200,
-            "end_stream": True,
-        }
+         assert True
+#        engine_running = threading.Event()
```
junr03, line 52: can we create an issue and assign to @crockeo?

junr03 (commented):
Alright, a couple clean up things. Then lets merge!

test/python/integration/test_send_headers.py
goaway: He's out for a week, so I'll try to fix on my own first. But I'll add a note.

test/common/http/client_test.cc
goaway: Yes

library/common/main_interface.cc
goaway: Sure.

library/common/event/BUILD
goaway: Good catch, thanks.

library/common/engine.cc
goaway: I'd prefer to leave it like this for now since I want to figure out the alternate path.

test/common/mocks/event/mocks.cc
goaway: default is default, forgot to delete, thanks!

test/common/mocks/event/mocks.cc
```diff
@@ -16,8 +16,5 @@ using testing::SaveArg;
 namespace Envoy {
```
junr03, line 16: is this file necessary?

library/common/engine.cc
junr03: I don't think this panic will log properly. This is the main reason I am pushing for RELEASE_ASSERT. Or even better ENVOY_BUG. tioli

junr03 (approved):

test/common/mocks/event/mocks.cc
goaway: No, but will be adding to it shortly when I add unit tests, so it's a placeholder.

junr03 (approved):

PR #1298, opened 2021-03-11
swift/core: add direct response functionality

rebello95:
Adds the ability to configure Envoy Mobile using typed direct responses. This unlocks a new type of testing for end-consumers of the library by allowing users to actually _have Envoy Mobile connect to itself over a socket and return its own set of direct responses to the original caller_. Effectively, this allows for running full integration tests with the library.

```swift
  let engine = TestEngineBuilder()
    .addDirectResponse(
      .init(
        matcher: RouteMatcher(
          fullPath: "/v1/abc", headers: [
            .init(name: "x-foo", value: "123", mode: .exact),
          ]
        ),
        status: 200, body: "hello world",
        headers: ["x-response-foo": "aaa"]
      )
    )
    .build()
```

One gotcha we ran into when implementing these changes is that Envoy actually performs an initial route resolution before invoking the filter chain. This route is cached, and is subsequently used by the router filter for directing the request. This implementation detail means that if a filter in the request chain mutates the request (i.e., by modifying its headers) in a way that causes it to match a direct response's configuration, the request will not be routed correctly because the cached route will be used by the router.

Envoy provides the ability to clear the route cache for a given stream from within request filters as a way to signal the filter manager that the request's route should be re-resolved by the router. To accomplish this, we decided to add a new `RouteCacheReset` filter to Envoy Mobile's filter chain when one or more direct responses are registered with the configuration. This allows request filters to mutate headers in a way that is guaranteed to properly route the request later on.

Direct responses are being added only to a new `TestEngineBuilder` type, rather than to the production `EngineBuilder` class. There are a couple reasons for this:

1. Direct responses are not something that are typically useful in production environments on mobile clients.
2. To fully support direct response routing (per the route cache discussion above), we would either have to a) force platform-level consumers to manually clear the route cache when they modify headers that are known to affect routing, b) always clear the route cache in production (which would result in unnecessarily re-routing in the vast majority of cases), c) do this automatically whenever filters modify headers (which would still result in unnecessarily re-routing in most cases since direct responses are not really used in production). By only supporting direct responses in the test builder, we can simply add the `RouteCacheReset` filter in the test builder if one or more direct responses are specified.
3. The default stats domain specified does not resolve. For tests, we can provide a valid IP that resolves to Envoy Mobile.

Note that the tests in this PR are split into one-per-target because each integration test requires the Envoy engine to actually start up. Because Envoy Mobile does not yet support multiple concurrent instances, an assertion gets tripped if each test function attempts to create an engine within one `XCTestCase`):

```[2021-03-16 17:35:13.561][1467292][critical][assert] [bazel-out/ios-x86_64-min11.0-applebin_ios-ios_x86_64-dbg/bin/external/envoy/source/common/singleton/_virtual_includes/threadsafe_singleton/common/singleton/threadsafe_singleton.h:64] assert failure: loader_ == nullptr. Details: InjectableSingleton initialized multiple times.```

As a way to test multiple engine support, these should be merged into one test case in https://github.com/envoyproxy/envoy-mobile/issues/332.

This PR also adds the ability to disable stats entirely by omitting a stats domain when starting the envoy engine, since those stats get flushed to an invalid domain as-is without this configuration **and** it is useful to be able to disable stats for testing.

Resolves https://github.com/envoyproxy/envoy-mobile/issues/1292.
Handles the iOS portion of https://github.com/envoyproxy/envoy-mobile/issues/1291, and the Android changes are being tracked there.

Risk Level: Medium, new filter being added to the core chain
Testing: Added in PR
Docs Changes: N/A

### CODE REVIEW

**Comment by rebello95** [comment]
@goaway @junr03 @buildbreaker rebased on top of the dispatcher changes from https://github.com/envoyproxy/envoy-mobile/pull/1272 - ready for review!

File: library/swift/RouteMatcher.swift [hunk 1]
```diff
@@ -0,0 +1,81 @@
+import Foundation
+
+/// Typed representation of a route matcher that may be specified when starting the engine.
+/// https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/
+/// v3/route_components.proto#envoy-v3-api-msg-config-route-v3-routematch
```
buildbreaker, line 5: This is too bad haha

library/common/main_interface.h
```diff
@@ -29,6 +29,27 @@ extern const char* platform_filter_template;
...
+ * filter's invocation on the request path. Typically only used for enabling
+ * direct responses to mutate headers which are then later used for routing.
+ */
+extern const char* route_cache_reset_filter_template;
+
+/**
+ * Template configuration used for creating "fake" remote clusters which enable
+ * local responses to be returned via direct response configurations.
+ */
+extern const char* fake_remote_cluster_template;
```
buildbreaker, line 45: It is a bit of a bummer that we can't hide/exclude this from the main target since we'll need to do the resolution at the configuration file :(

buildbreaker (approved):
I'll let @junr03 or @goaway take a look too but I'm fine with the changes

Thanks @rebello95!

library/swift/RouteMatcher.swift
rebello95: yeahhh

library/common/main_interface.h
rebello95: What do you mean hide it from the main target?

examples/java/hello_world/MainActivity.java
```diff
@@ -60,6 +60,7 @@ protected void onCreate(Bundle savedInstanceState) {
 
     engine = new AndroidEngineBuilder(getApplication())
                  .addLogLevel(LogLevel.DEBUG)
+                 .addStatsDomain("0.0.0.0")
```
goaway, line 63: We don't need to add this. It actually slows example app startup because the DNS resolver blocks on trying to resolve "0.0.0.0" - which is neither a valid IP address, nor a valid domain name - and inevitably fails while the rest of the app waits around.

rebello95: oh right. updated

goaway (dismissed):
This looks good to me.

It would be nice if there were an option to compile out all the test stuff for release. And the config stuff is a growing concern as we've discussed. But I think the value it brings for near-term coverage outweighs those concerns.

Suggest that @junr03 does a pass as well.

junr03:
> It would be nice if there were an option to compile out all the test stuff for release.

@goaway what parts are you referring to? The filters used for testing?

buildbreaker (approved):

Issue #1225, opened 2020-12-19
bridge: unify string accessor


junr03:
After discussions in #1202 I changed the API for string accessors. Android should be updated to match iOS.

PR #1301, opened 2021-03-12
android: parity with iOS string accessor

junr03:
Description: work toward https://github.com/envoyproxy/envoy-mobile/issues/1225. Missing Swift StringAcessor class.
Risk Level: low. Reworks API but functionality is the same. Consumers will need to update their usage.
Testing: builds - functionality remains the same.


library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -162,8 +164,8 @@ open class EngineBuilder(
    *
    * @return this builder.
    */
-  fun addStringAccessor(name: String, accessor: EnvoyStringAccessor): EngineBuilder {
-    this.stringAccessors.put(name, accessor)
+  fun addStringAccessor(name: String, accessor: () -> ByteBuffer): EngineBuilder {
```
rebello95, line 167: Why is this configured to accept a `ByteBuffer` instead of a `String` like iOS has?

junr03: Because we have existing utility functions to bridge ByteBuffer across the JNI.

rebello95: IMO we should keep consistency between the public iOS/Android interfaces here (especially since the `name` parameter on this line also accepts a `String`). It would also help give consumers some guardrails to ensure they don't pass a dataset of something that is not a string. Thoughts?

library/java/io/envoyproxy/envoymobile/engine/JvmStringAccessorContext.java
```diff
@@ -14,5 +12,5 @@
    *
    * @return ByteBuffer, the string retrieved from the platform.
```
rebello95, line 13: This is outdated

rebello95 (approved):
LGTM - @buildbreaker may want to take a final pass

library/java/io/envoyproxy/envoymobile/engine/JvmStringAccessorContext.java
buildbreaker:
```suggestion
   * @return String, the string retrieved from the platform.
```

buildbreaker (dismissed):
One comment. Looks good otherwise!

buildbreaker (approved):
Thanks @junr03 !

PR #1305, opened 2021-03-16
jni utility: create utility function from envoy_data to jbyteArray

junr03:
Description: notice that this direction was missing as a utility function.
Risk Level: low - functionality is the same, just dispatched to function.
Testing: builds - successful example app.


library/common/jni/jni_interface.cc
```diff
@@ -496,15 +474,7 @@ jvm_http_filter_on_resume(const char* method, envoy_headers* headers, envoy_data
...
-    j_in_data = env->NewByteArray(data->length);
-    // TODO: check if copied via isCopy.
-    // TODO: check for NULL.
-    // https://github.com/lyft/envoy-mobile/issues/758
-    void* critical_data = env->GetPrimitiveArrayCritical(j_in_data, nullptr);
-    memcpy(critical_data, data->bytes, data->length);
-    // Here '0' (for which there is no named constant) indicates we want to commit the changes back
-    // to the JVM and free the c array, where applicable.
-    env->ReleasePrimitiveArrayCritical(j_in_data, critical_data, 0);
+    j_in_data = native_data_to_array(env, *data);
```
junr03, line 477: There is no call to `env->DeleteLocalRef(j_in_data)` below. 

Should there be a:

```c
if (data) {
  env->DeleteLocalRef(j_in_data);
}
```

block below?

library/common/jni/jni_utility.cc
```diff
@@ -58,6 +58,19 @@ envoy_data array_to_native_data(JNIEnv* env, jbyteArray j_data) {
   return {data_length, native_bytes, free, native_bytes};
 }
 
+jbyteArray native_data_to_array(JNIEnv* env, envoy_data data) {
+  jbyteArray j_data = env->NewByteArray(data.length);
```
buildbreaker, line 62: I feel this would be confusing since the user of this method is responsible for deleting the local reference created here, right?

library/common/jni/jni_interface.cc
buildbreaker: I think we do need that deletion

library/common/jni/jni_utility.cc
goaway: That's actually the convention we adopt at the bridge layer for our c types. There's no precedent for JNI LocalRefs but it's not much of a departure.

goaway: It should at least be mentioned in the documentation for the function.

goaway:
What I think would make this change really useful is if it actually did the checks mentioned in the comments. Can we make that a part of this PR?

junr03:
> Can we make that a part of this PR?
@goaway of course. I wanted to resolve that now that the logic is centralized. I have added the assertion for checking null. 

re: isCopy. What is the behavior you were looking for? From my read of the docs I am not sure what we would want to be different here? According to the docs `mode has no effect if elems is not a copy of the elements in array.` otherwise we have a choice of:
- copy `critical_data` into `j_data` and free `critical_data` with mode `0`
- copy `critical_data` into `j_data` but _don't_ free `critical_data` with mode `JNI_COMMIT`
- _don't_ copy `critical_data` into `j_data` but free `critical_data` with mode `JNI_ABORT`

So if I am understanding correctly, with `0` we are already doing what we want whether `isCopy` is true or not?

library/common/jni/jni_utility.cc
```diff
@@ -58,6 +60,18 @@ envoy_data array_to_native_data(JNIEnv* env, jbyteArray j_data) {
...
+  RELEASE_ASSERT(j_data != NULL, "unable to allocate memory in jni_utility");
+  // TODO: check if copied via isCopy.
+  void* critical_data = env->GetPrimitiveArrayCritical(j_data, nullptr);
```
goaway, line 67: This is the null check that's important.

library/common/jni/jni_utility.cc
```diff
@@ -58,6 +60,18 @@ envoy_data array_to_native_data(JNIEnv* env, jbyteArray j_data) {
   return {data_length, native_bytes, free, native_bytes};
 }
 
+jbyteArray native_data_to_array(JNIEnv* env, envoy_data data) {
+  jbyteArray j_data = env->NewByteArray(data.length);
+  RELEASE_ASSERT(j_data != NULL, "unable to allocate memory in jni_utility");
```
goaway, line 65: This can be removed - IIRC it's not how you'd check for this.

goaway:
`mode` does have an effect in certain cases where the a copy hasn't been made, but reading this closer, I agree it's not relevant here. At present, we can safely use `0` and ignore `isCopy`, since we're simply moving a buffer to the JVM once.

In the future, there are potential optimizations we can do if we check `isCopy` (e.g. if we know it's a copy, depending on the size of the buffer, there may be a more efficient way to pass it to the JVM), but you can ignore the TODO for now.

goaway (approved):
Thank you for the cleanup!

PR #1314, opened 2021-03-22
python: reproducible-ish python build

crockeo:
Description: Adds steps to the build to install dependencies necessary to run certain parts of the python library (e.g. gevent_utils).

Risk Level: LOW

Testing: enables `//test/python/integration:test_send_headers` to be run by everyone else

Docs Changes: N/A
Release Notes: N/A


crockeo:
A couple notes on this PR meant to spark discussion:

1. It looks like envoy uses an old version of [rules_python](https://github.com/bazelbuild/rules_python) that includes a deprecated method of sourcing Python dependencies. According to the [recent updates](https://github.com/bazelbuild/rules_python#recent-updates) header, as of 2020-10-15 we're supposed to use `pip_install` not `pip_import`. This will likely break the build when envoy updates + envoy-mobile bumps the submodule SHA.

1. The [pip_import docs](https://github.com/bazelbuild/rules_python/tree/c8c79aae9aa1b61d199ad03d5fe06338febd0774#importing-pip-dependencies) call out the fact that pip runs at WORKSPACE-time, and so it doesn't know what python toolchain you're using unless you explicitly provide it with a path. This probably makes the build non-hermetic, but that may not be a big deal.

1. The docs also don't explicate whether or not `pip_install` runs under a virtualenv. If it doesn't, developers will have the contents of `requirements.txt` installed system-wide, which could cause version conflicts with other installed packages.

1. There's [no documented way](https://github.com/bazelbuild/rules_python/tree/c8c79aae9aa1b61d199ad03d5fe06338febd0774#consuming-pip-dependencies) to recursively depend on a python library, so you have to include dependencies' dependencies in the relevant python target, so you have to manually enumerate transitive dependencies, see: `//library/python/gevent_util:gevent_util`.

I'm starting to wonder if it makes sense to put the python code inside this repo, or in one managed by a python-specific build tool like setuptools.

crockeo:
Looks like it's failing because it doesn't have a Python runtime to use 🙁

crockeo:
closing this for now since dependency management will change in rules_python 0.2.0 and won't be needed until upstreaming envoy_requests (which will include its own dependency specification)

PR #1316, opened 2021-03-23
string accessor: move back to ByteBuffer for Android

junr03:
Description: in general java String usage accross the JNI layer is precarious. Reverting usage back to ByteBuffer.
Risk Level: low 
Testing: tested with modified dispatcher code in example app that accessed the installed string accessor in MainActivity.kt

### DIFF
--- a/examples/kotlin/hello_world/MainActivity.kt
+++ b/examples/kotlin/hello_world/MainActivity.kt
@@ -18,6 +18,7 @@ import io.envoyproxy.envoymobile.shared.Failure
 import io.envoyproxy.envoymobile.shared.ResponseRecyclerViewAdapter
 import io.envoyproxy.envoymobile.shared.Success
 import java.io.IOException
+import java.nio.ByteBuffer
 import java.util.concurrent.Executors
 import java.util.concurrent.TimeUnit
 
@@ -48,7 +49,7 @@ class MainActivity : Activity() {
       .addPlatformFilter { DemoFilter() }
       .addPlatformFilter { BufferDemoFilter() }
       .addPlatformFilter { AsyncDemoFilter() }
-      .addStringAccessor("demo-accessor", { "PlatformString" })
+      .addStringAccessor("demo-accessor", { ByteBuffer.wrap("PlatformString".toByteArray(Charsets.UTF_8)) })
       .addNativeFilter("envoy.filters.http.buffer", "{\"@type\":\"type.googleapis.com/envoy.extensions.filters.http.buffer.v3.Buffer\",\"max_request_bytes\":5242880}")
       .setOnEngineRunning { Log.d("MainActivity", "Envoy async internal setup completed") }
       .build()
--- a/examples/kotlin/hello_world/MainActivity.kt
+++ b/examples/kotlin/hello_world/MainActivity.kt
@@ -49,7 +49,7 @@ class MainActivity : Activity() {
       .addPlatformFilter { DemoFilter() }
       .addPlatformFilter { BufferDemoFilter() }
       .addPlatformFilter { AsyncDemoFilter() }
-      .addStringAccessor("demo-accessor", { ByteBuffer.wrap("PlatformString".toByteArray(Charsets.UTF_8)) })
+      .addStringAccessor("demo-accessor", { "PlatformString" })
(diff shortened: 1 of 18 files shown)


examples/kotlin/hello_world/MainActivity.kt
```diff
@@ -48,7 +49,7 @@ class MainActivity : Activity() {
       .addPlatformFilter { DemoFilter() }
       .addPlatformFilter { BufferDemoFilter() }
       .addPlatformFilter { AsyncDemoFilter() }
-      .addStringAccessor("demo-accessor", { "PlatformString" })
+      .addStringAccessor("demo-accessor", { ByteBuffer.wrap("PlatformString".toByteArray(Charsets.UTF_8)) })
```
buildbreaker, line 52: What's fun is that this doesn't catch the existing issue... :(

buildbreaker (dismissed):
Do we have an issue for patching up the test scenarios here @junr03 ?

If not, I can create one for us

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -162,7 +163,7 @@ open class EngineBuilder(
    *
    * @return this builder.
    */
-  fun addStringAccessor(name: String, accessor: () -> String): EngineBuilder {
+  fun addStringAccessor(name: String, accessor: () -> ByteBuffer): EngineBuilder {
```
goaway, line 166: IMO the public interface should still return a String. We can map it to ByteBuffer behind the scenes.

goaway: Also why not use the ByteArray directly? We can't predict the lifetime of its usage, so we have to do a copy anyways - might as well just do it via the critical array copy that we already have.

buildbreaker: @goaway I feel like we can revisit our ByteBuffer versus ByteArray conversation again. Though I do agree if we can have a String interface but internally, it just transforms to a ByteBuffer, that would be nice

goaway: My point though is that we're converting to a byte array and then wrapping with a byte buffer when we already support passing byte array directly.

examples/kotlin/hello_world/MainActivity.kt
junr03: Yeah, this only tests the install path, not the usage path.

When I originally implemented the change I modified the Http::Dispatcher to access the string, and obviously also tested using our internal filters. I failed to do that when I modified to String before :/, which is why I didn't catch the issues I am fixing now.

We could create a filter in Envoy Mobile that tests the example accessors. I will do a subsequent PR for that. We will just not install the filter in our private build.

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
junr03: @buildbreaker @goaway updated. Pretty pleased with how it came out. Easy and ergonomic for the user, and consistent with prior JNI experience under the hood. Lmk what you think.

library/java/io/envoyproxy/envoymobile/engine/types/EnvoyStringAccessor.java
```diff
@@ -1,5 +1,7 @@
 package io.envoyproxy.envoymobile.engine.types;
 
+import java.nio.ByteBuffer;
```
goaway, line 3: Looks like this can be dropped. :)

goaway (approved):
Looks good - see minor comment above. Thanks for updating!

examples/kotlin/hello_world/MainActivity.kt
buildbreaker: https://github.com/envoyproxy/envoy-mobile/issues/1317!

junr03: @buildbreaker already have the test code written. Will add in a subsequent PR. Thanks for creating the issue.

buildbreaker (approved):

PR #1318, opened 2021-03-25
python: add in a headers iterator

crockeo:
Description: Adds an iterator to the `Headers` classes to iterate through headers. Note that this violates C++ design by iterating through the header key, not the header key/value pair. I chose this design because `pybind11` was giving me invalid memory access errors while trying to iterate through pairs. That said, I'm happy to make the key/value -> key downgrade inside of a shim if preferred.

Risk Level: Low

Testing: N/A

Docs Changes: N/A

Release Notes: N/A

### DIFF
--- a/library/cc/headers.cc
+++ b/library/cc/headers.cc
@@ -3,6 +3,14 @@
 namespace Envoy {
 namespace Platform {
 
+Headers::const_iterator Headers::begin() const {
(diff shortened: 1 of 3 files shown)


goaway (approved):

goaway:
Seems fine, though would be nice to understand the memory access errors.

PR #1335, opened 2021-03-31
ci: log app run if job failed or cancelled

junr03:


goaway (approved):

junr03:
@goaway FYI it seems that the assert we were discussing _is_ sometimes (hit on CI)[https://github.com/envoyproxy/envoy-mobile/pull/1335/checks?check_run_id=2240562793#step:10:89]. I have noticed the CI jobs flake sometimes. 

Interestingly internal reporting has **never** seen this crash out on the wild in real clients.

PR #1336, opened 2021-03-31
python: draft envoy_requests library

crockeo:
_NOTE: this PR is not meant to be merged as-is, but rather provide an over all view of work i've been doing here. I want to break this off into smaller, more manageable PRs and merge them one at a time._

## Context

I'm publishing this absolute chonker of a PR to put my work on an envoy-backed [requests](https://pypi.org/project/requests/)-like library up to a public location. This is super, super, super WIP but I wanted to get feedback on it now that it's more or less coherent.

The primary interface seeks to replicate requests. That said, it is an explicit non-goal to make this a drop-in replacement for requests. Instead, I just want this to be familiar to a Python engineer who has worked with requests before.

## Arch Comments

Explaining some thoughts about this so far...

### Main API

This API is presented in two places:

* `envoy_requests.asyncio` provides an `async`/`await` based interface.
* `envoy_requests.gevent` provides a "synchronous" interface, which is transparently async via gevent.

Both of these are backed by a single request implementation inside of `envoy_requests.common.core`. This file lets you build a stream for your async platform (will touch on that later) and then send a request with that stream. This was done to mitigate long-term maintenance burden of having two different front-ends.

### Why Is the Core Structured So Weird?

A couple of reasons:

1. We need to provide an `Executor` that lets our event loops know about envoy-mobile callbacks.

1. `async`/`await` requires that you use its special syntax when interacting with asynchronous functions. The control of when a stream completes must be put into the function calling `send_request`, or else we won't block appropriately.

1. We also have to keep the `Stream` alive throughout the duration of the request, or else it gets cancelled.

There are probably better solutions (please suggest them!) but this is the best I could come up with.

### Executors

By default envoy-mobile calls callbacks on their own thread. Async libraries either don't like multithreading (e.g. `gevent`) or have a single event loop per thread (e.g. `asyncio`). When work comes in from a non-main thread, we have to move that work to the main thread's event loop. `Executor`s do exactly this!

Their interface contains exactly one function, `wrap`, which translates a callback from a raw "call me on the callback thread" to a "send this work over to the event hub." 

### Testing

Testing creates a simple HTTP server using the Python stdlib. This server just sends a JSON response containing the relevant parts of the request. At first glance these tests are a little too close to a tautology test, but since we're not mocking anything they verify that we actually do send what we intend to send.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,6 +10,7 @@ repos:
     - id: check-merge-conflict
     - id: check-yaml
       exclude: '.clang-format'
+      exclude: 'test_envoy_config_template.yaml'
     - id: detect-private-key
     - id: mixed-line-ending
 - repo: https://github.com/codespell-project/codespell
--- a/library/python/envoy_requests/__init__.py
+++ b/library/python/envoy_requests/__init__.py
@@ -1,0 +1,7 @@
+# TODO: don't automatically prefer gevent
+# maybe use threading by default
+# let the user explicitly choose between
+# envoy_requests.gevent and envoy_requests.asyncio
+# for async libraries
+from envoy_requests.gevent import *
+from envoy_requests.response import Response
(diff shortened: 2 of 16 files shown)


crockeo:
I just realized I've accidentally made this PR against the upstream repo instead of my own. Sorry about that 🙂

PR #1338, opened 2021-04-01
engine: set concurrency to 0

goaway:
Description: Until now, we've relied upon the implicit behavior that Envoy will start at most one worker without a connection-based listener. This explicitly sets concurrency to 0 to prevent multiple workers from starting.
Risk Level: Moderate
Testing: Local and CI


rebello95 (approved):

goaway:
Confirmed spurious failing checks locally.

PR #1339, opened 2021-04-01
python: allow Engine to take a custom config template

crockeo:
Description: Pipes through a `config_template` argument that can be used to override the default config template provided OOTB. This will be useful for 

Risk Level: Low

Testing: Migrated `test_send_headers.py` to use config template + be more like Swift test.

Docs Changes: N/A

Release Notes: N/A


goaway (approved):
LGTM

PR #1343, opened 2021-04-02
ci: cache build artifact across runs

goaway:
Description: Updates build job to use cached library artifact when run on the same SHA, allowing retries for connectivity failure or flakiness to run much faster.
Risk Level: Low
Testing: CI


buildbreaker (approved):
Cool! Thanks @goaway

PR #1344, opened 2021-04-02
python: wheel building for envoy_engine

crockeo:
Description: Bumps `rules_python` to version `0.2.0` and uses that new version to build a wheel for `envoy_engine.so`.  This is a test run for building an envoy_requests wheel down the road when it's upstreamed.

Risk Level: Low

Testing: `build //library/python:envoy_engine_whl`

Docs Changes: N/A

Release Notes: N/A


keith (approved):
i don't see any other way you could get a string with the python version other than doing this repository rule. Ideally you could pass a file or something as the tag to the py_wheel rule, but that could be a future improvement

goaway (approved):

PR #1345, opened 2021-04-02
ci: split iOS CI and add artifact caching

goaway:
Description: Splits iOS CI workflows into test and build, and updates build job to use cached library artifact when run on the same SHA.
Risk Level: Low
Testing: CI


buildbreaker (approved):

goaway:
Merging despite the failure on main so CI reqs can be updated.

PR #1346, opened 2021-04-02
main: fix argc

junr03:
Description: #1338 changed argv, but kept argc the same. This PR updates command line args to no longer depend on a hardcoded value.
Risk Level: low
Testing: local build to make sure all command line args were being respected.


goaway (approved):
Good catch, thanks!

rebello95:
Does the original PR result in any functional breakage?

junr03:
@rebello95, yeah, changing log levels would not take effect.

rebello95:
Got it. What is the default used by Envoy (what did we end up shipping to production if our overrides do not take effect?)?

junr03:
info should be default

PR #1348, opened 2021-04-02
logging: add callback-based logging delegate

junr03:


junr03:
broke up in #1354 #1355 and #1356

PR #1353, opened 2021-04-05
examples: add baseline debug logging to example apps

goaway:


goaway:
I've had to redo (and remember what to specify and how to import) this countless times when debugging locally, and there doesn't seem to be a good reason not to mainline it. It demonstrates how to set logging in the examples, and is useful when using the example for local testing.

buildbreaker (approved):

PR #1354, opened 2021-04-06
logging: add LambdaDelegate

junr03:
Description: adds a new SinkDelegate called LambdaDelegate which uses user-provided callbacks to drive `log` functionality.
Risk Level: low
Testing: new unit tests and integration test in main_interface_test.

### DIFF
--- a/library/common/common/lambda_logger_delegate.cc
+++ b/library/common/common/lambda_logger_delegate.cc
@@ -2,8 +2,6 @@
 
 #include <iostream>
 
-#include "common/common/lock_guard.h"
-
 namespace Envoy {
 namespace Logger {
 
@@ -16,12 +14,10 @@ LambdaDelegate::LambdaDelegate(LogCb log_callback, FlushCb flush_callback,
 LambdaDelegate::~LambdaDelegate() { restoreDelegate(); }
 
 void LambdaDelegate::log(absl::string_view msg) {
-  Thread::LockGuard lock(mutex_);
   log_callback_(msg);
 }
 
 void LambdaDelegate::flush() {
-  Thread::LockGuard lock(mutex_);
   flush_callback_();
 }
 
--- a/library/common/common/lambda_logger_delegate.cc
+++ b/library/common/common/lambda_logger_delegate.cc
@@ -13,13 +13,9 @@ LambdaDelegate::LambdaDelegate(LogCb log_callback, FlushCb flush_callback,
 
 LambdaDelegate::~LambdaDelegate() { restoreDelegate(); }
 
-void LambdaDelegate::log(absl::string_view msg) {
-  log_callback_(msg);
-}
+void LambdaDelegate::log(absl::string_view msg) { log_callback_(msg); }
 
-void LambdaDelegate::flush() {
-  flush_callback_();
-}
+void LambdaDelegate::flush() { flush_callback_(); }
 
 } // namespace Logger
 } // namespace Envoy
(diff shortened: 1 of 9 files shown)


test/common/common/lambda_logger_delegate_test.cc
```diff
@@ -0,0 +1,38 @@
...
+namespace Envoy {
+namespace Logger {
+
+TEST(LambdaDelegate, LogCb) {
+  uint i{0};
+  std::string expected_msg = "Hello LambdaDelegate";
+
+  LambdaDelegate delegate = LambdaDelegate(
+      [&i, &expected_msg](absl::string_view msg) -> void {
+        EXPECT_THAT(msg, HasSubstr(expected_msg));
```
buildbreaker, line 17: This is a nit. I generally prefer the expectations to be at the bottom of the test so maybe instead of incrementing an uint, could we capture the value passed and assert this captured value instead of asserting the uint?

junr03: I like that, updated.

buildbreaker (dismissed):
Thanks @junr03 !

library/common/common/lambda_logger_delegate.cc
```diff
@@ -0,0 +1,29 @@
...
+LambdaDelegate::LambdaDelegate(LogCb log_callback, FlushCb flush_callback,
+                               DelegatingLogSinkSharedPtr log_sink)
+    : SinkDelegate(log_sink), log_callback_(log_callback), flush_callback_(flush_callback) {
+  setDelegate();
+}
+
+LambdaDelegate::~LambdaDelegate() { restoreDelegate(); }
+
+void LambdaDelegate::log(absl::string_view msg) {
+  Thread::LockGuard lock(mutex_);
```
goaway, line 19: What's the purpose for this mutex?

library/common/common/lambda_logger_delegate.h
```diff
@@ -0,0 +1,34 @@
...
+#include "common/common/thread.h"
+
+#include "absl/strings/string_view.h"
+
+namespace Envoy {
+namespace Logger {
+
+class LambdaDelegate : public SinkDelegate {
+public:
+  using LogCb = std::function<void(absl::string_view)>;
```
goaway, line 15: Having a lambda at this layer isn't necessarily problematic, but it's a layer of indirection we haven't introduced for other callback cases (engine callbacks, request callbacks, PBF, etc.). Why not leverage the bridge struct directly here?

junr03: I originally wrote this class in upstream and then realized I could write it in Envoy Mobile, so the extra lambdas are an artifact of that. Good catch, will update.

library/common/common/lambda_logger_delegate.cc
junr03: If the delegate is used in a multi-threaded environment log calls could come from multiple threads. This is also an artifact of having written this class originally in Envoy. I can remove this given that Envoy Mobile is (for now) single threaded.

library/common/common/lambda_logger_delegate.h
junr03: I will wait for #1355 to land.

junr03: updated

goaway (approved):
Looks good to me. Had a nit, but decided it was too insignificant to spend time on. Let's roll.

PR #1355, opened 2021-04-06
logging: expose envoy_logging_callbacks

junr03:
Description: exposes new functionality in the bridge layer that will leverage #1354.
Risk Level: low
Testing: updated tests

### DIFF
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -9,10 +9,9 @@
 
 namespace Envoy {
 
-Engine::Engine(envoy_engine_callbacks callbacks, envoy_logging_callbacks logging_callbacks,
-               const char* config, const char* log_level,
-               std::atomic<envoy_network_t>& preferred_network)
-    : callbacks_(callbacks), logging_callbacks_(logging_callbacks) {
+Engine::Engine(envoy_engine_callbacks callbacks, envoy_logger, const char* config,
+               const char* log_level, std::atomic<envoy_network_t>& preferred_network)
+    : callbacks_(callbacks) {
   // Ensure static factory registration occurs on time.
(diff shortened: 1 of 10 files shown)


library/common/engine.cc
```diff
@@ -4,13 +4,15 @@
...
 #include "library/common/stats/utility.h"
 
 namespace Envoy {
 
-Engine::Engine(envoy_engine_callbacks callbacks, const char* config, const char* log_level,
+Engine::Engine(envoy_engine_callbacks callbacks, envoy_logging_callbacks logging_callbacks,
+               const char* config, const char* log_level,
                std::atomic<envoy_network_t>& preferred_network)
-    : callbacks_(callbacks) {
+    : callbacks_(callbacks), logging_callbacks_(logging_callbacks) {
```
goaway, line 15: logger_, imo

goaway:
A few high level comments:

I’d advocate for renaming types representing “logger_callbacks” to just be “logger”, and “on_log” calls to just be “log”.

I don't really see the purpose for exposing a flush callback. Since it's ultimately up to the logger implementation to decide what it does with the logs, I don't think it really cares when Envoy decides to call flush in its internal implementations.

Not I'm a little concerned about the increase in things appearing in the Engine constructor. Another option would be to have this be runtime configurable. We don't need it right now, but we should maybe also think about the case where people also might want to add more than one logger.

junr03:
> I don't really see the purpose for exposing a flush callback. Since it's ultimately up to the logger implementation to decide what it does with the logs, I don't think it really cares when Envoy decides to call flush in its internal implementations.

Discussed this. In addition, given that afaict Envoy only flushes on exit, we have decided to not expose that part of the API for now.

goaway (approved):
This looks good, thanks for making the changes!

PR #1356, opened 2021-04-06
logging: ios API for callback logging

junr03:
Description: iOS API for supporting #1355 
Risk Level: low
Testing: example apps. Pending e2e tests when all PRs land.

### DIFF
--- a/library/common/common/lambda_logger_delegate.cc
+++ b/library/common/common/lambda_logger_delegate.cc
@@ -12,7 +12,10 @@ LambdaDelegate::LambdaDelegate(envoy_logger logger, DelegatingLogSinkSharedPtr l
   setDelegate();
 }
 
-LambdaDelegate::~LambdaDelegate() { restoreDelegate(); }
+LambdaDelegate::~LambdaDelegate() {
+  restoreDelegate();
+  logger_.release(logger_.context);
+}
 
 void LambdaDelegate::log(absl::string_view msg) {
   logger_.log(Data::Utility::copyToBridgeData(msg), logger_.context);
--- a/library/common/types/c_types.h
+++ b/library/common/types/c_types.h
@@ -293,7 +293,7 @@ typedef void (*envoy_on_engine_running_f)(void* context);
  * @param context, contains the necessary state to carry out platform-specific dispatch and
  * execution.
  */
-typedef void (*envoy_log_f)(envoy_data data, void* context);
+typedef void (*envoy_logger_log_f)(envoy_data data, void* context);
 
 /**
  * Called when Envoy is done with the logger.
@@ -336,7 +336,7 @@ typedef struct {
  * Interface for logging.
  */
 typedef struct {
-  envoy_log_f log;
+  envoy_logger_log_f log;
   envoy_logger_release_f release;
   // Context passed through to callbacks to provide dispatch and execution state.
   void* context;
--- a/library/common/types/c_types.h
+++ b/library/common/types/c_types.h
@@ -295,6 +295,14 @@ typedef void (*envoy_on_engine_running_f)(void* context);
  */
 typedef void (*envoy_log_f)(envoy_data data, void* context);
 
+/**
+ * Called when Envoy is done with the logger.
+ *
+ * @param context, contains the necessary state to carry out platform-specific dispatch and
+ * execution.
+ */
+typedef void (*envoy_logger_release_f)(void* context);
+
 #ifdef __cplusplus
 } // function pointers
 #endif
@@ -329,6 +337,7 @@ typedef struct {
  */
 typedef struct {
   envoy_log_f log;
+  envoy_logger_release_f release;
   // Context passed through to callbacks to provide dispatch and execution state.
   void* context;
 } envoy_logger;
--- a/library/objective-c/EnvoyEngine.h
+++ b/library/objective-c/EnvoyEngine.h
@@ -313,14 +313,12 @@ extern const int kEnvoyFailure;
  @param onEngineRunning Closure called when the engine finishes its async startup and begins
  running.
  @param onEngineLog Closure to be called when the engine logs.
- @param onEngineFlush Closure to be called when the engine's logger flushes.
  @return A status indicating if the action was successful.
  */
 - (int)runWithConfig:(EnvoyConfiguration *)config
             logLevel:(NSString *)logLevel
      onEngineRunning:(nullable void (^)())onEngineRunning
-         onEngineLog:(nullable void (^)(NSString *))onEngineLog
-       onEngineFlush:(nullable void (^)())onEngineFlush;
+         onEngineLog:(nullable void (^)(NSString *))onEngineLog;
 
 /**
  Run the Envoy engine with the provided yaml string and log level.
@@ -331,15 +329,14 @@ extern const int kEnvoyFailure;
  @param onEngineRunning Closure called when the engine finishes its async startup and begins
  running.
  @param onEngineLog Closure to be called when the engine logs.
- @param onEngineFlush Closure to be called when the engine's logger flushes.
  @return A status indicating if the action was successful.
  */
 - (int)runWithTemplate:(NSString *)yaml
                 config:(EnvoyConfiguration *)config
               logLevel:(NSString *)logLevel
        onEngineRunning:(nullable void (^)())onEngineRunning
-           onEngineLog:(nullable void (^)(NSString *))onEngineLog
-         onEngineFlush:(nullable void (^)())onEngineFlush;
+           onEngineLog:(nullable void (^)(NSString *))onEngineLog;
+;
 
 /**
  Opens a new HTTP stream attached to this engine.
@@ -409,7 +406,6 @@ extern const int kEnvoyFailure;
 
 @property (nonatomic, copy, nullable) void (^onEngineRunning)();
 @property (nonatomic, copy, nullable) void (^onEngineLog)(NSString *);
-@property (nonatomic, copy, nullable) void (^onEngineFlush)();
 
 @end
 
--- a/library/objective-c/EnvoyEngine.h
+++ b/library/objective-c/EnvoyEngine.h
@@ -312,13 +312,13 @@ extern const int kEnvoyFailure;
  @param logLevel The log level to use when starting Envoy.
  @param onEngineRunning Closure called when the engine finishes its async startup and begins
  running.
- @param onEngineLog Closure to be called when the engine logs.
+ @param Logger Logging interface.
  @return A status indicating if the action was successful.
  */
 - (int)runWithConfig:(EnvoyConfiguration *)config
             logLevel:(NSString *)logLevel
      onEngineRunning:(nullable void (^)())onEngineRunning
-         onEngineLog:(nullable void (^)(NSString *))onEngineLog;
+              Logger:(nullable void (^)(NSString *))Logger;
 
 /**
  Run the Envoy engine with the provided yaml string and log level.
@@ -328,14 +328,14 @@ extern const int kEnvoyFailure;
  @param logLevel The log level to use when starting Envoy.
  @param onEngineRunning Closure called when the engine finishes its async startup and begins
  running.
- @param onEngineLog Closure to be called when the engine logs.
+ @param Logger Logging interface.
  @return A status indicating if the action was successful.
  */
 - (int)runWithTemplate:(NSString *)yaml
                 config:(EnvoyConfiguration *)config
               logLevel:(NSString *)logLevel
        onEngineRunning:(nullable void (^)())onEngineRunning
-           onEngineLog:(nullable void (^)(NSString *))onEngineLog;
+                Logger:(nullable void (^)(NSString *))Logger;
 ;
 
 /**
@@ -405,7 +405,7 @@ extern const int kEnvoyFailure;
 @interface EnvoyEngineImpl : NSObject <EnvoyEngine>
 
 @property (nonatomic, copy, nullable) void (^onEngineRunning)();
-@property (nonatomic, copy, nullable) void (^onEngineLog)(NSString *);
+@property (nonatomic, copy, nullable) void (^Logger)(NSString *);
 
 @end
 
--- a/library/objective-c/EnvoyEngine.h
+++ b/library/objective-c/EnvoyEngine.h
@@ -399,13 +399,28 @@ extern const int kEnvoyFailure;
 
 @end
 
+#pragma mark - EnvoyLogger
+
+// Logging interface.
+@interface EnvoyLogger : NSObject
+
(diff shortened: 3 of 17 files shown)


library/objective-c/EnvoyEngine.h
```diff
@@ -312,11 +312,15 @@ extern const int kEnvoyFailure;
...
  running.
+ @param onEngineLog Closure to be called when the engine logs.
+ @param onEngineFlush Closure to be called when the engine's logger flushes.
  @return A status indicating if the action was successful.
  */
 - (int)runWithConfig:(EnvoyConfiguration *)config
             logLevel:(NSString *)logLevel
-     onEngineRunning:(nullable void (^)())onEngineRunning;
+     onEngineRunning:(nullable void (^)())onEngineRunning
+         onEngineLog:(nullable void (^)(NSString *))onEngineLog
```
goaway, line 322: nit: Let's call just call this a Logger

library/swift/EngineBuilder.swift
```diff
@@ -169,6 +171,28 @@ public final class EngineBuilder: NSObject {
...
+  /// - parameter closure: The closure to be called.
+  ///
+  /// - returns: This builder.
+  @discardableResult
+  public func setonEngineLog(closure: @escaping (String) -> Void) -> EngineBuilder {
+    self.onEngineLog = closure
+    return self
+  }
+
+  /// Set a closure to be called when the engine's logger flushes.
```
goaway, line 185: What's the purpose of this callback if we're (presumably) logging each statement immediately?

library/objective-c/EnvoyEngine.h
```diff
@@ -326,12 +330,16 @@ extern const int kEnvoyFailure;
  @param logLevel The log level to use when starting Envoy.
  @param onEngineRunning Closure called when the engine finishes its async startup and begins
  running.
+ @param onEngineLog Closure to be called when the engine logs.
+ @param onEngineFlush Closure to be called when the engine's logger flushes.
```
rebello95, line 334: I think it could be useful to provide more information here as to when it flushes (and what the difference between a log and flush is in this context)

library/swift/EngineBuilder.swift
```diff
@@ -169,6 +171,28 @@ public final class EngineBuilder: NSObject {
     return self
   }
 
+  /// Set a closure to be called when the engine's logger logs.
+  ///
+  /// - parameter closure: The closure to be called.
+  ///
+  /// - returns: This builder.
+  @discardableResult
+  public func setonEngineLog(closure: @escaping (String) -> Void) -> EngineBuilder {
```
rebello95, line 180:
```suggestion
  public func setOnEngineLog(closure: @escaping (String) -> Void) -> EngineBuilder {
```

library/swift/EngineBuilder.swift
rebello95: I tend to agree with @goaway here unless there is additional context as to why we would want to observe this specific behavior

library/swift/EngineBuilder.swift
```diff
@@ -169,6 +171,28 @@ public final class EngineBuilder: NSObject {
...
+    return self
+  }
+
+  /// Set a closure to be called when the engine's logger flushes.
+  ///
+  /// - parameter closure: The closure to be called.
+  ///
+  /// - returns: This builder.
+  @discardableResult
+  public func setonEngineFlush(closure: @escaping () -> Void) -> EngineBuilder {
```
rebello95, line 191:
```suggestion
  public func setOnEngineFlush(closure: @escaping () -> Void) -> EngineBuilder {
```
If we decide to keep this

rebello95 (commented):
Looks good overall with a couple minor comments/questions

library/objective-c/EnvoyEngine.h
junr03: Given that this is just a block I want to keep the name for consistency. However, I do share the sentiment you wrote in for both the bridge, and at this level too:

> Not I'm a little concerned about the increase in things appearing in the Engine constructor. Another option would be to have this be runtime configurable. We don't need it right now, but we should maybe also think about the case where people also might want to add more than one logger. [source](https://github.com/envoyproxy/envoy-mobile/pull/1355#issuecomment-814389318)

And would like better separation of concerns (engine callbacks, config, logging, pulse api) at this level. I can carve this out here to begin that process and not add more to the pile that already exists, or we can land as is, and then start a series of PRs to clean up.

library/swift/EngineBuilder.swift
junr03: Discussed in https://github.com/envoyproxy/envoy-mobile/pull/1355. Not going to expose flushing.

rebello95 (approved):

library/objective-c/EnvoyEngine.h
goaway: Everything's in the same header due to the objective-c bridge header clang issue - if that's what you mean by separation of concerns.

library/swift/EngineBuilder.swift
```diff
@@ -169,6 +170,17 @@ public final class EngineBuilder: NSObject {
     return self
   }
 
+  /// Set a closure to be called when the engine's logger logs.
+  ///
+  /// - parameter closure: The closure to be called.
+  ///
+  /// - returns: This builder.
+  @discardableResult
+  public func setOnEngineLog(closure: @escaping (String) -> Void) -> EngineBuilder {
```
goaway, line 179: Mapping in Objective-C I can kind of take it or leave it with my above comment. But here, I really do object to this naming scheme. I feel pretty strongly we should expose the public interface as "Logger" or some variant thereof.

library/objective-c/EnvoyEngine.h
```diff
@@ -312,11 +312,13 @@ extern const int kEnvoyFailure;
...
  @param onEngineRunning Closure called when the engine finishes its async startup and begins
  running.
+ @param Logger Logging interface.
  @return A status indicating if the action was successful.
  */
 - (int)runWithConfig:(EnvoyConfiguration *)config
             logLevel:(NSString *)logLevel
-     onEngineRunning:(nullable void (^)())onEngineRunning;
+     onEngineRunning:(nullable void (^)())onEngineRunning
+              Logger:(nullable void (^)(NSString *))Logger;
```
rebello95, line 321: These property names should be lowercased (`logger`)

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -14,11 +14,20 @@ static void ios_on_engine_running(void *context) {
   if (engineImpl.onEngineRunning) {
     engineImpl.onEngineRunning();
   }
+  CFBridgingRetain(engineImpl);
```
rebello95, line 17: Would like @goaway to double check this and `ios_on_exit`

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -384,23 +398,34 @@ - (int)runWithTemplate:(NSString *)yaml
...
                                                (__bridge void *)(self)};
-    // TODO(junr03): wire up once https://github.com/envoyproxy/envoy-mobile/pull/1356 lands.
-    envoy_logger logger = {NULL, NULL};
-    return (int)run_engine(_engineHandle, native_callbacks, logger, configYAML.UTF8String,
+
+    envoy_logger native_logger = {ios_on_log, (__bridge void *)(self)};
+
+    if (!self.logger) {
+      native_logger.log = NULL;
+    }
```
rebello95, line 426: Can you elaborate on why this is necessary?

test/swift/integration/SetLoggerTest.swift
```diff
@@ -0,0 +1,69 @@
...
+        }
+      }
+      .setOnEngineRunning {
+        engineExpectation.fulfill()
+      }
+      .build()
+      .streamClient()
+
+    XCTAssertEqual(XCTWaiter.wait(for: [engineExpectation], timeout: 1), .completed)
+    XCTAssertEqual(XCTWaiter.wait(for: [loggingExpectation], timeout: 1), .completed)
```
rebello95, line 67: you can combine these into one array and add `enforceOrder: true`

rebello95 (approved):

library/objective-c/EnvoyEngine.h
```diff
@@ -394,12 +399,28 @@ extern const int kEnvoyFailure;
...
+// Logging interface.
+@interface EnvoyLogger : NSObject
+
+@property (nonatomic, copy, nullable) void (^log)(NSString *);
+
+/**
+ Create a new instance of the logger.
+ */
+- (instancetype)initWithLog:(nullable void (^)(NSString *))log;
+;
```
rebello95, line 413:
```suggestion
```

library/objective-c/EnvoyEngine.h
```diff
@@ -394,12 +399,28 @@ extern const int kEnvoyFailure;
 
 @end
 
+#pragma mark - EnvoyLogger
+
+// Logging interface.
+@interface EnvoyLogger : NSObject
+
+@property (nonatomic, copy, nullable) void (^log)(NSString *);
```
rebello95, line 407: Would it make sense to have the `EnvoyLogger` itself be `nil` instead of having an empty object with a `nil` closure?

library/objective-c/EnvoyEngine.h
```diff
@@ -394,12 +399,28 @@ extern const int kEnvoyFailure;
...
+
+@end
+
 #pragma mark - EnvoyEngineImpl
 
 // Concrete implementation of the `EnvoyEngine` interface.
 @interface EnvoyEngineImpl : NSObject <EnvoyEngine>
 
 @property (nonatomic, copy, nullable) void (^onEngineRunning)();
+@property (nonatomic, strong, nullable) EnvoyLogger *logger;
```
rebello95, line 423: Hmm, both are nullable. We probably only need either the logger or its closure to be nullable, no?

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -16,11 +16,15 @@ static void ios_on_engine_running(void *context) {
   }
 }
 
-static void ios_on_exit(void *context) {
-  // Currently nothing needs to happen in iOS on exit. Just log.
-  NSLog(@"[Envoy] library is exiting");
+static void ios_on_exit(void *context) { NSLog(@"[Envoy] library is exiting"); }
```
rebello95, line 19: Would it make sense to also log this to the logger now?

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -384,23 +393,35 @@ - (int)runWithTemplate:(NSString *)yaml
...
   @try {
     envoy_engine_callbacks native_callbacks = {ios_on_engine_running, ios_on_exit,
                                                (__bridge void *)(self)};
-    // TODO(junr03): wire up once https://github.com/envoyproxy/envoy-mobile/pull/1356 lands.
-    envoy_logger logger = {NULL, NULL};
-    return (int)run_engine(_engineHandle, native_callbacks, logger, configYAML.UTF8String,
+
+    envoy_logger native_logger = {NULL, NULL, NULL};
+    if (logger) {
+      EnvoyLogger *Objclogger = [[EnvoyLogger alloc] initWithLog:logger];
```
rebello95, line 418:
```suggestion
      EnvoyLogger *objcLogger = [[EnvoyLogger alloc] initWithLog:logger];
```

rebello95 (approved):
Looks good overall with a couple comments

library/common/types/c_types.h
```diff
@@ -329,6 +337,7 @@ typedef struct {
  */
 typedef struct {
   envoy_log_f log;
```
goaway, line 339: For parity, might want to consider calling this envoy_logger_log_f, but not a big deal.

library/objective-c/EnvoyEngineImpl.m
goaway, line 418: `initWithLogClosure` or `initWithLogCallback` maybe.

goaway (approved):
Nice! Looks good to me.

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -415,10 +415,10 @@ - (int)runWithConfigYAML:(NSString *)configYAML
 
     envoy_logger native_logger = {NULL, NULL, NULL};
     if (logger) {
-      EnvoyLogger *Objclogger = [[EnvoyLogger alloc] initWithLog:logger];
+      EnvoyLogger *objclogger = [[EnvoyLogger alloc] initWithLogClosure:logger];
```
rebello95, line 418:
```suggestion
      EnvoyLogger *objcLogger = [[EnvoyLogger alloc] initWithLogClosure:logger];
```

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -415,10 +415,10 @@ - (int)runWithConfigYAML:(NSString *)configYAML
...
       native_logger.log = ios_on_log;
       native_logger.release = ios_on_logger_release;
-      native_logger.context = CFBridgingRetain(Objclogger);
+      native_logger.context = CFBridgingRetain(objclogger);
```
rebello95, line 421:
```suggestion
      native_logger.context = CFBridgingRetain(objcLogger);
```

rebello95 (approved):

library/common/common/lambda_logger_delegate.cc
```diff
@@ -12,7 +12,10 @@ LambdaDelegate::LambdaDelegate(envoy_logger logger, DelegatingLogSinkSharedPtr l
   setDelegate();
 }
 
-LambdaDelegate::~LambdaDelegate() { restoreDelegate(); }
+LambdaDelegate::~LambdaDelegate() {
+  restoreDelegate();
+  logger_.release(logger_.context);
```
goaway, line 17: Excellent. Exactly the right place to put this.

library/objective-c/EnvoyEngineImpl.m
goaway: The logger isn't implicitly available here, nor is it even a given that we have a logger. I think this is fine as a placeholder, but we could also simply remove it.

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -438,10 +425,6 @@ - (int)runWithConfigYAML:(NSString *)configYAML
     //   native_logging_callbacks.on_log = NULL;
     // }
```
rebello95: Does this need to be updated?

goaway (approved):

rebello95 (approved):

library/common/common/lambda_logger_delegate.cc
junr03: Yep, very clean this way.

PR #1357, opened 2021-04-07
Cronvoy test

carloseltuerto:

Description: Testing why envoylib.so JNI does not allow the JVM to shutdown properly
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


library/common/jni/android_jni_interface.cc
```diff
@@ -1,5 +1,9 @@
+#ifndef UNIT_TEST
```
buildbreaker, line 1: I understand our directory organization is a little flat here but the `android_jni_interface.cc` is used in conjunction with `jni_interface.cc` which has only the java specific dependencies. We likely would not need to use this file for the local unit testing.

library/common/jni/BUILD
```diff
@@ -17,27 +17,64 @@ envoy_cc_library(
     ],
     repository = "@envoy",
     deps = [
+        "//:jni_headers",
         "//library/common/types:c_types_lib",
         "@envoy//source/common/common:assert_lib",
     ],
 )
 
+# Produces two binaries: one for Android, and one for tests.
```
buildbreaker, line 26: I was able tohandle this bazel gymnastics for you now!

https://github.com/envoyproxy/envoy-mobile/blob/main/library/common/jni/BUILD#L47-L71 Is exactly what you're looking for. You can rebase your fork and branch from envoy-mobile (if you're having trouble, I can help you too)

library/java/org/cronvoy/CronvoyEngine.java
```diff
@@ -0,0 +1,229 @@
...
+    // of background processes) are put in a cgroup that is allowed to
+    // consume up to 5% of CPU - these worker threads spend the vast
+    // majority of their time waiting on I/O, so making them contend with
+    // background applications for a slice of CPU doesn't make much sense.
+    // We want to hurry up and get idle.
+    final int threadPriority =
+        builder.threadPriority(THREAD_PRIORITY_BACKGROUND + THREAD_PRIORITY_MORE_FAVORABLE);
+    this.mUserAgent = builder.getUserAgent();
+    EngineBuilder engineBuilder = new EngineBuilder();
+    engine = engineBuilder.build();
```
buildbreaker, line 56:
```suggestion
    engine = engineBuilder.addLogLevel(LogLevel.TRACE).build();
```

Let's use trace logs and see what's going on internally

buildbreaker:
It might be helpful for me to go over some of our design decisions around the Android/Java side. Our goal is to have a fully functional java networking library (so we can exercise code paths across the java/jni/cpp layers). To do that, we decided to shift and isolate the android logic into a separate ("companion") file which the Android targets end up using (while the test targets can use the Java layer).

Is there a similar distinction or separation within Cronet?

carloseltuerto:
From an integration test perspective, it is preferable if all pieces are
present. The JNI Cronet library only works for Android, so the Android
context must work.

In other words, all the Java/Kotlin classes prefixed with "Android" must
work too in the context of an integration test involving Cronet.

For sure, we can avoid those ugly "ifdef" in .cc and .h. Then this implies
that the BUILD dependencies must point to different BUILD targets when
compiling for a Cronet Integration test. How to do that with elegance is
something I would need to further explore.

library/common/jni/android_jni_interface.cc
carloseltuerto: Unfortunately, as far as I can tell, android_jni_interface.cc is needed. The Cronet API for Android deals with Android Objects such as `android.content.Context` See [CronetEngine](https://chromium.googlesource.com/chromium/src/+/main/components/cronet/android/api/src/org/chromium/net/CronetEngine.java#75) . The Cronvoy will use the Android library provided by Envoy Mobile, which properly deals with said Android objects. True, this Android object will be ignored, since the libenvoy_jni.so library will be crafted to avoid any Android references for tests. Also, I tested my theory by using the Android Kotlin library (AndroidEngineBuilder). Surprisingly, the underlying AndroidNetworkMonitor class did not complain.

library/common/jni/BUILD
carloseltuerto: This I fear is not helping moving forward with Cronvoy. I can't use the "java_jni_lib" BUILD target to construct the Cronvoy tests. The Cronvoy tests depends on the libenvoy_jni.so BUILD target.

buildbreaker:
Gotcha. Let me get you a split up android jni file where the c-ares dependency is disjointed from the vanilla android (for Cronet).

I haven't gotten a chance to figure out your issue yesterday @carloseltuerto but I'll be focusing in on that today

test/java/org/cronvoy/CronvoyEngineTest.java
```diff
@@ -0,0 +1,20 @@
...
+@RunWith(RobolectricTestRunner.class)
+public class CronvoyEngineTest {
+
+  private final Context appContext = ApplicationProvider.getApplicationContext();
+
+  @Test
+  public void test() {
+    System.out.println("Creating CronvoyEngine...");
+    new CronvoyEngine(new CronvoyEngineBuilderImpl(appContext));
+    System.out.println("CronvoyEngine created.");
```
buildbreaker, line 18:
```suggestion
    System.out.println("CronvoyEngine created.");
        while(true) {
      try {
        Thread.sleep(1000);
      } catch (InterruptedException e) {
        throw new RuntimeException();
      }
    }
```

buildbreaker:
From slack:
a few things:
1. android robolectric testing also uses the same convention for the so file as the vanilla unit tests for osx -- this means we need to have both an .so and .jnilib artifacts for unit test execution (luckily i think this is going to be very similar to the vanilla java)
2. the unit test you wrote starts and immediately kills envoy ~immediately. this is likely the reason for your error. i tried with a thread sleep and i do get more trace logs but it fails for a different reason (investigating now)
[external/envoy/source/server/server.cc:810] exiting
yaml-cpp: error at line 1, column 1: end of map flow not found
3. i will need to split out the android jni file to separate out the c-ares from pure android

buildbreaker:
new error deals with the JRE:
```#
#
#  SIGSEGV (0xb) at pc=0x000000010cf18ba9, pid=15047, tid=0x0000000000006603
#
# JRE version: Java(TM) SE Runtime Environment (8.0_241-b07) (build 1.8.0_241-b07)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.241-b07 mixed mode bsd-amd64 compressed oops)
# Problematic frame:
# V  [libjvm.dylib+0x318ba9]
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /private/var/tmp/_bazel_achiu/720efad541ec2d99e2beac8a75575ed8/execroot/envoy_mobile/bazel-out/darwin-fastbuild/bin/test/java/org/cronvoy/cronvoy_test.runfiles/envoy_mobile/hs_err_pid15047.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
#
prio: 2 tag: [Envoy] text: jvm_on_engine_running
================================================================================```

Seems like this is the line that crashes (?):https://github.com/envoyproxy/envoy-mobile/blob/66b060965c5512fd07397278c871271062aaa957/library/common/jni/jni_interface.cc#L39

File: library/common/jni/jni_utility.cc [hunk 5]
```diff
@@ -24,8 +24,8 @@ JNIEnv* get_env() {
   jint result = static_jvm->GetEnv(reinterpret_cast<void**>(&local_env), JNI_VERSION);
   if (result == JNI_EDETACHED) {
     // Note: the only thread that should need to be attached is Envoy's engine std::thread.
-    JavaVMAttachArgs args = {JNI_VERSION, "EnvoyMain", NULL};
-    result = static_jvm->AttachCurrentThread(&local_env, &args);
+    JavaVMAttachArgs args = {JNI_VERSION, const_cast<char*>("EnvoyMain"), NULL};
+    result = static_jvm->AttachCurrentThread(reinterpret_cast<void**>(&local_env), &args);
```
buildbreaker, line 28: Conflict resolution:
```suggestion
    JavaVMAttachArgs args = {JNI_VERSION, "EnvoyMain", NULL};
    result = attach_jvm(static_jvm, &local_env, &args);
```

library/common/jni/jni_interface.cc
```diff
@@ -1,5 +1,8 @@
+#ifndef UNIT_TEST
 #include <android/log.h>
-#include <ares.h>
+#else
+#include "library/common/jni/android_log.h"
+#endif
```
buildbreaker, line 5:
```suggestion
#include <ares.h>
```

PR #1361, opened 2021-04-08
python: clean up python library in preparation for envoy_requests

crockeo:
Description: There were some vestigial pieces of code under the python directory from earlier prototyping. This PR: 

- Removes `e2e_test`, which is just the old version of the test that now lives under `//test/python/integration:test_send_headers`. 
- Removes `gevent_util`, which is the old version of integrating with gevent that's now built into envoy_requests, coming soon™
- Namespaces `envoy_engine.so` dependencies under `library/python/envoy_engine` in preparation of upstreaming envoy_requests
- Cleans up a dead dependency in `test_send_headers` on `gevent_utils`

Risk Level: Low

Testing: N/A

Docs Changes: N/A

Release Notes: N/A


library/python/BUILD
```diff
@@ -1,33 +1,20 @@
...
-)
-
 pybind_library(
     name = "envoy_engine_lib",
     srcs = [
-        "bytes_view.cc",
-        "engine_builder_shim.cc",
-        "stream_prototype_shim.cc",
-        "stream_shim.cc",
+        "envoy_engine/bytes_view.cc",
```
goaway, line 8: Generally we try to avoid prefixing targets/classes/files with "envoy" unless they represent a top-level interface. Any reason not to simply go with `engine/` here?

crockeo: >unless they represent a top-level interface

This might fall into that category, depending on its exact definition. These files collectively define the `envoy_engine.so` target, which could be imported without using `envoy_requests`. That said, there's no reason folder has to be named `envoy_engine`.

goaway: Sorry, to clarify, let's make the folder be `engine` unless there's a reason not to go this way?

crockeo:
i actually don't 100% think this is necessary. just some code twiddling that's not really worth the effort right now. we can do organization stuff when i'm more confident in the end state of code structure

PR #1366, opened 2021-04-09
python: upstream envoy_requests

crockeo:
**Description:** This PR upstreams envoy_requests, a pythonic wrapper around the raw Python platform implementation, to envoy-requests.

**Risk Level:** Low

**Testing:** `//library/python:runtest_unit` and `//library/python:runtest_integration`, but these are broken until https://github.com/envoyproxy/envoy-mobile/pull/1339 is merged. 

**Docs Changes:** N/A

**Release Notes:** N/A


crockeo:
This is frankly too large to have a suitable review at +1100 LOC. I'm going to chat with @goaway about review strategies--either breaking this up or having an interactive review session.

crockeo:
Aaaaand some known concerns, to be fixed before or after merge:

- Will break when https://github.com/envoyproxy/envoy-mobile/pull/1344 is merged, because it will change how you source dependencies
- Need to separate dev dependencies (for testing) from prod dependencies (for building + publishing)
- Need to move away from a separate python-defined HTTP server for testing, and instead use envoy HTTP filters

crockeo:
One thing I forgot to cover inside of our 1:1, @goaway, is details about how the testing works. Tl;dr:

* Create a `runner.py` file that just executes pytest
* Have a top-level `conftest.py`, which registers a `http.server` from inside of Python. This exposes an echo server that responds with a structured JSON response documenting the request it received.
* Have pytest execute either `test/python/unit/...` or `test/python/integration/...` 

A couple of things I'd like to change about this:

* `http.server` doesn't do SSL OOTB, so I made an envoy config w/o SSL. I don't like registering a custom config template so I want to remove this
* Move away from an echo server via `http.server` to an echo server based on an HTTP filter

crockeo:
That last merge should give this PR CI checks as well 🙂

test/python/test_envoy_config_template.yaml
```diff
@@ -0,0 +1,220 @@
+# this is just @envoy-mobile//library/common:config_template.c
+# except it doesn't use TLS
```
goaway, line 2: Cleartext support is now present in the main config. Can we get rid of this and use that instead?

goaway (changes requested):
This is great!

goaway (approved):
Thank you @crockeo!

PR #1376, opened 2021-04-12
android: fix permissions with network state access

buildbreaker:
Adding runtime permissions check for [Android](https://developer.android.com/training/permissions/requesting). Without this, when the user decides to decline the permission, the app crashes with the following stacktrace:
```java.lang.SecurityException: Package android does not belong to 10404
        at android.os.Parcel.createExceptionOrNull(Parcel.java:2385)
        at android.os.Parcel.createException(Parcel.java:2369)
        at android.os.Parcel.readException(Parcel.java:2352)
        at android.os.Parcel.readException(Parcel.java:2294)
        at android.net.IConnectivityManager$Stub$Proxy.listenForNetwork(IConnectivityManager.java:4703)
        at android.net.ConnectivityManager.sendRequestForNetwork(ConnectivityManager.java:4191)
        at android.net.ConnectivityManager.registerNetworkCallback(ConnectivityManager.java:4594)
        at android.net.ConnectivityManager.registerNetworkCallback(ConnectivityManager.java:4564)
        at io.envoyproxy.envoymobile.engine.AndroidNetworkMonitor.<init>(AndroidNetworkMonitor:72)
        at io.envoyproxy.envoymobile.engine.AndroidNetworkMonitor.load(AndroidNetworkMonitor:42)
        at io.envoyproxy.envoymobile.engine.AndroidEngineImpl.<init>(AndroidEngineImpl:17)
```

The API 21 min SDK is fine as that is Android L and the target SDK is what we have specified in the `WORKSPACE`.

Description: android: fix permissions with network state access 
Risk Level: low
Testing: ci/manual
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/library/java/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
+++ b/library/java/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
@@ -23,7 +23,8 @@ import androidx.core.content.ContextCompat;
  */
 @TargetApi(Build.VERSION_CODES.LOLLIPOP)
 public class AndroidNetworkMonitor extends BroadcastReceiver {
-  private static final String PERMISSION_DENIED_STATS_ELEMENT = "android_access_network_state_denied";
+  private static final String PERMISSION_DENIED_STATS_ELEMENT =
+      "android_access_network_state_denied";
   private static final int ENVOY_NET_GENERIC = 0;
   private static final int ENVOY_NET_WWAN = 1;
   private static final int ENVOY_NET_WLAN = 2;
--- a/library/java/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
+++ b/library/java/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
@@ -55,7 +55,7 @@ public class AndroidNetworkMonitor extends BroadcastReceiver {
         ContextCompat.checkSelfPermission(context, Manifest.permission.ACCESS_NETWORK_STATE);
     if (permission == PackageManager.PERMISSION_DENIED) {
       try {
-        envoyEngine.recordCounterInc(PERMISSION_DENIED_STATS_ELEMENT, Collections.emptyMap(),1);
+        envoyEngine.recordCounterInc(PERMISSION_DENIED_STATS_ELEMENT, Collections.emptyMap(), 1);
       } catch (Throwable t) {
         // no-op if this errors out and return
       }
--- a/library/java/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
+++ b/library/java/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
@@ -16,6 +16,8 @@ import android.net.NetworkRequest;
 import android.os.Build;
 import androidx.core.content.ContextCompat;
 
+import java.util.Collections;
+
 /**
  * This class makes use of some deprecated APIs, but it's only current purpose is to attempt to
  * distill some notion of a preferred network from the OS, upon which we can assume new sockets will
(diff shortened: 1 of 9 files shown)


library/kotlin/io/envoyproxy/envoymobile/EnvoyManifest.xml
```diff
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="io.envoyproxy.envoymobile">
     <uses-sdk
-            android:minSdkVersion="9"
-            android:targetSdkVersion="27" />
+            android:minSdkVersion="21"
```
junr03, line 5: is this the min version where this exists?

buildbreaker: The min version here is correct. We do want API 21 (Android L) as our min

buildbreaker:
@junr03 This is actually quite strange because I'm trying to create the crash but I'm not able to with our Lyft app (same with the examples)

What I was able to test is that the network monitoring is still working as expected but there is a permissions check which will allow us to avoid the problematic line: https://github.com/envoyproxy/envoy-mobile/blob/main/library/java/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java#L72

buildbreaker:
Let's add a stat for when the permissions is denied

library/java/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
```diff
@@ -39,11 +43,22 @@ public static void load(Context context) {
...
     }
   }
 
-  private AndroidNetworkMonitor(Context context) {
+  private AndroidNetworkMonitor(Context context, EnvoyEngine envoyEngine) {
+    int permission =
+        ContextCompat.checkSelfPermission(context, Manifest.permission.ACCESS_NETWORK_STATE);
+    if (permission == PackageManager.PERMISSION_DENIED) {
+      try {
+        envoyEngine.recordCounterInc(PERMISSION_DENIED_STATS_ELEMENT, 1);
```
buildbreaker, line 55: From a conversation with @goaway and @junr03, the suggestion was to add a stat for tracking when a user has their permissions denied. This is a bit stranger of a usage area because this class is actually used within the various `*EngineImpl` so we're kind of forced to use the primitive `EnvoyEngine` which exposes the primitive stats methods.

cc @jingwei99

library/java/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
```diff
@@ -20,6 +23,7 @@
  */
 @TargetApi(Build.VERSION_CODES.LOLLIPOP)
 public class AndroidNetworkMonitor extends BroadcastReceiver {
+  private static final String PERMISSION_DENIED_STATS_ELEMENT = "android_access_network_state_denied";
```
buildbreaker, line 26: Happy to change the naming here

buildbreaker:
@junr03 @jingwei99 could you ptal?

jingwei99:
lgtm for what it does

jingwei99 (dismissed):

library/java/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
```diff
@@ -20,6 +23,8 @@
  */
 @TargetApi(Build.VERSION_CODES.LOLLIPOP)
 public class AndroidNetworkMonitor extends BroadcastReceiver {
+  private static final String PERMISSION_DENIED_STATS_ELEMENT =
+      "android_access_network_state_denied";
```
goaway, line 27: Maybe something along the lines of

```suggestion
      "android_permissions.network_state_denied";
```

would let us use this namespace to track other instances we (might) care about in the future.

goaway:
This is our first use case for an externally-generated, but internally-used stat. I think this is fine, but we can think more about a general solution is we encounter more cases.

goaway (dismissed):
Approved, but see suggestion above.

goaway (approved):

goaway (approved):

goaway (approved):

PR #1379, opened 2021-04-14
python: keep stream and engine alive for callbacks

crockeo:
**Description:** Modifies the ownership model of objects so that callbacks keep their relevant parent alive (`EngineCallbacks` -> `Engine`; `StreamCallbacks` -> `Stream`) instead of the other way around.

This PR assumes that one of `StreamCallbacks` terminal callbacks will be called exactly once (`on_complete`, `on_error`, `on_cancel`) and that `on_exit` will be called exactly once. If these are called more than once there will be a double free, if they are not called once we'll leak memory. This is obviously not good 🙂 

**Risk Level:** Low

**Testing:** `bazel test //test/python/integration:test_lifetimes`

**Docs Changes:** N/A

**Release Notes:** N/A


crockeo:
Marking as ready for review because I want feedback on this strategy. I'm kind of fumbling around in the dark, so I'm sure there's a better way.

library/cc/engine_builder.cc
```diff
@@ -89,8 +89,23 @@ EngineSharedPtr EngineBuilder::build() {
...
+  auto engine_ptr = EngineSharedPtr(engine);
+
+  auto callbacks = this->callbacks_.release();
+  callbacks->parent = engine_ptr;
+
+  envoy_logger null_logger{
+      .log = nullptr,
+      .release = envoy_noop_release,
+      .context = nullptr,
+  };
```
goaway, line 103: Would need to verify, but by convention, it should be safe to simply calloc this (or any other envoy mobile struct) if you're not going to use it.

goaway (approved):
Nice work, lgtm.

library/cc/engine.cc
```diff
@@ -6,23 +6,26 @@
...
-  this->stream_client_ = std::make_shared<StreamClient>(this->engine_);
-  this->pulse_client_ = std::make_shared<PulseClient>();
+Engine::Engine(envoy_engine_t engine) : engine_(engine), terminated_(false) {}
+
+// we lazily construct the stream and pulse clients
+// because they either require or will require a weak ptr
+// which can't be provided from inside of the constructor
+// because of how std::enable_shared_from_this works
+StreamClientSharedPtr Engine::streamClient() {
+  if (!this->stream_client_) {
```
goaway, line 16: This looks like it's not going to be thread-safe. If we need to use lazy clients, we should maybe just create a new one every time someone asks for one. It's not an especially heavyweight object anyways.

library/cc/engine.cc
```diff
@@ -6,23 +6,26 @@
...
+    this->stream_client_ = std::make_shared<StreamClient>(this->weak_from_this());
```
goaway, line 17: I think StreamClient should probably take a shared reference to the Engine rather than a weak one. We made stats weak so that all the stats objects that might be distributed around a program wouldn't all be retaining the engine... but then again, it's up for debate whether they should be, as well.

goaway: (Also FWIW, it _does_ effectively take a shared ref in the other platforms.)

crockeo: This is a weak ptr to avoid a cycle (`Engine` -> `StreamClient` -> `Engine`), but I don't think this cycle needs to exist anymore. It's just a byproduct of the original design of this PR where `Stream`s kept `Engine`s alive.

I can address this comment along with the earlier comment at the same time, I think, by just removing the `EngineSharedPtr` references throughout `StreamClient` and `StreamPrototype` and just move them back to `envoy_engine_t`

goaway (approved):

PR #1380, opened 2021-04-14
platform bridge: additional coverage and fix leaks in edge cases onResume

junr03:
Description: there was missing coverage in the platform bridge filter that highlighted some memory leaks.
Risk Level: low
Testing: additional testing that produced asan failures, and then fixed.


library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -487,6 +487,18 @@ void PlatformBridgeFilter::FilterBase::onResume() {
...
+      free(result.pending_headers);
+    }
+    if (result.pending_data != nullptr) {
+      result.pending_data->release(result.pending_data->context);
+      free(result.pending_data);
+    }
+    if (result.pending_trailers != nullptr) {
+      release_envoy_headers(*result.pending_trailers);
+      free(result.pending_trailers);
+    }
```
goaway, line 501: I don't really think these releases should be here - if anything, these should be assertions that the pending entities _are_ nullptr.

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -519,6 +531,7 @@ void PlatformBridgeFilter::FilterBase::onResume() {
     free(result.pending_trailers);
   } else if (result.pending_trailers) {
     addTrailers(*result.pending_trailers);
+    free(result.pending_trailers);
```
goaway, line 534: Good catch here, though.

library/common/extensions/filters/http/platform_bridge/filter.cc
junr03: @goaway, sure. I wasn't sure what the intended behavior was. If the pending entities should be null in this branch, I can switch to assertions.

goaway (approved):

PR #1388, opened 2021-04-19
ci: always run coverage

goaway:
Description: This needs to run since CI requires it to pass; as a future optimization we can short-circuit if there's no changes to these directories.
Risk Level: Low
Testing: CI


buildbreaker (approved):
@carloseltuerto This should fix the issue you were observing!

PR #1401, opened 2021-04-21
Add mandatory "lib" prefix to test JNI library

carloseltuerto:
This makes the test test/java/integration:android_engine_start_test to pass when running on Linux

Risk Level: none
Testing: N/A
Docs Changes: N/A
Release Notes: N/A


test/java/integration/BUILD
```diff
@@ -6,8 +6,7 @@ envoy_mobile_android_test(
         "AndroidEnvoyEngineStartUpTest.java",
     ],
     native_deps = [
-        "//library/common/jni:ndk_envoy_lib.so",
-        "//library/common/jni:ndk_envoy_lib.jnilib",
```
buildbreaker, line 10: I think we'll need to keep this for mac os x :(

carloseltuerto: Oops - resurrected. Thanks.

buildbreaker:
@carloseltuerto If this works, let's go with it!

buildbreaker (dismissed):

carloseltuerto:
> @carloseltuerto If this works, let's go with it!

Well maybe this won't work for IOS.

Stripping the first 3 char with bazel is probably wrong.

carloseltuerto:
> > @carloseltuerto If this works, let's go with it!
> 
> Well maybe this won't work for IOS.
> 
> Stripping the first 3 char with bazel is probably wrong.

If running this test with a Mac works without stripping the first 3 char, maybe the solution is to try to load the JNI with the provided name first, and if this does not work, then try with the first 3 char of the name removed.

buildbreaker:
> If running this test with a Mac works without stripping the first 3 char, maybe the solution is to try to load the JNI with the provided name first, and if this does not work, then try with the first 3 char of the name removed.

Our tests are currently running on mac os x: https://github.com/carloseltuerto/envoy-mobile/blob/checkout12/.github/workflows/android_tests.yml#L12

Here's the issue for running the jvm tests on ubuntu for ci also: https://github.com/envoyproxy/envoy-mobile/issues/1402

library/java/io/envoyproxy/envoymobile/engine/JniLibrary.java
```diff
@@ -17,7 +17,9 @@
   // WARNING: This should only be used for testing.
   public static void loadTestLibrary() {
     if (System.getProperty("envoy_jni_library_name") != null) {
-      envoyLibraryName = System.getProperty("envoy_jni_library_name");
+      envoyLibraryName = System.getProperty("os.name").startsWith("Linux")
```
buildbreaker, line 20: Not feeling good about having this in code but since this works and unblocks you, let's go with it. I can async try to understand why bazel is not cooperating like I expect it to.

buildbreaker (approved):

buildbreaker:
Let's go with this, @carloseltuerto !

PR #1412, opened 2021-04-26
Facititate CA Certs configuration

carloseltuerto:
Make the CA Certs configuration somewhat easier by exposing the default certificates at the JNI level, and rather substitute the certificates at the Java level (EnvoyConfiguration)

Risk Level: Negligeable
Testing: //test/java/integration:android_engine_start_test still works
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/library/common/BUILD
+++ b/library/common/BUILD
@@ -16,7 +16,7 @@ envoy_cc_library(
 envoy_cc_library(
     name = "envoy_main_interface_lib_no_stamp",
     srcs = [
-        "certificates.inc",
+        "certificates.cc",
         "config_template.cc",
         "engine.cc",
         "engine.h",
--- a/library/common/certificates.cc
+++ b/library/common/certificates.cc
@@ -1,4 +1,8 @@
-R"certs(
+/**
+ * Default certificates.
+ */
+const char* certificates = R"certs(
+              inline_string: |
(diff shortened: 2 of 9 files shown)


goaway:
Hey @carloseltuerto, just wanted to acknowledge that I see the PR and I'm just thinking this through a little bit.

goaway:
Sorry for the delay, today! Generally I'd classify the risk of _anything_ that has to do with TLS as somewhat more than negligible.

In general, I don't think this approach is necessarily worse that what we have in place, but it would be nice to do better and there are some notable shortcomings:

- It exposes more raw configuration to the functional config API which I think is something to be avoided.
- The default cert chain is quite large, and this moves it across the JNI and copies it several times (that's happening anyways, but is also something that could be readily improved).
- We strive to maintain parity across stacks, and the PR as is only introduces new functionality to one of them; we should at least have a plan for how we will maintain parity when it comes to new features like this.

I actually think setting up a plaintext call path would be quite straightforward and not significantly more work, and it sounds like that's really what you're looking for. I'm pretty confident I could set up a plaintext path for you tomorrow if that would meet your needs - which I'd prefer I think, especially if it will *better* meet your needs.

Overall open to discuss though.

goaway (commented):
(Forgot to submit above comment as review.)

PR #1419, opened 2021-04-27
core: set up cleartext http request path

goaway:
Description: Adds built-in support for cleartext HTTP requests, without requiring custom config. TLS will be negotiated or not based off the request scheme. If cleartext is chosen, h2 protocol options will be ignored, and h1 will be used regardless.
Risk Level: Lowish
Testing: Local, Unit, CI

### DIFF
--- a/library/common/http/client.cc
+++ b/library/common/http/client.cc
@@ -367,26 +367,31 @@ void Client::setDestinationCluster(Http::RequestHeaderMap& headers) {
   // - Use TLS by default.
   // - Use http/2 if requested explicitly via x-envoy-mobile-upstream-protocol.
   // - Force http/1.1 if request scheme is http (cleartext).
-  std::string cluster;
+  const char* cluster{};
   auto h2_header = headers.get(H2UpstreamHeader);
+  auto network = preferred_network_.load();
+  ASSERT(network >= 0 && network < 3, "preferred_network_ must be valid index into cluster array");
 
   if (headers.getSchemeValue() == Headers::get().SchemeValues.Http) {
-    cluster = ClearTextClusters[preferred_network_.load()];
+    cluster = ClearTextClusters[network];
   } else if (!h2_header.empty()) {
     ASSERT(h2_header.size() == 1);
     const auto value = h2_header[0]->value().getStringView();
     if (value == "http2") {
-      cluster = H2Clusters[preferred_network_.load()];
+      cluster = H2Clusters[network];
     } else {
       RELEASE_ASSERT(value == "http1", fmt::format("using unsupported protocol version {}", value));
-      cluster = BaseClusters[preferred_network_.load()];
+      cluster = BaseClusters[network];
     }
-    headers.remove(H2UpstreamHeader);
   } else {
-    cluster = BaseClusters[preferred_network_.load()];
+    cluster = BaseClusters[network];
   }
 
-  headers.addReference(ClusterHeader, cluster);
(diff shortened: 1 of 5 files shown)


carloseltuerto:
FYI, I patched the PR, but unfortunately all requests end up in the same manner:

```
EnvoyError.errorCode: 2 // likely means Unknown
EnvoyError.message: ""
EnvoyError.attemptCount: -1
EnvoyError.cause: null
```

library/common/http/client.cc
```diff
@@ -342,46 +342,51 @@ void Client::removeStream(envoy_stream_t stream_handle) {
...
-    headers.addReference(ClusterHeader, use_h2_cluster ? BaseWlanClusterH2 : BaseWlanCluster);
-    break;
-  case ENVOY_NET_WWAN:
-    headers.addReference(ClusterHeader, use_h2_cluster ? BaseWwanClusterH2 : BaseWwanCluster);
-    break;
-  case ENVOY_NET_GENERIC:
-  default:
-    headers.addReference(ClusterHeader, use_h2_cluster ? BaseClusterH2 : BaseCluster);
-  }
+  headers.addReference(ClusterHeader, cluster);
```
carloseltuerto, line 389: If instead this line is:

`headers.addCopy(ClusterHeader, cluster);`

Then this PR works.

Note: last time I coded in C/C++ was in previous millennium. Still, there is one thing that seems to have remained constant: using a pointer to a structure residing in the stack will end up badly if that pointer is still used once the method has returned. Editorial note: C++ is truly a counterproductive language.

goaway: Yeah I missed that it was `addReference` in the initial implementation (though technically the source c string is constant, it's just the string intermediary ending up on the stack that's problematic).

goaway:
This PR is just pending a test fix and some additional coverage that I'll have up tomorrow morning.

I have confirmed testing locally that cleartext requests appear to work, so I think the issue you're seeing now may be a separate one @carloseltuerto. But I'll be available to sync on it tomorrow.

carloseltuerto:
> This PR is just pending a test fix and some additional coverage that I'll have up tomorrow morning.
> 
> I have confirmed testing locally that cleartext requests appear to work, so I think the issue you're seeing now may be a separate one @carloseltuerto. But I'll be available to sync on it tomorrow.

Everything is fine now - feel free to submit (as soon as you are fine with the tests you intend to update)

Thanks!

library/common/http/client.cc
```diff
@@ -342,46 +342,56 @@ void Client::removeStream(envoy_stream_t stream_handle) {
...
-    headers.addReference(ClusterHeader, use_h2_cluster ? BaseWlanClusterH2 : BaseWlanCluster);
-    break;
-  case ENVOY_NET_WWAN:
-    headers.addReference(ClusterHeader, use_h2_cluster ? BaseWwanClusterH2 : BaseWwanCluster);
-    break;
-  case ENVOY_NET_GENERIC:
-  default:
-    headers.addReference(ClusterHeader, use_h2_cluster ? BaseClusterH2 : BaseCluster);
+  if (!h2_header.empty()) {
+    headers.remove(H2UpstreamHeader);
```
buildbreaker, line 391: How come we didn't put this in the above if-else block instead since we have this explicit check?

goaway: Because we need to remove it regardless of whether the first condition resolves to true.

buildbreaker (approved):

Issue #1418, opened 2021-04-27
android: java integration tests are not run in linux


buildbreaker:
This is likely due to some android dependency not being in the envoy image

PR #1421, opened 2021-04-27
ci: fix kotlin/java tests

buildbreaker:

Description: ci: fix kotlin/java tests 
Risk Level: low
Testing: ci
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.github/workflows/android_tests.yml
+++ b/.github/workflows/android_tests.yml
@@ -23,8 +23,7 @@ jobs:
       - run: ./ci/mac_ci_setup.sh
         name: 'Install dependencies'
       - run: |
-          # /test/kotlin/integration is excluded since it is run on a separate job
-          bazelisk query //test/kotlin/...
+          bazelisk test --test_output=all --build_tests_only //test/kotlin/...
         name: 'Run Kotlin library tests'
   javatestsmac:
     name: java_tests_mac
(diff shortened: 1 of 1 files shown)


.github/workflows/android_tests.yml
```diff
@@ -42,8 +41,7 @@ jobs:
       - run: ./ci/mac_ci_setup.sh
         name: 'Install dependencies'
       - run: |
-          # /test/kotlin/integration is excluded since it is run on a separate job
-          bazelisk query //test/java/...
+          bazelisk test --test_output=all --build_tests_only  //test/java/...
         name: 'Run Java library tests'
   kotlintestslinux:
```
junr03, line 46: How come ^^ those jobs run in mac instead of linux? Seems like we should run on linux (which will probably get better machines and more concurrency easier).

Also the below target for integration tests is not running the integration test target? https://github.com/envoyproxy/envoy-mobile/blob/b767e175155f9e6135d1b33a82229ed9e0e3c686/.github/workflows/android_tests.yml#L69

junr03: Talked offline. Related to #1418.

junr03 (approved):

.github/workflows/android_tests.yml
goaway: The integration target is `//test/kotlin/integration/...` so it's included in the above target. I suggested running them combined (in this and other cases) due to fixed startup overhead and instance queueing.

PR #1424, opened 2021-04-28
kotlin: add cancel stream test and fix custom yaml path

buildbreaker:
Adding a cancel stream test to mirror swift's cancel stream test: https://github.com/envoyproxy/envoy-mobile/blob/102b40d0da174200b4308a2677f6788a1df2a0aa/test/swift/integration/CancelStreamTest.swift

Found that the custom yaml configuration also has a bug in it where it does not appropriately register platform and native filters when a custom yaml is used. Luckily this is the first usage of the custom yaml. The JNI structure is different from the objective-c structure but I think I was able to mimic it as close as possible by just requiring an `EnvoyConfiguration`.

The API changes are here: https://github.com/envoyproxy/envoy-mobile/pull/1424/files#diff-c4d0f50e81c31efc078d029708e09a001ea3e466c6b38141c027fad21c6b70abR30-R34

Description: kotlin: add cancel stream test and fix custom yaml path
Risk Level: low
Testing: integration test
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]


library/java/io/envoyproxy/envoymobile/engine/EnvoyEngineImpl.java
```diff
@@ -55,13 +55,23 @@ public void terminate() {
    * @return A status indicating if the action was successful.
    */
   @Override
-  public int runWithConfig(String configurationYAML, String logLevel) {
-    try {
-      return JniLibrary.runEngine(this.engineHandle, configurationYAML, logLevel);
-    } catch (Throwable throwable) {
-      // TODO: Need to have a way to log the exception somewhere.
-      return ENVOY_FAILURE;
+  public int runWithConfig(EnvoyConfiguration envoyConfiguration, String configurationYAML,
```
carloseltuerto, line 58: nit: please update the the above javadoc. 

Also the configurationYAML parameter can now have arguments substituted. Was not the case before.

buildbreaker: let me update with a blurb on the java docs

library/java/io/envoyproxy/envoymobile/engine/EnvoyEngineImpl.java
```diff
@@ -50,18 +50,31 @@ public void terminate() {
...
```
junr03, line 61: Do we care about not exposing a way to run with a resolved yaml file and no configuration object?

test/kotlin/integration/CancelStreamTest.kt
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  @Test
+  fun `cancel stream calls onCancel callback`() {
+    val engine = EngineBuilder(Custom(config))
+      .addPlatformFilter(
+        name = "cancel_validation_filter",
+        factory = { CancelValidationFilter(filterExpectation) }
+      )
+      .setOnEngineRunning {}
```
junr03, line 127: this is strange. I will look into this. I'd rather us not have this mystery.

test/kotlin/integration/CancelStreamTest.kt
```diff
@@ -0,0 +1,155 @@
...
+
+    client.newStreamPrototype()
+      .setOnCancel {
+        runExpectation.countDown()
+      }
+      .start(Executors.newSingleThreadExecutor())
+      .sendHeaders(requestHeaders, false)
+      .cancel()
+
+    engine.terminate()
```
junr03, line 149: ? How come we are terminating before the expectations are fulfilled?

test/kotlin/integration/CancelStreamTest.kt
buildbreaker: Let me add an issue: https://github.com/envoyproxy/envoy-mobile/issues/1429

test/kotlin/integration/CancelStreamTest.kt
buildbreaker: Let me change it to be after the expectation awaits

library/java/io/envoyproxy/envoymobile/engine/EnvoyEngineImpl.java
buildbreaker: We're currently not exposing this functionality to the Kotlin consumers. As the API design, the `EngineBuilder` manages the `EnvoyConfiguration` detail internally

goaway: Generally, I think this is fine; nothing stops someone from providing resolved yaml and not bothering to configure anything via the builder. That said, I'm curious - why not mirror the iOS API directly?

junr03: Yeah, I was mostly wondering about parity.

buildbreaker: Doesn't this mirror https://github.com/envoyproxy/envoy-mobile/blob/3a2fdb74d67f7ccc60e5fd8bd34aa2ae47e72003/library/objective-c/EnvoyEngineImpl.m#L454-L456? --  I can change the method name and signature to better reflect that.

junr03: I meant in the absence of running with a resolved yaml and no config object. Or do we not expose that in iOS?

test/kotlin/integration/CancelStreamTest.kt
junr03: thanks!

library/java/io/envoyproxy/envoymobile/engine/EnvoyEngineImpl.java
junr03: like: https://github.com/envoyproxy/envoy-mobile/blob/3a2fdb74d67f7ccc60e5fd8bd34aa2ae47e72003/library/objective-c/EnvoyEngineImpl.m#L474

buildbreaker: If we're missing that method I can add it but I would prefer to keep this PR as purely just to make things works for the test

junr03: sg, can you open an issue?

junr03 (dismissed):
lgtm, pending issue for missing method, and main merge. Thanks @buildbreaker! More tests, more better :)

library/java/io/envoyproxy/envoymobile/engine/EnvoyEngineImpl.java
buildbreaker: This is actually what we have now https://github.com/envoyproxy/envoy-mobile/blob/3a2fdb74d67f7ccc60e5fd8bd34aa2ae47e72003/library/objective-c/EnvoyEngine.h#L312-L331

goaway (approved):
Looks great, nice job @buildbreaker!

PR #1434, opened 2021-05-01
quic: introduce working e2e test

goaway:
Description: Introduces an end-to-end test that successfully sends a request and receives a response over the local network with QUIC/H3, originating in the public iOS library API and exercising all layers of the stack.
Risk Level: Low
Testing: Local


buildbreaker (approved):
Cool! Let's get an issue up for Android.

PR #1443, opened 2021-05-05
config: update keepalive settings

goaway:
Description: An alternate keepalive approach.
Risk Level: Low
Testing: In progress.

### DIFF
--- a/library/common/config_template.cc
+++ b/library/common/config_template.cc
@@ -166,9 +166,9 @@ static_resources:
                               R"(
     upstream_connection_options: &upstream_opts
       tcp_keepalive:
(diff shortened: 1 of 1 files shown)


junr03 (commented):
Do you mind updating the description to explain some of the thought behind the new approach?

Did we change Lyft's polling loop interval out of 5 secs, I can't remember? If we didn't, doesn't this approach go back to having the potential risk of resonating with the polling interval?

goaway:
Lyft's polling loop is no longer a multiple of 5 seconds - will discuss more in our meeting today.

junr03 (approved):

PR #1446, opened 2021-05-06
Add "Flow" Integration Test for Android

carloseltuerto:
This test helps documenting what is the Envoy-Mobile flow logic. This is useful in regards with Cronvoy development, however, this is not testing anything related to Cronvoy - it is a pure Envoy-Mobile test.

Description: Add tests exercising some basic HTTP flows with the Android library
Risk Level: Inexistent
Testing: This PR only adds a new test - no prod code is being changed
Docs Changes: N/A
Release Notes: N/A


test/java/integration/AndroidEnvoyFlowTest.java
```diff
@@ -0,0 +1,356 @@
...
+  @BeforeClass
+  public static void loadJniLibrary() {
+    AndroidJniLibrary.loadTestLibrary();
+  }
+
+  @AfterClass
+  public static void shutdownEngine() {
+    if (engine != null) {
+      engine.terminate();
+    }
```
buildbreaker, line 61: Do we need to reset the engine to null here since we are doing a null check for initialization in the `@Before` step?

carloseltuerto: Well, for sure, the tests are fast: no start & stop after each @Test. 

I assumed that doing a new `AndroidEngineBuilder(appContext)....build();` for each test would not work if we call `engine.terminate();` after each @Test.

buildbreaker (approved):
Looks good, thanks for the insights/clarification on some details @carloseltuerto !

PR #1462, opened 2021-05-12
Add "HTTP Redirect" support to Cronvoy

carloseltuerto:
Cronet requirements to support "HTTP redirect" seem fairly simple: always wait for the caller to confirm redirection acceptance. If the caller rejects the redirection, then the caller does not need to do anything else - the UrlRequest must be in a state where everything has been "closed" already.

There is one API subtlety: in the event that an HTTP Redirect response contains a body, that body must be ignored. Implementation wise, this implies that `setOnResponseData` callbacks must become a no-op.

Description: Add "HTTP Redirect" support, implemented within Cronvoy shim.
Risk Level: N/A
Testing: N/A
Docs Changes: N/A
Release Notes: N/A
Signed-off-by: Charles Le Borgne cleborgne@google.com

### DIFF
--- a/library/java/org/cronvoy/CronvoyUrlRequest.java
+++ b/library/java/org/cronvoy/CronvoyUrlRequest.java
@@ -27,6 +27,7 @@ import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
 import org.chromium.net.CronetException;
 import org.chromium.net.InlineExecutionProhibitedException;
 import org.chromium.net.UploadDataProvider;
@@ -47,7 +48,6 @@ final class CronvoyUrlRequest extends UrlRequestBase {
 
   private final AsyncUrlRequestCallback callbackAsync;
   private final PausableSerializingExecutor cronvoyExecutor;
-  private final PausableSerializingExecutor envoyCallbackExecutor;
   private final String mUserAgent;
   private final Map<String, String> requestHeaders = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
   private final List<String> urlChain = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


RyanTheOptimist (approved):
I've not programmed in Java for a while so I'm not the best reviewer, but this looks good to me!

carloseltuerto:
> I've not programmed in Java for a while so I'm not the best reviewer, but this looks good to me!
Thanks, Ryan.

The truth here is I don't expect anyone to fully understand the logic being implemented in CronvoyUrlRequest without a thorough explanation - don't worry.

What I rather wanted from you is to assess my understanding of the Cronet API regarding "HTTP Redirect". If I misconstrued something, or if there are peculiar details about HTTP redirection that I should consider, please let me know.

goaway:
One question I have here - is this meant to just be an expedient solution, or is this actually intended to be long-term? I feel handling redirects internally in Envoy is the best path forward generally speaking. If a confirmation hook needs to be added upstream, that's something we can look into.

Even in the interim, the buffering, api callback/confirmation, and re-dispatch could be implemented in an Envoy filter, which to me possibly still seems preferable to implementing the behavior as a shell around the library.

But if all this is simply in service of coming up with the quickest path to API compatibility, then perhaps this is fine for now?

carloseltuerto:
> One question I have here - is this meant to just be an expedient solution, or is this actually intended to be long-term? I feel handling redirects internally in Envoy is the best path forward generally speaking. If a confirmation hook needs to be added upstream, that's something we can look into.
> 
> Even in the interim, the buffering, api callback/confirmation, and re-dispatch could be implemented in an Envoy filter, which to me possibly still seems preferable to implementing the behavior as a shell around the library.
> 
> But if all this is simply in service of coming up with the quickest path to API compatibility, then perhaps this is fine for now?

TL;DR: This would probably be an expedient solution.

I did somehow misjudge the situation. I looked a the native Chromium implementation of Cronet, and it turns out that is does leave the C++ layer handle the redirection:

- [Java](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/components/cronet/android/java/src/org/chromium/net/impl/CronetUrlRequest.java#294)
- [Android JNI Entry point](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/components/cronet/android/cronet_url_request_adapter.cc#166)
- [Cronet layer](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/components/cronet/cronet_url_request.cc#135)
- [Net layer](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/net/url_request/url_request_job.cc#234)

Most callers do invoke "cancel" when they reject the redirection. Still, there is at least one exception in Google codebase - "cancel" is not invoked in such case. But I guess that the outcome will be similar with Cronet or Envoy-Mobile: the Engine will be dangling for a while, and then will clean up itself upon a timeout.

goaway (approved):
LGTM. Thanks for the discussion on this @carloseltuerto!

goaway:
@alyssawilk do you want to do a pass before merging?

alyssawilk (commented):
@carloseltuerto and I discussed cross-scheme redirects "offline" and I think he was going to disallow changing URL scheme in the redirects until we make sure we're happy with the security of how DFP code handles it and we confirm they're needed.

PR #1465, opened 2021-05-13
cc, python: populate EnvoyError.message

crockeo:
Description: populates EnvoyError.message instead of populating it with an empty string
Risk Level: Low
Testing: `bazel test //test/python:runtest_unit` to run `test_envoy_error`
Docs Changes: N/A
Release Notes: N/A


test/python/unit/test_basics.py
```diff
@@ -70,3 +70,10 @@ def test_send_data_form_urlencoded(http_server_url: str, data):
         == "application/x-www-form-urlencoded"
     )
     assert json.get("headers", {}).get("charset") == "utf8"
+
+
+def test_envoy_error():
+    response = envoy_requests.get("http://127.0.0.1:0/fake-url")
+    assert response.envoy_error is not None
+    assert response.envoy_error.error_code != 0
+    assert response.envoy_error.message != ""
```
goaway, line 79: Not to split hairs, but probably we could assert that error_code and message are actually something specific?

crockeo: If you think it's appropriate! I kept them non-zero and non-empty since I didn't know how stable envoy-mobile's errors are, and wanted to keep maintenance burden down.

goaway:
At a glance, I think the test failure looks legitimate.

goaway (approved):
Nice!

PR #1466, opened 2021-05-14
engine: add configurable stream idle timeout

goaway:
Description: Exposes HTTP stream idle timeout as a configurable option via the Swift and Kotlin EngineBuilders.
Risk Level: Low
Testing: Local and CI


buildbreaker (dismissed):
Thanks @goaway !

rebello95 (commented):
LGTM but can you add a test to ensure the template resolves? We should also update the docs

rebello95 (approved):

PR #1473, opened 2021-05-14
engine: attempt server shutdown during off-thread destruction

goaway:


junr03:
let's re-open when picking back up

PR #1474, opened 2021-05-15
common: main interface shared library target

crockeo:
Description: This PR introduces two build targets, `//library/common:envoy_main_interface_lib_shared` and `//:envoy_main_interface_lib_shared` (the latter being a `genrule` that renames the `.so`) which can be used to build a shared `libenvoy_mobile.so`. 

I was working on [envoy-mobile-rs](https://github.com/crockeo/envoy-mobile-rs) in my free time a while back and ran into issues with building a Bazel target under Cargo. After a lot of reading, I don't know if I can resolve it without changing Cargo code, so I want to stop building envoy-mobile as part of the Rust build and instead just require that `libenvoy_mobile.so` exists somewhere discoverable.

This is totally orthogonal to my py-envoy-mobile work, so consider this super low priority.

Risk Level: Low

Testing: `bazel build //:envoy_main_interface_lib_shared`

Docs Changes: N/A

Release Notes: N/A


crockeo:
Looks like this doesn't work as expected OOTB on macOS, because it [sets -install_name to the original library path](https://github.com/bazelbuild/bazel/blob/275c54c982468b97611dc764c68daed8d7c45004/tools/osx/crosstool/cc_toolchain_config.bzl#L2919-L2938) which results in an `LC_ID_DYLIB` load command like so:

```
Load command 4
          cmd LC_ID_DYLIB
      cmdsize 112
         name bazel-out/darwin-fastbuild/bin/library/common/libenvoy_main_interface_lib_shared.so (offset 24)
   time stamp 1 Wed Dec 31 16:00:01 1969
      current version 0.0.0
compatibility version 0.0.0
```

Aka libenvoy_mobile.so needs to be located on the `LD_LIBRARY_PATH` and then exist at `bazel-out/darwin-fastbuild/bin/library/common/libenvoy_main_interface_lib_shared.so`, which is not what I intend with this PR. I need to figure out how to handle this correctly.

crockeo:
Looks like a quick `install_name_tool -id libenvoy_mobile.so` works--at least with my intended use-case. Going to experiment a little bit more to see if this would work more generally as well.

junr03:
re-open if needed

PR #1475, opened 2021-05-16
[WIP/TESTING] extensions: Update config

phlax:
Description: extensions: Update config
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

PR #1481, opened 2021-05-19
envoy: update submodule 46b646c -> 33cf6cd

phlax:
Signed-off-by: Ryan Northey <ryan@synca.io>

Description:

When testing #1475 tests were failing as the envoy submodule was behind envoy main

This PR also:

- updates the protobuf patch to make it work with v3.16.0
- adds the `envoy.http.original_ip_detection.xff` extension as tests seemed to indicate it as necessary

Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


phlax:
not sure if failing tests are flakes - im struggling to get any info when clicking on the "details" link

buildbreaker:
The python one might be a flake that @crockeo was looking into but the core test likely is unexpected. I just reran both the jobs and am hoping that makes it green

phlax:
@buildbreaker checking the swift_helloworld test it seems to be failing with:

```
[2021-05-19 17:15:15.447][31938][critical][main] [external/envoy/source/server/server.cc:113] error initializing configuration '': Original IP detection extension not found: 'envoy.http.original_ip_detection.xff'
```

this was the reason that i added in the xff extension in the build config.

its not clear to me why this is still failing, where the config is coming from, or why the envoy binary doesnt have the extension added

im guessing that the problem is related to the ip detection extension being listed as required in envoy's `all_extensions.bzl` but its just a guess

any ideas ?

envoy_build_config/extensions_build_config.bzl
```diff
@@ -7,6 +7,7 @@ EXTENSIONS = {
     "envoy.filters.http.buffer":                      "//source/extensions/filters/http/buffer:config",
     "envoy.filters.http.dynamic_forward_proxy":       "//source/extensions/filters/http/dynamic_forward_proxy:config",
     "envoy.filters.http.local_error":                 "@envoy_mobile//library/common/extensions/filters/http/local_error:config",
+    "envoy.http.original_ip_detection.xff":           "//source/extensions/http/original_ip_detection/xff:config",
```
phlax, line 10: this is probably not necessary as its listed in envoy's `all_extensions.bzl` 

if it is its in the wrong alpha order

goaway:
@phlax sorry I didn't see this sooner - if you're adding an extension, it will need to be registered here:
https://github.com/envoyproxy/envoy-mobile/blob/c8ef0d13643e54141e1bfd5ad4fd9018885a0101/envoy_build_config/extension_registry.cc

goaway:
This has to do with differences in static initialization order when running as a process vs. running as a statically-linked library.

phlax:
great, thanks - ill update shortly

phlax:
@goaway ive tried registering the extension - unfortunately it causes an error - something to do with namespacing and factories

im afraid this is about my limit of c++ knowledge - if you can lmk what else is required i would be happy to add

goaway (approved):

PR #1482, opened 2021-05-21
Adopt original Cronet java package structure

carloseltuerto:
This PR realigns the subpar pre-prototype of Cronvoy to the original Cronet structure. This will save us lot of pointless efforts moving forward. More original code was copied over - this helps exposing the functional gap we are facing.

Description: To ease Cronet test porting, and the unfortunate GSMcore direct references to some implementation classes, move all classes to the original Cronet java package structure, and adopt original Class names.
Risk Level: Inexistent - code under development
Testing: existing test was moved too
Docs Changes: N/A
Release Notes: N/A


buildbreaker (approved):
Let's do it. I am in!

PR #1486, opened 2021-05-26
expose flushStats function on engine

snowp:
Exposes a flushStats function on the Engine, allowing the app to trigger flushes at will. This should allow hooking flushes up
to app lifecycle events.

Risk Level: Medium
Testing: Integration test
Docs Changes: n/a
Release Notes: n/a


snowp:
This has the Android side working, as demonstrated by the test that verifies that we're flushing stats to a simple StatsdServer.

Haven't looked at the iOS part yet, and one open question here is whether it makes sense to try to do some work here to allow us to reuse this stats server between the platforms (either via native code or via an executable/wrapping socat) or if this approach where we make a language specific server that we can write more ergonomic helper functions on is preferred. Maybe @goaway has thoughts here since we talked about these kind of integration tests?

test/kotlin/integration/StatFlushIntegrationTest.kt
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  @After
+  fun teardown() {
+    engine?.terminate()
+    engine = null
+  }
+
+  @Test
+  fun `concurrent flushes with histograms`() {
```
jingwei99, line 28: im assuming the goal for this test case is to verify it's safe to concurrently flushing, could you confirm?

snowp: It's to verify that we can have multiple stats flushes "queued up" on the dispatcher at the same time, regression for the issue we ran into previously. The contains thing is just to ensure that Envoy has booted up before we start hitting it with flushes

jingwei99: 👍

bazel/envoy_mobile_dependencies.bzl
```diff
@@ -23,6 +23,7 @@ def swift_dependencies():
 def kotlin_dependencies():
     maven_install(
         artifacts = [
+            "com.google.code.findbugs:jsr305:3.0.2",
```
jingwei99, line 26: i didn't find where this library is used in the code, could you point it for me 😬

snowp: It's used to provide the `javax.annotation.Nullable` annotation to let me mark a Java function as being able to accept a Kotlin nullable in the jvm-only code (so no android Nullable)

I couldn't find any other Nullable annotations so I went with this, lmk if there is a better way to go about this :)

library/common/config_template.cc
```diff
@@ -57,7 +57,7 @@ const char* fake_remote_cluster_template = R"(
               socket_address: { address: 127.0.0.1, port_value: 10101 }
 )";
 
-const char* stats_sink_template = R"(
+const char* grpc_stats_sink_template = R"(
```
jingwei99, line 60: it probably worth mentioning in the pr description the addition of the `statsd_sink_template`

library/swift/Engine.swift
```diff
@@ -9,4 +9,10 @@ public protocol Engine: AnyObject {
 
   /// A client for recording time series metrics.
   func pulseClient() -> PulseClient
+
+  ///
+  /// Flush the stats sinks outside of a flushing interval.
+  /// Note: stats flushing may not be synchronous.
```
jingwei99, line 15: this is just out of curiosity, since i read "may not be synchronous" a few times in the docs: when is the stats flushing synchronous and when is it async?

jingwei99 (approved):
lgtm!

library/swift/Engine.swift
snowp: The way it works is that if a flushStats is issued before the server has initialized it will flush directly, otherwise it will schedule a histogram merge on all the worker threads and do the flush once this has completed. That said, since we issue this flush request by putting it on the EM dispatcher, this top level function will never block even if we do the synchronous flush code path.

Maybe I should update this comment (I took it from the revert) to be clearer here?

jingwei99: 👍

just a follow up clarifying question so im clear on it:
> if a flushStats is issued before the server has initialized it will flush directly

i might be wrong, but im assuming there won't be "real" stats flush before the server has initialized?

snowp: So the Envoy handler has logic for doing a flush before the server has been initialized: https://github.com/envoyproxy/envoy/blob/main/source/server/server.cc#L208-L229

This makes it possible to do a stat merge that doesn't rely on merging histograms (since worker threads may not have been created yet).

If a merge comes in before `server_` has been set on the EM engine it will be a noop since we never dispatch a flushStats at all

So there are three results: 1) before EM/envoy's startup process has started (noop) 2) during EM/envoy init phase (direct, no histogram merge) and 3) after EM/envoy init (histogram merge, async flush)

snowp:
/ptal @junr03

Mind taking a look as well?

library/swift/Engine.swift
jingwei99: 👌🙏

snowp:
ptal @goaway @junr03 (or tag whoever is more appropriate to review this :) )

library/common/main_interface.cc
```diff
@@ -146,6 +146,12 @@ envoy_status_t record_histogram_value(envoy_engine_t, const char* elements, envo
   return ENVOY_FAILURE;
 }
 
+void flush_stats(envoy_engine_t) {
```
goaway, line 149: Please move dispatch into here.

library/common/engine.cc
```diff
@@ -233,4 +233,16 @@ Http::Client& Engine::httpClient() {
   return *http_client_;
 }
 
+void Engine::flushStats() {
+  // The server will be null if the server has not been initialized as part of main.
+  // In this case, we can simply ignore the flush.
+
+  Thread::LockGuard lock(mutex_);
```
goaway, line 240: These checks/guards are unnecessary once dispatch is moved to main_interface.

goaway: (You can ASSERT isThreadSafe if you want to be defensive/paranoid though.)

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -62,6 +65,21 @@ open class EngineBuilder(
...
+   * Add a loopback port to emit statsd stats to.
+   * Passing nil disables stats emission via the statsd stat sink.
+   *
+   * Only one of the statsd and gRPC stat sink can be enabled.
+   *
+   * @param port The port to send statsd UDP packets to via loopback
+   *
+   * @return this builder.
+   */
+  fun addStatsdPort(port: Int): EngineBuilder {
```
goaway, line 78:
```suggestion
  fun addStatsDPort(port: Int): EngineBuilder {
```

Maybe, just to make the "D" a little more obvious? (I misread the first time.)

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -239,7 +257,7 @@ open class EngineBuilder(
         EngineImpl(
           engineType(),
           EnvoyConfiguration(
-            statsDomain, connectTimeoutSeconds,
+            statsDomain, statsdPort, connectTimeoutSeconds,
```
goaway, line 260: Can we rename public references to statsDomain to be grpcStatsDomain, since we're distinguishing between sinks/protocols now?

Also can we have some programmatic guard/indication that you're doing it wrong if both are set?

snowp: Done and done

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
snowp: Done

library/common/engine.cc
snowp: Done

library/kotlin/io/envoyproxy/envoymobile/Engine.kt
```diff
@@ -20,4 +20,11 @@ interface Engine {
    * Terminates the running engine.
    */
   fun terminate()
+
+  /**
+   * Flush the stats sinks outside of a flushing interval.
+   * Note: stat flushing is done asynchronously, this function will never block.
+   * This is a noop if called before the underlying EnvoyEngine has started.
+   */
+  fun flushStats()
```
jingwei99, line 29: Just realized this, it'll make more sense to fold this method to `PulseClient`, no changes needed as for now, can be planned as a clean-up/follow-up change.

PR #1487, opened 2021-05-26
http client: select from a pair of clusters for each connection class

goaway:
Description: The objective here is to make it less likely for a given connection class to suffer "catastrophic" failure of all outbound requests due to a network blip, by distributing requests across a minimum of two potential connections per connection class.
Risk Level: Low
Testing: Local, In-App, Unit, CI

### DIFF
--- a/library/common/http/client.cc
+++ b/library/common/http/client.cc
@@ -360,44 +360,38 @@ const LowerCaseString H2UpstreamHeader{"x-envoy-mobile-upstream-protocol"};
 // Long-term we will be working to generally provide more responsive connection handling within
 // Envoy itself.
 
-const char* BaseClusters[][3] = {
-  {
-    "base",
-    "base_wlan",
-    "base_wwan",
-  },
-  {
-    "base_alt",
-    "base_wlan_alt",
-    "base_wwan_alt",
-  }
-};
+const char* BaseClusters[][3] = {{
+                                     "base",
+                                     "base_wlan",
+                                     "base_wwan",
+                                 },
+                                 {
+                                     "base_alt",
+                                     "base_wlan_alt",
+                                     "base_wwan_alt",
+                                 }};
 
-const char* H2Clusters[][3] = {
-  {
-    "base_h2",
-    "base_wlan_h2",
-    "base_wwan_h2",
-  },
-  {
-    "base_h2_alt",
-    "base_wlan_h2_alt",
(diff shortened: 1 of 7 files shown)


goaway:
Depends on a minor Envoy change to allow mocking of the `bernoulli(p)` function.

library/common/http/client.cc
```diff
@@ -377,20 +401,21 @@ void Client::setDestinationCluster(Http::RequestHeaderMap& headers) {
   auto h2_header = headers.get(H2UpstreamHeader);
   auto network = preferred_network_.load();
   ASSERT(network >= 0 && network < 3, "preferred_network_ must be valid index into cluster array");
+  ASSERT(!alternate, "every time");
```
junr03, line 404: I might be being dense, but I am not sure I understand this assertion?

RyanTheOptimist:
> I'm +1 on this as a stopgap, but I would recommend that you leave a comment trail to the effect of: this is a stopgap until we have better dead connection detection, especially something like [envoyproxy/envoy#16313](https://github.com/envoyproxy/envoy/issues/16313). cc @alyssawilk @DavidSchinazi @RyanTheOptimist. I discussed briefly with @alyssawilk and the way I would imagine this might work is that if a connection hasn't been used for some period of time, we use a new connection while we liveness check the old connection before putting it back in rotation, though there are many possible policies.

Agreed that this is a fine stopgap and we should explore a long term solution. FWIW, in Chome when attempting to re-use connection the request is immediately sent, along with an HTTP/2 PING. Chrome then closes connection early and tries a new connection if the PING is not responded to "quickly". In general it's a good idea to reuse sockets as much as possible so Chrome almost never opens more than one HTTP/2 connection to a given server. I suspect something like this is the right approach here, but it's worth further discussion.

library/common/http/client.cc
goaway: Just leftover from testing. :)

goaway:
@mattklein123 @RyanTheOptimist thanks for your thoughts! Definitely agree there are better ways long-term that we can handle this. Will comment to that effect.

junr03 (dismissed):
lgtm

junr03:
@goaway format is still unhappy

goaway:
@junr03 the script is still broken on my machine, so I have to run in manually for each file. -__-

buildbreaker (approved):
🙌

PR #1488, opened 2021-05-26
filters: ensure PBFs pending destruction can't be resumed async

goaway:
Description: Previously asynchronous filter resumption of a bridged platform filter was guarded in the PBF by a weak_ptr that ensured the object was still around. It turns out this is not sufficient, as filters are submitted for delayed destruction and the presence of the object isn't enough to ensure it's still associated with a functional stream. This fix adds an additional check to ensure we don't attempt to resume iteration on a PlatformBridgeFilter that is pending destruction.
Risk Level: Low
Testing: Unit and CI

### DIFF
--- a/library/common/extensions/filters/http/platform_bridge/filter.h
+++ b/library/common/extensions/filters/http/platform_bridge/filter.h
@@ -78,6 +78,8 @@ public:
   Http::FilterDataStatus encodeData(Buffer::Instance& data, bool end_stream) override;
   Http::FilterTrailersStatus encodeTrailers(Http::ResponseTrailerMap& trailers) override;
 
+  bool isAlive() { return alive_; }
+
 private:
   /**
    * Internal delegate for managing logic and state that exists for both the request (decoding)
@@ -160,6 +162,7 @@ private:
   envoy_http_filter_callbacks platform_request_callbacks_{};
   envoy_http_filter_callbacks platform_response_callbacks_{};
   bool error_response_{};
+  bool alive_{true};
 };
 
 using PlatformBridgeFilterSharedPtr = std::shared_ptr<PlatformBridgeFilter>;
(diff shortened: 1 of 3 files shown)


goaway:
@mattklein123 It's late and I may have missed something, but I think we still need weak_ptrs to the PBF because we actually do need to have a handle to it both in the platform callbacks, and in the dispatcher postcb. It's already managed as a shared_ptr so share_from_this also seems required to ensure we don't mess up the ownership graph. (But it's entirely possible I've misunderstood or overlooked something.)

Anyways, because of that I ended up going with the boolean check, which was trivial, guaranteed to be on the main thread, and seems like it should close the door on this issue (we use a similar check to ensure we don't accidentally resume a filter where iteration is already ongoing).

goaway:
And of course, @mattklein123, thanks again for your help tracking this down!

library/common/extensions/filters/http/platform_bridge/filter.h
```diff
@@ -160,6 +162,7 @@ class PlatformBridgeFilter final : public Http::PassThroughFilter,
   envoy_http_filter_callbacks platform_request_callbacks_{};
   envoy_http_filter_callbacks platform_response_callbacks_{};
   bool error_response_{};
+  bool alive_{true};
```
mattklein123, line 165: This is fine and should work. Just for illustrative/educational purposes, I was suggesting the following:
1) `std::shared_ptr<BridgeFilterHandle> still_alive_`;
2) `BridgeFilterHandle` contains a reference to the filter.
3) Return a weak `BridgeFilterHandle` everywhere you now return a weak filter using `weak_from_this`.
4) Locking the weak `BridgeFilterHandle` is sufficient to understand liveiness if you `still_alive.reset()` on `onDestroy()`.

The mild benefit here is I think it's a bit more intentional what you are doing with less boolean logic, but again this is fine.

mattklein123 (approved):
Nice!

PR #1496, opened 2021-05-28
configs: config overhaul phase 1

goaway:
Description: This change creates a config header that is skipped over by Envoy's config parser but can contain dynamic and/or heavyweight yaml anchors to be referenced by alias in the config template. This results in significantly less overhead when composing config. Phase 1 of the config overhaul initially moves TLS certs to the header, since they are by far the most costly component of interpolation at present.
Risk Level: Moderate
Testing: Local & CI


mattklein123 (dismissed):
Very nice! :)

goaway:
thanks @mattklein123!

test/common/engine_test.cc
```diff
@@ -12,21 +12,41 @@ typedef struct {
   absl::Notification on_exit;
 } engine_test_context;
 
+
```
buildbreaker, line 15:
```suggestion
```

library/common/BUILD
```diff
@@ -4,6 +4,19 @@ licenses(["notice"])  # Apache 2
 
 envoy_package()
 
+envoy_cc_library(
+    name = "config_lib",
+    srcs = [
+        "certificates.inc",
```
junr03, line 10: additional clean up: should we move these files to a subdir?

library/common/config_template.h
```diff
@@ -0,0 +1,46 @@
...
+extern const char* native_filter_template;
+
+/**
+ * Template that enables the route cache reset filter in the chain.
+ * Should only be added when the route cache should be cleared on every request
+ * going through the filter chain between initial route resolution and the router
+ * filter's invocation on the request path. Typically only used for enabling
+ * direct responses to mutate headers which are then later used for routing.
+ */
+extern const char* route_cache_reset_filter_template;
```
junr03, line 29: maybe we should move the templates that are intended for test usage to a different header.

library/common/engine.cc
```diff
@@ -32,23 +33,31 @@ envoy_status_t Engine::run(const std::string config, const std::string log_level
 envoy_status_t Engine::main(const std::string config, const std::string log_level) {
   // Using unique_ptr ensures main_common's lifespan is strictly scoped to this function.
   std::unique_ptr<EngineCommon> main_common;
+  const std::string name = "envoy";
+  const std::string config_flag = "--config-yaml";
+
+  size_t header_size = config_header.size();
```
junr03, line 39: up to you about this vs using std:string and abseils concat utils. My paranoid stance is to go for the latter.

cc @mattklein123 in case he has a preference.

library/common/engine.cc
```diff
@@ -59,10 +68,13 @@ envoy_status_t Engine::main(const std::string config, const std::string log_leve
 
       cv_.notifyAll();
     } catch (const Envoy::NoServingException& e) {
+      free(composed_config);
```
junr03, line 71: the above comment would also save us from these frees :)

library/common/config_template.cc
```diff
@@ -70,6 +70,19 @@ const char* stats_sink_template = R"(
           cluster_name: stats
 )";
 
+const std::string config_header = R"(
```
junr03, line 73: I dont think I fully understand why this become the config header? And its relation to the config_template.

library/common/BUILD
goaway: There's significant cleanup coming in phases 2/3, I propose we defer till then since a lot of this changes.

library/common/config_template.h
goaway: same as above

library/common/engine.cc
goaway: I switched to MemBlockBuilder which is Envoy-provided and avoids raw memcpy.

library/common/engine.cc
goaway: Honestly we don't need them at all with the PANIC happening below, but I put it here in case we change the behavior again. Actually now that I think about it, I handle the memory via a unique_ptr so I'll do that.

library/common/config_template.cc
goaway: Let me expand in the PR description.

goaway: Updated.

library/common/engine.cc
goaway: Updated.

library/common/engine.cc
goaway: Switching to absl::StrCat since I think the docs make it sound trustworthy enough. :p

library/common/config_template.h
goaway: To expand on this - a number of these templates are going away entirely. Interpolation itself will be going away entirely.

goaway:
We've been creeping up on the size warning threshold for a while, so bumping the limit here, since we've finally hit it.
Actually this PR reduces the source size by about 20K, so I don't know why the compiled binary is ending up larger. Maybe we're losing some compression efficiency with our gzip settings, and they can be tweaked.

Overall, I'd be surprised if the config overhaul doesn't end up net reducing binary size with phases 2 and 3, since we'll be able to remove a bunch of templating and interpolation code.

junr03 (approved):

Issue #1497, opened 2021-05-31
signal handling: allow Envoy Logger to be ostream for signal handling


junr03:
Currently Event::SignalAction installs std::cerr as the only ostream passed during signal handling on termination. For Envoy Mobile it would be far more useful to allow Envoy Logger to act as ostream so crash scopes could be output to platform-specific crash APIs.

junr03:
This could be done by having the fatal action pass an std::stringstream and then logging that. However, overriding `callFatalErrorHandlers` is not entirely straightforward. Another alternative would be to derive SignalAction and override `sigHandler`

PR #1498, opened 2021-05-31
scope tracker: attach scope trackers to Http:Client and PlatformBridgeFilter

junr03:
Description: scope tracker - attach scope trackers to Http:Client and PlatformBridgeFilter
Risk Level: med - changing crash clean up routine
Testing: example app. Integration test pending.


library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -151,8 +152,17 @@ void PlatformBridgeFilter::onDestroy() {
   platform_filter_.instance_context = nullptr;
 }
 
+void PlatformBridgeFilter::FilterBase::dumpState(std::ostream&, int) const {
+  // TODO(junr03): output to ostream - https://github.com/envoyproxy/envoy-mobile/issues/1497.
+  ENVOY_LOG(error, "PlatformBridgeFilter: {} state={} stream_complete={} error_response={}",
```
junr03, line 157: @goaway anything else you would want to see related to the PBF state?

goaway: Status of the buffer (does it exist? is anything buffered? how much?) would be great.

goaway: Also whether headers/data/trailers have been forwarded.

junr03: sg. Will unearth.

junr03: @goaway this is updated!

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -153,8 +154,42 @@ void PlatformBridgeFilter::onDestroy() {
...
+    auto buffer = request_filter_base_->buffer();
+    ENVOY_LOG(
+        error,
+        "  Request Filter state={} on_headers_called={} on_data_called={} on_trailers_called={} "
+        "on_resume_called={} buffer={} stream_complete={}",
+        (request_filter_base_->iteration_state_ == IterationState::Ongoing ? "ongoing" : "stopped"),
+        request_filter_base_->on_headers_called_, request_filter_base_->on_data_called_,
+        request_filter_base_->on_trailers_called_, request_filter_base_->on_resume_called_,
+        (buffer ? fmt::format("{} bytes", buffer->length()) : "no buffer"),
+        request_filter_base_->stream_complete_);
```
goaway, line 170:
```suggestion
    request_filter_base_->dumpState(the_ostream, that_int);
```

library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -153,8 +154,42 @@ void PlatformBridgeFilter::onDestroy() {
...
+  } else {
+    ENVOY_LOG(error, "  Request Filter absent");
```
goaway, line 172: "absent" sort of makes it sound like it's _supposed_ to be there, which it isn't necessarily. I'd probably just leave this blank, or if you feel it's still helpful, say something like "No request filter."

library/common/extensions/filters/http/platform_bridge/filter.h
```diff
@@ -114,9 +119,13 @@ class PlatformBridgeFilter final : public Http::PassThroughFilter,
     IterationState iteration_state_;
     PlatformBridgeFilter& parent_;
     envoy_filter_on_headers_f on_headers_;
+    bool on_headers_called_{};
```
goaway, line 122: I'd actually be more curious if these have been _forwarded_ than called.

In other words, you'd have to have continued/resumed iteration and have them not be pending.

library/common/extensions/filters/http/platform_bridge/filter.cc
junr03: good call.

junr03:
@goaway updated! re: using the ostream arg, I opened https://github.com/envoyproxy/envoy-mobile/issues/1497. I want to land this and then think about that if you are ok with that.

library/common/extensions/filters/http/platform_bridge/filter.h
junr03: I think having iteration_state + on_resume_called + the state of buffer/pending_<headers/trailer> does the trick. wdyt?

junr03:
note: this PR depends on https://github.com/envoyproxy/envoy/pull/16709

test/common/extensions/filters/http/platform_bridge/platform_bridge_filter_test.cc
```diff
@@ -95,6 +126,10 @@ TEST_F(PlatformBridgeFilterTest, NullImplementation) {
   Http::TestResponseTrailerMapImpl response_trailers{{"x-test-trailer", "test trailer"}};
   EXPECT_EQ(Http::FilterTrailersStatus::Continue, filter_->encodeTrailers(response_trailers));
 
+  checkFilterState("NullImplementation", "0",
```
junr03, line 129: Will add throughout the rest of the test suite once we are finalized on all the state we want to keep track of and dump.

library/common/event/provisional_dispatcher.cc
```diff
@@ -49,5 +49,21 @@ void ProvisionalDispatcher::deferredDelete(DeferredDeletablePtr&& to_delete) {
   event_dispatcher_->deferredDelete(std::move(to_delete));
 }
 
+void ProvisionalDispatcher::pushTrackedObject(const ScopeTrackedObject* object) {
+  if (event_dispatcher_) {
```
goaway, line 53: This access is not inherently threadsafe, though it will likely only be called from a valid context. Still, we might want to have a safety check in these calls (like we do with deferredDelete).

library/common/event/provisional_dispatcher.cc
```diff
@@ -49,5 +49,21 @@ void ProvisionalDispatcher::deferredDelete(DeferredDeletablePtr&& to_delete) {
...
+  }
+}
+
+void ProvisionalDispatcher::popTrackedObject(const ScopeTrackedObject* expected_object) {
+  if (event_dispatcher_) {
+    event_dispatcher_->popTrackedObject(expected_object);
+  }
+}
+
+bool ProvisionalDispatcher::trackedObjectStackIsEmpty() const {
```
goaway, line 64: When does this get called?

goaway:
Overall looks good, just some concerns around threadsafety when accessing the underlying dispatcher.

library/common/event/provisional_dispatcher.cc
goaway: Based on your check below, we might want to lock here, depending on the usage.

library/common/event/provisional_dispatcher.cc
junr03: currently scopes are only created in the threading context of the event dispacther. I removed all if statements and put in asserts like in deferred delete

library/common/event/provisional_dispatcher.cc
junr03: This actually never gets currently called. I could change usage to NOT_REACHED. But for now I changed to isThreadSafe assertion.

test/common/extensions/filters/http/platform_bridge/platform_bridge_filter_integration_test.cc
```diff
@@ -68,67 +68,67 @@ class PlatformBridgeIntegrationTest : public testing::TestWithParam<Network::Add
...
-  upstream_request_->encodeData(100, false);
-  upstream_request_->encodeData(100, true);
-
-  // Wait for frames to arrive downstream.
-  ASSERT_TRUE(response->waitForEndStream());
-
-  EXPECT_TRUE(response->complete());
-  EXPECT_EQ("200", response->headers().Status()->value().getStringView());
-}
+// TEST_P(PlatformBridgeIntegrationTest, MultipleFilters) {
```
junr03, line 71: @goaway given how request and responses are passed in threading contexts in the integration text framework, I am having trouble getting the filters to run in the thread of their own dispatcher without stalling the test. 

If you are ok with it, could we maybe land this PR, and I can untangle the threading issue in this integration test in a subsequent PR?

junr03: I can open a ticket to track

goaway: Yes, please open a tracking issue/ticket.

test/common/http/client_test.cc
```diff
@@ -725,11 +800,15 @@ TEST_F(ClientTest, EncodeMetadata) {
...
   // Send request headers.
+  EXPECT_CALL(dispatcher_, pushTrackedObject(_));
+  EXPECT_CALL(dispatcher_, popTrackedObject(_));
   envoy_headers c_headers = defaultRequestHeaders();
   EXPECT_CALL(request_decoder_, decodeHeaders_(_, true));
   http_client_.sendHeaders(stream_, c_headers, true);
 
   // Encode response headers.
+  EXPECT_CALL(dispatcher_, pushTrackedObject(_));
+  EXPECT_CALL(dispatcher_, popTrackedObject(_));
```
goaway, line 811: Don't need to block on this, but we might just want to adopt `NiceMock` behavior for these calls so we don't have to count them everywhere in tests.

goaway (dismissed):

goaway (approved):

goaway:
Nice work!

test/common/extensions/filters/http/platform_bridge/platform_bridge_filter_integration_test.cc
junr03: https://github.com/envoyproxy/envoy-mobile/issues/1533

PR #1501, opened 2021-06-01
python: expose the option to pre-build an engine in envoy_requests

crockeo:
**Description:** This PR does 2 things:

1. Has individual implementations create the Engine using their method of asynchrony (i.e. asyncio uses asyncio, gevent uses gevent, threading uses threading--instead of everything using threading for Engine construction)
2. Exposes a `pre_build_engine`, which lets users of `envoy-requests` pay the cost of building an engine _before_ the first request. This will help smooth out the first couple of requests a newly deployed service makes by paying the cost of booting the engine _before_ it's ready to accept connections.

**Risk Level:** Low

**Testing:** CI checks include testing via existing unit/integration tests + standalone tests that check each implementation's `pre_build_engine`. Also manual testing locally on my computer.

**Docs Changes:** N/A
**Release Notes:** N/A


crockeo:
I merged 7c4f0a5 to try to get more visibility into why this was failing on CI (couldn't repro locally) but was unable to make it fail after merging that commit. Reverted so this is the intended state of the PR for review.

goaway (approved):

PR #1505, opened 2021-06-03
engine: convert Envoy exceptions to crashes

goaway:
Description: More than once we've been bitten by a complete failure to start up manifesting as a simple log statement and then broken software that simply continues to run without networking. There's maybe an argument to be made for making this configurable, but for now, making exceptions fatal seems like a saner default.
Risk Level: Moderate
Testing: Local & CI


rebello95 (approved):

goaway:
This will generally allow broken integration tests to fail faster as well, as opposed to them often waiting out the timeout.

goaway:
thanks @rebello95!

PR #1506, opened 2021-06-03
python: hide symbols except for PyInit on macOS

crockeo:
Description: This PR modifies the build recipe for `envoy_engine.so` to hide all symbols except for `_PyInit_envoy_engine` by default on maOS, so that envoy_engine can be loaded alongside a [protobuf](http://pypi.org/project/protobuf) wheel without breaking.

Loading in protobuf alongside `envoy_engine.so` would register two sets of the same symbols and, on macOS, the runtime linker chooses the first symbol it finds causing problems. See the [similary protobuf change](https://github.com/protocolbuffers/protobuf/pull/8346) and its [sister change in grpc](https://github.com/grpc/grpc/pull/24992) for more information on what's going on.

Risk Level: Low
Testing: See the now-closed #1504 for how I've been testing this.
Docs Changes: N/A
Release Notes: N/A


crockeo:
reverting the minimum version bump since this doesn't affect prod and we don't need to pull it in until we also have other important changes

goaway (approved):

PR #1510, opened 2021-06-04
testing/ignore: break example app

rebello95:


rebello95:
Verified that the example iOS app fails on CI with this change

PR #1513, opened 2021-06-09
explicit flow control mode for cronvoy

alyssawilk:
Description: adding explicit flow control mode for cronvoy
Risk Level: high (changing stream lifetime)
Testing: new unit tests
Docs Changes: n/a
Release Notes: n/a


alyssawilk:
well that's fascinating, git amend -S isn't working.
I'll go debug and fix, but shouldn't be blocking

carloseltuerto:
There is one another detail: when a connection failure is detected by Envoy-Mobile, the onError() callback is invoked immediately. What Cronet does however is stalling the failure callback until the caller re-enables the callbacks. So this is one scenario:

Erroneous:
- EnvoyMobile invokes callback onHeaders()
- Internet interrupted / socket closed without remission
- EnvoyMobile invokes callback onError()
- Caller does not need to re-enable callbacks - Cronet breach of contract

Correct:
- EnvoyMobile invokes callback onHeaders()
- Internet interrupted / socket closed without remission
- EnvoyMobile does nothing, waiting for the caller to re-enable callbacks
- Caller re-enables callbacks
- EnvoyMobile invokes callback onError()

Short term this can be circumvented with some Java code - the EnvoyMobile callback of onError() can be put in an "internal" queue until the caller re-enables the callbacks, and then the onError() callback gets invoked. And maybe this is fine long term too, since the main objective of this PR is enabling flow control for the response body.

Now, for the Request Body, flow control is also needed. The caller can easily overflow EnvoyMobile - there is even a test called "send one Terabyte" to make sure that the C++ layer pauses the next callback for "more data". Surely another PR.

alyssawilk:
yeah, definitely want to do request push back in another PR, once we have a good handle on the response path :-)

The error case is interesting.  If there's an error before headers, is that sent up immediately or do all errors wait for a "resume" call?  I could defer them in async mode, at which point I think we'd want resumeData to be more of a resumeCallbacks though still with max bytes.

I think the shim layer is also going to need to do some permanent handling of trailers because AFIK Cronet doesn't handle trailers and Envoy does.  I think that means that functionally trailers should be converted in the Cronvoy layer to empty data + end stream.   Right now, Envoy sends trailers up immediately after body data, but would you prefer I switch it so again the caller has to ask for "data" (where here it'd be always waiting for a FIN)?  That way you don't have to do queuing there.

cc @DavidSchinazi @RyanTheOptimist for the discussion

RyanTheOptimist:
> 2. never sending data up in async mode without a call to resumeData(max_bytes) at which point up to max_bytes will be sent up once (immediately if available, else next time there is data)

One question about this. If data is available immediately, is the data passed up via a return value or via a callback? I assume the latter, and I *think* that's what I see in resumeData(). This is probably fine and envoy does seem to invoke callback synchronously in cases like this so it definitely seems par for the course. But it makes me wonder if the callback invoked by resumeData() is allowed to call resumeData()? I could imagine the OnData handler pushing the data up to the app and then calling resumeData() again to keep the data flowing. But I'm not sure how well resumeData() would work in that case? Is this something we should think about/document?

alyssawilk:
I think but would ask @goaway or @junr03 to confirm, that when we send data to the bridge that's already an async call, so I don't think you can end up in a stack loop if that's the concern.  I can always post a callback if I'm misunderstanding the model though.

carloseltuerto:
The flow control dance for the response body only starts once the onHeaders() callback has been invoked. Before that, getting an onError() callback invocation is a valid outcome. In other words, for an HTTP GET, once the request header has been sent, there are two mutually exclusive outcomes: onHeaders() callback is invoked, or onError() callback is invoked.

Avoiding queuing is always better. So for the trailer headers, it is fine for the Cronvoy shim to handle the onTrailerHeaders callback, and bonus points if no queuing is needed. But one sure thing, whatever we do here, this won't break the "flow control" contract, so this PR is a step in the right direction.

FYI Mike and José: the Cronet API at the JNI level never exposes the body data is got already when invoking the callback with the response header. With Cronet, the caller must bring its own ByteBuffers. And the way for the caller to provide its ByteBuffer is by invoking [UrlRequest.read(ByteBuffer b)](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/components/cronet/android/api/src/org/chromium/net/UrlRequest.java#369).  The "UrlRequest.read" invocation by the caller is the one re-enabling the next callback. EnvoyMobile must play that game. With this PR, the only difference at this point is that the caller only provides the max amount of data it is ready to handle, not a ByteBuffer. That's fine for now.

alyssawilk:
Ok, in that case I'll definitely push back on trailers and take a look at delaying error callbacks too.  I'll ping when it's ready for another pass :-)

alyssawilk:
just unit tests - still haven't tweaked push back

/wait

alyssawilk:
Ok, I think this is where we want it - PTAL at your convenience

alyssawilk:
cc @goaway

carloseltuerto:
This looks good to me. I suspect that providing a buffer instead of a "length" would not be much more efforts if we just copy copy the data over (no attempt to plumb this though at first). But let's do that later, if needed.

Thanks.

library/common/data/utility.cc
```diff
@@ -20,8 +28,9 @@ Buffer::InstancePtr toInternalData(envoy_data data) {
   return buf;
 }
 
-envoy_data toBridgeData(Buffer::Instance& data) {
-  envoy_data bridge_data = copyToBridgeData(data);
+envoy_data toBridgeData(Buffer::Instance& data, uint32_t max_bytes) {
```
RyanTheOptimist, line 31: This is an odd name for a type. Does Envoy Mobile use a different style guide? (not that this has anything to do with your PR)

RyanTheOptimist (approved):
Looks great! two tiny questions.

library/common/http/client.cc
```diff
@@ -95,14 +104,53 @@ void Client::DirectStreamCallbacks::encodeData(Buffer::Instance& data, bool end_
...
+
+  // If not all the bytes have been sent up, buffer any remaining data in response_data.
+  if (data.length() != 0) {
+    ASSERT(async_mode_);
+    if (!response_data_) {
+      response_data_ = std::make_unique<Buffer::WatermarkBuffer>(
+          [this]() -> void { this->bufferedDataDrained(); },
+          [this]() -> void { this->hasBufferedData(); }, []() -> void {});
+      // Default to 64K per stream.
+      response_data_->setWatermarks(1000000);
```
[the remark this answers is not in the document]
alyssawilk: Will fix comment for accuracy.  it's pretty arbitrary based off cronet 6m defaults.
H2 will handle watermark + stream-window-update worth of data, where H1 will stop reading remotely so do watermark + tcp buffer limit. if we set this to 6m we'd buffer way more, so this felt like a reasonable compromise.

library/common/data/utility.cc
alyssawilk: AFIK the bridge is c structures so yeah different style for that part

library/common/http/client.cc
```diff
@@ -112,19 +162,62 @@ void Client::DirectStreamCallbacks::encodeTrailers(const ResponseTrailerMap& tra
...
+    onError();
+    deferred_error_ = false;
+  }
+}
+
 void Client::DirectStreamCallbacks::closeStream() {
   // Envoy itself does not currently allow half-open streams where the local half is open
   // but the remote half is closed. Note that if local is open, Envoy will reset the stream.
   http_client_.removeStream(direct_stream_.stream_handle_);
+  end_stream_read_ = true;
```
snowp, line 216: is closeStream equivalent to seeing an end_stream? If not, should this be called `closed_stream_` or something like that?

library/common/http/client.cc
```diff
@@ -112,19 +162,62 @@ void Client::DirectStreamCallbacks::encodeTrailers(const ResponseTrailerMap& tra
...
+  // Make sure to send end stream with data only if
+  // 1) it has been received from the peer and
+  // 2) there are no trailers
+  if ((response_data_.get() && response_data_->length() != 0) ||
+      (end_stream_read_ && !end_stream_communicated_)) {
+    sendDataToBridge(*response_data_, end_stream_read_ && !response_trailers_.get());
+  }
+
+  // If all buffered data has been sent, send and free up trailers.
+  if ((!response_data_.get() || response_data_->length() == 0) && response_trailers_.get()) {
```
snowp, line 201: nit: maybe add a function `hasBufferedResponseData` or something to make these if blocks easier to read?

snowp, line 201: I think we can possibly send the trailers immediately following the last bit of data, is that intended? Seems like we gate errors on bytes_to_send_ but not trailers?

snowp (commented):
Looks pretty good to me, a few questions/nits

library/common/http/client.cc
```diff
@@ -95,14 +104,55 @@ void Client::DirectStreamCallbacks::encodeData(Buffer::Instance& data, bool end_
...
+
+  // If not all the bytes have been sent up, buffer any remaining data in response_data.
+  if (data.length() != 0) {
+    ASSERT(async_mode_);
+    if (!response_data_) {
+      response_data_ = std::make_unique<Buffer::WatermarkBuffer>(
+          [this]() -> void { this->bufferedDataDrained(); },
+          [this]() -> void { this->hasBufferedData(); }, []() -> void {});
+      // Default to 1M per stream. This is fairly arbitrary and will result in
+      // Envoy buffering up to 1M + flow-control-window for HTTP/2 and HTTP/2,
```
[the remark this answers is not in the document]
alyssawilk: Haha, off by one error - HTTP/2 and HTTP/3 :p-P

library/common/http/client.cc
alyssawilk: yeah trailers should be gated too - will regression test.

library/common/http/client.cc
alyssawilk: close stream isn't a thing in HTTP, so from a spec perspective it's reading the end stream bit, and communicating it.  added some comments in the .h file.  WDYT?

test/common/http/client_test.cc
```diff
@@ -384,17 +399,9 @@ TEST_F(ClientTest, BasicStreamTrailers) {
...
   cc_.end_stream_with_headers_ = false;
 
-  bridge_callbacks_.on_data = [](envoy_data data, bool, void* context) -> void* {
-    // TODO: assert end_stream and contents of c_data for multiple calls of on_data.
-    callbacks_called* cc = static_cast<callbacks_called*>(context);
-    cc->on_data_calls++;
-    data.release(data.context);
-    return nullptr;
-  };
-
```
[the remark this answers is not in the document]
alyssawilk: correct - the default function now accumulates the body and I added content checks below

library/common/data/utility.cc
goaway: That's correct - we currently have standardized on the C bridge types (both structs and aliases) using snake case. It's not set in stone, but it helps to distinguish them.

library/common/http/client.cc
```diff
@@ -117,19 +164,63 @@ void Client::DirectStreamCallbacks::encodeTrailers(const ResponseTrailerMap& tra
...
+  bytes_to_send_ = bytes_to_send;
+
+  ENVOY_LOG(debug, "[S{}] received resume data call for {} bytes", direct_stream_.stream_handle_,
+            bytes_to_send_);
+
+  // If there is bufferd data, send up to bytes_to_send bytes.
+  // Make sure to send end stream with data only if
+  // 1) it has been received from the peer and
+  // 2) there are no trailers
+  if (hasBufferedData() || (end_stream_read_ && !end_stream_communicated_ && !response_trailers_)) {
```
goaway, line 197: Should some or all of this maybe just be an ASSERT? For example, we should never have received end_stream on a data frame if we received trailers. Similarly, I saw that you're guarding elsewhere that you don't forward end_stream if you need to buffer data.

goaway:
Regarding docs changes we've discussed elsewhere, I'd suggest putting them here https://envoy-mobile.github.io/docs/envoy-mobile/latest/api/api.html under a new 'Explicit Buffering' section after the callbacks are detailed. I also think it's fine if we add it in a separate PR.

library/common/http/client.cc
alyssawilk: I think these are legit conditions :-)
we get "end stream" either with a data FIN or with trailers so this is basically saying "if we got end stream not via trailers" send up empty-data-with-fin.
For sendDataToBridge if we have 12k+fin buffered (no trailers) and we're only allowed to send 8k up we need to not send the end stream ahead of the last 4k

alyssawilk:
ohhh, this asan bug is a real thing.
So currently any time end stream is communicated or an error is communicated, there's an immediate call to removeStream which deferred deletes all the memory (before we're actually done with it, yuck), we do the up calls, and next dispatcher loop clears.

For Cronvoy mode, we need things to stick around until end stream / error has been communicated up.  Defering the remove isn't enough though because presumably the remove stops down-calls, so we need to guard all the down-calls against "deferred deferred delete".  Thoughts on if there's a more elegant way to do that or should I just take a whack at it?

alyssawilk:
Amusingly the test fail was something completely different, haha, but anyway I'm now defering delete until after end_stream/reset has been communicated up.  I think this makes it a high risk refactor since it's no longer a no-op for E-M so tagging accordingly.

library/common/http/client.cc
goaway: Got it - I initially misread this as more of a guard than valid branch.

library/common/http/client.cc
```diff
@@ -99,14 +108,52 @@ void Client::DirectStreamCallbacks::encodeData(Buffer::Instance& data, bool end_
     return;
   }
```
goaway, line 110: How would you feel about this?

```suggestion
if (!explicit_buffering_) {
  bytes_to_send_ = data.length();
}

```

Then we could always rely on bytes_to_send_, we could always do a simple `std::min` before sending, and you wouldn't need to treat 0 as a special value (which feels a little bit risky to me, even though I see where it's guarded).

library/common/http/client.h
```diff
@@ -135,16 +141,53 @@ class Client : public Logger::Loggable<Logger::Id::http> {
...
+    void setExplicitlyBuffering() { explicit_buffering_ = true; }
+    void sendDataToBridge(Buffer::Instance& data, bool end_stream);
+    void sendTrailersToBridge(const ResponseTrailerMap& trailers);
+
     DirectStream& direct_stream_;
     const envoy_http_callbacks bridge_callbacks_;
     Client& http_client_;
     absl::optional<envoy_error_code_t> error_code_;
     absl::optional<envoy_data> error_message_;
     absl::optional<int32_t> error_attempt_count_;
```
goaway, line 175: These optionals actually already represents a pending error - can we use this state instead of introducing `deferred_error_`?

library/common/http/client.h
```diff
@@ -135,16 +141,53 @@ class Client : public Logger::Loggable<Logger::Id::http> {
...
+    ResponseTrailerMapPtr response_trailers_;
+    // True if the bridge should operate in explicitly buffering mode, and only send
+    // data when it is requested by the caller.
+    bool explicit_buffering_{};
+    bool response_headers_sent_{};
+    // Called in closeStream() to communicate that the end of the stream has
+    // been received by the DirectStreamCallbacks.
+    bool end_stream_read_{};
+    // Set true when the end stream has been communicated up the bridge.
+    bool end_stream_communicated_{};
```
goaway, line 188: nit:
`response_headers_forwarded_`
`end_stream_received_`
`end_stream_forwarded_`

would align slightly better with terminology we've used elsewhere.

library/common/http/client.h
```diff
@@ -221,10 +264,21 @@ class Client : public Logger::Loggable<Logger::Id::http> {
   Event::ProvisionalDispatcher& dispatcher_;
   HttpClientStats stats_;
   absl::flat_hash_map<envoy_stream_t, DirectStreamSharedPtr> streams_;
+  absl::flat_hash_map<envoy_stream_t, DirectStreamSharedPtr> draining_streams_;
```
goaway, line 267: Is this being used?

library/common/http/client.cc
```diff
@@ -117,19 +164,60 @@ void Client::DirectStreamCallbacks::encodeTrailers(const ResponseTrailerMap& tra
...
+    bytes_to_send_ = 0;
+  }
+
+  if (deferred_error_ && bytes_to_send_ > 0) {
+    onError();
+    deferred_error_ = false;
+  }
 }
 
+void Client::DirectStreamCallbacks::closeStream() { end_stream_read_ = true; }
```
goaway, line 216: Okay, so I'm guessing the intent of `draining_streams_` was that the stream should be moved there if `explicit_buffering_` is enabled in this function. We've had a few bugs around the `streams_` map becoming out of sync with Envoy's view of the world, so that still seems like it might be worthwhile. That way we can preserve the convention that we update `streams_` to be in sync with Envoy at the beginning of every callback.

library/common/http/client.cc
```diff
@@ -117,19 +164,60 @@ void Client::DirectStreamCallbacks::encodeTrailers(const ResponseTrailerMap& tra
...
+
 void Client::DirectStreamCallbacks::onComplete() {
+  http_client_.removeStream(direct_stream_.stream_handle_);
```
goaway, line 219: If you make the change above in `closeStream`, don't forget update this to remove the stream from `draining_streams_` iff `explicit_buffering_` is enabled.

goaway:
@alyssawilk this is looking good. I did a close pass, and I think the comments I've left above cover any remaining thoughts/concerns I'd have.

goaway (approved):

library/common/http/client.cc
goaway: I think I'm maybe missing something. Why would this introduce an upcall for cronvoy or introduce additional context switches? I'm just suggesting adding these lines so that later we don't need extra checks on `bytes_to_send_` - we can rely on it being the correct value regardless of whether `explicit_buffering_` is enabled.

library/common/http/client.h
goaway: It's okay to clear them. Let's remove the bool.

goaway (changes requested):
Sorry, I jumped the gun a little bit on the previous approval.

I'm still a little bit concerned about the change to the timing the stream remains in the stream map. We've had some tricky crashes when that map stopped aligning 100% with Envoy's internal state. I do think the draining map approach you started on would be safer. It keeps the stream alive, but prevents any further interaction with internal Envoy state (which is no longer safe).

library/common/http/client.cc
goaway: I think it still is safer to take the draining approach - see the checks in the request path that all confirm if the stream is still there. There are some weird edge cases with timing around hanging requests, gRPC streaming, local replies, filters stopping iteration, etc. where it's been problematic if the map doesn't align strictly with Envoy's internal view on the stream's existence.

alyssawilk:
Yeah so I'd started down that path and backed out, because fundamentally we use the stream map for 2 things, getting the stream for downstream->upstream communication, and getting the stream for upstream->downstream communication.

When we're in this "closed but not propagated" mode, the latter is still necessary while the former is largely nonsensical. If we want to guard against the latter, I think we need to tweak getStream to have a directionality, and either have 2 maps (only consult one for downstream->upstream) or have 2 map but check "half closed" for downstream to upstream.  I'd be inclined to do the latter, but I'll start the directional split and the move would be pretty easy if you prefer the former.

library/common/http/client.h
```diff
@@ -42,11 +43,13 @@ struct HttpClientStats {
 class Client : public Logger::Loggable<Logger::Id::http> {
 public:
   Client(ApiListener& api_listener, Event::ProvisionalDispatcher& dispatcher, Stats::Scope& scope,
-         std::atomic<envoy_network_t>& preferred_network, Random::RandomGenerator& random)
+         std::atomic<envoy_network_t>& preferred_network, Random::RandomGenerator& random,
+         bool explicit_buffering = false)
```
mattklein123, line 47: I realize there are no API docs previously, but can you add some comments on what "explicit buffering" is? It's not clear to from a quick skim what this parameter does.

Edit: after reading I think I understand what this does. I might actually call this `explicit_flow_control`? IMO that would be a bit more clear (along with some API docs)? WDYT?

library/common/http/client.h
```diff
@@ -135,16 +141,55 @@ class Client : public Logger::Loggable<Logger::Id::http> {
...
+    // True if the bridge should operate in explicitly buffering mode, and only send
+    // data when it is requested by the caller.
+    bool explicit_buffering_{};
+    // Set true when the response headers have been forwarded to the bridge.
+    bool response_headers_forwarded_{};
+    // Called in closeStream() to communicate that the end of the stream has
+    // been received by the DirectStreamCallbacks.
+    bool end_stream_received_{};
+    // Set true when the end stream has been forwarded to the bridge.
+    bool end_stream_forwarded_{};
```
mattklein123, line 190: I might make these `remote_end_stream_`... for clarity?

library/common/http/client.h
```diff
@@ -213,7 +258,20 @@ class Client : public Logger::Loggable<Logger::Id::http> {
 
   using DirectStreamWrapperPtr = std::unique_ptr<DirectStreamWrapper>;
 
-  DirectStreamSharedPtr getStream(envoy_stream_t stream_handle);
+  enum Permissions {
```
mattklein123, line 261: nit: I would probably call this `GetStreamFilters` or something like that so it's more specific to the function and also "permissions" has ACL connotations.

library/common/http/client.cc
```diff
@@ -99,37 +108,117 @@ void Client::DirectStreamCallbacks::encodeData(Buffer::Instance& data, bool end_
...
+void Client::DirectStreamCallbacks::resumeData(int32_t bytes_to_send) {
+  ASSERT(explicit_buffering_);
+  ASSERT(bytes_to_send > 0);
+
+  bytes_to_send_ = bytes_to_send;
+
+  ENVOY_LOG(debug, "[S{}] received resume data call for {} bytes", direct_stream_.stream_handle_,
+            bytes_to_send_);
+
+  // If there is bufferd data, send up to bytes_to_send bytes.
```
mattklein123, line 194:
```suggestion
  // If there is buffered data, send up to bytes_to_send bytes.
```

mattklein123 (changes requested):
Cool stuff. A few small comments.

> When we're in this "closed but not propagated" mode, the latter is still necessary while the former is largely nonsensical. If we want to guard against the latter, I think we need to tweak getStream to have a directionality, and either have 2 maps (only consult one for downstream->upstream) or have 2 map but check "half closed" for downstream to upstream. I'd be inclined to do the latter, but I'll start the directional split and the move would be pretty easy if you prefer the former.

All things being equal I think I mildly prefer 2 maps for clarity, but I think what you have with the "permissions" guard is fine also and pretty clear to me. (I agree with @goaway that this code is pretty tricky and hard to reason about, so whatever we can do to improve clarity is probably worth it.)

library/common/http/client.cc
alyssawilk: yeah this is why I actually held off - moving things around and having 2 maps felt more risky to me, but happy to give it a shot and if you two think it's cleaner the way it was we can always undo it :-)

library/common/http/client.h
goaway: @mattklein123, we had called it `flow_control_enabled` previously, but @RyanTheOptimist noted that that could be misinterpreted as enabling/disabling HTTP flow control. I actually think `explicit_flow_control` is an improvement because it maybe carries less of that implication, while "flow control" still somewhat better captures the intent.

RyanTheOptimist: Yeah I have angst about flow control being an ambiguous term in this space. That being said, none of the names are ideal and I agree with @goaway that `explicit_flow_control` is an improvement from `flow_control_enabled`, so it works for me.

goaway (approved):
Thanks @alyssawilk!

PR #1515, opened 2021-06-10
configs: config overhaul phase 2

goaway:
Description: Removes most config interpolation, excepting custom listeners, clusters, filters, and routes. In normal usage only custom filters will generally be present. Library defaults are baked into the base configuration, and passing empty config to the engine is now valid.
Risk Level: Moderate
Testing: Local & CI


goaway:
(Phase 3 will introduce json-based extensions and remove remaining interpolation.)

library/common/config/config.cc
```diff
@@ -0,0 +1,387 @@
...
+          typed_config: {{ native_filter_typed_config }}
+)";
+
+const char* route_cache_reset_filter_insert = R"(
+        - name: envoy.filters.http.route_cache_reset
+          typed_config:
+            "@type": type.googleapis.com/envoymobile.extensions.filters.http.route_cache_reset.RouteCacheReset
+)";
+
+const int custom_cluster_indent = 2;
```
junr03, line 23: I think I am missing how this is used

library/common/config/config.cc
```diff
@@ -0,0 +1,387 @@
...
+const int custom_listener_indent = 2;
+const int custom_filter_indent = 8;
+const int custom_route_indent = 14;
+
+const int fake_remote_response_indent = 14;
+const char* fake_remote_cluster_insert = "  - *fake_remote_cluster\n";
+const char* fake_remote_listener_insert = "  - *fake_remote_listener\n";
+const char* fake_remote_route_insert = "              - *fake_remote_route\n";
+
+const std::string config_header = R"(
```
junr03, line 33: Is there any way for us to verify that the loaded config to the engine is the same? Maybe diff between startup logs, which print out the config being loaded?

junr03: Just as a sanity check

junr03 (commented):
Have you been able to measure improved start up time? I am curious about it.

library/common/config/config.cc
goaway: It's not used yet, but will be used in p3 to help avoid brittle assumptions about indentation in remote code locations (see my TODO elsewhere in this PR).

library/common/config/config.cc
goaway: I'm not sure I completely understand the suggestion - but the config would be very obviously broken if the aliases weren't working.

junr03: I am less concerned about a full break, and more about more subtle problems that might be cause by bad indentation or copy paste. 

If you run the app in debug it will print out the final bootstrap it is loading. So we can diff main and this PR to make sure that the config is 1:1.

goaway:
@snowp fyi

library/common/config/config.cc
```diff
@@ -39,11 +39,29 @@ const std::string config_header = R"(
...
 - &stats_flush_interval 60s
+- &stats_sinks []
+- &statsd_host 127.0.0.1
+- &statsd_port 8125
 - &stream_idle_timeout 15s
 - &virtual_clusters []
 
 !ignore stats_defs:
-  stats_sinks_key: &stats_sinks !ignore stats_sinks
+  base_metrics_service: &base_metrics_service
```
junr03, line 49: cool

library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -113,9 +113,16 @@ String resolveTemplate(final String templateYAML, final String platformFilterTem
         .append(virtualClusters)
         .append("\n");
 
-    if (statsDomain != null) {
-      configBuilder.append("- &stats_domain ").append(statsDomain).append("\n");
+    // TODO(goaway): enable support for both types of sinks, since it's now much easier.
```
junr03, line 116: yeah, this wouldn't be hard now.

test/common/engine_test.cc
```diff
@@ -6,6 +6,40 @@
 
 namespace Envoy {
 
+// This config is the minimal envoy mobile config that allows for running the engine.
```
junr03, line 9: don't forget to cut me a branch to debug the timing issue!

junr03 (approved):

junr03 (dismissed):

goaway:
thanks @junr03!

buildbreaker (approved):
Minor fixes from @junr03 's approval


Thanks @goaway

goaway:
Thanks @junr03 and @buildbreaker!

PR #1527, opened 2021-06-17
http client: use void return type for external API

goaway:
Description: These calls are always dispatched asynchronously, so nothing can consume a return value, and always returned ENVOY_SUCCESS, besides.
Risk Level: Low
Testing: CI


buildbreaker (dismissed):
Nice

PR #1535, opened 2021-06-21
config: add prefetch hostnames

junr03:
Description: adding API to allow adding hostnames to preresolve via the EngineBuilder. 
Risk Level: low
Testing: updated testing


rebello95 (commented):
I think it's worthwhile to avoid hardcoding api.lyft.com into this configuration. Would it be possible to move this to Lyft's private wrapper/repo? If not, what do you think about updating the builders here?

junr03:
> I think it's worthwhile to avoid hardcoding api.lyft.com into this configuration

@rebello95 for sure I agree 100%. My intention with this change is to be able to gather concrete evidence that this config is worth it, and then based on that result come back and delete or open it up to the API layer. I felt that the hardcoding was worth it to not waste cycle and create more API surface for a feature that we haven't definitively proved is needed.

rebello95:
Yea, I saw the original description (thanks for writing that up). I'm concerned about adding this even temporarily though, for the following reasons:
- There are other companies running tests with Envoy Mobile today, and this would affect them adversely/unexpectedly
- Adding Lyft-specific code like this is a slippery slope, even if we intend for it to be temporary

Thoughts?

goaway:
I think I'm with @rebello95 on this - I don't think we should go down this path.

As a shortcut, you could probably just point the internal build at this branch though, since it's here.

junr03:
I mostly want to gather data in a expedient manner to verify that it is worth adding a kotlin/swift API for this.

@goaway is there a way with the new config management scheme to declare an anchor publicly that we define in our private build?

goaway:
Yes, absolutely. If you look in config.cc, you can see that numerous anchors are declared and defaulted in the header.

rebello95 (approved):
Feel free to do it in another PR, but we should also update the docs for this

PR #1536, opened 2021-06-21
pbf: update builder to allow both shorthand and block form factories

goaway:
Description: Updates builder to allow Kotlin constructor shorthand for specifying filter factory methods, while still supporting block long-form if needed.
Risk Level: Low
Testing: Local & CI


buildbreaker (dismissed):
Cool! Thanks @goaway

buildbreaker (dismissed):

PR #1537, opened 2021-06-22
http: add wiring for explicit flow control along the response path

goaway:
Description: Adds public interface support for explicit flow control mode on Android and iOS, including all wiring down to the Http::Client.
Risk Level: Moderate
Testing: Unit & pending E2E (#1649)


goaway:
Still needs updates to tests and add additional coverage, but Android and iOS artifacts both build.

library/common/http/client.h
```diff
@@ -69,6 +69,13 @@ class Client : public Logger::Loggable<Logger::Id::http> {
    */
   void sendHeaders(envoy_stream_t stream, envoy_headers headers, bool end_stream);
 
+  /**
+   * Notify the stream that the caller is ready to receive more data. Only used in explicit
+   * buffering mode.
+   * @param bytes_to_read, the quantity of data the caller is prepared to process.
+   */
+  void readData(envoy_stream_t /*stream*/, size_t /*bytes_to_read*/) {}
```
alyssawilk, line 77: maybe clarify if this is request or response data for all of these, since we'll eventually have calls for both?

alyssawilk (commented):
Wohoo! One high level comment but otherwise will let the experts review :-)

goaway:
Pretty sure this one will be green.

library/common/http/client.h
```diff
@@ -234,6 +241,15 @@ class Client : public Logger::Loggable<Logger::Id::http> {
     Client& parent_;
     // Response details used by the connection manager.
     absl::string_view response_details_;
+    // True if the bridge should operate in explicit flow control mode.
+    //
+    // In this mode only one callback can be sent to the bridge until more is
+    // asked for. When a response is started this will either allow headers or an
+    // error to be sent up. Body, trailers, or further errors will not be sent
+    // until resumeData is called. This, combined with standard Envoy flow control push
```
junr03, line 249: resumeData, or readData?

library/common/jni/jni_interface.cc
```diff
@@ -800,6 +800,12 @@ Java_io_envoyproxy_envoymobile_engine_EnvoyHTTPFilterCallbacksImpl_callReleaseCa
 
 // EnvoyHTTPStream
 
+extern "C" JNIEXPORT jint JNICALL Java_io_envoyproxy_envoymobile_engine_JniLibrary_readData(
```
junr03, line 803: I can't remember if we have to add these functions to proguard?

library/common/main_interface.h
```diff
@@ -25,9 +25,11 @@ envoy_stream_t init_stream(envoy_engine_t engine);
  * can occur.
  * @param stream, handle to the stream to be started.
  * @param callbacks, the callbacks that will run the stream callbacks.
+ * @param explicit_flow_control, whether request/response buffers will be managed explicitly.
```
junr03, line 28: along the line of Alyssa's comment, lets make it clear everywhere that right now this is only wired for the response path, but request path is coming.

Although, that being said I am confused what the request path means for this type of flow control if the client is the one sending request bytes down to envoy mobile with the send* API, so it is _always_ explicitly managed by the client.

library/java/io/envoyproxy/envoymobile/engine/JniLibrary.java
```diff
@@ -97,6 +100,16 @@ public static void load() {
    */
   protected static native int sendData(long stream, ByteBuffer data, boolean endStream);
 
+  /**
+   * Read data from the response stream. Returns immediately.
```
junr03, line 104: nit: about the comment this function does not read data right? it merely informs the envoy mobile client that the platform client is ready for more callbacks on the response path.

As a larger question that just occurred to me, should the function be called something else? readyForData?

library/objective-c/EnvoyEngine.h
```diff
@@ -198,6 +201,14 @@ extern const int kEnvoyFilterResumeStatusResumeIteration;
  */
 - (void)sendHeaders:(EnvoyHeaders *)headers close:(BOOL)close;
 
+/**
+ Read data from the response stream. Returns immediately.
```
junr03, line 205: same comment as above

junr03 (commented):
I am guessing there is no functional tests or API docs because there is no implementation in client.cc yet.

library/common/jni/jni_interface.cc
goaway: I think just new clases on the JVM side.

library/common/main_interface.h
goaway: I believe most (all?) of my docstring comments call out the response path specifically. Alyssa's PR implementing request path is posted.

library/java/io/envoyproxy/envoymobile/engine/JniLibrary.java
goaway: I think the eventual function will actually take a buffer (and it did in an earlier version of this PR). Also of note, for stream libraries in general, "read" is not unprecedented for an API that simply takes byte count.

library/common/http/client.h
goaway: readData was what I landed on. See my comment below.

goaway:
@junr03 the client.cc code _is_ implemented; it was just in an earlier PR by @alyssawilk. I updated the unit coverage, but planned to do end-to-end coverage in an upcoming PR.

library/common/http/client.h
goaway: I didn't update client.h and client.cc, just because wanted to err on the side of caution and not cause unnecessary conflicts with any branches @alyssawilk might have to land on top.

library/common/main_interface.h
goaway: Updated this comment, which was the only one that I think mentioned request buffers.

library/java/io/envoyproxy/envoymobile/engine/JniLibrary.java
goaway: I'm open to reconsidering/discussing, but for what it's worth, I did consider the naming for a while, and felt that sticking to a stream-focused interface seemed the best model moving forward.

library/common/http/client.h
goaway: Just realized you are completely right about there being a missing step here. Updating.

goaway:
@junr03 can you give this another pass please?

junr03 (approved):
alright, looks good now. Looking forward to e2e tests

PR #1538, opened 2021-06-22
Port CronetUrlRequestTest

carloseltuerto:
This PR is the starting point to port Cronet tests - it includes the main Cronet test: CronetUrlRequestTest. However, most the the code is for the test infrastructure itself: the CronetTestRule, NativeTestServer, MockUrlRequestJobFactory and some others. At this point, most of them are "fake" implementations. For example, the NativeTestServer has nothing "native" - it is just a wrapper around MockWebServer, which was coded to behave as much as possible like the original C++ Cronet Test Server.

With this PR, rhe CronetUrlRequestTest only tests the Java implementation (65 tests). Testing with Envoy Mobile incurs 16 test failures, so it is disabled for the moment. This will be mitigated with other PRs.

Description:  Port enough Cronet code to have test CronetUrlRequestTest running
Risk Level: None
Testing: this PR objective is to port a test
Docs Changes: N/A
Release Notes: N/A


library/java/org/chromium/net/impl/CronetUrlRequest.java
```diff
@@ -774,6 +777,121 @@ void onFailed(final UrlResponseInfo urlResponseInfo, final CronetException e) {
     }
   }
 
+  private static String textResponseCode(int code) {
```
buildbreaker, line 780: Would this make sense to be put in a separate file/class?

library/java/org/chromium/net/impl/JavaUrlRequest.java
```diff
@@ -119,778 +120,796 @@
...
-  mExecutor.execute(new Runnable() {
+  private void closeResponseChannel() {
+    mExecutor.execute(new Runnable() {
+      @Override
+      public void run() {
+        if (mResponseChannel != null) {
+          try {
+            mResponseChannel.close();
+          } catch (IOException e) {
+            e.printStackTrace();
```
buildbreaker, line 844: Is this still needed?

test/java/org/chromium/net/testing/CronetTestRule.java
```diff
@@ -0,0 +1,359 @@
...
+  public void assertResponseEquals(UrlResponseInfo expected, UrlResponseInfo actual) {
+    Assert.assertEquals(expected.getAllHeaders(), actual.getAllHeaders());
+    Assert.assertEquals(expected.getAllHeadersAsList(), actual.getAllHeadersAsList());
+    Assert.assertEquals(expected.getHttpStatusCode(), actual.getHttpStatusCode());
+    Assert.assertEquals(expected.getHttpStatusText(), actual.getHttpStatusText());
+    Assert.assertEquals(expected.getUrlChain(), actual.getUrlChain());
+    Assert.assertEquals(expected.getUrl(), actual.getUrl());
+    // Transferred bytes and proxy server are not supported in pure java
+    if (!testingJavaImpl()) {
+      Assert.assertEquals(expected.getReceivedByteCount(), actual.getReceivedByteCount());
```
buildbreaker, line 234: Static import the `assertEquals`?

test/java/org/chromium/net/testing/TestUploadDataProvider.java
```diff
@@ -0,0 +1,257 @@
...
+import java.util.ArrayList;
+import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.chromium.net.UploadDataProvider;
+import org.chromium.net.UploadDataSink;
+
+/**
+ * An UploadDataProvider implementation used in tests.
+ */
+public class TestUploadDataProvider extends UploadDataProvider {
```
buildbreaker, line 16: `UploadDataProviderTest`?

test/java/org/chromium/net/testing/TestUrlRequestCallback.java
```diff
@@ -0,0 +1,389 @@
...
+import org.chromium.net.InlineExecutionProhibitedException;
+import org.chromium.net.UrlRequest;
+import org.chromium.net.UrlResponseInfo;
+
+/**
+ * Callback that tracks information from different callbacks and and has a method to block thread
+ * until the request completes on another thread. Allows to cancel, block request or throw an
+ * exception from an arbitrary step.
+ */
+public class TestUrlRequestCallback extends UrlRequest.Callback {
```
buildbreaker, line 29: `UrlRequestCallbackTest`?

test/java/org/chromium/net/testing/data/content_length_mismatch.html
```diff
@@ -0,0 +1 @@
+Response that lies about content length.
```
buildbreaker, line 1: Do all of these need to be in a file (versus just a string in the code)?

buildbreaker (commented):
Just a few comments

test/java/org/chromium/net/testing/CronetTestRule.java
carloseltuerto: This was like that in the original code - maybe because this is not a test. 

But I agree with you - let's be systematic about import static "assert" methods.

Done.

library/java/org/chromium/net/impl/JavaUrlRequest.java
carloseltuerto: Well, this is like that in the original code:

https://chromium.googlesource.com/chromium/src/+/refs/heads/main/components/cronet/android/java/src/org/chromium/net/impl/JavaUrlRequest.java#900

Given that the Java implementation is a fifth wheel, just ignoring silently is good enough.

Done.

library/java/org/chromium/net/impl/CronetUrlRequest.java
carloseltuerto: It's a hack - my expectation is this will be short lived - the response message should come from the first line of the HTTP response, example:

```HTTP/1.1 404 Not Found```

As far as I can tell, Envoy Mobile does not expose the first line it receives for the Java layer, so the text had to be "reconstructed". 

I will open an issue for that, or maybe there is somewhere this is available - then I will yank this immediately.

test/java/org/chromium/net/testing/data/content_length_mismatch.html
carloseltuerto: Well, as you can imagine, this is exactly like that with [Cronet](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/components/cronet/testing/test_server/data).

At first sight, it seemed easier to just port this "as is": the content of the files is easy to understand, and there are other tests to be ported relying on this.

test/java/org/chromium/net/testing/TestUploadDataProvider.java
carloseltuerto: In Cronet, only CronetTestRule has its test. TBH, I suspect this will stay like that...

test/java/org/chromium/net/testing/TestUrlRequestCallback.java
carloseltuerto: In Cronet, only CronetTestRule has its test. TBH, I suspect this will stay like that...

carloseltuerto:
Thanks, Alan.

buildbreaker (approved):
Looks good @carloseltuerto !

PR #1541, opened 2021-06-22
headers: not overwriting the server header

### LINKED ISSUE #1539: Envoy mobile shouldn't overwrite the server header
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#envoy-v3-api-enum-extensions-filters-network-http-connection-manager-v3-httpconnectionmanager-serverheadertransformation

Should be doable for the server_header_transformation to be PASS_THROUGH

cc @carloseltuerto

alyssawilk:
Fixes #1539


alyssawilk:
not sure where config tests are - feel free to point me their way!

goaway (approved):
Thanks @alyssawilk!

goaway:
To test this at present, we'd probably need to use an integration test. The tests under //test/common might work, but it depends where Envoy actually does the stripping.

Otherwise //test/swift/integration or //test/kotlin/integration are end-to-end and would definitely be able to cover it. That said, this is just config leveraging functionality that's already covered under Envoy main, so I'm comfortable merging as is.

alyssawilk:
SG.  We can merge and let Charles verify with the cronvoy tests :-)

PR #1545, opened 2021-06-23
http:  request path push back

alyssawilk:
Description: Adding APIs for request path push back
Risk Level: low
Testing: tweaked unit tests
Docs Changes: n/a
Release Notes: n/a


library/common/http/client.cc
```diff
@@ -176,6 +176,19 @@ void Client::DirectStream::resetStream(StreamResetReason reason) {
...
 
+void Client::DirectStream::readDisable(bool disable) {
+  if (disable) {
+    ++read_disable_count_;
+  } else {
+    ASSERT(read_disable_count_ > 0);
+    --read_disable_count_;
+    if (read_disable_count_ == 0 && wants_write_notification_) {
+      //      bridge_callbacks_.on_can_send_data(bridge_callbacks_.context);
+      wants_write_notification_ = false;
```
RyanTheOptimist, line 187: I wonder if this should be done before the call to `on_can_send_data()` so that if `on_can_send_data()` ends up calling `sendData()` we don't then end up clearing `wants_write_notification_` erroneously?

carloseltuerto:
Now we know what happens to Envoy Mobile when request body chunks are sent unbridled: Envoy Mobile silently drops data - this [test](https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/android/test/javatests/src/org/chromium/net/urlconnection/CronetBufferedOutputStreamTest.java;l=157) fails with data mixup. If we put a Thread.sleep(100) after sending each chunk, then the test passes...

alyssawilk:
Ok synced with head and updated async -> explicit flow control. pending CI should be ready for a review pass.

alyssawilk:
nope, cc broken, please ignore until the next push

library/cc/stream_callbacks.cc
```diff
@@ -87,6 +87,17 @@ void* c_on_cancel(envoy_stream_intel, void* context) {
   return nullptr;
 }
 
+void* c_on_can_send_data(envoy_stream_intel, void* context) {
```
goaway, line 90: Should this contain some indication of how much data can be sent on the next call?

goaway: If it did, it could probably plug directly into some of the stream APIs on the platform stacks.

goaway: Or is the idea not so much to avoid queueing/buffering as to simply inform when whatever was previously queued/buffered has been sent?

goaway: Okay, reading the comments here, I think I understand better. It looks like the underlying API doesn't really support passing through any notion of available send window.

goaway:
WDYT about `onSendWindowAvailable()`?

goaway (commented):
Overall, I think this looks good. I am still curious if there's some way the caller could infer how much data to send next. Also, one bikeshedding comment above.

alyssawilk:
Happy to rename.  on the send data, I mean Envoy will infinitely buffer, so I think the sender just sends the chunk size it's willing to buffer and Envoy informs if there's buffer space left

goaway (approved):
Thanks @alyssawilk! This looks good.

PR #1560, opened 2021-06-30
filters: prevent spurious cancellation callbacks from the gRPC error path

goaway:
Description: Platform filters could fail to correctly transition to a completed state due to the lack of a response body on gRPC error responses. This resulted in spurious cancellation callbacks when the filter was scheduled for destruction due to the invalid "incomplete" status of the stream.
Risk Level: Low
Testing: Integration, Local & CI

### DIFF
--- a/library/common/extensions/filters/http/platform_bridge/filter.cc
+++ b/library/common/extensions/filters/http/platform_bridge/filter.cc
@@ -392,42 +392,42 @@ Http::FilterHeadersStatus PlatformBridgeFilter::encodeHeaders(Http::ResponseHead
   // Presence of internal error header indicates an error that should be surfaced as an
   // error callback (rather than an HTTP response).
   const auto error_code_header = headers.get(Http::InternalHeaders::get().ErrorCode);
-  if (!error_code_header.empty()) {
-    // This must be set here, since we won't be delegating to FilterBase.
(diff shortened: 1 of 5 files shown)


library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -393,6 +393,10 @@ Http::FilterHeadersStatus PlatformBridgeFilter::encodeHeaders(Http::ResponseHead
   // error callback (rather than an HTTP response).
   const auto error_code_header = headers.get(Http::InternalHeaders::get().ErrorCode);
   if (!error_code_header.empty()) {
+    // This must be set here, since we won't be delegating to FilterBase.
+    response_filter_base_->state_.stream_complete_ = end_stream;
```
mattklein123, line 397: IMO it would be cleaner to set this universally at the beginning of the function, so you could either allow it to be reset in the common function below or have some type of preamble function that is split out?

goaway: Updated. MK what you think.

goaway:
Hilariously, it looks like the Kotlin error integration tests have been relying on this behavior to pass. Fix incoming.

mattklein123 (approved):
Nice, thanks!

goaway:
Thanks @mattklein123!

PR #1567, opened 2021-07-02
core: add passive healthchecking to http connections

goaway:
Description: The objective of this change is to leverage Envoy's host outlier detection to passively healthcheck connections. When a host is ejected due to outlier triggers (in this case, 3 consecutive request failures), connections to the host are drained. This allows the library to reap potentially dead/stale connections that whose status hasn't been otherwise detected through socket events or TCP keep-alive, etc.
Risk Level: Low
Testing: Local


library/common/config/config.cc
```diff
@@ -228,83 +228,98 @@ const char* config_template = R"(
             budget_percent:
               value: 100
             min_retry_concurrency: 0xffffffff # uint32 max
+    outlier_detection: &base_outlier_detection
+      consecutive_5xx: 3
+      base_ejection_time: 0.001s
+      max_ejection_time: 0.001s
```
mattklein123, line 234: At a high level this seems reasonable to me, however I'm not sure if there are any implications of such a short base/max ejection time. At minimum I would add more comments here as it's unclear to the reader why it's configured this way. Also, please do a bunch of manual testing, especially in totally offline modes.

goaway: Yep, makes sense. I did some very rudimentary validation going into the weekend, but I'll do some more thorough testing today/tomorrow.

PR #1569, opened 2021-07-03
cast fix: incorrect cast in EnvoyConfiguration.m

junr03:

### DIFF
--- a/library/objective-c/EnvoyConfiguration.m
+++ b/library/objective-c/EnvoyConfiguration.m
@@ -101,8 +101,7 @@
                             (unsigned long)self.dnsFailureRefreshSecondsBase];
   [definitions appendFormat:@"- &dns_fail_max_interval %lus\n",
                             (unsigned long)self.dnsFailureRefreshSecondsMax];
-  [definitions
-      appendFormat:@"- &dns_preresolve_hostnames %@\n", self.dnsPreresolveHostnames];
+  [definitions appendFormat:@"- &dns_preresolve_hostnames %@\n", self.dnsPreresolveHostnames];
   [definitions
       appendFormat:@"- &stream_idle_timeout %lus\n", (unsigned long)self.streamIdleTimeoutSeconds];
   [definitions appendFormat:@"- &metadata { device_os: %@, app_version: %@, app_id: %@ }\n", @"iOS",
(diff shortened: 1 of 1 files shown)


rebello95 (dismissed):

goaway (approved):

goaway:
Unit test should be able to catch this, no?

rebello95 (approved):

junr03:
> Unit test should be able to catch this, no?

Yeah, I am confused about that. But I am not sure how string interpolation works in objc w.r.t casting. i.e, does that fact that I am interpolating using `%@` invalidates the cast?

rebello95:
> Unit test should be able to catch this, no?

Looks like the behavior still works with this because of `%@`:

```
int main(int argc, const char * argv[]) {
    @autoreleasepool {
        NSString *dnsPreresolveHostnames = @"someHostNames";
        NSMutableString *str = [@"foo" mutableCopy];
        [str appendFormat:@"- &dns_preresolve_hostnames %@\n", (unsigned long)dnsPreresolveHostnames];
        NSLog(@"%@", str);
    }
    return 0;
}
```

Output:

```
2021-07-03 07:22:16.208244-0700 casttest[5727:55446] foo- &dns_preresolve_hostnames someHostNames
Program ended with exit code: 0
```

It does, however, warn:
> Format specifies type 'id' but the argument has type 'unsigned long'
> Replace '%@' with '%lu'

If we were to change it to `%lu`, it actually would cast:

```
int main(int argc, const char * argv[]) {
    @autoreleasepool {
        NSString *dnsPreresolveHostnames = @"someHostNames";
        NSMutableString *str = [@"foo" mutableCopy];
        [str appendFormat:@"- &dns_preresolve_hostnames %@\n", (unsigned long)dnsPreresolveHostnames];
        NSLog(@"%@", str);
    }
    return 0;
}
```

Output:
```
2021-07-03 07:20:12.533229-0700 casttest[5692:53455] foo- &dns_preresolve_hostnames 4294983704
Program ended with exit code: 0
```

The [logs on CI](https://github.com/envoyproxy/envoy-mobile/runs/2975715341) actually do indicate that this warning was there, but builds succeeded:
```
INFO: From Compiling library/objective-c/EnvoyConfiguration.m:
library/objective-c/EnvoyConfiguration.m:105:57: warning: format specifies type 'id' but the argument has type 'unsigned long' [-Wformat]
      appendFormat:@"- &dns_preresolve_hostnames %@\n", (unsigned long)self.dnsPreresolveHostnames];
                                                 ~~     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                 %lu
1 warning generated.
```

Do you know if there's a reason we're building without warnings? IIRC it was because core Envoy was generating warnings for us

Issue #1558, opened 2021-06-30
Keep header name capitalization for HTTP1


carloseltuerto:
For HTTP1, unfortunately, keeping the original capitalization for the header names can't be avoided.

alyssawilk:
reopened - needs to be applied globally not just to alpn pool :-(

alyssawilk:
ditto on this one - hopefully fixed

carloseltuerto:
Unfortunately,  [PR1710](https://github.com/envoyproxy/envoy-mobile/pull/1710") resuscitated this bug.

PR #1579, opened 2021-07-08
http: preserving casing for HTTP/1.1

alyssawilk:
Description: Changing envoy mobile to preserve header casing for HTTP/1.1 header keys
Risk Level: Medium
Testing: new integration test
Part of https://github.com/envoyproxy/envoy-mobile/issues/1558

The other half is applying the preserve_case rules to library/common/config/config.cc which is in #1588 so it only affects the new cluster.


library/common/http/header_utility.cc
```diff
@@ -9,11 +10,16 @@ namespace Http {
...
+  transformed_headers->setFormatter(
+      std::make_unique<
+          Extensions::Http::HeaderFormatters::PreserveCase::PreserveCaseHeaderFormatter>());
+  Envoy::Http::StatefulHeaderKeyFormatter& formatter = transformed_headers->formatter().value();
   for (envoy_map_size_t i = 0; i < headers.length; i++) {
-    transformed_headers->addCopy(
-        LowerCaseString(Data::Utility::copyToString(headers.entries[i].key)),
-        Data::Utility::copyToString(headers.entries[i].value));
+    std::string key = Data::Utility::copyToString(headers.entries[i].key);
+    formatter.processKey(key);
```
RyanTheOptimist, line 20: So is the idea here that a copy of the non-lower-cased header name (key) is being saved in the formatter where it can be retrieved down in toBridgeHeaders? If so, perhaps add a comment to this effect. When I read `processKey()` I expected that the key was being transformed but I think that's not quite the case?

test/common/integration/client_integration_test.cc
```diff
@@ -19,11 +21,17 @@ namespace {
 
 // Based on Http::Utility::toRequestHeaders() but only used for these tests.
 Http::ResponseHeaderMapPtr toResponseHeaders(envoy_headers headers) {
-  Http::ResponseHeaderMapPtr transformed_headers = Http::ResponseHeaderMapImpl::create();
+  std::unique_ptr<Http::ResponseHeaderMapImpl> transformed_headers =
```
RyanTheOptimist, line 24: There's an unfortunate bit of cut/paste from header_utility.cc here. Is there any way to share some code, or is that a non-starter?

test/common/integration/client_integration_test.cc
```diff
@@ -211,54 +236,31 @@ TEST_P(ClientIntegrationTest, BasicNon2xx) {
...
-    return nullptr;
-  };
-  bridge_callbacks.on_error = [](envoy_error error, void* context) -> void* {
-    error.message.release(error.message.context);
-    callbacks_called* cc = static_cast<callbacks_called*>(context);
-    cc->on_error_calls++;
-    return nullptr;
-  };
-
+  // Setup bridge_callbacks_ to handle the response.
```
RyanTheOptimist, line 239: Is this comment still needed now that much of the setup logic is in Initialize()?

RyanTheOptimist (commented):
Wow, this is clever!

It might be nice to mention somewhere that this case preservation applies to header names (keys) and not header values. This is probably obvious to folks who have thought about this in any detail, but it'd be a nice hint for folks that are new to the notion of lower-cased header names.

goaway (approved):
This looks good to me, and I'm fine with merging as is. If you'd like to make further updates in response to @RyanTheOptimist's comments, that's fine, too - just ping me and I'll re-approve.

envoy_build_config/extensions_build_config.bzl
```diff
@@ -16,5 +16,6 @@ EXTENSIONS = {
     "envoy.stat_sinks.metrics_service":               "//source/extensions/stat_sinks/metrics_service:config",
     "envoy.transport_sockets.raw_buffer":             "//source/extensions/transport_sockets/raw_buffer:config",
     "envoy.transport_sockets.tls":                    "//source/extensions/transport_sockets/tls:config",
+    "envoy.http.stateful_header_formatters.preserve_case":       "//source/extensions/http/header_formatters/preserve_case:preserve_case_formatter",
```
junr03, line 19: nit: fmt

test/common/integration/client_integration_test.cc
junr03: +1 I am guessing that this direction was not written in prod code because we never go from c->c++ there. But I think we can probably have a shared function that does the transformation in the common map type and wrapper functions for the derived types?

test/common/integration/client_integration_test.cc
```diff
@@ -59,7 +68,34 @@ class ClientIntegrationTest : public BaseIntegrationTest,
       bootstrap.mutable_static_resources()->add_listeners()->MergeFrom(
           Server::parseListenerFromV3Yaml(api_listener_config()));
     });
-    BaseIntegrationTest::initialize();
+
```
junr03, line 71: always appreciate your clean up :) thanks!

test/common/integration/client_integration_test.cc
alyssawilk: yeah, the problem is the base class doesn't have setFormatter, so then we end up having to set in the request/response specific subclasses and assert that work is done in the shared class.  kinda ugly - maybe I'll fix the APIs upstream so I can do a cleanup PR later.

goaway (approved):

PR #1580, opened 2021-07-09
envoy: 3876d7c0f

junr03:
Includes three relevant commits:
https://github.com/envoyproxy/envoy/pull/17226
https://github.com/envoyproxy/envoy/pull/17274
https://github.com/envoyproxy/envoy/pull/17207

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -36,6 +36,7 @@ const std::string config_header = R"(
 - &dns_refresh_rate 60s
 - &dns_fail_base_interval 2s
 - &dns_fail_max_interval 10s
+- &dns_query_timeout 25s
 - &dns_preresolve_hostnames []
 - &metadata {}
 - &stats_domain 127.0.0.1
@@ -183,7 +184,7 @@ static_resources:
               dns_failure_refresh_rate:
                 base_interval: *dns_fail_base_interval
                 max_interval: *dns_fail_max_interval
-              dns_query_timeout: 25s
+              dns_query_timeout: *dns_query_timeout
         # TODO: make this configurable for users.
         - name: envoy.filters.http.decompressor
           typed_config:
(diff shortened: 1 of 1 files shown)


library/common/config/config.cc
```diff
@@ -183,6 +183,7 @@ const char* config_template = R"(
               dns_failure_refresh_rate:
                 base_interval: *dns_fail_base_interval
                 max_interval: *dns_fail_max_interval
+              dns_query_timeout: 25s
```
junr03, line 186: hard coding this in this current PR so no one is burnt by the default. I am doing a PR to make it configurable.

library/common/config/config.cc
```diff
@@ -36,6 +36,7 @@ const std::string config_header = R"(
 - &dns_refresh_rate 60s
 - &dns_fail_base_interval 2s
 - &dns_fail_max_interval 10s
+- &dns_query_timeout 25s
```
junr03, line 39: Putting a default so no one is burnt by the upstream 5s default. I am working on a parallel PR to make it configurable.

library/common/config/config.cc
```diff
@@ -36,6 +36,7 @@ const std::string config_header = R"(
 - &dns_refresh_rate 60s
 - &dns_fail_base_interval 2s
```
junr03, line 37: @goaway in slack we discussed 25s timeout + 5s failure refresh rate. The rate as is here starts a 2s and backoffs to 10s. Are we ok with that policy or do you want to reduce max to 5s?

library/common/config/config.cc
junr03: https://github.com/envoyproxy/envoy-mobile/pull/1583

rebello95 (approved):

library/common/config/config.cc
goaway: I think that policy is fine for now.

Issue #1548, opened 2021-06-24
Envoy mobile should ALPN by default


alyssawilk:
Related to https://github.com/envoyproxy/envoy-mobile/issues/1546 Envoy mobile should likely be using auto-http (use ALPN to determine HTTP/1.1 or HTTP/2) by default.

There's a couple of behavioral differences, some which are fixed at HEAD (e.g. https://github.com/envoyproxy/envoy/issues/16881) and some which are not (https://github.com/envoyproxy/envoy/issues/15947)

I'm adding this to the meeting agenda for next week and we can chat about blockers

alyssawilk:
so @junr03  / @goaway I'd like to pick up this one and I need a bit of context from y'all

looking at library/common/config/config.cc it looks like we have clusters for HTTP/1.1 (base, base_alt, base_wlan, base_wlan_alt,  ditto the 4 base clear clusters) and HTTP/2 (base_h2 and friends)  Is that correct?

If so would the thing to do to add base_alpn, have you guys verify moving the non-clear traffic over to base_alpn works, and then deleting the non-clear traffic?

alyssawilk:
Tossing over to Lyft to test our migration.  Once you give me the thumbs up that you're happy with ALPN I'm happy to remove the 12 legacy "http-only" and "http2 only" clusters.

PR #1588, opened 2021-07-13
http: adding alpn clusters accessed via x-envoy-mobile-upstream-protocol: alpn

alyssawilk:
Risk Level: low
Testing: unit tests
Docs Changes: n/a
Release Notes: n/a
Part of #1548
Part of #1558


alyssawilk:
CI failure is legit - please ignore this PR for now :-P

goaway:
Re-running iOS app tests.

alyssawilk:
No, as said over in the propercase PR, I update Envoy to handle propercasing there but had added it to the "prod config" here

Initially I'd had some syntax errors, and the cc tests using config failed.  I'm pretty confident with those passing now this is valid config, but it's _only_ valid config for new Envoy-mobile builds with the propercasing library linked in.

I thought adding propercasing to library/common/config/BUILD would be sufficient but as this is still failing I'm wondering if something in this test uses prebuilt envoy-mobile, and if I have to submit a PR with the new linking rule, then submit a follow-up PR using the new link rule.  Thoughts?

alyssawilk:
ok in the absence of advice I'll try checking in the build changes for preserving case, and see if a follow-up PR applying the config sticks better.

goaway (approved):
Looks good. Thank you @alyssawilk!

PR #1589, opened 2021-07-13
jni: null terminate strings before passing to NewStringUTF

snowp:
The data coming in from the logging interface doesn't seem to be null terminated in all cases. This constructs
a std::string and uses c_str() to get a null terminated version of the data that honors the specified length.

Risk Level: Medium, JNI changes
Testing: Existing tests
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/library/common/jni/jni_utility.cc
+++ b/library/common/jni/jni_utility.cc
@@ -62,7 +62,9 @@ envoy_data array_to_native_data(JNIEnv* env, jbyteArray j_data) {
 }
 
 jstring native_data_to_string(JNIEnv* env, envoy_data data) {
-  jstring jstrBuf = env->NewStringUTF(const_cast<char*>(reinterpret_cast<const char*>(data.bytes)));
+  // Ensure we get a null-terminated string, the data coming in via envoy_data might not be.
+  std::string str(reinterpret_cast<const char*>(data.bytes), data.length);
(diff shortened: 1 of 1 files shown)


jingwei99 (dismissed):

snowp:
ptal @buildbreaker @goaway @junr03 

I think this is pretty safe to do at the cost of another string copy/allocation, and seems to fix the logging issue @jingwei99 ran into

goaway:
Will look today.

goaway:
Re-running test.

goaway (approved):
What's unfortunate is that we're copying a null-terminated string to a non-null terminated buffer back to a new null-terminated string and then to a totally unrelated representation in the JVM. We should maybe consider an envoy_string type that's guaranteed to be null-terminated just to avoid these shenanigans (at least as long as we're in C, since C strings are expected to be null terminated anyways).

Oh well. This is only used in one place, and it's unlikely to be a problem where it's used. Hopefully it doesn't come back to bite us.

snowp:
At what point is it null terminated prior to this? The string coming out of spdlog is not when using formatting (it allocates into a buffer and provides the size of the formatted string), after which point I think it's all string_view from there on, which also doesn't guarantee null termination. Maybe I'm missing a step?

PR #1590, opened 2021-07-14
stats: add event tracker

Augustyniak:
Description: Add an interface for tracking events that can be used by `EnvoyMobile` to inform platform code (i.e., Android or iOS) about events inside of envoy platform bridge. There are currently no events that are emitted. At Lyft we have a plan to use this interface to emit events from within one of our private extensions to Envoy Mobile.
Testing: Manual
Docs Changes: Done
Release Notes: None


library/common/types/c_types.h
```diff
@@ -176,10 +179,16 @@ void release_envoy_error(envoy_error error);
 
 /**
  * Called by a receiver of envoy_stats_tags to indicate memory/resources can be released.
- * @param headers, envoy_stats_tags to release.
+ * @param stats_tags, envoy_stats_tags to release.
  */
 void release_envoy_stats_tags(envoy_stats_tags stats_tags);
 
+/**
+ * Called by a receiver of envot_event to indicate that a resource can be released.
```
snowp, line 187: typo

library/common/types/c_types.h
```diff
@@ -362,3 +380,13 @@ typedef struct {
   // Context passed through to callbacks to provide dispatch and execution state.
   const void* context;
 } envoy_logger;
+
+/**
+ * Interface for event tracking.
+ */
+
```
snowp, line 387: remove newline

library/swift/EngineBuilder.swift
```diff
@@ -216,6 +217,17 @@ public class EngineBuilder: NSObject {
     return self
   }
 
+  /// Set a closure to be called when the engine emits an event.
+  ///
+  /// - parameter closure: The closure to be called.
+  ///
+  /// - returns: This builder.
+  @discardableResult
+  public func setEventTracker(closure: @escaping ([String: String]) -> Void) -> Self {
```
Augustyniak, line 226: @rebello95 @goaway @junr03 

I tried to come up with an integration test for `setEventTracker` that would look similar to [SetLoggerTest](https://github.com/envoyproxy/envoy-mobile/blob/main/test/swift/integration/SetLoggerTest.swift) test. 

There are some issues here since by default Envoy Engine does not emit any events - we will emit them in our private extensions to Envoy Mobile. That means that for the integration test to work we would need to "simulate" an emission of an event using `Api::External::retrieveApi("envoy_event_tracker_api_name")` that's written in C++ and not accessible in Swift (a language the integration test would be written in).

Any ideas for how I could test `setEventTracker` functionality that's being added in this PR?

snowp: How about adding a native test filter that makes use of this and then add that via `addNativeFilter`? That way you can invoke the event tracker from there and verify that sending a request through EM triggers an event

Augustyniak: That could work - I assume I would need to perform some request for the filter chain to be executed. Does it sound about right or is there an easier way to do this?

library/common/extensions/filters/http/test_event_tracker/filter.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+package envoymobile.extensions.filters.http.test_event_tracker;
+
+import "validate/validate.proto";
+
+message TestEventTracker {
+  map<string, string> attributes = 1;
```
rebello95, line 8: Might be nice to add some docs here

library/objective-c/EnvoyBridgeUtility.h
```diff
@@ -100,6 +100,23 @@ static inline NSString *to_ios_string(envoy_data data) {
...
+                                                   length:header.key.length
+                                                 encoding:NSUTF8StringEncoding];
+    NSString *headerValue = [[NSString alloc] initWithBytes:header.value.bytes
+                                                     length:header.value.length
+                                                   encoding:NSUTF8StringEncoding];
+    newMap[headerKey] = headerValue;
+  }
+
+  release_envoy_map(map);
+  return [newMap copy];
```
rebello95, line 117: Do we actually need a copy here? Looks like `to_ios_headers` does not copy // @goaway

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -365,12 +365,26 @@ static envoy_data ios_get_string(const void *context) {
   return toManagedNativeString(accessor.getEnvoyString());
 }
 
+static void ios_track_event(envoy_map map, const void *context) {
+  // This code block runs inside the Envoy event loop. Therefore, an explicit autoreleasepool block
+  // is necessary to act as a breaker for any Objective-C allocation that happens.
+  @autoreleasepool {
+    EnvoyEventTracker *eventTracker = (__bridge EnvoyEventTracker *)context;
+    // release memory even if the tracking closure is not provided.
```
rebello95, line 373: Is this docstring relevant to a specific line of code? I don't see an explicit release here

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -365,12 +365,26 @@ static envoy_data ios_get_string(const void *context) {
...
+    EnvoyEvent *platformMap = to_ios_map(map);
```
rebello95, line 374: Why not move this into the `if` statement to avoid creating this if there is no `track` closure?

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -432,6 +449,16 @@ - (int)registerStringAccessor:(NSString *)name accessor:(EnvoyStringAccessor *)a
   return register_platform_api(name.UTF8String, accessorStruct);
 }
 
+- (int)registerEventTracker:(EnvoyEventTracker *)eventTracker {
+  // TODO(Augustyniak): Everything here leaks, but it's all tied to the life of the engine.
+  // THis will need to be updated for https://github.com/lyft/envoy-mobile/issues/332.
```
rebello95, line 454:
```suggestion
  // This will need to be updated for https://github.com/lyft/envoy-mobile/issues/332.
```

library/objective-c/EnvoyEventTracker.m
```diff
@@ -0,0 +1,14 @@
...
+- (instancetype)initWithEventTrackingClosure:(nullable void (^)(EnvoyEvent *))track {
+  self = [super init];
+  if (!self) {
+    return nil;
+  }
+
+  self.track = track;
+  return self;
+}
+@end
```
rebello95, line 14:
```suggestion

@end
```

test/swift/integration/SetEventTrackerTest.swift
```diff
@@ -0,0 +1,39 @@
...
+        engineExpectation.fulfill()
+      }
+      .setEventTracker { event in
+        XCTAssertEqual("bar", event["foo"])
+        eventTrackingExpectation.fulfill()
+      }
+
+      .addNativeFilter(name: "envoy.filters.http.test_event_tracker",
+      // swiftlint:disable:next line_length
+                       typedConfig: "{\"@type\":\"type.googleapis.com/envoymobile.extensions.filters.http.test_event_tracker.TestEventTracker\",\"attributes\":{\"foo\":\"bar\"}}")
```
rebello95, line 23:
```suggestion
      .addNativeFilter(
          name: "envoy.filters.http.test_event_tracker",
          // swiftlint:disable:next line_length
          typedConfig: "{\"@type\":\"type.googleapis.com/envoymobile.extensions.filters.http.test_event_tracker.TestEventTracker\",\"attributes\":{\"foo\":\"bar\"}}"
      )
```

rebello95 (commented):
LGTM overall, but a few questions

library/objective-c/EnvoyBridgeUtility.h
Augustyniak: we do not need it per se - I just decided to make a copy so that we return an instance of `NSDictionary` as opposed to `NSMutableDictionary`. Without copying here it's possible for a caller of this method to take the returned value, cast it to `NSMutableDictionary` and modify it... that being said, it's highly unlike and copying costs resources so I will just get rid of this copying.

library/objective-c/EnvoyEngineImpl.m
Augustyniak: the release happens inside of `to_ios_map` method in [here](https://github.com/envoyproxy/envoy-mobile/pull/1590/files#diff-271bc2a427a66a65657bc509a808ff32e2536938f9a6365a8731e5ed50d4b9adR116). Not  obvious but it seems to be following a pattern established by other methods such as `to_ios_headers`. @goaway maybe we should rename this method to make their memory-related behavior more obvious? `bridge_to_ios_map` is an alternate name that comes to my mind here.

library/objective-c/EnvoyEngineImpl.m
Augustyniak: See the comment above, `to_ios_map` method call releases the memory so we want to ensure that it's always called.

docs/root/api/starting_envoy.rst
```diff
@@ -259,6 +259,22 @@ Specify a closure to be called when Envoy's engine emits a log message.
...
+~~~~~~~~~~~~~~~~~~~
+``setEventTracker``
+~~~~~~~~~~~~~~~~~~~
+
+Specify a closure to be called when Envoy's engine emits an event.
+
+**Example**::
+
+  // Kotlin
+  // This interface is pending for Kotlin
```
murki, line 271: do you need help with this? Or is this for a separate PR?

library/common/extensions/filters/http/test_event_tracker/config.h
```diff
@@ -0,0 +1,36 @@
...
+      const envoymobile::extensions::filters::http::test_event_tracker::TestEventTracker& config,
+      const std::string& stats_prefix, Server::Configuration::FactoryContext& context) override;
+};
+
+DECLARE_FACTORY(TestEventTrackerFilterFactory);
+
+} // namespace TestEventTracker
+} // namespace HttpFilters
+} // namespace Extensions
+} // namespace Envoy
```
murki, line 36: what are these?

library/objective-c/EnvoyBridgeUtility.h
rebello95: I'd definitely bias towards reducing the copy here since external consumers would be misusing the API by trying to mutate it

library/common/extensions/filters/http/test_event_tracker/config.h
Augustyniak: These comments explain what namespace a given `}` refers to. A few lines higher you should see how all of these namespaces are opened using `namespace Envoy` etc. It's a common practice to have these namespace related comments in Envoy Mobile repo I think.

docs/root/api/starting_envoy.rst
Augustyniak: It's a separate PR - we are going to sync with Jingwei and/or Don on Android work. Hopefully, they will be able to help with Android changes but these iOS/common changes need to land first .

Augustyniak:
Closing since I lost write access to the repo and cannot continue to push my changes here. Going to continue my work in https://github.com/envoyproxy/envoy-mobile/pull/1602.

PR #1597, opened 2021-07-16
PoC: stop gap for pausing/resuming timeouts

buildbreaker:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/library/kotlin/io/envoyproxy/envoymobile/filters/RequestFilterCallbacks.kt
+++ b/library/kotlin/io/envoyproxy/envoymobile/filters/RequestFilterCallbacks.kt
@@ -22,4 +22,9 @@ interface RequestFilterCallbacks {
    * on the stream.
    */
   fun resetIdleTimer()
+
+
+  fun pauseRequestTimeout()
+
+  fun resumeRequestTimeout()
 }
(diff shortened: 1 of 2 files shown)


library/kotlin/io/envoyproxy/envoymobile/filters/RequestFilterCallbacks.kt
```diff
@@ -22,4 +22,9 @@ interface RequestFilterCallbacks {
    * on the stream.
    */
   fun resetIdleTimer()
+
+
+  fun pauseRequestTimeout()
```
goaway, line 27: It seems worth calling out that there are many timeouts that this will not pause, and that this would only affect the idle timer/timeout.

goaway:
I know this is only a PoC, but just wanted to call out that resetting the timeout on an interval can be done in C++ as well, and if we go that route, that has the advantage of only requiring a single implementation of the timing aspect. Naturally, not a requirement though.

buildbreaker:
@goaway Yeah, I recognize that could be a solution here. I'm still kind of weighing the cost/benefit of implementing this. For now, I think we can hold off on a change like this since it is primarily to help improve the testability of the application code

buildbreaker:
Going to close this PR for now

PR #1602, opened 2021-07-20
stats: add event tracker

Augustyniak:
This PR is a continuation of work from https://github.com/envoyproxy/envoy-mobile/pull/1590. I needed to move off from working from my branch in EM repo since I lost write access to envoy-mobile repo.

Description: Add an interface for tracking events that can be used by `EnvoyMobile` to inform platform code (i.e., Android or iOS) about events inside of envoy platform bridge. There are currently no events that are emitted. At Lyft we have a plan to use this interface to emit events from within one of our private extensions to Envoy Mobile.
Testing: Manual, integration test
Docs Changes: Done
Release Notes: None


Augustyniak:
@snowp @jingwei99 @rebello95 @goaway (I cannot assign you to this PR so just pinging instead). See the original PR https://github.com/envoyproxy/envoy-mobile/pull/1590 for previous comments related to my changes.

library/common/extensions/filters/http/test_event_tracker/config.h
```diff
@@ -0,0 +1,36 @@
...
+ * Config registration for the EventTracker filter. @see NamedHttpFilterConfigFactory.
+ */
+class TestEventTrackerFilterFactory
+    : public Common::FactoryBase<
+          envoymobile::extensions::filters::http::test_event_tracker::TestEventTracker> {
+public:
+  TestEventTrackerFilterFactory() : FactoryBase("test_event_tracker") {}
+
+private:
+  ::Envoy::Http::FilterFactoryCb createFilterFactoryFromProtoTyped(
```
jingwei99, line 26: what's `Cb`?

library/common/types/utility.cc
```diff
@@ -0,0 +1,27 @@
+#include <string>
+
+#include "library/common/data/utility.h"
+
+namespace Envoy {
+namespace Types {
+
+envoy_map makeEnvoyMap(std::vector<std::pair<std::string, std::string>> pairs) {
```
jingwei99, line 8: this method probably belong here: https://github.com/envoyproxy/envoy-mobile/blob/main/library/common/data/utility.h

library/common/types/utility.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <string>
+
+#include "library/common/types/c_types.h"
+
+namespace Envoy {
+namespace Types {
+
+envoy_map makeEnvoyMap(std::vector<std::pair<std::string, std::string>> pairs);
```
jingwei99, line 10: this method probably belong here: https://github.com/envoyproxy/envoy-mobile/blob/main/library/common/data/utility.h

test/common/types/utility_test.cc
```diff
@@ -0,0 +1,24 @@
+#include "gtest/gtest.h"
+#include "library/common/data/utility.h"
+#include "library/common/types/c_types.h"
+#include "library/common/types/utility.h"
+
+namespace Envoy {
+namespace Types {
+
+TEST(EnvoyMapConvenientInitializerTest, FromCppToCEmpty) {
```
jingwei99, line 9: this file probably belong here: https://github.com/envoyproxy/envoy-mobile/tree/main/test/common/data

library/common/extensions/filters/http/test_event_tracker/BUILD
```diff
@@ -0,0 +1,45 @@
...
+envoy_proto_library(
+    name = "filter",
+    srcs = ["filter.proto"],
+    deps = [
+        "@envoy_api//envoy/config/common/matcher/v3:pkg",
+    ],
+)
+
+envoy_cc_extension(
+    name = "test_event_tracker_filter_lib",
```
jingwei99, line 21: im curious about the `test` prefix in the name `test_event_tracker_`...

docs/root/api/starting_envoy.rst
```diff
@@ -259,6 +259,22 @@ Specify a closure to be called when Envoy's engine emits a log message.
     NSLog("Envoy log: \(msg)")
   }
 
+~~~~~~~~~~~~~~~~~~~
```
jingwei99, line 262: maybe save the doc for when the function is ready end to end?

jingwei99 (commented):
looks good overall 

a few comments/questions

library/common/extensions/filters/http/test_event_tracker/config.h
Augustyniak: This is just an alias on a function that comes from Envoy repo [here](https://github.com/envoyproxy/envoy/blob/main/envoy/http/filter.h#L1049-L1056). The function is used to create filter - I think it's something like a factory but implemented with the use of a function that's returned here and should be called every time a new instance of a filter is needed. 

Frankly speaking, the only reason why I know about this type is because I used `TestAccessor` as an example for my changes and it uses `Cb` in [here](https://github.com/envoyproxy/envoy-mobile/blob/main/library/common/extensions/filters/http/test_accessor/config.h#L26)

library/common/extensions/filters/http/test_event_tracker/BUILD
Augustyniak: The prefix is because this filter is intended to be used in tests only  I needed to add this test filter since without it there is no place in Envoy Mobile that emits any events and I want to be test that events emissions works as expected. The addition of a filter was suggested in my [previous PR](https://github.com/envoyproxy/envoy-mobile/pull/1590#discussion_r670649498) and I decided to call it `test_...` since we already have `test_accessor` filter. There may be a better way to do this //cc @junr03 who worked on test_accessor filter in case he has opinions.

library/common/types/utility.h
Augustyniak: mhm, this sounds like a good idea - @goaway mind chiming in since I talked with you about the current location of this file (which I know is not perfect)? I think that `common/data/utility` may be a good place for my utility function too - do you have any preference here?

rebello95 (commented):
Overall LGTM but would like @goaway or @junr03 to take a final pass on the C++

library/common/extensions/filters/http/test_event_tracker/BUILD
Augustyniak: will add some comments that will make it clearer that the filter is suppose to be for testing purposes only.

library/common/extensions/filters/http/test_event_tracker/config.h
jingwei99: ah Cb probably means callbacks

goaway: It means callbacks :)

library/common/extensions/filters/http/test_event_tracker/config.cc
```diff
@@ -0,0 +1,30 @@
+#include "library/common/extensions/filters/http/test_event_tracker/config.h"
+
+#include "library/common/extensions/filters/http/test_event_tracker/filter.h"
```
goaway, line 3: Obviously, you're following an existing pattern (established by the assertion filter, among other things), but I'd strongly prefer we not be building test infrastructure into the core library. Not going to ask you to solve this in this PR (though if you did, that would be great), but did want to mention I don't think this is a long-term tenable approach.

library/common/types/utility.cc
```diff
@@ -0,0 +1,27 @@
+#include <string>
+
+#include "library/common/data/utility.h"
+
+namespace Envoy {
+namespace Types {
```
goaway, line 6: I'd suggest the namespace "Bridge" here.

library/common/types/utility.h
goaway: I actually think the Data namespace might be overspecified. I agree there's an opportunity for cleanup, but I'd suggest limiting the scope of this PR. (I think a Bridge namespace alongside these type definitions might make more sense moving forward, and have been thinking a little bit about how to restructure this layer in general.)

library/objective-c/EnvoyBridgeUtility.h
```diff
@@ -100,6 +100,23 @@ static inline NSString *to_ios_string(envoy_data data) {
   return platformString;
 }
 
+static inline EnvoyEvent *to_ios_map(envoy_map map) {
+  NSMutableDictionary *newMap = [NSMutableDictionary new];
```
goaway, line 104: The last time I looked at bytecode, I think `alloc`/`init` was still marginally preferable, but obviously this isn't a terrific concern.

goaway:
Overall I think this looks good. What's the story with the Android side of things?

library/common/extensions/filters/http/test_event_tracker/config.h
Augustyniak: haha, I thought that we had `Ca` somewhere `Cb` was just next available name - lol

library/common/types/utility.h
Augustyniak: Moved to Bridge named  in https://github.com/envoyproxy/envoy-mobile/pull/1602/commits/2f00eae81101ae1b94c16e5f15ee84d78bfed3a5

test/common/types/utility_test.cc
Augustyniak: moved to `test/common/bridge` per suggestion above.

library/common/types/utility.cc
Augustyniak: Done in https://github.com/envoyproxy/envoy-mobile/pull/1602/commits/2f00eae81101ae1b94c16e5f15ee84d78bfed3a5

library/objective-c/EnvoyBridgeUtility.h
Augustyniak: we use `new` in multiple places in this file already - for consistency sake will keep `new` here. This is something that will be hard to ensure unless we lint against the usage of `new` (if such linters exist)

library/common/extensions/filters/http/test_event_tracker/BUILD
Augustyniak: > Overall I think this looks good. What's the story with the Android side of things?

We are going to start working on Android changes after iOS/common changes from this PR land. @jingwei99 offered to help with them :)

library/common/extensions/filters/http/test_event_tracker/config.cc
Augustyniak: Discussed offline and created a GH issue to track this work https://github.com/envoyproxy/envoy-mobile/issues/1610

library/objective-c/EnvoyBridgeUtility.h
goaway: Fair enough.

Augustyniak:
> Overall I think this looks good. What's the story with the Android side of things?

We are going to start working on Android changes sometimes after iOS/common changes from this PR lands. @jingwei99 offered to help with the changes :)

goaway (dismissed):
Nice work, thanks @Augustyniak!

goaway (approved):

PR #1603, opened 2021-07-20
config p3

goaway:


goaway:
More coming, just split this off for reference to the interfaces.

jingwei99:
makes sense to me

the only thing i was thinking about is that in this way we'll duplicate the config between ios and android, but i didn't come up with a good approach to de-dupe, i thought about having a separate config file, but it feels too much - a config file for configs lol

junr03:
re-open when picking back up

PR #1609, opened 2021-07-20
cc: remove designated initializer

crockeo:
Description: Remove the designated initializer I used in `engine_builder.cc` and instead populate `null_logger` field-by-field. Turns out this syntax is one of this parts of C++ that's _not_ the same as C.
Risk Level: Low
Testing: `bazel test //test/...`
Docs Changes: N/A
Release Notes: N/A


goaway:
Thanks @crockeo!

goaway (approved):

PR #1626, opened 2021-07-23
ci: fix workflow name

goaway:


goaway:
Have to re-enable required checks, so might as well fix this.

rebello95 (approved):

goaway:
The only failing check is the shadow check - there's something going on with the RBE instances, but it was never a required check anyways.

PR #1634, opened 2021-07-27
ci: increase timeout for some jobs

junr03:
Description: I've noticed that these jobs are timing out making CI red on PRs and main. Being conservative, specially with the App runs because we don't want them to run _too_ long.


goaway (approved):
Thanks @junr03!

PR #1635, opened 2021-07-27
register a no-op event tracker on Android

Augustyniak:
Description: Register a no-op event tracker so that we can implement Envoy extensions that assume that event tracker API is always registered. The tracker has `track` method equal to NULL which makes it effective a no-op thing.
Risk Level: Small (additive change that's effectively not doing anything)
Testing: Manual
Docs Changes: N/A
Release Notes: None
[Optional Fixes #Issue]
[Optional Deprecated:]


goaway (approved):
Looks good!

PR #1636, opened 2021-07-28
dev: add method for enabling admin interface

Augustyniak:
Add method for enabling Envoy admin interface.

Description:
Risk Level: Low (new feature, disabled by default)
Testing: Unit, Manual
Docs Changes: Added
Release Notes: None


library/common/config/config.cc
```diff
@@ -46,6 +46,7 @@ const std::string config_header = R"(
 - &statsd_port 8125
 - &stream_idle_timeout 15s
 - &virtual_clusters []
+- &admin_interface {}
```
goaway, line 49: I'm not certain this is safe. In fact, the interpretation of this when mapping YAML to Proto may result in an admin interface initialized to all (proto) default values, which could lead to crashes, security issues or unpredictable behavior.

Augustyniak: good point, removed

goaway (approved):
Thanks for addressing feedback @Augustyniak! This looks good.

Issue #1546, opened 2021-06-24
Cronvoy: Negotiated Protocol


carloseltuerto:
As part of the Cronet API, the Cornet Engine exposes the "Negotiated Protocol".

/**
 * Returns the protocol (for example 'quic/1+spdy/3') negotiated with the server.
 * Returns an empty string if no protocol was negotiated, the protocol is
 * not known, or when using plain HTTP or HTTPS.
 */
[public String getNegotiatedProtocol()](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/components/cronet/android/api/src/org/chromium/net/UrlResponseInfo.java#92)

This somehow will need to be resurfaced by the Envoy Mobile Engine. I guess that adding a ```x-envoy-negotiated-protocol``` response header would suffice.

alyssawilk:
yeah, I think this needs to go in StreamInfo somewhere (I can take that) then we have to expose up to mobile.  @junr03 / @goaway is there a generic way for mobile to get access to streaminfo fields or a example code you can point me to where we do that now?

alyssawilk:
Question for @carloseltuerto / @DavidSchinazi / @RyanTheOptimist if no protocol is negotiated (ALPN not attempted) is this still filled in with the protocol used?
AFIK Cronvoy is still hard-coded to use HTTP/2 by default so isn't actually negotiating.  I should fix that, though that'll need more coordination with the lyfties becasue [implementation details]

RyanTheOptimist:
Based on my reading of the code, if ALPN is not use then the negotiated protocol will be unknown.
```
  // Returns the protocol negotiated via ALPN for this socket, or
  // kProtoUnknown will be returned if ALPN is not applicable.
  virtual NextProto GetNegotiatedProtocol() const = 0;```

https://source.chromium.org/chromium/chromium/src/+/main:net/socket/stream_socket.h;l=137?q=getNegotiatedProtocol&ss=chromium

https://source.chromium.org/chromium/chromium/src/+/main:net/socket/ssl_client_socket_impl.cc;l=560;drc=d53e1d68f1d6a20877b5e845cd88dd393bf7dd75;bpv=1;bpt=1

**Comment by carloseltuerto** [comment]
The description of the getNegotiatedProtocol() states that the returned String is empty when using plain https or http. I browsed chromium for tests, and here are the strings I got (maybe some are bogus):

```
"h2"
"h3"
"http/2+quic"
"quic/1+spdy/3"
```

[...]

PR #1638, opened 2021-07-28
Adjust code and test since HTTP1 capitalization has been fixed

carloseltuerto:
HTTP1 capitalization has been fixed, so all the ```toLowerCase``` were removed.
The "number of byte received" check as been disabled in order to enable encompassing tests

Description: Adjust code and test since HTTP1 capitalization has been fixed
Risk Level: None
Testing: yes
Docs Changes: N/A
Release Notes: N/A
Signed-off-by: Charles Le Borgne cleborgne@google.com


library/java/org/chromium/net/impl/CronetUrlRequest.java
```diff
@@ -403,11 +403,11 @@ private void onResponseHeaders(ResponseHeaders responseHeaders, boolean lastCall
       return;
     }
     final List<Map.Entry<String, String>> headerList = new ArrayList<>();
-    String selectedTransport = "http/1.1"; // TODO(carloseltuerto) looks dubious
```
alyssawilk, line 406: Is this the ALPN issue? Should we leave it with a TODO?

alyssawilk (approved):
Yay progress!

library/java/org/chromium/net/impl/CronetUrlRequest.java
carloseltuerto: I added a todo further down, line 431

``` // TODO(https://github.com/envoyproxy/envoy-mobile/issues/1546) negotiated protocol```

For HTTP1, the logic here is fine: the expected value is ```"unknown"``` - that's what the original test is looking for.

For HTTP2/3, I don't have tests yet, as I need to make TLS to work for tests.

buildbreaker (approved):
🎉

PR #1640, opened 2021-07-29
test: ensure we properly shut down the engine thread in test

snowp:
Ensures that we shut down and join the engine thread in engine_test, preventing a data race from happening when the test runner
writes to Envoy::Logger::current_log_context during shutdown. 

Also updates the tsan-dev option to specify the same overrides as upstream Envoy, without this we see failures when using absl::Notification.

Description: n/a
Risk Level: Low, test only
Testing: Existing tests
Docs Changes: n/a
Release Notes: n/a


goaway (approved):
Thanks @snowp!

PR #1645, opened 2021-07-29
CI: disable shadow build for upgrade

ulfjack:
Description: This is in preparation for changing the remote execution
    service authentication configuration. We could theoretically do
    it without downtime, but since this isn't on the critical path yet,
    let's go the easier route.
Risk Level: Low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a


goaway (commented):
Approved/merged.

Issue #1215, opened 2020-12-18
ios: set up objective-c unit tests under bazel


goaway:
So that coverage can be added to library/objective-c/EnvoyBridgeUtility.h

Attempting to do this previously, ran into an issue where bazel wouldn't detect test (and thus, run) test targets.

goaway:
https://github.com/envoyproxy/envoy-mobile/pull/1216

goaway:
This was completed with the above PR.

PR #1654, opened 2021-07-31
Comma headers

junr03:
Description: Lyft uses x-location and sets them as `lat,long`. Currently Envoy Mobile incorrectly separates the header into two different values.
Risk Level: med
Testing: no existing test suite. Part of https://github.com/envoyproxy/envoy-mobile/issues/1215

### DIFF
--- a/library/java/io/envoyproxy/envoymobile/engine/JvmBridgeUtility.java
+++ b/library/java/io/envoyproxy/envoymobile/engine/JvmBridgeUtility.java
@@ -49,8 +49,10 @@ class JvmBridgeUtility {
     }
 
     // These headers may contain commas in single values, in contravention of the RFC.
-    if (headerKey.equalsIgnoreCase​("cookie") || headerKey.equalsIgnoreCase​("proxy-authenticate") ||
-        headerKey.equalsIgnoreCase​("set-cookie") || headerKey.equalsIgnoreCase​("www-authenticate") ||
+    if (headerKey.equalsIgnoreCase​("cookie") ||
+        headerKey.equalsIgnoreCase​("proxy-authenticate") ||
+        headerKey.equalsIgnoreCase​("set-cookie") ||
+        headerKey.equalsIgnoreCase​("www-authenticate") ||
         headerKey.equalsIgnoreCase​("x-location")) {
       values.add(headerValue);
     } else {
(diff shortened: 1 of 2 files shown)


junr03:
This PR introduced the behavior: https://github.com/envoyproxy/envoy-mobile/pull/1654. As far as I can see Envoy itself does not have logic to separate comma-delimited headers, and I also see empirically that after my PR x-location gets sent upstream without being split. @goaway I am not sure what was the context that you and @rebello95 had for the original commit in #1654.

goaway:
This is a non-standard header, so I'm not sure it makes sense to add it here.

goaway:
I can see a couple possibilities to move forward though:
1. Switch header-splitting to be an allow list (also can we reconfirm why we added this behavior?).
2. Make the exception list overridable/configurable.
3. Automatically except "x-" prefixed headers (but we should reconfirm the reasoning per option 1. again).

Here's the RFC: https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

goaway:
@junr03 I believe the original change was motivated by an issue @rebello95 encountered, but I can't recall the details offhand. @buildbreaker do you remember the context?

goaway (changes requested):
I think treating the prefix "x-" exceptionally here might be the most expedient option, but we still need to confirm the original context.

buildbreaker:
@goaway I don't remember the context for that issue

junr03:
@goaway I don't have much to go with respect to historic context from the PRs (https://github.com/envoyproxy/envoy-mobile/pull/1212 https://github.com/envoyproxy/envoy-mobile/pull/1213) or my searches in Lyft's or Envoy's slack. 

I propose for the time being that we allow this PR to return `x-location` to the behavior that URLSession and OkHttp have. And when @rebello95 is back loop back on what the intention of the original PRs was and work from there to put in place a long term solution.

pazabo (approved):

junr03:
@goaway @buildbreaker  coming back to this after my time OOO. Do you object to landing this to preserve legacy behavior and then work on a longer term solution?

goaway:
To clarify my stance above, I do not think Envoy Mobile should have special handling for a Lyft-specific header. I think any of the suggestions I mention above could be implemented in short order as a reasonable path forward.

Having thought about the issue some more now, IIRC the issue that prompted this was that Envoy was actually coalescing some headers which had repeated header names, and _that_ was what was different from the native stack. Of course, what Envoy was doing was valid per the RFC, and Envoy Mobile "undoing" it is in somewhat more of a grey area, so one could argue that this behavior should go altogether.

PR #1656, opened 2021-08-02
dev: fix enabling of admin interface

Augustyniak:
Change the way we do string manipulations to avoid the following error during Envoy startup:
```
error initializing configuration '': yaml-cpp: error at line 6078, column 1: end of map not found
```

The error was not caught during manual testing since it occurs only when both `addGrpcStatsDomain` and `enableAdminInterface` methods are called on `EngineBuilder` prior to Envoy start up.

Risk Level: Low (fixes Envoy crash)
Testing: Manual
Docs Changes: N/A
Release Notes: None

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -455,7 +455,6 @@ static_resources:
     upstream_connection_options: *upstream_opts
     circuit_breakers: *circuit_breakers_settings
     typed_extension_protocol_options: *base_protocol_options_defs
-#{admin_interface}
 stats_flush_interval: *stats_flush_interval
 stats_sinks: *stats_sinks
 stats_config:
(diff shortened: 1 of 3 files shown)


library/common/config/config.cc
```diff
@@ -455,6 +455,7 @@ const char* config_template = R"(
     upstream_connection_options: *upstream_opts
     circuit_breakers: *circuit_breakers_settings
     typed_extension_protocol_options: *base_protocol_options_defs
+#{admin_interface}
```
goaway, line 458: This is a pattern we're actually looking to avoid introducing more of. Can we find an alternative closer to your original approach?

Augustyniak:
It turns out that this was broken on iOS only

library/common/config/config.cc
Augustyniak: Done

goaway (approved):
Thanks @Augustyniak!

PR #1660, opened 2021-08-04
stats: add event tracker for android

jingwei99:
Adds event tracker for android. 

The ios change was done a few days ago: https://github.com/envoyproxy/envoy-mobile/pull/1602

This pr also adds an event tracker to the kotlin example app. Sample logs:
Screen Shot 2021-08-03 at 5 03 17 PM

Description:
Risk Level: low, extensively tested
Testing: example app, unit testing & ci
Docs Changes: done
Release Notes: Adds event tracker for android

### DIFF
--- a/library/common/jni/jni_interface.cc
+++ b/library/common/jni/jni_interface.cc
@@ -873,18 +873,41 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_registerStringAccessor(JNIEnv* 
   return result;
 }
 
+static void jvm_on_track(envoy_map events, const void* context) {
+  jni_log("[Envoy]", "jvm_on_track");
+  if (context == nullptr) {
+    return;
+  }
+
+  JNIEnv* env = get_env();
+  jobject events_hashmap = native_map_to_map(env, events);
+
+  jobject j_context = static_cast<jobject>(const_cast<void*>(context));
+  jclass jcls_EnvoyEventTracker = env->GetObjectClass(j_context);
+  jmethodID jmid_onTrack = env->GetMethodID(jcls_EnvoyEventTracker, "track", "(Ljava/util/Map;)V");
+  env->CallVoidMethod(j_context, jmid_onTrack, events_hashmap);
+
+  env->DeleteLocalRef(events_hashmap);
+  env->DeleteLocalRef(jcls_EnvoyEventTracker);
+}
+
 // EnvoyEventTracker
 
 extern "C" JNIEXPORT jint JNICALL
-Java_io_envoyproxy_envoymobile_engine_JniLibrary_registerEventTracker(JNIEnv* env, jclass) {
-  // TODO(Augustyniak): The event_tracker leaks, but it's tied to the life of the engine.
-  // This will need to be updated for https://github.com/lyft/envoy-mobile/issues/332.
-  envoy_event_tracker* event_tracker =
-      (envoy_event_tracker*)safe_malloc(sizeof(envoy_event_tracker));
-  // TODO(Augustyniak): Allow for the registration of a "real" (no no-op) event tracker.
-  event_tracker->track = nullptr;
-  event_tracker->context = nullptr;
+Java_io_envoyproxy_envoymobile_engine_JniLibrary_registerEventTracker(JNIEnv* env, jclass,
+                                                                      jobject j_event_tracker) {
+  jni_log("[Envoy]", "register event tracker");
 
+  // TODO(goaway): The retained_context leaks, but it's tied to the life of the engine.
+  // This will need to be updated for https://github.com/lyft/envoy-mobile/issues/332.
+  jobject retained_context = env->NewGlobalRef(j_event_tracker);
+  jni_log_fmt("[Envoy]", "retained_context: %p", retained_context);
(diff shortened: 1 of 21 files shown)


library/common/jni/jni_interface.cc
```diff
@@ -873,18 +873,41 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_registerStringAccessor(JNIEnv*
...
-  // TODO(Augustyniak): Allow for the registration of a "real" (no no-op) event tracker.
-  event_tracker->track = nullptr;
-  event_tracker->context = nullptr;
+Java_io_envoyproxy_envoymobile_engine_JniLibrary_registerEventTracker(JNIEnv* env, jclass,
+                                                                      jobject j_event_tracker) {
+  jni_log("[Envoy]", "register event tracker");
 
+  // TODO(goaway): The retained_context leaks, but it's tied to the life of the engine.
+  // This will need to be updated for https://github.com/lyft/envoy-mobile/issues/332.
+  jobject retained_context = env->NewGlobalRef(j_event_tracker);
```
buildbreaker, line 903: Do we delete this ref anywhere?

jingwei99: _i think_  the `retained_context` should be kept, since `event_tracker->context` is pointed to it, i might be wrong though

library/java/io/envoyproxy/envoymobile/engine/EnvoyEngineImpl.java
```diff
@@ -14,18 +15,17 @@
...
-  private EnvoyOnEngineRunning onEngineRunning;
-  private EnvoyLogger logger;
+  private final EnvoyEventTracker eventTracker;
 
   /**
    * @param runningCallback Called when the engine finishes its async startup and begins running.
    * @param logger          The logging interface.
    */
-  public EnvoyEngineImpl(EnvoyOnEngineRunning runningCallback, EnvoyLogger logger) {
+  public EnvoyEngineImpl(EnvoyOnEngineRunning runningCallback, EnvoyLogger logger,
```
Augustyniak, line 24: nit: add doc strings for newly added `eventTracker` argument

test/kotlin/integration/SetEventTrackerTest.kt
```diff
@@ -0,0 +1,108 @@
...
+
+  @Test
+  fun `set eventTracker`() {
+    val engine = EngineBuilder(Custom(config))
+      .addNativeFilter(
+        "envoy.filters.http.test_event_tracker",
+        "{\"@type\":\"type.googleapis.com/envoymobile.extensions.filters.http.test_event_tracker.TestEventTracker\",\"attributes\":{\"foo\":\"bar\"}}"
+      )
+      .setEventTracker { it ->
+        for (entry in it) {
```
Augustyniak, line 69: Can we make this test fail for cases when an event tracker closure/lambda is not called to verify that event tracker is used?

library/common/jni/jni_interface.cc
```diff
@@ -873,18 +873,41 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_registerStringAccessor(JNIEnv*
   return result;
 }
 
+static void jvm_on_track(envoy_map events, const void* context) {
```
Augustyniak, line 876: definitely not familiar with how we are dealing with memory in Java bridging code etc so I may be wrong here but I think that we are missing the release of `events` memory?

Augustyniak (dismissed):
Looks great! One nit comment, 2 suggestions/questions with a note that we should fix the one that relates to the memory (the lack of release) if we think that it really leaks.

library/common/jni/jni_interface.cc
jingwei99: it sounds that it should be the case, but im not sure, given my limit on memory mamangement.

for this method i followed the way `jvm_on_log` method releases its resources

library/common/jni/jni_interface.cc
goaway: It could be deleted during shutdown for hygiene's sake, but it looks like it's meant to be around for the lifetime of the engine, and we leak other such references today.

library/common/jni/jni_interface.cc
goaway: Correct, it will need to be released somehow. In other code (iirc) the native_x_to_y functions handle it, but that seems to not be the case here. Regardless, `release_envoy_map` needs to be called when we're done with it.

goaway:
Overall looks good; see comments.

library/common/jni/jni_interface.cc
jingwei99: i see 👌

library/java/io/envoyproxy/envoymobile/engine/types/EnvoyEventTracker.java
```diff
@@ -0,0 +1,7 @@
+package io.envoyproxy.envoymobile.engine.types;
+
+import java.util.Map;
+
+public interface EnvoyEventTracker {
```
goaway, line 5: We also need a new proguard rule for this to keep its methods from getting stripped.

goaway (approved):
Thanks @jingwei99!

PR #1663, opened 2021-08-05
Use self-provided xcode_config target instead of relying on @local_config_xcode

Yannic:
Description:

With remote execution, envoy mobile spends a significant fraction
of its built time in loading/analysis phase, and specificaly, in a
7.5 minutes build, around 2.7 minutes in evaluating @local_config_xcode
which auto configures Bazel with the locally installed Xcode versions.
Since envoy mobile passes `--xcode_version=12.4` in its `.bazelrc`, thus
requires a specific Xcode version, we can make the build significantly
faster by creating our own `xcode_config` target instead of relying
on Bazel's auto configuration.

Example invocation: https://envoy.cluster.engflow.com/invocation/a23bd20b-f0b9-472b-9a07-b9e5bc629b30#performance

Risk Level: low
Testing: local/CI
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -16,7 +16,6 @@ build --ios_simulator_version=14.4
 build --spawn_strategy=local
 build --verbose_failures
 build --workspace_status_command=envoy/bazel/get_workspace_status
-build --xcode_version_config=//ci:xcode_config
 build --xcode_version=12.4
 build --javabase=@bazel_tools//tools/jdk:jdk
 build --define disable_known_issue_asserts=true
@@ -96,6 +95,7 @@ build:coverage --javabase=@bazel_tools//tools/jdk:remote_jdk11
 # however, the MacOS remote execution cluster doesn't use GCP auth
 # so we have to override it to 'false' here.
 build:remote-ci-macos --config=remote
+build:remote-ci-macos --xcode_version_config=//ci:xcode_config
 build:remote-ci-macos --auth_enabled=false
 build:remote-ci-macos --remote_executor=grpcs://envoy.cluster.engflow.com
 build:remote-ci-macos --bes_backend=grpcs://envoy.cluster.engflow.com/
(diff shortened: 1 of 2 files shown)


.bazelrc
```diff
@@ -16,6 +16,7 @@ build --ios_simulator_version=14.4
 build --spawn_strategy=local
 build --verbose_failures
 build --workspace_status_command=envoy/bazel/get_workspace_status
+build --xcode_version_config=//ci:xcode_config
```
keith, line 19: to avoid requiring folks having 12.4 locally can you move this to the `build:remote-ci-macos` config?

keith (dismissed):

.bazelrc
Yannic: the project already requires everyone to have 12.4 because of `--xcode_version=12.4` below, but sure will do

keith (approved):

Yannic:
Looks like https://envoy.cluster.engflow.com/invocation/29b3f4f8-1fac-439b-99c1-5bd44c0e8402 took longer with this change, but I suspect that it's because of less cache hits during the build (Bazel sends the exact Xcode version in the action Proto and we intentionally only specify 12.4 instead of 12.4.0.<buildid> in `xcode_version` so this works with both the AppStore Xcode and the dev portal Xcode which have different builder's).

Without this , Bazel spend ~3min in analysis, which this change seems to cut roughly in half.

goaway:
Just approved CI run.

keith (approved):

Yannic:
not sure why swift_tests fail, looks unrelated

Yannic:
thanks! looks like kotlin_tests_linux also flaked

goaway:
Rerunning.

goaway (approved):

PR #1668, opened 2021-08-06
CI: Create configuration for running Linux CI jobs on remote execution

lfpino:
Description: Some of the presubmit jobs run on Linux and as part of migrating CI to use remote execution we need to add a Linux-only config to the bazelrc that can be referenced by the corresponding Github actions.

Risk Level: Low
Testing: Locally
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -82,15 +82,31 @@ build:coverage --instrumentation_filter="//library/common[/:]"
 build:coverage --host_javabase=@bazel_tools//tools/jdk:remote_jdk11
 build:coverage --javabase=@bazel_tools//tools/jdk:remote_jdk11
 
-# Experimental EngFlow Remote Execution Config
-# --config=remote also sets --auth_enabled=true to enable GCP authentication.
-# We are not using GCP authentication for the MacOS remote execution cluster,
-# so we have to override it to false here.
-build:remote-ci-macos --config=remote
+#############################################################################
+# Experimental EngFlow Remote Execution Configs                             
+#############################################################################
+# remote-ci: These options are valid for any platform, use the configs below
+# to add platform-specific options
+#############################################################################
+build:remote-ci --config=remote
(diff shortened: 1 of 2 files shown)


.bazelrc
```diff
@@ -82,15 +82,31 @@ build:coverage --instrumentation_filter="//library/common[/:]"
...
+build:remote-ci --remote_executor=grpcs://envoy.cluster.engflow.com
+build:remote-ci --bes_backend=grpcs://envoy.cluster.engflow.com/
+build:remote-ci --bes_results_url=https://envoy.cluster.engflow.com/invocation/
+build:remote-ci --bes_lifecycle_events
+build:remote-ci --jobs=40
+build:remote-ci --verbose_failures
+#############################################################################
+# remote-ci-linux: These options are linux-only
+#############################################################################
+build:remote-ci-linux --config=remote
```
Yannic, line 101:
```suggestion
build:remote-ci-linux --config=remote-ci
```

.bazelrc
```diff
@@ -82,15 +82,31 @@ build:coverage --instrumentation_filter="//library/common[/:]"
...
+#############################################################################
+# remote-ci-macos: These options are macOS-only
+#############################################################################
+build:remote-ci-macos --config=remote-ci
+# --config=remote sets --auth_enabled=true to enable GCP authentication,
+# however, the MacOS remote execution cluster doesn't use GCP auth
+# so we have to override it to 'false' here.
 build:remote-ci-macos --auth_enabled=false
```
Yannic, line 109: I beliebe this should also be moved to the generic `remote-ci`, it's not Mac specific

lfpino: I wasn't sure, @ulfjack can confirm

ulfjack: Correct, it's not Mac specific. We just need to be careful not to conflict with the other .bazelrc settings here or in envoy.

PR #1673, opened 2021-08-09
stats: noop during registration when event tracker not set

jingwei99:
Description: do not register platform api when event tracker is nullptr
Risk Level: low
Testing: ci, local build

### DIFF
--- a/library/common/jni/jni_interface.cc
+++ b/library/common/jni/jni_interface.cc
@@ -920,6 +920,11 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_registerEventTracker(JNIEnv* en
                                                                       jobject j_event_tracker) {
   jni_log("[Envoy]", "register event tracker");
 
+  if (j_event_tracker == nullptr) {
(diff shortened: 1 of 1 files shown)


library/common/jni/jni_interface.cc
```diff
@@ -920,6 +920,11 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_registerEventTracker(JNIEnv* en
                                                                       jobject j_event_tracker) {
   jni_log("[Envoy]", "register event tracker");
 
+  if (j_event_tracker == nullptr) {
```
Augustyniak, line 923: for cases when `j_event_tracker` is `nullptr` we need to register an empty `envoy_event_tracker ` (with its `track` and `context` properties equal to `nullptr`). Otherwise the app will crash when lyne sink attempts to retrieve the registered api in [here](https://github.com/lyft/envoy-mobile-private/blob/master/extensions/stat_sinks/lyne/grpc_metrics_service_impl.cc#L26) and because of [this assertion](https://github.com/envoyproxy/envoy-mobile/blob/main/library/common/api/external.cc#L24)

PR #1682, opened 2021-08-10
build: bumping boring to HTTP/3 compatible version

alyssawilk:
Risk Level: low  (Envoy upstream uses this version)
Testing: assuming extra tests validate
Docs Changes: n/a
Release Notes: n/a


goaway (approved):
Thanks @alyssawilk!

PR #1703, opened 2021-08-13
assert: emit events for assertions

Augustyniak:
Description: Send an envoy event with the use of envoy event tracker every time an assertion is hit. The event contains two fields: `name` equal to `assertion` and location which contains the information about the name of the file and line number for where the assertion was hit.
Risk Level: Low. Optional feature that's active only when `DENVOY_LOG_DEBUG_ASSERT_IN_RELEASE` compilation flag is specified.
Testing: Unit/Integration
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/library/cc/engine_builder.cc
+++ b/library/cc/engine_builder.cc
@@ -120,8 +120,13 @@ EngineSharedPtr EngineBuilder::build() {
   null_logger.release = envoy_noop_const_release;
   null_logger.context = nullptr;
 
+  envoy_event_tracker null_tracker;
+  null_tracker.track = nullptr;
+  null_tracker.context = nullptr;
+
   auto config_str = this->generateConfigStr();
-  auto envoy_engine = init_engine(this->callbacks_->asEnvoyEngineCallbacks(), null_logger);
+  auto envoy_engine =
+      init_engine(this->callbacks_->asEnvoyEngineCallbacks(), null_logger, null_tracker);
   run_engine(envoy_engine, config_str.c_str(), logLevelToString(this->log_level_).c_str());
 
   // we can't construct via std::make_shared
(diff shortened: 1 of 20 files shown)


snowp (changes requested):
Some comments, mostly this looks good

library/cc/engine_builder.cc
```diff
@@ -120,8 +120,13 @@ EngineSharedPtr EngineBuilder::build() {
   null_logger.release = envoy_noop_const_release;
   null_logger.context = nullptr;
 
+  envoy_event_tracker null_tracker;
+  null_tracker.track = nullptr;
+  null_tracker.context = nullptr;
```
goaway, line 125:
```suggestion
  envoy_event_tracker null_tracker{};
```

library/common/main_interface.h
```diff
@@ -163,9 +163,11 @@ envoy_status_t register_platform_api(const char* name, void* api);
  * Initialize an engine for handling network streams.
  * @param callbacks, the callbacks that will run the engine callbacks.
  * @param logger, optional callbacks to handle logging.
+ * @param event_tracker, optional event tracker for the emission of events.
```
goaway, line 166: Is this optional?

goaway: (Asking because you're explicitly registering a null tracker in the cc path.)

goaway:
In principal I'm good with this. I have some nits that I've refrained from posting here - I'd be fine with merging and iterating. Depending on what you'd like to do though, I can post them and we could iterate here as well.

goaway (dismissed):

library/common/main_interface.h
Augustyniak: nah, it's not - used to be initially since it was a pointer but I changed it to be a struct so this no longer holds. Updating.

buildbreaker (dismissed):

buildbreaker (approved):

goaway (approved):

PR #1704, opened 2021-08-14
Revert "ci: revert EngFlow ios_build changes (#1700)"

lfpino:
Description:
This reverts commit 1adf02d1cb46357f6db11ad2370e2879d212d2ec.

The machine was running out of space because it was keeping large replicas of build artifacts for too long. We've since increased the disk space as well as decreased the timeout for evicting those large artifacts. Performance may be slightly degraded if those artifacts need to go to disk to be fetched but by looking at previous results it'll probably be ok

Risk Level: Low
Testing: See ios_build presubmit
Docs Changes: N/A
Release Notes: N/A


goaway (approved):
Thanks @lfpino! Let's give this another go.

PR #1709, opened 2021-08-16
CI: Add linux config for Engflow's remote execution

lfpino:

Description:
CI: Add linux config for Engflow's remote execution. No semantic changes. This PR is a no-op.

We add two things to the envoy-mobile build:
- A Linux-specific bazelrc configuration that sets all the necessary flags to run on a Linux Engflow worker
- Instruct envoy to use the latest envoyproxy/envoy-build-ubuntu docker container for the linux builds through the WORKSPACE rules

Notice that the envoy image is only necessary for the Linux builds since the MacOS builds are set somehow manually and don't use containers. Also notice that Engflow builds have a notion of "Pool" which for now only apply to Linux workers since the first workers in the Envoy cluster were Macs and they didn't have a Pool associated to it (e.g. no pool means Mac at the moment). Once the MacOS workers belong to a Pool we'll update the MacOS config.

Risk Level: Low
Testing: No-op
Docs Changes: N/A
Release Notes: N/A


.bazelrc
```diff
@@ -91,21 +91,50 @@ build:coverage --javabase=@bazel_tools//tools/jdk:remote_jdk11
...
-# config is used for MacOS-specific flags.
+# remote-ci-common: These options are valid for any platform, use the configs below
+# to add platform-specific options. Avoid using this config directly and
+# instead use a platform-specific config
 #############################################################################
 # --config=remote sets --auth_enabled=true to enable GCP authentication,
-# however, the MacOS remote execution cluster doesn't use GCP auth
-# so we have to override it to 'false' here.
-build:remote-ci-macos --config=remote
-build:remote-ci-macos --xcode_version_config=//ci:xcode_config
```
keith, line 101: looks like we lost this config in this change?

lfpino: Good catch! Done now

keith (approved):

goaway (approved):

PR #1621, opened 2021-07-22
extending propercasing to legacy clusters, header sanitization to all routes

alyssawilk:
Risk Level: medium
hopefully addresses https://github.com/envoyproxy/envoy-mobile/issues/1558 and  #1557

PR #1710, opened 2021-08-16
Revert "extending propercasing to legacy clusters, header sanitization to all routes (#1621)"

buildbreaker:
This is causing some TLS errors:
```upstream connect error or disconnect/reset before headers. reset reason: connection failure, transport failure reason: TLS error: 268435703:SSL routines:OPENSSL_internal:WRONG_VERSION_NUMBER```

This reverts commit 6e507d3e753a0725eef6398d10dd1503c5dc3010.

Description: Revert "extending propercasing to legacy clusters, header sanitization to all routes (#1621)"
Risk Level: medium
Testing: manual testing
Docs Changes: n/a
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -65,17 +65,6 @@ const std::string config_header = R"(
       address:
         socket_address: { address: *statsd_host, port_value: *statsd_port }
 
-!ignore http1_protocol_defs: &http1_protocol_options_defs
-    envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
-      "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
-      explicit_http_config:
-        http_protocol_options:
-          header_key_format:
-            stateful_formatter:
-              name: preserve_case
-              typed_config:
-                "@type": type.googleapis.com/envoy.extensions.http.header_formatters.preserve_case.v3.PreserveCaseFormatterConfig
-
 !ignore protocol_defs: &base_protocol_options_defs
     envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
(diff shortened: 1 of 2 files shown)


mattklein123:
cc @alyssawilk I can't see any scenario in which this config change causes a TLS wrong version error, but I'm actually able to repro this on Linux so I'm looking into it.

mattklein123:
So the issue is in the following code:

https://github.com/envoyproxy/envoy/blob/5e48a19b20c10f31750ce430721ea868e7b1c457/source/extensions/clusters/dynamic_forward_proxy/cluster.cc#L182-L193

We did not previously configure `upstream_http_protocol_options`. This meant that both auto_sni() and auto_san_validation() have been set to true. Note also that `upstream_http_protocol_options` is deprecated. When we configure `typed_extension_protocol_options` we actually overwrite the values of this deprecated field, thus ignoring it:

https://github.com/envoyproxy/envoy/blob/5e48a19b20c10f31750ce430721ea868e7b1c457/source/common/upstream/upstream_impl.cc#L683-L689

This change definitely needs to be reverted and a few action items I would recommend:
1) Make sure we have tests that would have caught this problem.
2) Make sure we are warning/failing on using deprecated fields in Envoy Mobile.
3) Make sure that the DFP cluster checking code properly understands how to check for whether auto_sni and auto_san_validation are set.

mattklein123 (dismissed):

buildbreaker:
@carloseltuerto the `//test/java/org/chromium/net:cronet_url_request_test` is failing here. I will have to revert it :(

mattklein123 (dismissed):

goaway (approved):

goaway (approved):

PR #1713, opened 2021-08-17
CI: Re-enable ios_build in Engflow's remote execution 

lfpino:
Description:
CI: Re-enable ios_build in Engflow's remote execution.

The builds were timing out because the machines were down in AWS due to a restart.

Risk Level: Low
Testing: See ios_build check
Docs Changes: N/A
Release Notes: N/A


goaway (approved):
Approving, hopefully this is the one that takes.

We discussed in Slack the need for a plan for failure scenarios and an escalation path, but it sounds like the underlying issue should be fixed.

PR #1718, opened 2021-08-17
config: add outlier detection config

junr03:
Description: This PR (which picks up work started on #1567) adds outlier detection to all configured clusters.

The main objective of this configuration is to drain connections that are perceived to be "dead" so that the cluster is forced to establish a new connection for future streams. 

Currently the outlier detector will not eject already ejected hosts. And host ejection is what causes connection drain. In mobile, the desire is not to keep hosts ejected, but rather to cycle the underlying upstream connections. Hence the short base/max ejection time and the the tight unejection interval. The unejection interval is not any tighter because to balance timer churn.

The idea with the current configuration is that most, if not all "dead" connections will be reaped without having any adverse performance issues. Lyft will deploy and analyze production data to further tune the configuration. If ultimately different detector behavior (drain connections even if host is ejected) is desired, we can work to add to upstream envoy.

Risk Level:
Testing: This configuration was tested with the existing example apps pointed to a server that continuously returns 503. Expected host ejection, connection drain, host unejection behavior was observed.

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -252,6 +252,10 @@ static_resources:
             budget_percent:
               value: 100
             min_retry_concurrency: 0xffffffff # uint32 max
+    # Used to reap dead connections. In mobile devices it is less important to keep a "host" ejected
+    # a long period and more important to be able to cycle connections assigned to given hosts.
+    # Therefore, the ejection time is short and the interval for unejection is tight, but not too
+    # tight to cause unnecessary churn.
     outlier_detection: &base_outlier_detection
       consecutive_5xx: 3
       base_ejection_time: 0.001s
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -252,6 +252,11 @@ static_resources:
             budget_percent:
               value: 100
             min_retry_concurrency: 0xffffffff # uint32 max
+    outlier_detection: &base_outlier_detection
+      consecutive_5xx: 3
+      base_ejection_time: 0.001s
+      max_ejection_time: 0.001s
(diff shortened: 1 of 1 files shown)


library/common/config/config.cc
```diff
@@ -252,83 +252,99 @@ const char* config_template = R"(
             budget_percent:
               value: 100
             min_retry_concurrency: 0xffffffff # uint32 max
+    outlier_detection: &base_outlier_detection
+      consecutive_5xx: 3
+      base_ejection_time: 0.001s
+      max_ejection_time: 0.001s
+      interval: 1s
```
mattklein123, line 259: Can you add comments here on how/why these are selected?

mattklein123 (changes requested):
Cool thanks LGTM with comment ask.

goaway (approved):
LGTM

buildbreaker (approved):

library/common/config/config.cc
```diff
@@ -252,83 +252,103 @@ const char* config_template = R"(
             budget_percent:
               value: 100
             min_retry_concurrency: 0xffffffff # uint32 max
+    # Used to reap dead connections. In mobile devices it is less important to keep a "host" ejected
```
mattklein123, line 255: Follow up nit: this comment gets compiled in and just wastes binary space. I would probably move the comment and/or just move this config stanza out also and then use YAML include to include it.

goaway: Here's a take on what that could look like:
https://github.com/envoyproxy/envoy-mobile/pull/1722

PR #1719, opened 2021-08-17
stats: add outlier detection stats

junr03:
Description: adds allow list for outlier detection stats to instrument #1718 
Risk Level: low - while the stats tree is somewhat wide, the current config only will emit stats for consecutive 5xx limiting the number of stats emitted.


junr03:
@snowp @goaway I will create the internal changes once this is merged.

goaway (approved):
Thanks @junr03!

PR #1722, opened 2021-08-18
config: compile out inline config comments

goaway:
Description: Compile out inline config comments.
Risk Level: Low
Testing: CI


goaway:
cc @mattklein123 

This is one take on compiling out yaml comments. I'm not sure it's the best since:
1) The syntax is maybe a bit esoteric.
2) All this stuff gets zipped anyways, so the comments probably don't amount to much by the time it's packaged for distribution.

One potential plus to this approach though is it keeps the comments inline.

mattklein123 (dismissed):
I don't feel strongly about it either way. This seems fine to me.

buildbreaker (approved):

PR #1732, opened 2021-08-20
envoy:  26dae5d6542

junr03:
includes: https://github.com/envoyproxy/envoy/pull/17775 and https://github.com/envoyproxy/envoy/pull/17780


Augustyniak (approved):

buildbreaker (approved):

junr03:
Going to merge regardless because the failure on the new commit is CI failure and the change does not affect any of the jobs that passed in the previous commit. The only job that mattered is the python tests one and that is passing now.

PR #1733, opened 2021-08-20
add engine API to dump stats

snowp:
This adds a dumpStats function to the Engine interfaces, allowing callers to retrieve the value of all active stats.

This is done by invoking an admin handler, and so this PR also sets up a pattern for synchronously calling the admin
handler. This should make it easier to add support for other features that would be based on invoking admin endpoints.

Risk Level: Medium
Testing: New integration tests
Docs Changes: n/a
Release Notes: n/a


library/java/io/envoyproxy/envoymobile/engine/AndroidEngineImpl.java
```diff
@@ -48,6 +48,11 @@ public void flushStats() {
     envoyEngine.flushStats();
   }
 
+  @Override
+  public String dumpStats(int timeout_ms) {
```
Augustyniak, line 52: On iOS it's a convention to use `TimeInterval` for cases like this (it's a double that expresses time in seconds) -> you could call it `timeout` then. 

That being said, I think that we should get rid of this argument altogether. Let's come up with a reasonable value and hardcode it in the engine itself so that it's consistent between iOS and Android. We can then document the method telling its users that it's a blocking one and can take significant amount of time (and hence it's reasonable to call it from a non-main thread).

Otherwise, I think that we will end up in a situation in which the caller needs to decide on a timeout that's not obvious (as in, should this be 5s as for network requests or 100ms because it's local).

Augustyniak: Maybe before you make a change (if you do not have a strong opinion) I will let @goaway share his opinion here in case he thinks we should keep this parameter

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -540,6 +540,20 @@ - (void)flushStats {
   flush_stats(_engineHandle);
 }
 
+- (NSString *)dumpStats:(NSUInteger)timeout_ms {
+  envoy_data data;
+  envoy_status_t status = dump_stats(_engineHandle, &data, timeout_ms);
+  if (status == ENVOY_SUCCESS) {
+    NSString *string_copy = [[NSString alloc] initWithBytes:data.bytes
```
Augustyniak, line 547: should be called `stringCopy` I think

library/kotlin/io/envoyproxy/envoymobile/Engine.kt
```diff
@@ -27,4 +27,11 @@ interface Engine {
    * This is a noop if called before the underlying EnvoyEngine has started.
    */
   fun flushStats()
+
+  /**
+   * Retrieve the value of all active stats. Note that this function may block for some time.
+   * @param timeout_ms the timeout for how long to wait when fetching stats.
+   * @return The list of active stats and their values, or empty string of the operation failed
+   */
+  fun dumpStats(timeout_ms: Int): String
```
Augustyniak, line 36: I think that we should get rid of this argument (per my previous comment) but if we want to keep it, it should probably follow camel case

library/common/main_interface.cc
```diff
@@ -157,6 +159,64 @@ envoy_status_t record_histogram_value(envoy_engine_t, const char* elements, envo
...
+namespace {
+struct AdminCallContext {
+  envoy_status_t status_{};
+  envoy_data response_{};
+
+  absl::Mutex mutex_{};
+  absl::Notification data_received_{};
+};
+
+absl::optional<envoy_data> blockingAdminCall(absl::string_view path, absl::string_view method,
```
Augustyniak, line 171: maybe a test that verifies that `POST` calls work too - not a blocker but would be nice to verify it (since POST is not used at all as of now). Or just hardcode it to `GET` 🤷

library/common/jni/jni_interface.cc
```diff
@@ -212,6 +212,25 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_flushStats(JNIEnv* env,
...
+  if (result == ENVOY_SUCCESS) {
+    native_data_to_string(env, data);
+    jstring str = native_data_to_string(env, data);
+    release_envoy_data(data);
+
+    return str;
+  }
+
+  jstring jstrBuf = env->NewStringUTF("");
+  return jstrBuf;
```
Augustyniak, line 231: nit: why not simply return in line 230?

Augustyniak (approved):
Not sure whether "dump" stats is the right name for the method - I assume that it's used in Envoy, if yes it makes sense to use it here too I gue. Other than that looks good to me :)

snowp:
ptal @goaway

PR #1736, opened 2021-08-20
WIP update cipher list, allow 1.3

snowp:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -89,6 +89,19 @@ const std::string config_header = R"(
   typed_config:
     "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext
     common_tls_context:
+      tls_params:
+        tls_maximum_protocol_version: TLSv1_3
+        cipher_suites:
+          - ECDHE-ECDSA-AES256-GCM-SHA384
+          - ECDHE-ECDSA-AES128-GCM-SHA256
+          - ECDHE-ECDSA-AES256-SHA
+          - ECDHE-ECDSA-AES128-SHA
+          - ECDHE-ECDSA-CHACHA20-POLY1305
+          - ECDHE-RSA-AES256-GCM-SHA384
+          - ECDHE-RSA-AES128-GCM-SHA256
+          - ECDHE-RSA-AES256-SHA
+          - ECDHE-RSA-AES128-SHA
+          - ECDHE-RSA-CHACHA20-POLY1305
       validation_context:
         trusted_ca:
           inline_string: |


mattklein123:
cc @davidben in case you have a sec to take a look. We are trying to match up the ciphers per our offline discussion.

mattklein123:
cc @RyanTheOptimist also

library/common/config/config.cc
```diff
@@ -89,6 +89,15 @@ const std::string config_header = R"(
   typed_config:
     "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext
     common_tls_context:
+      tls_params:
+        tls_maximum_protocol_version: TLSv1_3
+        cipher_suites: ["ALL:!aPSK:!ECDSA+SHA1:!3DES"]
```
snowp, line 94: This should be the BoringSSL with the same modifications that Chrome uses (https://source.chromium.org/chromium/chromium/src/+/main:net/socket/ssl_client_socket_impl.cc;drc=c361f8a4b30a6b9f289107b8085d63876114dba5;l=854)

assuming that 3DES is not enabled and RSA is allowed

RyanTheOptimist: Yup, this is exactly what I see in my instance of Chrome.

RyanTheOptimist: Do we need to change the code here, too? https://github.com/envoyproxy/envoy/blob/main/source/extensions/transport_sockets/tls/context_config_impl.cc#L319
I'm not sure I understand the relationship between these two?

mattklein123: IMO we probably should change the default, but that is outside of the scope of this PR which is to specifically configure it right now for EM.

mattklein123 (approved):
Awesome, thanks.

library/common/config/config.cc
RyanTheOptimist: *facepalm* I completely missed that this PR is an Envoy *Mobile* PR. Ok, that makes a lot more sense. Agreed that it's clearly out of scope for this PR :) (But would be good for an Envoy PR.)

RyanTheOptimist (approved):

library/common/config/config.cc
mattklein123: We probably need to also discuss GREASE for 1.3, but also outside the scope of this because AIUI that requires additional Envoy code to enable it in bssl.

library/common/config/config.cc
```diff
@@ -95,6 +95,15 @@ R"(
...
+        ecdh_curves:
+        - P-256
+        - X25519
+        # Non-default curves added that are used by Chrome
+        - CECPQ2
```
davidben, line 105: This is an experimental code point and not one you should be enabling just yet. We ourselves are still in the process of rolling it in and out.
https://www.imperialviolet.org/2018/12/12/cecpq2.html

library/common/config/config.cc
```diff
@@ -89,6 +89,15 @@ const std::string config_header = R"(
   typed_config:
     "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext
     common_tls_context:
+      tls_params:
+        tls_maximum_protocol_version: TLSv1_3
+        cipher_suites: ["ALL:!aPSK:!ECDSA+SHA1:!3DES"]
+        ecdh_curves:
+        - P-256
+        - X25519
```
davidben, line 103: X25519 should be ordered ahead of P-256.

davidben:
> cc @davidben in case you have a sec to take a look. We are trying to match up the ciphers per our offline discussion.

If the aim is to just align with what Chrome currently does, other than the comments I left, this all seems reasonable. If your goal is to persistently track Chrome, we will continue to adjust this in the future, as legacy cipher suites fall off, or we introduce new modes.

RyanTheOptimist:
Is this still a PR we'd like to land?

PR #1737, opened 2021-08-20
config: advertise h2 alpn string when forcing h2

goaway:
Description: The Envoy Mobile h2 clusters force h2 to be the underlying protocol regardless, but this change has them advertise h2 via ALPN to be closer aligned with the standard.
Risk Level: Moderate
Testing: Local & CI

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -84,17 +84,30 @@ const std::string config_header = R"(
         address: 127.0.0.1
         port_value: 9901
 
-!ignore tls_socket_defs: &base_tls_socket
+!ignore tls_root_ca_defs: &tls_root_certs |
+)"
+#include "certificates.inc"
+R"(
+
+!ignore tls_socket_defs:
+- &base_tls_socket
   name: envoy.transport_sockets.tls
   typed_config:
     "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext
     common_tls_context:
       validation_context:
         trusted_ca:
-          inline_string: |
-)"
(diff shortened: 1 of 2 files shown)


library/common/config/config.cc
```diff
@@ -84,17 +84,30 @@ const std::string config_header = R"(
...
-)"
-#include "certificates.inc"
-;
+          inline_string: *tls_root_certs
+- &base_tls_h2_socket
+  name: envoy.transport_sockets.tls
+  typed_config:
+    "@type": type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext
+    common_tls_context:
+      alpn_protocols: [h2]
```
mattklein123, line 106: I think this needs to be a string.

goaway: Repeated string in proto doesn't map to a set in YAML?

goaway: According to the docs the field is a repeated string: https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/transport_sockets/tls/v3/tls.proto.html?highlight=alpn_protocols

And whatever I can find on proto to yaml mappings seems to indicate repeated fields are mapped to sequences, e.g.: https://istio-releases.github.io/v0.1/docs/reference/writing-config.html

buildbreaker (approved):
if it doesn't need to be a string

library/common/config/config.cc
mattklein123: OK I thought it needed to be a string but maybe not. If it works SGTM.

PR #1745, opened 2021-08-24
CI: Move asan workflow to Engflow's remote execution

lfpino:
Description:
CI: Move asan workflow to Engflow's remote execution and cut build times by 5x

Before (~2hr):
Screenshot from 2021-08-27 18-03-41

After (~20min):
Screenshot from 2021-08-31 00-38-38

Due to https://github.com/google/sanitizers/issues/916 we add an extra capability (SYS_PTRACE) to the remote execution docker container, we do this by adding a new Clang ASAN toolchain. Otherwise, dangling threads can fail the test during teardown (example with a previous run: https://github.com/envoyproxy/envoy-mobile/runs/3443649963). 

Risk Level: Low
Testing: See asan workflow
Docs Changes: N/A
Release Notes: N/A


goaway (approved):
Thanks @lfpino!

goaway:
@lfpino I just noticed that leak sanitization has been  turned off. As much as I _really_ would like to merge this, we've relied on leak sanitization to catch issues in the past, and they have the potential to be quite disastrous on mobile devices. Is it possible to update our tests so that we can leave this enabled?

lfpino:
I understand and thanks for taking a look. Let me see if I can find an alternative and get back to you.

lfpino:
I found a different workaround that doesn't require deactivating leak sanitization. Please take a look!

lfpino:
Fully cached asan builds are down to 3 minutes!

goaway (approved):
Thank you @lfpino!

PR #1746, opened 2021-08-24
emit events based on ENVOY_LOG_EVENT 

snowp:
Wires up the EM custom logger such that we intercept logs with stable names emitted by
ENVOY_LOG_EVENT, translating them into an event passed to the event tracker.

To maintain the original behavior when a logger is not installed, we end up having to partially
reimplement the Stderr logger.

For integration testing, we add a new test filter that emits an event, verifying that the engine
properly emits this both with and without a logger added.

Risk Level: Medium
Testing: Integration test
Docs Changes: n/a
Release Notes: n/a


Augustyniak (dismissed):
let's make sure that we follow up with iOS integration tests once we start logging some events over there (in iOS codepath)

snowp:
I can actually do a swift test since I added the test filter, so let me just add that

PR #1752, opened 2021-08-25
header handling: delete splitting comma-separated header values and add specific logic to the RetryPolicy classes

junr03:
Description: reverts earlier logic that splits comma-separated headers introduced in #1212 #1213. Upon further inspection of the Envoy code we now understand that the Envoy behavior to coalesce headers with multiple appearances in the header map is limited to Envoy's inline headers. Thus this PR reverts the Envoy Mobile logic to split up headers which should have no semantic impact in headers per RFC. However, we do add logic to the Envoy Mobile RetryPolicy classes which need individual values in order to correctly map to the swift/kotlin enum cases specific to those classes.
Risk Level: med - changes header handling logic
Testing: added unit tests


goaway (approved):
Thanks for getting to the bottom of this @junr03!

PR #1757, opened 2021-08-27
CI: Move coverage workflow to Engflow's remote execution

lfpino:

Description:
CI: Move coverage workflow to Engflow's remote execution and cut build times by 6x.

A few considerations:
- It's critical to set the right environment variables for coverage builds to work (this is how Bazel controls how coverage is run), in particular we need to set CC, CXX, GCOV and BAZEL_LLVM_COV to point to the right tools and they need to exist in the docker container that's running the builds.
- BAZEL_USE_LLVM_NATIVE_COVERAGE is required for bazel to run LLVM coverage
- These environment variables affect the rbe_autoconfig toolchain generation (i.e. the generated packages will point to different places depending of those env variables), so we need a new toolchain just for coverage.

Some shortcomings / gotchas:
- Coverage builds depend on config=coverage from upstream Envoy, as such, the coverage actions run locally (probably because remote coverage builds are buggy), because of this we need test_strategy=standalone to run the tests locally and produce the right coverage artifacts.
- After my experimentation, coverage builds seem to be incompatible with remote_download_outputs=toplevel (or minimal) so we need to set it to remote_download_outputs=all. Otherwise, some coverage artifacts aren't present locally for the coverage report generation. This is a bug in Bazel and I'll file a bug upstream for that.
- Similar experimentation shows that remote caching is also incompatible with remote coverage. For some reason, caching makes prevents some of the coverage artifacts to be present locally and the report generation consistently fails. This is another bug from Bazel (I'll file another issue for this).

Last, I've added a remote-ci-debug config to print various different streams of data from different Bazel stages

Before (~1h30min):
Screenshot from 2021-09-20 22-48-51

After (~15min):
Screenshot from 2021-09-20 22-46-30

Result:
Screenshot from 2021-09-20 22-41-28

Risk Level: Low
Testing: See coverage workflow
Docs Changes: N/A
Release Notes: N/A


.bazelrc
```diff
@@ -175,11 +175,57 @@ build:remote-ci-linux-local-java8 --host_java_toolchain=@bazel_tools//tools/jdk:
...
+build:remote-ci-linux-coverage --extra_execution_platforms=@engflow_remote_config_clang_coverage//config:platform
+build:remote-ci-linux-coverage --extra_toolchains=@engflow_remote_config_clang_coverage//config:cc-toolchain
+build:remote-ci-linux-coverage --host_platform=@engflow_remote_config_clang_coverage//config:platform
+build:remote-ci-linux-coverage --platforms=@engflow_remote_config_clang_coverage//config:platform
+build:remote-ci-linux-coverage --extra_toolchains=@local_jdk//:all
+build:remote-ci-linux-coverage --host_java_toolchain=@bazel_tools//tools/jdk:toolchain_java11
+build:remote-ci-linux-coverage --java_toolchain=@bazel_tools//tools/jdk:toolchain_java11
+# Flags to run tests locally which are necessary since Bazel C++ LLVM coverage isn't fully supported for remote builds
+# TODO(lfpino): Reference upstream Bazel issue here on the incompatibility of remote test execution and LLVM coverage.
+build:remote-ci-linux-coverage --test_strategy=standalone
```
ulfjack, line 204: I think this flag doesn't do anything in Bazel. The `--strategy=TestRunner=local,remote` makes these actions run locally, not this flag.

.bazelrc
```diff
@@ -175,11 +175,57 @@ build:remote-ci-linux-local-java8 --host_java_toolchain=@bazel_tools//tools/jdk:
...
+build:remote-ci-linux-coverage --remote_download_outputs=all
+build:remote-ci-linux-coverage --strategy=TestRunner=local,remote
+build:remote-ci-linux-coverage --strategy=CoverageReport=local,remote
+# Bazel remote caching is incompatible with C++ LLVM coverage so we need to deactivate it for coverage builds
+# TODO(lfpino): Reference upstream Bazel issue here on the incompatibility of remote caching and LLVM coverage.
```
ulfjack, line 209: Can you repro with a small example?

.github/workflows/coverage.yml
```diff
@@ -11,31 +11,50 @@ jobs:
...
         run: |
-          if git rev-parse --abbrev-ref HEAD | grep -q ^main$ || git diff --name-only origin/main | grep -q common/ ; then
+          if git rev-parse --abbrev-ref HEAD | grep -q ^main$ || git diff --name-only origin/main | grep -qe common/ -e ^.github/workflows/coverage.yml ; then
             echo "Coverage will run."
             echo "::set-output name=run_coverage::true"
           else
             echo "Skipping coverage."
             echo "::set-output name=run_coverage::false"
           fi
+      - name: 'Java setup'
```
ulfjack, line 33: Why is this needed? Maybe add a comment?

WORKSPACE
```diff
@@ -103,3 +103,22 @@ rbe_autoconfig(
...
     create_java_configs = False,
 )
+
+rbe_autoconfig(
+    name = "engflow_remote_config_clang_coverage",
+    digest = "sha256:239015d203837f2bdcd4cfdd710d7db60acc3fb1f002f1c926b92b42c59afdd6",
+    registry = "docker.io",
+    repository = "envoyproxy/envoy-build-ubuntu",
+    use_legacy_platform_definition = False,
+    env = {
```
ulfjack, line 113: I find it surprising that you have the same env variables here and above. Are both required?

ulfjack (changes requested):

.bazelrc
lfpino: I think this flag predates my addition of `--strategy=TestRunner=local,remote`, I'll try without it

.bazelrc
lfpino: That was my plan when filing the upstream issue, haven't had time to do it just yet.

.github/workflows/coverage.yml
```diff
@@ -35,7 +33,12 @@ jobs:
       - name: 'Run coverage'
         if: steps.check_context.outputs.run_coverage == 'true'
         continue-on-error: true
-        run: "PATH=/opt/llvm/bin:${PATH} ./envoy/test/run_envoy_bazel_coverage.sh //test/common/..."
+        run: |
+          echo "build --remote_header=\"Authorization=Bearer ${{ secrets.GITHUB_TOKEN }}\"" > ~/.bazelrc
```
ulfjack, line 37: Doesn't this overwrite the checked-in `.bazelrc`? Shouldn't it use `>>`?

lfpino: From what I could tell the home bazelrc in the GH runner is always empty and I wanted to avoid adding flags on top of existing ones if the bazelrc exists. I can change it if you think it's better but I'm ok with it as it is.

ulfjack: Ah, didn't read that correctly.

ulfjack (approved):

goaway (approved):
Thank you!

PR #1759, opened 2021-08-27
CI: Move Kotlin workflow in format to Engflow's remote execution

lfpino:

Description:
CI: Move Kotlin workflow in format to Engflow's remote execution and cut format workflow build time by a half.

Before:
Screenshot from 2021-08-31 00-21-47

After:
Screenshot from 2021-08-31 00-21-54

Risk Level: Low
Testing: See format workflow
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -60,8 +60,11 @@ jobs:
       - run: ./ci/mac_ci_setup.sh
         name: 'Install dependencies'
       - name: 'Run Kotlin Lint (Detekt)'
+        env:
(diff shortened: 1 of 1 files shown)


goaway:
Thanks @lfpino this is great. I'm wondering though, would we want this to be done on linux ultimately?

lfpino:
Since it's currently running on MacOS I'm trying to not divert from the current semantics to avoid doing too many changes at the same time.

My suggestion would be to merge this and file an issue as a follow-up to move this to Linux. Sometimes it's trivial to move these flows around but I gotta test it out.

What do you think?

goaway (approved):

PR #1761, opened 2021-08-27
core: log and emit event listing available network interfaces

goaway:
Description: Only works on iOS and Android API > 23 currently.
Risk Level: Low
Testing: Local & CI

### DIFF
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -291,15 +291,15 @@ void Engine::drainConnections() {
 void Engine::logInterfaces() {
   auto v4_vec = Network::MobileUtility::enumerateV4Interfaces();
   std::string v4_names = std::accumulate(v4_vec.begin(), v4_vec.end(), std::string{},
-                                      [](std::string acc, std::string next) {
-                                        return acc.empty() ? next : std::move(acc) + "," + next;
-                                      });
+                                         [](std::string acc, std::string next) {
+                                           return acc.empty() ? next : std::move(acc) + "," + next;
+                                         });
 
   auto v6_vec = Network::MobileUtility::enumerateV6Interfaces();
   std::string v6_names = std::accumulate(v6_vec.begin(), v6_vec.end(), std::string{},
-                                      [](std::string acc, std::string next) {
-                                        return acc.empty() ? next : std::move(acc) + "," + next;
-                                      });
+                                         [](std::string acc, std::string next) {
+                                           return acc.empty() ? next : std::move(acc) + "," + next;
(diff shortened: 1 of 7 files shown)


goaway:
cc @Reflejo

library/common/engine.cc
```diff
@@ -284,4 +288,14 @@ void Engine::drainConnections() {
...
 }
 
+void Engine::logInterfaces() {
+  auto name_vec = Network::MobileUtility::enumerateInterfaces();
+  std::string names = std::accumulate(name_vec.begin(), name_vec.end(), std::string{},
+                                      [](std::string acc, std::string next) {
+                                        return acc.empty() ? next : std::move(acc) + "," + next;
+                                      });
+  ENVOY_LOG(debug, "available interfaces: {}", names);
+  ENVOY_LOG_EVENT(debug, "socket_selection_available_interfaces", names);
```
junr03, line 298: just need ENVOY_LOG_EVENT, it "dual writes" to the event tracker and the process logs

library/common/network/mobile_utility.cc
```diff
@@ -0,0 +1,42 @@
...
+#include "source/common/common/assert.h"
+
+#ifdef SUPPORTS_GETIFADDRS
+#include <ifaddrs.h>
+#endif
+
+namespace Envoy {
+namespace Network {
+
+std::vector<std::string> MobileUtility::enumerateInterfaces() {
```
junr03, line 14: can we write a quick test?

goaway: This isn't really very testable since it's mapping system output. Here's the extent of Envoy testing on this functionality: https://github.com/envoyproxy/envoy/blob/main/test/common/network/utility_test.cc#L228

Since this function doesn't return a pointer, there's not even a non-null assertion to be made here.

library/common/engine.cc
goaway: Yep, I do see that in the log output now, though it would be nice if the raw log outputted the name as well, I think.

library/common/network/mobile_utility.cc
junr03: If Envoy is expecting non null. Can we expect and non empty vector?

goaway: Discussed offline, but an empty vector is potentially a valid result.

Reflejo (dismissed):

buildbreaker (approved):

buildbreaker (dismissed):

PR #1769, opened 2021-08-31
ci: re-enable disabled event tracker tests

Augustyniak:
Description: Enable tests that were disabled as part of https://github.com/envoyproxy/envoy-mobile/pull/1761.
Risk Level: None
Testing: Unit


Augustyniak:
@goaway kotlin tests are red - I cannot restart them but my PR changes tests only so there is no way that it breaks Kotlin tests I think

PR #1770, opened 2021-08-31
engine: add H2 ping config API

junr03:
Description: adds engine builder and config plumbing to setup h2 ping support
Risk Level: low - exposed API. Default config is no-op in Envoy
Testing: added tests
Docs Changes: pending if reviewers agree with API

### DIFF
--- a/library/cc/engine_builder.cc
+++ b/library/cc/engine_builder.cc
@@ -109,7 +109,7 @@ std::string EngineBuilder::generateConfigStr() {
       {"h2_connection_keepalive_interval",
        fmt::format("{}s", this->h2_connection_keepalive_interval_seconds_)},
       {"h2_connection_keepalive_idle_interval",
-       fmt::format("{}s", this->h2_connection_keepalive_idle_interval_milliseconds_)},
+       fmt::format("{}s", this->h2_connection_keepalive_idle_interval_milliseconds_ / 1000.0)},
       {"h2_connection_keepalive_timeout",
        fmt::format("{}s", this->h2_connection_keepalive_timeout_seconds_)},
       {
--- a/library/cc/engine_builder.cc
+++ b/library/cc/engine_builder.cc
@@ -54,17 +54,21 @@ EngineBuilder::addDnsPreresolveHostnames(const std::string& dns_preresolve_hostn
   return *this;
 }
 
-EngineBuilder& EngineBuilder::addH2ConnectionKeepaliveIntervalSeconds(int h2_connection_keepalive_interval_seconds) {
+EngineBuilder& EngineBuilder::addH2ConnectionKeepaliveIntervalSeconds(
+    int h2_connection_keepalive_interval_seconds) {
   this->h2_connection_keepalive_interval_seconds_ = h2_connection_keepalive_interval_seconds;
   return *this;
 }
 
-EngineBuilder& EngineBuilder::addH2ConnectionKeepaliveIdleIntervalMilliseconds(int h2_connection_keepalive_idle_interval_milliseconds) {
-  this->h2_connection_keepalive_idle_interval_milliseconds_ = h2_connection_keepalive_idle_interval_milliseconds;
+EngineBuilder& EngineBuilder::addH2ConnectionKeepaliveIdleIntervalMilliseconds(
+    int h2_connection_keepalive_idle_interval_milliseconds) {
+  this->h2_connection_keepalive_idle_interval_milliseconds_ =
+      h2_connection_keepalive_idle_interval_milliseconds;
   return *this;
 }
 
-EngineBuilder& EngineBuilder::addH2ConnectionKeepaliveTimeoutSeconds(int h2_connection_keepalive_timeout_seconds) {
+EngineBuilder&
+EngineBuilder::addH2ConnectionKeepaliveTimeoutSeconds(int h2_connection_keepalive_timeout_seconds) {
   this->h2_connection_keepalive_timeout_seconds_ = h2_connection_keepalive_timeout_seconds;
   return *this;
 }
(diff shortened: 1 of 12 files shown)


library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -118,6 +131,12 @@ String resolveTemplate(final String templateYAML, final String platformFilterTem
         .append(String.format("- &dns_fail_max_interval %ss\n", dnsFailureRefreshSecondsMax))
         .append(String.format("- &dns_query_timeout %ss\n", dnsQueryTimeoutSeconds))
         .append(String.format("- &dns_preresolve_hostnames %s\n", dnsPreresolveHostnames))
+        .append(String.format("- &h2_connection_keepalive_interval %ss\n",
+                              h2ConnectionKeepaliveIntervalSeconds))
+        .append(String.format("- &h2_connection_keepalive_idle_interval %sms\n",
```
junr03, line 136: I dont think `ms` works automagically. Will update tomorrow.

goaway: Only `s` works with the proto parser.

library/cc/engine_builder.cc
```diff
@@ -54,6 +54,25 @@ EngineBuilder::addDnsPreresolveHostnames(const std::string& dns_preresolve_hostn
...
+  this->h2_connection_keepalive_idle_interval_milliseconds_ =
+      h2_connection_keepalive_idle_interval_milliseconds;
+  return *this;
+}
+
+EngineBuilder&
+EngineBuilder::addH2ConnectionKeepaliveTimeoutSeconds(int h2_connection_keepalive_timeout_seconds) {
+  this->h2_connection_keepalive_timeout_seconds_ = h2_connection_keepalive_timeout_seconds;
+  return *this;
+}
```
goaway, line 74: I'm actually not certain we want to expose these values at this layer. If this ends up being our primary mechanism to detect zombie connections, overriding this configuration seems like a pretty major footgun for a naive user of the library.

library/common/config/config.cc
```diff
@@ -39,6 +39,9 @@ const std::string config_header = R"(
 - &dns_fail_max_interval 10s
 - &dns_query_timeout 25s
 - &dns_preresolve_hostnames []
+- &h2_connection_keepalive_interval 0s
+- &h2_connection_keepalive_idle_interval 0s
+- &h2_connection_keepalive_timeout 5s
```
goaway, line 44: It looks like these still need to be referenced in the cluster config.

junr03: Yep, I didn't want to finish wiring until we settled on interface.

library/cc/engine_builder.cc
junr03: @goaway I wasn't aware that was your preference. I described that I was exposing all the way at the Engine API. What would your preference be? If we dont expose them at the engine builder layer then we wouldn't be able to experiment with settings internally at Lyft.

junr03: going to set up default API in a subsequent PR. per offline discussion.

library/common/config/config.cc
```diff
@@ -39,6 +39,8 @@ const std::string config_header = R"(
 - &dns_fail_max_interval 10s
 - &dns_query_timeout 25s
 - &dns_preresolve_hostnames []
+- &h2_connection_keepalive_idle_interval 100000s
```
goaway, line 42: Probably should be 0 here just so it's disabled in vanilla config.

library/cc/engine_builder.h
```diff
@@ -53,6 +57,8 @@ class EngineBuilder {
   int dns_failure_refresh_seconds_max_ = 10;
   int dns_query_timeout_seconds_ = 25;
   std::string dns_preresolve_hostnames_ = "[]";
+  int h2_connection_keepalive_idle_interval_milliseconds_ = 0;
+  int h2_connection_keepalive_timeout_seconds_ = 5;
```
goaway, line 61: I'd suggest 10 as a starting point.

library/common/config/config.cc
junr03: unfortunately the proto constraint to at least 1ms.

library/common/config/config.cc
```diff
@@ -39,6 +39,8 @@ const std::string config_header = R"(
...
+- &h2_connection_keepalive_timeout 5s
```
goaway, line 43: 10

library/cc/engine_builder.h
junr03: oh right, will update.

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -35,6 +35,8 @@ open class EngineBuilder(
   private var dnsFailureRefreshSecondsMax = 10
   private var dnsQueryTimeoutSeconds = 25
   private var dnsPreresolveHostnames = "[]"
+  private var h2ConnectionKeepaliveIdleIntervalMilliseconds = 0
+  private var h2ConnectionKeepaliveTimeoutSeconds = 5
```
goaway, line 39: 10

library/swift/EngineBuilder.swift
```diff
@@ -21,6 +21,8 @@ open class EngineBuilder: NSObject {
   private var dnsFailureRefreshSecondsMax: UInt32 = 10
   private var dnsQueryTimeoutSeconds: UInt32 = 25
   private var dnsPreresolveHostnames: String = "[]"
+  private var h2ConnectionKeepaliveIdleIntervalMilliseconds: UInt32 = 0
+  private var h2ConnectionKeepaliveTimeoutSeconds: UInt32 = 5
```
goaway, line 25: 10

goaway:
Other than some nits about the defaults, I think this looks good. An integration test would be amazing, but that doesn't *need* to be part of this PR.

goaway:
Also, suggest merging main, to make sure you've picked up the new RBE configs for CI.

goaway (approved):

goaway:
Thanks @junr03!

PR #1771, opened 2021-08-31
core: emit envoy event every time envoy bug macro is called

Augustyniak:
Description: Emit Envoy event every time an Envoy bug is detected. Envoy bug is similar to Envoy assert.
Risk Level: Low, new functionality
Testing: Unit


buildbreaker (approved):
thanks @Augustyniak

PR #1772, opened 2021-08-31
android: conditionally build internal getifaddrs support

goaway:
Description: The range of Android versions supported by Envoy Mobile don't all have native getifaddrs(3) support. In order to ship a single distributable package that can support all versions, this change conditionally bundles the custom Android Gingerbread implementation, which is (mostly) contained in a single header.
Risk Level: Moderate
Testing: Local & CI

### DIFF
--- a/library/common/network/BUILD
+++ b/library/common/network/BUILD
@@ -6,9 +6,21 @@ envoy_package()
 
 envoy_cc_library(
     name = "mobile_utility_lib",
-    srcs = ["mobile_utility.cc"],
+    srcs = [
+        "mobile_utility.cc",
+    ] + select({
+        "//bazel:include_ifaddrs": [
+            "//third_party:android/ifaddrs-android.h",
+            "//third_party:android/LocalArray.h",
+            "//third_party:android/ScopedFd.h",
+        ],
+        "//conditions:default": [],
+    }),
+    copts = select({
+        "//bazel:include_ifaddrs": ["-DINCLUDE_IFADDRS"],
+        "//conditions:default": [],
+    }),
     hdrs = ["mobile_utility.h"],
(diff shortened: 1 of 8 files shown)


goaway:
@keith would appreciate you taking a look to help verify I'm not doing anything crazy here.

library/common/network/BUILD
```diff
@@ -6,9 +6,21 @@ envoy_package()
...
+    ] + select({
+        "//bazel:include_ifaddrs": [
+            "//third_party:android/ifaddrs-android.h",
+            "//third_party:android/LocalArray.h",
+            "//third_party:android/ScopedFd.h",
+        ],
+        "//conditions:default": [],
+    }),
     hdrs = ["mobile_utility.h"],
-    copts = ["-Ithird_party/android"],
```
keith, line 11: don't still need this one?

third_party/BUILD
```diff
@@ -0,0 +1,7 @@
+licenses(["notice"])  # Apache 2
+
+exports_files([
```
keith, line 3: potentially nicer to expose this as a filegroup and depend on that, just so you get real visibility and you just have to add new headers in 1 place

keith (approved):

third_party/BUILD
goaway: Good thought. I'll try it out locally, though may delay on making the change, since we've made it through the CI gauntlet here.

goaway:
Thanks @keith!

library/common/network/BUILD
goaway: Yeah, it seems to only really be needed when your include paths are something funky that you can't change.

PR #1773, opened 2021-08-31
reinstating propercasing

alyssawilk:
Turning up propercasing for HTTP/1.1, now avoiding the SAN issues.

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
Fixes https://github.com/envoyproxy/envoy-mobile/issues/1558


buildbreaker (approved):
Thanks @alyssawilk ! I did a quick manual regression test with this config for hitting our CDN and it passed.

PR #1777, opened 2021-09-01
support configuring component log levels

snowp:
Adds support for configuring component log levels via the EngineBuilder interfaces, allowing for more granular control over
what logs are emitted.

Risk Level: Low, new functionality
Testing: Android and iOS integration tests
Docs Changes: inline
Release Notes: n/a
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/library/common/engine.h
+++ b/library/common/engine.h
@@ -133,6 +133,7 @@ private:
   envoy_logger logger_;
   envoy_event_tracker event_tracker_;
   Assert::ActionRegistrationPtr assert_handler_registration_;
+  Assert::ActionRegistrationPtr bug_handler_registration_;
   Thread::MutexBasicLockable mutex_;
   Thread::CondVar cv_;
   Http::ClientPtr http_client_;
--- a/library/common/engine.h
+++ b/library/common/engine.h
@@ -31,12 +31,18 @@ public:
    */
   ~Engine();
 
+  struct EngineRunOptions {
+    std::string config;
+    std::string log_level;
+    std::string component_log_level;
+  };
+
   /**
    * Run the engine with the provided configuration.
    * @param config, the Envoy bootstrap configuration to use.
    * @param log_level, the log level.
    */
-  envoy_status_t run(std::string config, std::string log_level);
+  envoy_status_t run(const EngineRunOptions& options);
 
   /**
    * Immediately terminate the engine, if running.
@@ -123,7 +129,7 @@ public:
   void drainConnections();
 
 private:
-  envoy_status_t main(std::string config, std::string log_level);
+  envoy_status_t main(const EngineRunOptions& run_options);
   void logInterfaces();
 
   Event::Dispatcher* event_dispatcher_{};
--- a/library/common/http/client.h
+++ b/library/common/http/client.h
@@ -228,7 +228,9 @@ private:
     void readDisable(bool disable) override;
     uint32_t bufferLimit() override { return 65000; }
     // Not applicable
-    void setAccount(Buffer::BufferMemoryAccountSharedPtr) override {}
+    void setAccount(Buffer::BufferMemoryAccountSharedPtr) override {
+      // Acounting became default in https://github.com/envoyproxy/envoy/pull/17702 but is a no=op.
+    }
     void setFlushTimeout(std::chrono::milliseconds) override {}
 
     // ScopeTrackedObject
--- a/library/common/http/client.h
+++ b/library/common/http/client.h
@@ -228,9 +228,7 @@ private:
     void readDisable(bool disable) override;
     uint32_t bufferLimit() override { return 65000; }
     // Not applicable
-    void setAccount(Buffer::BufferMemoryAccountSharedPtr) override {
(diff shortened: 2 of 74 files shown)


library/common/http/client.h
```diff
@@ -228,9 +228,7 @@ class Client : public Logger::Loggable<Logger::Id::http> {
     void readDisable(bool disable) override;
     uint32_t bufferLimit() override { return 65000; }
     // Not applicable
-    void setAccount(Buffer::BufferMemoryAccountSharedPtr) override {
-      PANIC("buffer accounts unsupported");
```
jingwei99, line 232: what's the reason/thoughts behind this?

(this is nothing critical, but just want to know)

library/common/engine.h
```diff
@@ -31,12 +31,18 @@ class Engine : public Logger::Loggable<Logger::Id::main> {
    */
   ~Engine();
 
+  struct EngineRunOptions {
```
jingwei99, line 34: 👍

library/common/http/client.h
jingwei99: 👍

test/common/engine_test.cc
```diff
@@ -79,12 +80,58 @@ TEST_F(EngineTest, EarlyExit) {
...
 
   engine_.reset();
   ASSERT_TRUE(test_context.on_exit.WaitForNotificationWithTimeout(absl::Seconds(3)));
 
   start_stream(0, {}, false);
 }
+
+TEST_F(EngineTest, ComponentLogLevel) {
+  const std::string level = "debug";
+  const std::string component_log_level = "assert:info,misc:trace";
```
jingwei99, line 94: the intent of this test as I assume is to verify engine will be running ok when the `component_log_level` is set correctly (or in the format that's expected), is this assumption right?

jingwei99 (dismissed):
looks good to me

test/common/engine_test.cc
snowp: I wrote this originally to verify that plumbing through the component log level without crashing (so yes your assumption is correct), though this test might be unnecessary now that we there are integration tests for Kotlin and iOS. The other test in here which validates that we don't crash when we get an invalid string is more valid.

jingwei99: 👌

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -26,6 +26,7 @@ open class EngineBuilder(
     EnvoyEngineImpl(onEngineRunning, logger, eventTracker)
   }
   private var logLevel = LogLevel.INFO
+  private var componentLogLevel = ""
```
buildbreaker, line 29: What are the options for component log level? If it's a finite set of strings, would making an enum configuration make more sense?

snowp: It's a string of the format `<component>:<level>,<comopnent>:<level>`, and the components needs to be statically registered in Envoy. I started this with a map, but was deterred by the fact that passing a map through the C layer is painful, and decomposing the map into the string in the platform layer felt repetitive.

Open to suggestions though :)

rebello95: The current form does feel a little magical. If we don't want to do a map of `component (string): LogLevel (enum)`, let's at least document the expected structure inline and in the documentation for the builder

buildbreaker: I'm fine with the proposed changes but I would like to understand where these values are registered in Envoy. 

We can fixate those values on the platform layer which would make the validation in the common layer more sane while providing a clear set of values for the API.

snowp: The components are statically registered here: https://github.com/envoyproxy/envoy/blob/main/source/common/common/logger.h#L29-L79

If we wanted to we could add a C API that let us query https://github.com/envoyproxy/envoy/blob/main/source/common/common/logger.h#L312 in order to get the list of registered log names. Another option would be to just capture a static list of valid log components that we manually keep in sync.

LMK which option ya'll prefer between the above and just documenting the expected log component string, happy to do whichever

snowp: @buildbreaker ping on this, lmk how you want to proceed

library/common/engine.h
goaway: I like this conceptually, but I do think perhaps config should be separate from it. There's probably more that can go in here down the road.

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
goaway: I think posting the above link to the components and having an example string in the public API comments/docstrings would be helpful.

PR #1781, opened 2021-09-02
CI: Move tsan workflow to Engflow's remote execution

lfpino:
Description:
CI: Move tsan workflow to Engflow's remote execution

Same as https://github.com/envoyproxy/envoy-mobile/pull/1745 but using a couple of tsan flags instead of asan. Using a separate config for modularity.

Risk Level: Low
Testing: See tsan workflow
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -12,11 +12,11 @@ build --swiftcopt=-Xfrontend --swiftcopt=-no-serialize-debugging-options
 build --host_force_python=PY3
 build --ios_minimum_os=11.0
 build --ios_simulator_device="iPhone 12"
-build --ios_simulator_version=14.4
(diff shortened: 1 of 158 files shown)


lfpino:
@goaway This is identical to ASAN but with a few different flags. Please take a look

I also cleaned up some redundant flags and once we finish more flows we can see how to unify better the configs.

goaway (commented):
Thanks @lfpino! LGTM, pending CI.

goaway (approved):
And since tsan already passed, looks like this will be good to go when it completes.

lfpino:
I think you can merge now.

goaway (approved):

PR #1782, opened 2021-09-03
stream intel: change sentinel values for conn and stream id to -1

junr03:
Description: while looking at sessions it became clear that a 0 sentinel value can be confused with C0. Setting to max int to disambiguate. 
Risk Level: low


goaway:
Max int is a pretty esoteric sentinel. If we need this, here's some options I think would be preferable:

1) Add 1 to all connection/stream IDs making 0 the sentinel. As long as stream intel is the source of truth and Envoy's value remains internal, this should be fine. The only risk is if the IDs somehow get surfaced via some alternative mechanism, it could be confusing. Trivial to implement.
2) Use a separate mask to indicate which values are set, basically making up to 64 values optional, and avoiding sentinels altogether.
3) Make all these values signed and use -1 as the sentinel. Not great, but Java's internal type is signed anyways, so it's at least aligned there. Still, I like this option least.

junr03:
> The only risk is if the IDs somehow get surfaced via some alternative mechanism, it could be confusing

I want to minimize that possibility.

> Use a separate mask to indicate which values are set, basically making up to 64 values optional, and avoiding sentinels altogether.

You mean adding a bitmask to stream intel? I like that but keeping that up to date at all the layers seems cumbersome? maybe not too much.

> -1

simple? not confusing? not too error prone? I updated with this but lmk if you feel strongly.

library/common/http/client.cc
```diff
@@ -245,10 +245,10 @@ void Client::DirectStreamCallbacks::onCancel() {
 
 envoy_stream_intel Client::DirectStreamCallbacks::streamIntel() {
   const auto& info = direct_stream_.request_decoder_->streamInfo();
-  envoy_stream_intel stream_intel{};
-  stream_intel.connection_id = info.upstreamConnectionId().value_or(0);
+  envoy_stream_intel stream_intel{-1, -1, -1};
+  stream_intel.connection_id = info.upstreamConnectionId().value_or(-1);
   stream_intel.stream_id = static_cast<uint64_t>(direct_stream_.stream_handle_);
-  stream_intel.attempt_count = info.attemptCount().value_or(0);
```
goaway, line 251: For attempt count I think 0 is still correct.

library/common/types/c_types.h
```diff
@@ -146,12 +146,12 @@ typedef struct {
...
-  uint64_t stream_id;
-  // An internal identifier for the connection carrying the stream.
-  uint64_t connection_id;
-  // The number of internal attempts to carry out a request/operation.
-  uint64_t attempt_count;
+  // An internal identifier for the stream. -1 if not preset.
+  int64_t stream_id;
+  // An internal identifier for the connection carrying the stream. -1 if not present.
+  int64_t connection_id;
+  // The number of internal attempts to carry out a request/operation. -1 if not present.
```
goaway, line 153: Can't think of any case where 0 would be incorrect though.

library/swift/StreamIntel.swift
```diff
@@ -4,14 +4,14 @@ import Foundation
...
 @objcMembers
 public final class StreamIntel: NSObject, Error {
-  // An internal identifier for the stream.
-  public let streamId: UInt64
-  // An internal identifier for the connection carrying the stream.
-  public let connectionId: UInt64
-  // The number of internal attempts to carry out a request/operation.
-  public let attemptCount: UInt64
+  // An internal identifier for the stream. -1 if not set.
+  public let streamId: Int64
```
goaway, line 8: Here, I would make connectionId and streamId optional (so we don't expose our sentinel - which is why I still slightly prefer the bitmask since it would purely be a core/bridge mechanism).

goaway:
This is fine I suppose, though I still slightly prefer the bitmask per my comment above. I would make attemptCount non-optional and 0 as default, because "no attempts" is still zero.

goaway (changes requested):
See above comments.

library/swift/StreamIntel.swift
junr03: The sentinel is already exposed, right? Is your suggestion to now hide it by changing this class? I wonder if that would require changes in the consuming code in our private analytics filter?

junr03:
@goaway updated

junr03:
@goaway any more comments here?

buildbreaker (approved):
Let's move forward. I believe these values are great to have since we can differentiate between an empty value versus not.

We can address issues moving forward :)

goaway:
Agreed, let's move forward, this is a minor detail to get hung up on.

PR #1786, opened 2021-09-07
dns: effectively disable host ttl

junr03:
Description: On mobile, backgrounding might cause the host to be past its TTL without good reason. Given the host would be deleted, and new streams for a given domain would have to wait for resolution, it is better to not delete existing hosts; especially since deletion only happens when re-resolving is already in progress. There is no way to disable so the value below is equivalent to 24 hours.
Risk Level: low
Testing: CI


mattklein123 (approved):
Thanks

PR #1789, opened 2021-09-07
outlier detection: expose consecutive 5xx API

junr03:


goaway:
Does setting this to 0 disable outlier detection?

junr03:
> Does setting this to 0 disable outlier detection?

@goaway it does not.

junr03:
closing for now as we might need to change the outlier detector mode used in Envoy Mobile.

PR #1797, opened 2021-09-13
network: create SocketSelectionFilter 

goaway:
TODO:
- [x] create selection filter
- [x] define (different) socket option for Android and iOS
- [x] central network preference handling
- [x] define dummy option to create non-default cluster
- [x] select default/non-default cluster based on reachability status
- [x] remove network clusters

Description: Removes network-specific clusters and replaces with a filter that sets unique socket options based on current preferred network (via Reachability) status. Envoy internally creates a different socket pool for a unique set of socket options, so the effect of this is that we still use different sets of connections per interface, but no longer rely on a large set of clusters to do so. This PR also lays the groundwork for us to experiment with interface-bound sockets.
Risk Level: High
Testing: Existing unit and integration coverage

### DIFF
--- a/library/common/extensions/filters/http/socket_selection/filter.cc
+++ b/library/common/extensions/filters/http/socket_selection/filter.cc
@@ -1,0 +1,39 @@
+#include "library/common/extensions/filters/http/socket_selection/filter.h"
+
+#include "envoy/server/filter_config.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace SocketSelection {
+
+Http::FilterHeadersStatus SocketSelectionFilter::decodeHeaders(Http::RequestHeaderMap&, bool) {
+  ASSERT(decoder_callbacks_);
+  ENVOY_LOG(debug, "SocketSelectionFilter::decodeHeaders");
+  //auto& info = decoder_callbacks_->streamInfo();
+
+  return Http::FilterHeadersStatus::Continue;
+}
+
+Http::LocalErrorStatus SocketSelectionFilter::onLocalReply(const LocalReplyData& reply) {
+  ENVOY_LOG(debug, "SocketSelectionFilter::onLocalReply({}, {})", reply.code_, reply.details_);
+  ASSERT(decoder_callbacks_);
+  auto& info = decoder_callbacks_->streamInfo();
+  // TODO(goaway): set responseCode in upstream Envoy when responseCodDetails are set.
+  // ASSERT(static_cast<uint32_t>(reply.code_) == info.responseCode());
+  // TODO(goaway): follow up on the underscore discrepancy between these values.
+  // ASSERT(reply.details_ == info.responseCodeDetails().value());
+  info.setResponseCode(static_cast<uint32_t>(reply.code_));
+
+  // Charge failures to the upstream cluster to allow for passive healthchecking.
+  if (auto upstream = info.upstreamHost()) {
+    ENVOY_LOG(trace, "SocketSelectionFilter::onLocalReply charging failure to upstream host");
+    upstream->outlierDetector().putHttpResponseCode(static_cast<uint64_t>(reply.code_));
+  }
+  return Http::LocalErrorStatus::ContinueAndResetStream;
+}
+
+} // namespace SocketSelection
+} // namespace HttpFilters
+} // namespace Extensions
+} // namespace Envoy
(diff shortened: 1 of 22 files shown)


library/common/extensions/filters/http/socket_selection/filter.cc
```diff
@@ -0,0 +1,33 @@
...
+
+#include "library/common/network/mobile_utility.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace SocketSelection {
+
+Http::FilterHeadersStatus SocketSelectionFilter::decodeHeaders(Http::RequestHeaderMap&, bool) {
+  ASSERT(decoder_callbacks_);
```
mattklein123, line 13: nit: not sure this assert adds much value, same below.

library/common/extensions/filters/http/socket_selection/filter.cc
```diff
@@ -0,0 +1,33 @@
...
+  auto connection_options = Network::MobileUtility::getUpstreamSocketOptions(network);
+  decoder_callbacks_->addUpstreamSocketOptions(connection_options);
+
+  return Http::FilterHeadersStatus::Continue;
+}
+
+Http::LocalErrorStatus SocketSelectionFilter::onLocalReply(const LocalReplyData&) {
+  ASSERT(decoder_callbacks_);
+  return Http::LocalErrorStatus::ContinueAndResetStream;
+}
```
mattklein123, line 28: Why is this implemented?

library/common/extensions/filters/http/socket_selection/filter.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace SocketSelection {
+
+/**
+ * Interface for socket selection controller. Shared by filter instances.
+ */
+class SocketSelectionController {
```
mattklein123, line 19: Unused?

library/common/http/client.cc
```diff
@@ -524,30 +525,10 @@ const LowerCaseString H2UpstreamHeader{"x-envoy-mobile-upstream-protocol"};
 // Long-term we will be working to generally provide more responsive connection handling within
 // Envoy itself.
 
-const size_t ClustersPerPool = 3;
```
mattklein123, line 527: Yay! So much simpler.

library/common/network/mobile_utility.cc
```diff
@@ -21,6 +53,43 @@ namespace {
 #define SUPPORTS_GETIFADDRS
 #endif
 
+namespace {
+
+// Internal SocketOptionImpl that permutes hash key based on name and value.
+class InternalOptionImpl : public SocketOptionImpl {
```
mattklein123, line 59: You don't need this anymore as long as you have an envoy that has my retry options predicate PR.

mattklein123 (changes requested):
Thanks generally looks great. A few small comments.

library/common/extensions/filters/http/socket_selection/filter.cc
goaway: Agreed, it's mostly a holdover from when we were less certain about the contract on when this will definitively be set. I'll go ahead and remove.

library/common/extensions/filters/http/socket_selection/filter.cc
goaway: It's here because some of accounting logic that plugs into here, but I punted that to a later PR. I can remove the override for this PR.

library/common/extensions/filters/http/socket_selection/filter.h
goaway: Ah, good catch, this is left over from an earlier approach. Will remove.

library/common/http/client.cc
goaway: :)

library/common/network/mobile_utility.cc
goaway: Thanks, will bump upstream and update.

goaway: https://github.com/envoyproxy/envoy-mobile/pull/1806

goaway:
Thank @mattklein123!

mattklein123:
LGTM pending the upstream bump and code removal.

mattklein123 (approved):
Thanks!

PR #1801, opened 2021-09-13
CI: Move remaining Linux android_test workflows to Engflow's remote execution

lfpino:
Description:
CI: Move remaining Linux android_test workflows to Engflow's remote execution and improve build times by 10x

As part of this change we need to setup the Java toolchain to use Java 8 (as specified by the java-setup GitHub action).

Also, switch from setOnResponseHeaders to setOnResponseData in //test/kotlin/integration:set_event_tracker_test which since the former was consistently failing when running remotely ([example](https://github.com/envoyproxy/envoy-mobile/runs/3604730195#step:6:232))

Before (~40min):
Screenshot from 2021-09-15 02-43-44

After (~4mins):
Screenshot from 2021-09-15 02-54-26

Risk Level: Low
Testing: See android_test
Docs Changes: N/A
Release Notes: N/A


goaway (approved):
Thanks @lfpino!

PR #1802, opened 2021-09-14
config: add protocol options to h2 clusters

junr03:

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -372,6 +372,7 @@ R"(
     upstream_connection_options: *upstream_opts
     circuit_breakers: *circuit_breakers_settings
     outlier_detection: *base_outlier_detection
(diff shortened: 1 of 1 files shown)


library/common/config/config.cc
```diff
@@ -136,6 +118,25 @@ R"(
 )";
 
 const char* config_template = R"(
+!ignore base_protocol_options_defs:
+    typed_extension_protocol_options: &base_protocol_options
```
goaway, line 122: You can actually delete this key and either make it a sequence (`-`) or move the whole thing up a level. But not a big deal. Not even a nit really, just an FYI. :)

goaway (dismissed):
Thanks @junr03! LGTM

goaway (approved):
Still good :)

PR #1808, opened 2021-09-16
Remove Kotlin dependency from Cronvoy

### LINKED ISSUE #1658: Cronvoy: remove Kotlin dependencies
There is the intention to include Cronvoy as part of the Android Framework. Unfortunately, imposing Kotlin dependencies at that level is a risky adventure that has not been done yet, and there is some strong reluctance to do so. This leads to the conclusion that for Android using Kotlin is probably a snake pit.

Cronvoy could be written to only use the pure Envoy Mobile Java libraries. Removing Kotlin from Envoy Mobile could be considered too.

carloseltuerto:
It turns out that removing Kotlin from Envoy Mobile is not really needed at this point. In reality, the Kotlin classes are a layer on top of the Java classes, and Cronvoy is a layer too: one level of indirection for the callbacks is enough. So instead Cronvoy uses the java classes directly. The main drawback is the CronvoyConfiguration - there is some duplication. 

Description: Have Cronvoy to only use the Envoy Mobile Java classes for its implementation  
Risk Level: None
Testing: CI
Docs Changes: N/A
Release Notes: N/A
Fixes #1658


goaway:
Rerunning some tests (we've had some spurious failures on RBE).

goaway:
Overall, I'm fine with this approach and I think it's great we found a lower-friction solution for the time being.

goaway (approved):
LGTM

buildbreaker (approved):
Wow! This was all we needed?

carloseltuerto:
Allan, I was surprised too by how few changes were required. And the good news it is now much easier for me to used the explicitFlowControl

PR #1816, opened 2021-09-21
network: introduce singleton configurator

goaway:
Description: Introduces a singleton Network::Configurator. This object tracks current network state and, as part of this PR, will refresh cached DNS entries on network updates. It also carries over previous functionality from the Network::MobileUtility that provides socket configuration for new connections.
Risk Level: Moderate
Testing: Existing coverage; adding unit test in followup.


goaway:
Depends on https://github.com/envoyproxy/envoy/pull/18208

library/common/network/configurator.cc
```diff
@@ -77,11 +79,13 @@ void Configurator::refreshDns(envoy_network_t network) {
...
   // Note this does NOT completely prevent parallel refreshes from being triggered in multiple
   // flip-flop scenarios.
-  if (network != preferred_network_.load() || !dns_cache_.has_value()) {
+  if (network != preferred_network_.load()) {
     return;
   }
   // TODO(goaway): track event here or are there existing signals we can use?
-  dns_cache_.value()->forceRefreshHosts();
+  if (auto dns_cache = dns_cache_manager_->lookUpCacheByName(BaseDnsCache)) {
+    dns_cache->forceRefreshHosts();
```
junr03, line 87: check for nullptr

goaway: Sorry, which pointer do you mean?

junr03: dns_cache. The returned pointer could be null

junr03: Oh wait I'm being totally dense. You have this in an if 🤦

goaway: no worries :)

library/common/engine.h
```diff
@@ -53,6 +54,12 @@ class Engine : public Logger::Loggable<Logger::Id::main> {
    */
   Http::Client& httpClient();
 
+  /**
+   * Accessor for the network configuraator. Must be called from the dispatcher's context.
```
junr03, line 58:
```suggestion
   * Accessor for the network configurator. Must be called from the dispatcher's context.
```

library/common/main_interface.cc
```diff
@@ -88,7 +88,15 @@ envoy_status_t reset_stream(envoy_stream_t stream) {
 }
 
 envoy_status_t set_preferred_network(envoy_network_t network) {
-  Envoy::Network::MobileUtility::setPreferredNetwork(network);
+  envoy_network_t previous = Envoy::Network::Configurator::setPreferredNetwork(network);
+  if (previous != network && previous != ENVOY_NET_GENERIC) {
+    if (auto e = engine()) {
+      e->dispatcher().post([network]() -> void {
+        if (auto e = engine())
+          e->networkConfigurator().refreshDns(network);
```
junr03, line 96: let's add a test with a mock cache to test that a flush happens on network change.

library/common/main_interface.cc
```diff
@@ -88,7 +88,15 @@ envoy_status_t reset_stream(envoy_stream_t stream) {
 }
 
 envoy_status_t set_preferred_network(envoy_network_t network) {
-  Envoy::Network::MobileUtility::setPreferredNetwork(network);
+  envoy_network_t previous = Envoy::Network::Configurator::setPreferredNetwork(network);
+  if (previous != network && previous != ENVOY_NET_GENERIC) {
+    if (auto e = engine()) {
+      e->dispatcher().post([network]() -> void {
+        if (auto e = engine())
```
junr03, line 95: I wonder if in the future we want to have Engine::networkChange(). If we want to have more things happen when a network changes. I am sure you are already thinking about that.

library/common/network/configurator.cc
```diff
@@ -60,21 +61,42 @@ namespace {
 #define SUPPORTS_GETIFADDRS
 #endif
 
-std::atomic<envoy_network_t> MobileUtility::preferred_network_{ENVOY_NET_GENERIC};
+SINGLETON_MANAGER_REGISTRATION(network_configurator);
 
-void MobileUtility::setPreferredNetwork(envoy_network_t network) { preferred_network_ = network; }
+constexpr absl::string_view BaseDnsCache = "base_dns_cache";
```
junr03, line 66: I wonder if we can tie this somehow to the config being loaded so that we don't have the possibility of the names mismatching.

library/common/network/configurator.cc
```diff
@@ -60,21 +61,42 @@ namespace {
...
+  // refreshDns is intended to be queued on Envoy's event loop, whereas preferred_network_ is
+  // updated synchronously. In the event that multiple refreshes become queued on the event loop,
+  // this avoids triggering a refresh for a non-current network.
+  // Note this does NOT completely prevent parallel refreshes from being triggered in multiple
+  // flip-flop scenarios.
+  if (network != preferred_network_.load()) {
+    return;
+  }
+  // TODO(goaway): track event here or are there existing signals we can use?
+  if (auto dns_cache = dns_cache_manager_->lookUpCacheByName(BaseDnsCache)) {
```
junr03, line 86: I wonder if it would be good to fire an event in the case:
- the cache is not present
- when we force refresh

library/common/main_interface.cc
junr03: We could also fire an event on network change in that function. Which I think we should do anyway to have a picture of when the engine was "aware" of a network change.

library/common/network/configurator.cc
goaway: I was wondering if we were firing an event in the actual refresh implementation (which might be preferable). But if we're not, I can fire it here.

library/common/network/configurator.cc
goaway: Well, that's why I made the handling optional/nullable. I also anticipated that we'd eventually be doing this per-interface, so I didn't want to overthink it.

library/common/main_interface.cc
goaway: Yes, I have thought about that. The current form felt reasonable enough to me for now, but we might be getting close to something like that making sense.

library/common/network/configurator.cc
junr03: sg. As long as we fire an event if this is null we should be able to observe if we are not getting a cache we are expecting.

library/common/main_interface.cc
goaway: That's actually exactly why I made this an object instead of static (and why initially I had it taking the cache in the constructor, but I'm sure we can mock the manager more or less as easily).

I have the start of a test class written.

If this goes green though, I wonder if it might be worth merging so we can start the release for testing over the weekend?

library/common/network/configurator.cc
junr03: It is not. I can add it there. Let's add one for the cache missing though per our other [thread](https://github.com/envoyproxy/envoy-mobile/pull/1816#discussion_r715211592)

library/common/main_interface.cc
junr03: Up to you. I think at least we should fire an (network_change with the new network) event.

junr03: Up to you. I think at least we should fire an (network_change with the new network) event.

library/common/main_interface.cc
junr03: Great. Looking forward to the test. I think there is a way to replace singleton's with mocks. I have a vague memory of that. Let me dig it up.


Well, I would like to have the events we are discussing below go out too. So at worst we merge tomorrow? I can review again once you push.

junr03: @goaway you can use `TestThreadsafeSingletonInjector` to inject a mock manager I believe.

library/common/network/configurator.cc
junr03: Adding the force refresh event here: https://github.com/envoyproxy/envoy/pull/18247/files

goaway: Thanks @junr03

junr03 (commented):
Thanks for the events. I'm still missing the test.

envoy_build_config/extensions_build_config.bzl
```diff
@@ -9,7 +9,7 @@ EXTENSIONS = {
     "envoy.filters.http.dynamic_forward_proxy":            "//source/extensions/filters/http/dynamic_forward_proxy:config",
     "envoy.filters.http.local_error":                      "@envoy_mobile//library/common/extensions/filters/http/local_error:config",
     "envoy.filters.http.platform_bridge":                  "@envoy_mobile//library/common/extensions/filters/http/platform_bridge:config",
-    "envoy.filters.http.socket_selection":                 "@envoy_mobile//library/common/extensions/filters/http/socket_selection:config",
+    "envoy.filters.http.network_configuration":                 "@envoy_mobile//library/common/extensions/filters/http/network_configuration:config",
```
snowp, line 12: indentation is off here, no need to block and rerun CI for this though. Follow up?

library/common/engine.h
snowp: Ping on this

library/common/extensions/filters/http/network_configuration/config.cc
```diff
@@ -0,0 +1,30 @@
...
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace NetworkConfiguration {
+
+Http::FilterFactoryCb NetworkConfigurationFilterFactory::createFilterFactoryFromProtoTyped(
+    const envoymobile::extensions::filters::http::network_configuration::NetworkConfiguration&,
+    const std::string&, Server::Configuration::FactoryContext& context) {
+
+  auto network_configurator = Network::ConfiguratorHandle{context}.get();
```
snowp, line 14: This pattern looks a little bit weird (at first thought I thought ConfigurationHandle was a RAII object, and that we threw it away immediately after getting the inner object). I probably would have opted for a single static function to construct/return the singleton instead of a temporary object

Non-blocking

library/common/network/configurator.cc
```diff
@@ -60,21 +61,46 @@ namespace {
...
+void Configurator::refreshDns(envoy_network_t network) {
+  // refreshDns is intended to be queued on Envoy's event loop, whereas preferred_network_ is
+  // updated synchronously. In the event that multiple refreshes become queued on the event loop,
+  // this avoids triggering a refresh for a non-current network.
+  // Note this does NOT completely prevent parallel refreshes from being triggered in multiple
+  // flip-flop scenarios.
+  if (network != preferred_network_.load()) {
+    return;
+  }
+  // TODO(goaway): track event here or are there existing signals we can use?
```
snowp, line 86: This seems to have been done with the below events? Or is this talking about something else?

library/common/network/configurator.h
```diff
@@ -0,0 +1,83 @@
...
+namespace Network {
+
+using DnsCacheManagerSharedPtr = Extensions::Common::DynamicForwardProxy::DnsCacheManagerSharedPtr;
+
+/**
+ * Network utility routines related to mobile clients.
+ */
+class Configurator : public Logger::Loggable<Logger::Id::upstream>, public Singleton::Instance {
+public:
+  Configurator(DnsCacheManagerSharedPtr dns_cache_manager)
```
snowp, line 23: `explicit` (per style guide)

snowp (approved):
LGTM besides some nits. Per offline discussion, we can merge this now and address non-blocking issues in a follow up

library/common/extensions/filters/http/network_configuration/config.cc
goaway: I lifted this pattern from some examples in Envoy, though in Envoy these tended to be called `Factory` rather than `Handle`. Would that be clearer?

library/common/network/configurator.cc
goaway: Yep the comment is out of date; thanks.

goaway:
Thanks @junr03 and @snowp! Will post followups.

library/common/extensions/filters/http/network_configuration/config.cc
snowp: Yeah I think that would make it clearer, thanks!

PR #1817, opened 2021-09-21
bazel: unify envoy_mobile_swift_test targets

keith:
Ideally, similar to cc_test targets, these would all be separate. But
with iOS tests this has a bunch of downsides:

1. You have to re-link every test bundle (same as cc_tests) which takes
   a huge amount of time for these tests (on the order of minutes per
   target)
2. You have to re-create and launch a fresh simulator for every test
   target which is a huge amount of overhead
3. You have to separately maintain a duplicative list of deps

Combining these does have some small downsides

1. You link a bit more when iterating on a test, this is nothing
   compared to the overall deps you end up linking from envoy, so it
   shouldn't meaningfully impact performance
2. You over-specify deps for some test cases, considering how few deps
   tests have today this shouldn't be an issue

Overall I think this is worth it for the time savings above. This should
meaningfully improve CI times.

### DIFF
--- a/test/swift/BUILD
+++ b/test/swift/BUILD
@@ -3,70 +3,19 @@ load("@envoy_mobile//bazel:apple_test.bzl", "envoy_mobile_swift_test")
 licenses(["notice"])  # Apache 2
 
 envoy_mobile_swift_test(
-    name = "engine_builder_tests",
+    name = "test",
     srcs = [
         "EngineBuilderTests.swift",
-    ],
-    deps = [
-        "//library/objective-c:envoy_engine_objc_lib",
-    ],
-)
-
-envoy_mobile_swift_test(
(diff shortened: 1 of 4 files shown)


rebello95:
I think the win in CI/test time reduction is worth doing 👍🏽 Should we also make this change on Android to have consistency at the platform layer? (cc @buildbreaker)

Also @goaway if you have an opinion here

goaway:
Thanks @keith. Overall I'm on board. I do find myself building/running individual integration tests with some frequency though. I think there's some sort of bazel --test-filter flag for that when it's under a single target. Do you happen to know what it is offhand?

keith:
theoretically `--test_filter`, that does only affect building and not running, it _should_ work in your setup

goaway (approved):

PR #1823, opened 2021-09-22
Add logging public api

jingwei99:
Description: Adding public logging API for kotlin and swift. The API does not function at the moment.
Risk Level: low
Testing: ci
Docs Changes: Will add docs when the API is wired up and start fuctioning

### DIFF
--- a/library/kotlin/io/envoyproxy/envoymobile/PulseClient.kt
+++ b/library/kotlin/io/envoyproxy/envoymobile/PulseClient.kt
@@ -1,7 +1,7 @@
 package io.envoyproxy.envoymobile
 
 /**
- * Client for Envoy Mobile's stats library, Pulse, used to record client time series metrics.
+ * Client for Envoy Mobile's stats and logging library, Pulse, used to record client time series metrics and loggs
  *
  * Note: this is an experimental interface and is subject to change The implementation has not been
  * optimized, and there may be performance implications in production usage.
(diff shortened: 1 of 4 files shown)


library/kotlin/io/envoyproxy/envoymobile/PulseClient.kt
```diff
@@ -41,4 +42,14 @@ interface PulseClient {
...
+
+  /**
+   * Logs with level, message and tags.
+   *
+   * @param level Level of the log
+   * @param message The log message
+   * @param tags Log tags, carries miscellaneous information about the log, such as file name, line
+   *             number etc.
+   */
+  fun log(level: LogLevel, message: String, tags: Map<String, String>?)
```
buildbreaker, line 54: Would it make sense to use `Tags` here as well? If not, would it make sense to call this something other than tags?

jingwei99: it's a difficult question haha

conceptually they are kind of different but the same - `Tags` currently are used in stats APIs, and there are certain expectations for stats "tags", such as: tag names and values are supposed to be short(er) and are supposed to be more limited in size (number of tags per stats).

but the data structure for stats tags and log tags are exactly the same

i think just the sake of not having consumers to mix these two concepts (stats tags and log tags) together, it's better to make the log tags some different

i'll rename the parameter to `logTags` lmk

buildbreaker: Sounds good. Thanks for the explanation!

buildbreaker (approved):

buildbreaker (approved):

Issue #1827, opened 2021-09-23
MacOS: can't get backtraces outside of Xcode


junr03:
Currently Envoy and Envoy Mobile do not print local stacktraces on MacOS. This is a long standing issue in the Envoy community and has not been resolved. Historical context [here](https://envoyproxy.slack.com/archives/C7B8QL52R/p1600798625006600).

Interestingly stack traces do get produced in release iOS apps with Envoy Mobile, or even when a crash happens locally in MacOS for an app running in XCode.

junr03:
cc @ggreenway @keith as people that have helped try to solve this in the past. I think @lizan might have looked at this in a previous life too?

keith:
In the short term you can view the reports here in Console.app https://support.apple.com/guide/console/reports-cnsl664be99a/mac

But it looks like this is just a macOS bug triggered by the signal handling configuration in envoy, working on that here https://github.com/envoyproxy/envoy/pull/18251

keith:
My PR is merged upstream, so next submodule update this should be fixed here

junr03:
Confirmed fixed. Thanks @keith

PR #1828, opened 2021-09-23
backtrace: no backtrace on macos

junr03:
For investigating #1827. 

Repro with `bazel test --test_output=streamed //test/common/buffer/...`


keith:
once you do the submodule update you should re-test this use case, should see the stack trace now

PR #1832, opened 2021-09-27
network: allow interface binding to be configured in programmatic API

goaway:
Description: Provides public API to configure interface binding; disabled by default.
Risk Level: Low
Testing: Updated tests & CI

### DIFF
--- a/library/swift/EngineBuilder.swift
+++ b/library/swift/EngineBuilder.swift
@@ -21,6 +21,7 @@ open class EngineBuilder: NSObject {
   private var dnsFailureRefreshSecondsMax: UInt32 = 10
   private var dnsQueryTimeoutSeconds: UInt32 = 25
   private var dnsPreresolveHostnames: String = "[]"
+  private var enableInterfaceBinding: Bool = false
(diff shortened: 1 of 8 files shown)


library/swift/EngineBuilder.swift
```diff
@@ -339,6 +352,7 @@ open class EngineBuilder: NSObject {
       dnsFailureRefreshSecondsMax: self.dnsFailureRefreshSecondsMax,
       dnsQueryTimeoutSeconds: self.dnsQueryTimeoutSeconds,
       dnsPreresolveHostnames: self.dnsPreresolveHostnames,
+      enableInterfaceBinding: self.enableInterfaceBinding,
```
Augustyniak, line 355: for consistency with the existing `adminInterfaceEnabled` we should consider calling it `interfaceBindingEnabled`. Alternatively, we can change the name of `adminInterfaceEnabled` in future PRs.

goaway: I guess I have a slight preference for the active voice, and a greater preference for addressing it later. :)

Augustyniak (approved):

PR #1833, opened 2021-09-27
network: wire interface binding config through core filter

goaway:
Description: Pass interface binding configuration and state through the NetworkConfiguration filter to the NetworkConfigurator. Down the road, we may want to have a dedicated extension for configuring this sort of thing.
Risk Level: Low
Testing: Existing integration.

### DIFF
--- a/library/common/extensions/filters/http/network_configuration/filter.cc
+++ b/library/common/extensions/filters/http/network_configuration/filter.cc
@@ -18,7 +18,8 @@ Http::FilterHeadersStatus NetworkConfigurationFilter::decodeHeaders(Http::Reques
   }
   ENVOY_LOG(debug, "will override interface: {}", override_interface_);
 
-  auto connection_options = network_configurator_->getUpstreamSocketOptions(network, override_interface_);
+  auto connection_options =
+      network_configurator_->getUpstreamSocketOptions(network, override_interface_);
   decoder_callbacks_->addUpstreamSocketOptions(connection_options);
 
   return Http::FilterHeadersStatus::Continue;
--- a/library/common/extensions/filters/http/network_configuration/filter.cc
+++ b/library/common/extensions/filters/http/network_configuration/filter.cc
@@ -13,7 +13,12 @@ Http::FilterHeadersStatus NetworkConfigurationFilter::decodeHeaders(Http::Reques
   envoy_network_t network = network_configurator_->getPreferredNetwork();
   ENVOY_LOG(debug, "current preferred network: {}", network);
 
-  auto connection_options = network_configurator_->getUpstreamSocketOptions(network);
+  if (enable_interface_binding_) {
+    override_interface_ = network_configurator_->overrideInterface(network);
+  }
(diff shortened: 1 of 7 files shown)


library/common/extensions/filters/http/network_configuration/filter.proto
```diff
@@ -3,4 +3,7 @@ syntax = "proto3";
 package envoymobile.extensions.filters.http.network_configuration;
 
 message NetworkConfiguration {
+  // If set to true, the filter will permit the NetworkConfigurator to provide upstream
+  // socket option that MAY bind a connection to a specific network interface.
```
Augustyniak, line 7: should we expand on what `MAY` means here? As in - what needs to happen for the binding to happen? Or maybe that would be to specific for a doc string?

library/common/network/configurator.h
```diff
@@ -53,7 +58,8 @@ class Configurator : public Logger::Loggable<Logger::Id::upstream>, public Singl
   /**
    * @returns the current socket options that should be used for connections.
    */
-  Socket::OptionsSharedPtr getUpstreamSocketOptions(envoy_network_t network);
+  Socket::OptionsSharedPtr getUpstreamSocketOptions(envoy_network_t network,
```
Augustyniak, line 61: lack of doc strings for params?

library/common/extensions/filters/http/network_configuration/filter.cc
```diff
@@ -13,7 +13,13 @@ Http::FilterHeadersStatus NetworkConfigurationFilter::decodeHeaders(Http::Reques
   envoy_network_t network = network_configurator_->getPreferredNetwork();
   ENVOY_LOG(debug, "current preferred network: {}", network);
 
-  auto connection_options = network_configurator_->getUpstreamSocketOptions(network);
+  if (enable_interface_binding_) {
+    override_interface_ = network_configurator_->overrideInterface(network);
+  }
+  ENVOY_LOG(debug, "will override interface: {}", override_interface_);
```
Augustyniak, line 19: Should we use `ENVOY_EVENT_LOG` here to get more visibility into this feature? If not, what's our plan to get instrumentation for interface binding?

Augustyniak: Should we include the information about the current/target interfaces in here (if possible)?

library/common/extensions/filters/http/network_configuration/filter.proto
goaway: I actually left it ambiguous intentionally since whether it does will change based on configuration and the heuristic we choose.

library/common/extensions/filters/http/network_configuration/filter.cc
goaway: I'm planning to add event logs last, once I'm sure of the shape of things.

library/common/network/configurator.h
goaway: Sure, I'll add in the next PR.

Augustyniak (approved):
looking forward to these future PRs :)

PR #1846, opened 2021-09-30
Enable Android debugging for libenvoy_jni.so

Reflejo:


keith (approved):

keith (approved):

keith (approved):
Now that I think about it I wonder if build ID is passed by default by bazel

keith:
it is not, although envoy does pass this in some cases for its binaries, theoretically we could use that, but this is probably fine

PR #1847, opened 2021-09-30
build: allow injecting additional targets into the build

snowp:
Adds a couple new injection points that allows a private build to add additional build targets into the overall EM build.

Risk Level: Low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a


keith (dismissed):

snowp:
@keith @goaway Any idea why this change is causing the JVM hello world apps to fail? I would have expected the build changes here to not change the build here at all unless the new sources/deps are specified

keith:
do the logs if you unfold the "start app" section mean anything to you? im not sure if they're normal

snowp:
I think I figured it out: by pulling in the JNI sources with a dep instead of via src it elided the link since the dependency was not referenced. Marking it as alwayslink seems to have done the trick locally at least

snowp:
All green! @keith if you want to take another look

keith (approved):

PR #1848, opened 2021-09-30
retry options predicate: create extension that is aware of the network configurator

junr03:
Description: builds a retry options extension that has access to the Network::Configurator.
Risk Level: low, not currently used
Testing: added testing scaffolding that makes sure all initial assertions are correct.


library/common/extensions/retry/options/interface_binding/predicate.h
```diff
@@ -0,0 +1,26 @@
...
+
+#include "library/common/extensions/retry/options/interface_binding/predicate.pb.h"
+#include "library/common/extensions/retry/options/interface_binding/predicate.pb.validate.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace Retry {
+namespace Options {
+
+class InterfaceBindingRetryOptionsPredicate : public Upstream::RetryOptionsPredicate {
```
goaway, line 13: Suggest we `s/InterfaceBinding/NetworkConfiguration/g`, as that's the namespacing I've landed on using elsewhere.

library/common/extensions/retry/options/network_configuration/predicate.cc
```diff
@@ -0,0 +1,25 @@
...
+namespace Extensions {
+namespace Retry {
+namespace Options {
+
+NetworkConfigurationRetryOptionsPredicate::NetworkConfigurationRetryOptionsPredicate(
+    const envoymobile::extensions::retry::options::network_configuration::
+        NetworkConfigurationOptionsPredicate&,
+    Upstream::RetryExtensionFactoryContext& context) {
+  network_configurator_ = Network::NullableConfiguratorHandle{context.singletonManager()}.get();
+  RELEASE_ASSERT(network_configurator_ != nullptr, "unexpected nullptr network configurator");
```
junr03, line 13: I made this an assert given that it would be a programming error for the singleton to not be present at the time we build the retry predicate.

library/common/extensions/retry/options/network_configuration/predicate.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
+
+package envoymobile.extensions.retry.options.network_configuration;
+
+import "validate/validate.proto";
+
+message NetworkConfigurationOptionsPredicate {
```
junr03, line 7: Nothing in config right now, right?

library/common/network/configurator.h
```diff
@@ -85,5 +85,22 @@ class ConfiguratorHandle {
   Server::Configuration::FactoryContextBase& context_;
 };
 
+/**
+ * Provides nullable access to the singleton Configurator.
+ */
+class NullableConfiguratorHandle {
```
junr03, line 91: Created this rather than having naked access via the singleton manager. Lmk what you think.

test/common/extensions/retry/options/network_configuration/predicate_test.cc
```diff
@@ -0,0 +1,46 @@
...
+
+using namespace testing;
+
+namespace Envoy {
+namespace Extensions {
+namespace Retry {
+namespace Options {
+namespace {
+
+TEST(NetworkConfigurationRetryOptionsPredicateTest, PredicateTest) {
```
junr03, line 22: Initial test that confirms proper behavior of:
- factory registration.
- death on predicate creation without a configurator.
- correct predicate creation.

junr03:
@goaway ready for review! (CI will fail until I merge https://github.com/envoyproxy/envoy/pull/18362)

library/common/extensions/retry/options/network_configuration/predicate.cc
goaway: Yep, this was my thinking as well.

library/common/extensions/retry/options/network_configuration/predicate.proto
goaway: Right.

library/common/network/configurator.h
goaway: Sure. Nit: I'd call this one the `ConfiguratorHandle` and the other the `ConfiguratorFactory`. (The reason the other is called "Handle" today is because I actually used a similar non-constructing mechanism for it at one point.)

goaway:
Thanks @junr03! One minor suggestion regarding naming, otherwise this looks great!

library/common/network/configurator.h
junr03: Updated!

goaway (approved):
Looks great - thank you @junr03!

PR #1850, opened 2021-09-30
bazelrc: update to xcode 13

junr03:

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -11,12 +11,12 @@ build --features=swift.cacheable_swiftmodules
 build --swiftcopt=-Xfrontend --swiftcopt=-no-serialize-debugging-options
 build --host_force_python=PY3
 build --ios_minimum_os=11.0
-build --ios_simulator_device="iPhone 12"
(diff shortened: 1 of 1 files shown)


docs/root/start/building/building.rst
```diff
@@ -55,7 +55,7 @@ See `ci/mac_ci_setup.sh` for the specific NDK version used during builds.
 iOS requirements
 ----------------
 
-- Xcode 12.5.1
+- Xcode 13.0
 - Swift 5.x
```
jpsim, line 59: Should this be Swift 5.5 considering Xcode 13.0+ is required?

Swift 5.0 was last supported in Xcode 10.2.1 so I'm assume we don't validate that it still compiles with that language version?

keith: we might as well just remove this line since they're tied together

keith (approved):

junr03:
Waiting on RBE machines to have Xcode 13 installed.

PR #1852, opened 2021-10-01
network: add basic unit coverage for dns refresh

goaway:

### DIFF
--- a/test/common/network/configurator_test.cc
+++ b/test/common/network/configurator_test.cc
@@ -1,0 +1,41 @@
+#include "test/extensions/common/dynamic_forward_proxy/mocks.h"
+
+#include "gtest/gtest.h"
+#include "library/common/network/configurator.h"
+
+using testing::_;
+using testing::Return;
+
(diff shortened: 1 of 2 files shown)


test/common/network/configurator_test.cc
```diff
@@ -0,0 +1,41 @@
...
+#include "gtest/gtest.h"
+#include "library/common/network/configurator.h"
+
+using testing::_;
+using testing::Return;
+
+namespace Envoy {
+namespace Network {
+
+class ConfiguratorTest : public testing::Test {
```
junr03, line 12: Do we also want an integration test?

junr03 (commented):
wonderful thank you so much!

test/common/network/configurator_test.cc
goaway: I have one, will push a separate branch in a bit.

junr03 (approved):
Talked offline. Integration test will come in a subsequent PR

PR #1854, opened 2021-10-01
android: logcat if connectivity does not succeed in example apps

junr03:


goaway:
Is there a non-interactive mode of adb logcat that's being used here?

junr03:
> Is there a non-interactive mode of adb logcat that's being used here?

Totally forgot that logcat is interactive. Going to look more into this.

PR #1856, opened 2021-10-02
Fix getifaddrs to return all addresses

Reflejo:
Description:

As of today the third party implementation of `getifaddrs` we are using
returns interfaces but doesn't iterate over all addresses so the v6_vec
is always empty given that the kernel always returns v4 first.

Risk Level: Low
Docs Changes: No

### DIFF
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -302,13 +302,15 @@ void Engine::drainConnections() {
 
 void Engine::logInterfaces() {
   auto v4_vec = network_configurator_->enumerateV4Interfaces();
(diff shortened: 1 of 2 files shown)


goaway:
Thank you @Reflejo! I suspect this implementation was written before v6 was really widely deployed anywhere, and it was replaced soon after.

Do you mind amending your commit with `-s` to add DCO?

Reflejo:
> Do you mind amending your commit with `-s` to add DCO?

Done

goaway (approved):

PR #1858, opened 2021-10-04
network: implement initial heuristic for binding alternate interface

goaway:

### DIFF
--- a/library/common/main_interface.cc
+++ b/library/common/main_interface.cc
@@ -88,14 +88,12 @@ envoy_status_t reset_stream(envoy_stream_t stream) {
 }
 
 envoy_status_t set_preferred_network(envoy_network_t network) {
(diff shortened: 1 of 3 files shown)


goaway:
Marking ready for review, but with some caveats:
- Additional testing is needed; either here or in a follow-up.
- The AuxStreamInfo should probably switch to one of the built-in metadata mechanisms in StreamInfo, which I missed when looking for someplace to put these values initially.

goaway:
cc @Reflejo

goaway:
@mattklein123 I resolved your comments as I addressed them (hopefully you don't mind), since I like to use the GH facility as a checklist of sorts during large updates. Feel free to re-open any of them though if you feel they need further discussion/work.

buildbreaker:
Thank you @mattklein123

PR #1863, opened 2021-10-05
bazel: Use hermetic JDK 11

brentleyjones:
Description: Using hermetic JDK 11 with Bazel. This improves caching, while also removing the need to manage a local JDK.
Risk Level: Low.
Testing: Built locally.
Docs Changes: N/A
Release Notes: N/A

I feel the risk is low, as this only changes the Java Runtime used, but not the toolchain used for compiling (which would still be target version of 8).


bazel/dokka.bzl
```diff
@@ -28,6 +28,7 @@ def _sources_javadocs_impl(ctx):
         unzip $sources_jar -d $sources_dir > /dev/null
 
         $java \
+            --add-opens java.base/java.util=ALL-UNNAMED \
```
brentleyjones, line 31: https://youtrack.jetbrains.com/issue/KT-43704

brentleyjones:
Failures seem to be related to the Java 8 `ClassLoader` change. Looking into a fix.

mattklein123:
:clap: @brentleyjones next up hermetic NDK??? :)

brentleyjones:
I do have an idea for that, it's just less out of the box.

mattklein123:
> I do have an idea for that, it's just less out of the box.

I will buy you a bottle of your beverage of choice if you can get us to a fully hermetic build. :)

PR #1864, opened 2021-10-05
tests: Fix `EnvoyMobileTestSuite` for JDK 11

brentleyjones:
Description: Fixed `EnvoyMobileTestSuite` for JDK 11. In JDK 9 you can no longer cast `ClassLoader` to `URLClassLoader`.
Risk Level: Low. Affects tests only.
Testing: Ran locally.
Docs Changes: N/A
Release Notes: N/A

Thanks to @bencodes for helping me resolve this.

Signed-off-by: Brentley Jones <brentleyjones@lyft.com>

### DIFF
--- a/bazel/EnvoyMobileTestSuite.kt
+++ b/bazel/EnvoyMobileTestSuite.kt
@@ -1,68 +1,31 @@
 package io.envoyproxy.envoymobile.bazel
 
-import java.io.File
-import java.net.URLClassLoader
-import java.util.zip.ZipFile
+import io.github.classgraph.ClassGraph
 import junit.framework.JUnit4TestAdapter
 import junit.framework.TestSuite
 import org.junit.runner.RunWith
 
-/**
- * This is class is taken from https://stackoverflow.com/questions/46365464/how-to-run-all-tests-in-bazel-from-a-single-java-test-rule
- *
- * Translated to Kotlin and slightly modified
- *
- * Requirements:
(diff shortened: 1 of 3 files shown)


bazel/EnvoyMobileTestSuite.kt
```diff
@@ -1,68 +1,31 @@
...
 
-    if (testAdapters.isEmpty()) {
-      throw NoTestFoundException("Unable to find any tests in test target")
-    }
+    val scan = ClassGraph()
+      .disableModuleScanning()
+      .enableAnnotationInfo()
+      .enableMethodInfo()
+      .ignoreClassVisibility()
+      .acceptPackages("io.envoyproxy")
```
Bencodes, line 21: Double check that this catches all of the test classes that the original implementation used to find.

brentleyjones: It's not catching them all, thanks for pointing it out. Digging in.

brentleyjones: I'm removing the filtering, as it needs to pick up a wide variety of them.

brentleyjones: I'll add the other ones as well.

brentleyjones: Should be good now.

buildbreaker (approved):
Thanks for your help @brentleyjones !

PR #1866, opened 2021-10-06
outlier detection: remove consecutive 5xx outlier detection

junr03:
Description: this PR removes outlier detection behavior on Envoy Mobile. Consecutive 5xx outlier detection has a few issues:
- It triggers on server 5xx(s) which has the potential of breaking clients on a service outage. 
- Not all dead connection situations end up on 5xx. This means outlier detection is not even triggering correctly on local cases. e.g.,  a stream idle timeout before a response header is received will end up on 4xx which won't trigger it for example. 

We believe passive health checking with [Local Origin outlier detection](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier.html?highlight=outlier%20detection#consecutive-local-origin-failure) _will be_ beneficial. We will experiment with that behavior in the future. 

Risk Level: med - behavior change.
Testing: device testing.

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -319,11 +319,6 @@ R"(
     // Therefore, the ejection time is short and the interval for unejection is tight, but not too
     // tight to cause unnecessary churn.
 R"(
-    outlier_detection: &base_outlier_detection
-      consecutive_5xx: 3
-      base_ejection_time: 0.001s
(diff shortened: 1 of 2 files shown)


junr03:
cc @Reflejo @mattklein123

mattklein123 (approved):

PR #1867, opened 2021-10-07
explicit flow control: wire onSendWindowAvailable through JVM

goaway:
Description: Support for explicit flow control on the request path was not yet wired through the JVM public interface. This adds the needed callback to indicate when more data may be sent.
Risk Level: Low
Testing: Existing (though we should add further coverage for this feature specifically)


carloseltuerto:
Looks like this is returning null: `env->GetMethodID(jcls_JvmObserverContext, "onSendWindowAvailable", "([J)Ljava/lang/Object;")`

PR #1900, opened 2021-10-28
envoy: c926499ed

junr03:
Additional Info:
- v3.10.4 of  nlohmann/json introduced [CPP 17 specific functionality](https://github.com/nlohmann/json/compare/v3.10.2..v3.10.4#diff-8477168a53cf806fd7b9fba0ee3b7f4e28e378a5b0b2737cdddd8cf49f99b318R396). Envoy Mobile itself is compatible with CPP 17. However, there is specific functionality not available on iOS <13. This means that this dependency cannot be updated until Envoy Mobile moves the [minimum iOS version to 13](https://github.com/envoyproxy/envoy-mobile/blob/69a697d9052d8f6842e147d5e8b9d7a34e8fdf64/.bazelrc#L12).

### DIFF
--- a/bazel/envoy_mobile_repositories.bzl
+++ b/bazel/envoy_mobile_repositories.bzl
@@ -76,7 +76,7 @@ def upstream_envoy_overrides():
         sha256 = "081ed0f9f89805c2d96335c3acfa993b39a0a5b4b4cef7edb68dd2210a13458c",
         strip_prefix = "json-3.10.2",
         urls = ["https://github.com/nlohmann/json/archive/v3.10.2.tar.gz"],
-    ),
+    )
 
 def swift_repos():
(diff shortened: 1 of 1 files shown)


goaway (approved):
thanks @junr03

jpsim:
This probably doesn't matter much, but I think the PR that introduced this CPP 17 functionality not present in the iOS 12 libraries was first introduced in v3.10.4 (https://github.com/nlohmann/json/pull/3073), meaning we could still bump the dependency up from our current v3.10.2 to v3.10.3. There doesn't appear to be any significant improvements or bug fixes in v3.10.3 for this to matter much though: https://github.com/nlohmann/json/compare/v3.10.2...v3.10.3

PR #1904, opened 2021-11-01
deps: patch nlohmann/json

junr03:
Description: creates a patch forn nlohmann/json to deal with incompatibilities in it's code with use of std::filesystem in iOS<13.
Risk Level: low - compile patch
Testing: build


buildbreaker (approved):

jpsim:
We can remove this patch once https://github.com/nlohmann/json/pull/3101 makes it into a release

PR #1908, opened 2021-11-02
[examples] Add Swift async/await example app

jpsim:
I built this while debugging an issue a user had reported, and thought it would be useful to keep as an example app.

This example showcases a few integrations that weren't being exercised elsewhere I could find:

* Swift 5.5's new async/await & structured concurrency features
* Concurrent uploads to https://httpbin.org
* Starting streams on concurrent GCD queues
* SwiftUI interface

Finally, also add a CI job to validate that this example builds and logs the expected output.

swift-async-await-example

### DIFF
--- a/examples/swift/async_await/ContentView.swift
+++ b/examples/swift/async_await/ContentView.swift
@@ -1,0 +1,116 @@
+import Envoy
+import Foundation
+import SwiftUI
+
+// MARK: - Content View
+
+private struct ContentRow: Identifiable {
+    let id = UUID()
+    let content: String
+}
+
+private extension EngineBuilder {
+    static let demoEngine = EngineBuilder()
+        .addLogLevel(.debug)
+        .build()
+}
+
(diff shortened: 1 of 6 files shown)


examples/swift/async_await/ContentView.swift
```diff
@@ -0,0 +1,116 @@
+import Envoy
+import Foundation
+import SwiftUI
+
+// MARK: - Content View
+
+private struct ContentRow: Identifiable {
```
jpsim, line 7: This file could easily be split up, but is more helpful as an example when you can read the whole thing from top to bottom. Given that this fits in 116 lines, I think that's the right tradeoff.

buildbreaker (approved):
Cool async await example @jpsim !

jpsim:
@buildbreaker can you merge this? I don't have merge rights on this repo.

goaway:
Thanks @jpsim!

PR #1910, opened 2021-11-02
Exposes a hook into the Engine to run callbacks along with getClusterManager fn on the Engine

kathan24:
Description: 
- expose a getClusterManager function on the Engine, allowing callers access to EM's cluster manager from the engine
- exposes a way for C++ consumers to hook into the Engine to run callbacks onto the dispatcher, which is required in order to achieve thread safe access to the CM exposed above
- refactors main_interface to make use of this helper. This DRYs up the code in main_interface quite a bit, and is also a consequence of wanting to expose the helpers for interacting with the engine in a C++ header separate from main_interface.h (in order to keep main_interface.h C compatible)

### DIFF
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -104,7 +104,10 @@ envoy_status_t Engine::main(const std::string config, const std::string log_leve
 
           network_configurator_ =
               Network::ConfiguratorFactory{server_->serverFactoryContext()}.get();
-          logInterfaces();
+          auto v4_interfaces = network_configurator_->enumerateV4Interfaces();
+          auto v6_interfaces = network_configurator_->enumerateV4Interfaces();
+          logInterfaces("netconf_get_v4_interfaces", v4_interfaces);
+          logInterfaces("netconf_get_v6_interfaces", v6_interfaces);
           client_scope_ = server_->serverFactoryContext().scope().createScope("pulse.");
           // StatNameSet is lock-free, the benefit of using it is being able to create StatsName
           // on-the-fly without risking contention on system with lots of threads.
@@ -300,22 +303,19 @@ void Engine::drainConnections() {
   server_->clusterManager().drainConnections();
 }
 
-void Engine::logInterfaces() {
-  auto v4_vec = network_configurator_->enumerateV4Interfaces();
-  auto v4_vec_unique_end = std::unique(v4_vec.begin(), v4_vec.end());
-  std::string v4_names = std::accumulate(v4_vec.begin(), v4_vec_unique_end, std::string{},
-                                         [](std::string acc, std::string next) {
-                                           return acc.empty() ? next : std::move(acc) + "," + next;
-                                         });
+void Engine::logInterfaces(absl::string_view event,
+                           std::vector<Network::InterfacePair>& interfaces) {
+  std::vector<std::string> names;
+  names.resize(interfaces.size());
+  std::transform(interfaces.begin(), interfaces.end(), names.begin(),
+                 [](Network::InterfacePair& pair) { return std::get<0>(pair); });
 
-  auto v6_vec = network_configurator_->enumerateV6Interfaces();
-  auto v6_vec_unique_end = std::unique(v6_vec.begin(), v6_vec.end());
-  std::string v6_names = std::accumulate(v6_vec.begin(), v6_vec_unique_end, std::string{},
-                                         [](std::string acc, std::string next) {
-                                           return acc.empty() ? next : std::move(acc) + "," + next;
-                                         });
-  ENVOY_LOG_EVENT(debug, "netconf_get_v4_interfaces", v4_names);
-  ENVOY_LOG_EVENT(debug, "netconf_get_v6_interfaces", v6_names);
+  auto unique_end = std::unique(names.begin(), names.end());
+  std::string all_names = std::accumulate(names.begin(), unique_end, std::string{},
+                                          [](std::string acc, std::string next) {
+                                            return acc.empty() ? next : std::move(acc) + "," + next;
+                                          });
+  ENVOY_LOG_EVENT(debug, event, all_names);
 }
 
 } // namespace Envoy
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -303,6 +303,10 @@ void Engine::drainConnections() {
   server_->clusterManager().drainConnections();
 }
 
+Upstream::ClusterManager& Engine::getClusterManager() {
+  return server_->clusterManager();
+}
+
 void Engine::logInterfaces(absl::string_view event,
                            std::vector<Network::InterfacePair>& interfaces) {
   std::vector<std::string> names;
--- a/library/common/engine.cc
+++ b/library/common/engine.cc
@@ -304,6 +304,8 @@ void Engine::drainConnections() {
 }
 
 Upstream::ClusterManager& Engine::getClusterManager() {
+  ASSERT(dispatcher_->isThreadSafe(),
+         "getClusterManager must be called from the dispatcher's context");
   return server_->clusterManager();
 }
 
--- a/library/common/engine_handle.cc
+++ b/library/common/engine_handle.cc
@@ -1,5 +1,7 @@
 #include "library/common/engine_handle.h"
 
+// NOLINT(namespace-envoy)
+
 envoy_status_t EngineHandle::runOnEngineDispatcher(envoy_engine_t,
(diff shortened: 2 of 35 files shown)


library/common/main_interface.h
```diff
@@ -200,6 +200,8 @@ void terminate_engine(envoy_engine_t engine);
  */
 envoy_status_t drain_connections(envoy_engine_t engine);
 
+Envoy::Upstream::ClusterManager& get_cluster_manager(envoy_engine_t engine);
```
snowp, line 203: This is the C interface, we can't return C++ types through it. I'd just remove this, callers will have to get it via the Engine handle directly

library/common/engine.cc
```diff
@@ -303,6 +303,10 @@ void Engine::drainConnections() {
   server_->clusterManager().drainConnections();
 }
 
+Upstream::ClusterManager& Engine::getClusterManager() {
+  return server_->clusterManager();
```
snowp, line 307: add an ASSERT like the other functions have

library/common/main_interface.h
kathan24: and how do we access the engine object ? by calling this - https://github.com/envoyproxy/envoy-mobile/blob/e41c2b65c7d1c523b99675174a60c8f77393841a/library/common/main_interface.cc#L19 
Again, not sure if we can call it.

snowp: Huh I was under the impression that this was accessible, seems like not. Some options:
- declare `engine()` as an exported function in a .h so that other code can pull out the internal engine handle
- similar to above, but return a cc/Engine that wraps the engine(). This would retain a bit more separation between internal and external APIs. We'd essentially let C++ get a handle to a C++ engine that points to the same underlying engine that was instantiated by the platform initialization call.
- have the C function return a void* which can be cast to ClusterManager* by calling code. This should work fine but is a bit gross in terms of APIs
- Rethink this in a way where we're not directly exposing the CM via the engine

Looping in @goaway for thoughts here

The context here is having external code integrate with the EM cluster manager, specifically in order to set up xDS subscriptions using the built in xDS subscription subsystem.

mattklein123: I think allowing the engine to be exposed is fine for now, but I don't feel strongly about it.

snowp:
Could we add a test to engine_test where we access the engine after initializing it?

library/common/main_interface.h
```diff
@@ -4,11 +4,21 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#include "engine.h"
```
snowp, line 7: Prefer full include paths, not relative. In this case though I think you can just forward declare Engine and avoid this include:

```
namespace Envoy {
class Engine;
}
```

test/common/engine_test.cc
```diff
@@ -87,4 +87,24 @@ TEST_F(EngineTest, EarlyExit) {
...
+                                     engine_running->on_engine_running.Notify();
+                                   } /*on_engine_running*/,
+                                   [](void*) -> void {} /*on_exit*/, &test_context /*context*/};
+
+  engine_ = std::make_unique<EngineHandle>(callbacks, level);
+  ASSERT_TRUE(test_context.on_engine_running.WaitForNotificationWithTimeout(absl::Seconds(3)));
+
+  bool isRun = runOnEngineDispatcher(
+      1, [](std::shared_ptr<Envoy::Engine> engine) { engine->getClusterManager(); });
+  ASSERT_TRUE(isRun);
```
snowp, line 107: Can you add a call to this after we reset the engine? That way we get coverage for the code path that returns false

snowp (changes requested):

library/common/main_interface.cc
```diff
@@ -21,6 +21,19 @@ static std::shared_ptr<Envoy::Engine> engine() {
   return engine_.lock();
 }
 
+bool runOnEngineDispatcher(envoy_engine_t,
+                           std::function<void(std::shared_ptr<Envoy::Engine>)> func) {
```
snowp, line 25: The function should pass the Engine by ref (`Envoy::Engine&`), not by shared pointer, to avoid lifetime extension

snowp:
This PR now does a few things:
* expose a `getClusterManager` function on the Engine, allowing callers access to EM's cluster manager from the engine
* exposes a way for C++ consumers to hook into the Engine to run callbacks onto the dispatcher, which is required in order to achieve thread safe access to the CM exposed above
* refactors main_interface to make use of this helper. This DRYs up the code in main_interface quite a bit, and is also a consequence of wanting to expose the helpers for interacting with the engine in a C++ header separate from main_interface.h (in order to keep main_interface.h C compatible)

junr03:
> This PR now does a few things:

@kathan24 please update the PR description and title.

library/common/main_interface.cc
```diff
@@ -258,21 +202,13 @@ envoy_status_t run_engine(envoy_engine_t, const char* config, const char* log_le
 
 void terminate_engine(envoy_engine_t) {
   // Reset the primary handle to the engine, but retain it long enough to synchronously terminate.
-  auto e = strong_engine_;
-  strong_engine_.reset();
+  auto e = EngineHandle::strong_engine_;
```
junr03, line 205: instead of making these friend accesses to the EngineHandle class, why don't we add that API to the public EngineHandle class?

library/common/engine_handle.cc
```diff
@@ -0,0 +1,18 @@
+#include "library/common/engine_handle.h"
+
+// NOLINT(namespace-envoy)
+
+envoy_status_t EngineHandle::runOnEngineDispatcher(envoy_engine_t,
```
junr03, line 5: instead of exposing this as C++ couldn't we have a `run_on_engine_dispatcher` c function that takes the envoy_engine_t and function pointer. Which in turn captures the engine ref instead of in the arg list?

What I am trying to get to is that now we are exposing functionality that is only available to C++ consumers.

junr03 (commented):
initial comments plus a request to update title and description.

library/common/engine_handle.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include "library/common/engine.h"
+#include "library/common/main_interface.h"
+#include "library/common/types/c_types.h"
+
+// TODO(snowp): No good reason for this to not be in Envoy::, but I had issues declaring the friend
+// functions when using it, so disable for now.
+
+// NOLINT(namespace-envoy)
```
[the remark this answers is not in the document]
snowp: I suspect when I put this into Envoy it was trying to befriend Envoy::run_engine. When I tried to specify ::run_engine I ran into an issue about envoy_status_t not being defined, so not quite sure what was going on there

library/common/main_interface.cc
snowp: My thinking was that we wanted to prevent consumers from being able to interfere with the init/termination flow, so by using friend functions we could limit access to only the main_interface.h functions. If we're okay with opening up access to init/start/terminate to the "public" (any additional C++ code that may pull this in) then I don't mind doing that and getting rid of all the friends

library/common/engine_handle.cc
snowp: I think exposing C++ only functionality is kinda the point: it provides safe access to interacting with Envoy::Engine which is a C++ interface and wouldn't work in C.

How were you thinking a C function would interact with the engine? We either need to pass the Engine to the function or expose some way of retrieving a handle to the engine

I guess we could have a version that doesn't interact with the Engine but just runs some C code on the dispatcher, though I'm not sure what the use case would be

snowp:
@junr03 ptal, happy to make more changes if necessary

library/common/engine_handle.h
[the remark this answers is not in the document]
junr03: I wonder if just `run_engine` without the global namespace prefix would work?

library/common/engine_handle.cc
junr03: I see. That makes sense.

library/common/main_interface.cc
snowp: Ok thanks for clarifying, that makes sense to me

library/common/engine_handle.h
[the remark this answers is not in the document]
snowp: Got this working, the magic incantation was `envoy_status_t (::run_engine(...))`

snowp:
Ok this is green, hopefully I've addressed all the comments @junr03

junr03 (approved):
lgtm

cc @goaway

PR #1917, opened 2021-11-03
CI: Add "macos" pool explicitly to fix CI failures

lfpino:
Description:
We're using --remote_default_exec_properties=Pool=macos (https://github.com/envoyproxy/envoy-mobile/blob/main/.bazelrc#L214) to set the pool name for the MacOS builds on Engflow, however, we also set some other exec_properties in some tests https://github.com/lfpino/envoy-mobile/blob/20caa5a61508b1abd8aa3578cc3eae528d5274b0/test/java/integration/BUILD#L51 and Bazel doesn't merge the exec properties set in the bazelrc with the ones in the rule (i.e. the ones in the rule takes precedence). Because of this we end up sending the MacOS builds to a non-existent "_default_" pool which triggers build failures (https://github.com/envoyproxy/envoy-mobile/runs/4095145581?check_suite_focus=true#step:7:1373).

This PR fixes this by setting the pool name explicitly (until we get rid of the sandbox exec property in the next PR)

Risk Level: Low
Testing: See workflows
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/test/java/integration/BUILD
+++ b/test/java/integration/BUILD
@@ -11,8 +11,9 @@ envoy_mobile_android_test(
         "AndroidEnvoyEngineStartUpTest.java",
     ],
     exec_properties = {
-        # TODO(lfpino): Remove this once the sandboxNetwork=off works for ipv4 localhost addresses.
+        # TODO(lfpino): Remove these once the sandboxNetwork=off works for ipv4 localhost addresses.
         "sandboxNetwork": "standard",
+        "Pool": "macos",
     },
     native_deps = [
         "//library/common/jni:libndk_envoy_jni.so",
@@ -29,8 +30,9 @@ envoy_mobile_android_test(
         "AndroidEnvoyFlowTest.java",
     ],
     exec_properties = {
-        # TODO(lfpino): Remove this once the sandboxNetwork=off works for ipv4 localhost addresses.
+        # TODO(lfpino): Remove these once the sandboxNetwork=off works for ipv4 localhost addresses.
         "sandboxNetwork": "standard",
+        "Pool": "macos",
     },
     native_deps = [
         "//library/common/jni:libndk_envoy_jni.so",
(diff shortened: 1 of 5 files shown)


keith (approved):

keith:
```
ERROR: /__w/envoy-mobile/envoy-mobile/test/kotlin/integration/BUILD:116:25: Compiling Kotlin to JVM //test/kotlin/integration:receive_data_test { kt: 64, java: 0, srcjars: 0 } failed: (Exit 1): builder failed: error executing command 
  (cd /github/home/.cache/bazel/_bazel_root/e390684146b86f31340021c5495bb190/execroot/envoy_mobile && \
  exec env - \
    LC_CTYPE=en_US.UTF-8 \
  bazel-out/host/bin/external/io_bazel_rules_kotlin/src/main/kotlin/builder '--flagfile=bazel-out/k8-fastbuild/bin/test/kotlin/integration/receive_data_test.jar-0.params')
Execution platform: @engflow_remote_config_clang//config:platform
bazel-out/host/bin/external/io_bazel_rules_kotlin/src/main/kotlin/builder: line 366: /private/var/lib/engflow/worker/work/4/exec/bazel-out/host/bin/external/io_bazel_rules_kotlin/src/main/kotlin/builder.runfiles/remotejdk11_linux/bin/java: cannot execute binary file
bazel-out/host/bin/external/io_bazel_rules_kotlin/src/main/kotlin/builder: line 366: /private/var/lib/engflow/worker/work/4/exec/bazel-out/host/bin/external/io_bazel_rules_kotlin/src/main/kotlin/builder.runfiles/remotejdk11_linux/bin/java: Undefined error: 0
```

🤔

lfpino:
I forgot we were using the same targets in both linux and mac. https://github.com/envoyproxy/envoy-mobile/pull/1922/files is better to solve this.

PR #1920, opened 2021-11-03
update tooling: bazel, and android sdk

junr03:


.bazelrc
```diff
@@ -15,7 +15,7 @@ build --ios_simulator_version=15.0
 build --spawn_strategy=local
 build --verbose_failures
 build --workspace_status_command=envoy/bazel/get_workspace_status
-build --xcode_version=13.0
+build --xcode_version=13.1
```
jpsim, line 18: Do you also need to update `ci/mac_ci_setup.sh`? https://github.com/envoyproxy/envoy-mobile/blob/e41c2b65c7d1c523b99675174a60c8f77393841a/ci/mac_ci_setup.sh#L57

jpsim: I see that wasn't sufficient. Is this because the compilation actually happens on remote execution on EngFlow and we need to add Xcode 13.1-enabled workers there?

junr03: yep!

WORKSPACE
```diff
@@ -44,7 +44,7 @@ load("//bazel:python.bzl", "declare_python_abi")
 declare_python_abi(name = "python_abi", python_version = "3")
 
 # Note: proguard is failing for API 30+
```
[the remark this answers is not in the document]
buildbreaker: I remember this would break our hello world app testing but if the CI passes, we should be good

buildbreaker (approved):

buildbreaker (approved):

PR #1921, opened 2021-11-03
make functions as c functions

jingwei99:
Description: declare functions as c functions (as opposed to c++ functions)
Risk Level: low
Testing: ci
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/library/common/jni/jni_support.h
+++ b/library/common/jni/jni_support.h
@@ -2,8 +2,8 @@
 
 // NOLINT(namespace-envoy)
 
-int jni_log_fmt(const char* tag, const char* fmt, void* value);
+extern "C" int jni_log_fmt(const char* tag, const char* fmt, void* value);
 
-int jni_log(const char* tag, const char* str);
+extern "C" int jni_log(const char* tag, const char* str);
 
-jint attach_jvm(JavaVM* vm, JNIEnv** p_env, void* thr_args);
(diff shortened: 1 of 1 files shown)


jpsim (dismissed):
This fixes the case you mentioned earlier where symbols were defined both with and without C++ mangling, right?

jingwei99:
yup. thank you!

mattklein123:
Note the actual problem is that we are including both the JVM and NDK jni.h in different places, which is pretty bad. This is just a bandaid over that. I think @snowp @keith @buildbreaker are discussing other potential better solutions.

snowp:
For posterity: the problem this tries to solve is that when building on top of EM and depending on the EM JVM targets, we may run into cases where both the NDK jni.h and JDK jni.h are being consumed, which results in two different versions of JNIEnv and JavaVM being referenced, resulting in two different overloads for attach_jvm. Only one of these has an implementation, so we end up with an undefined reference and the loader barfs at the .so.

This change removes C++ mangling from the functions, meaning that even with the two different jni.h at play we don't generate multiple functions, so the loader will now happily accept the .so. This seems to be fine, but is a bit of a band aid to solve the immediate issue: if there ever are substantial differences between what's defined by the two jni.h files we may end up in undefined behavior territory, which will likely lead to strange behavior or crashes. I'll spend some time looking into what we can do here to make things a bit more sane

@jingwei99 Can you merge main to pick up the CI fix?

buildbreaker (approved):

PR #1923, opened 2021-11-04
jni: only import JVM jni on linux, add validation that we never import JVM jni on Android

snowp:
Risk Level: Low
Testing: Existing tests
Docs Changes: n/a
Release Notes: n/a


library/common/jni/jni_import.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include <jni.h>
+
+// This validates that the jni.h header that we include is *not* the jni.h provided by the JVM. This
+// helps ensure that the build is using a consistent jni.h header.
+#if defined(__ANDROID_API__) && defined(_JAVASOFT_JNI_H_)
```
keith, line 7: surprised you also need the latter here

library/common/jni/BUILD
```diff
@@ -1,13 +1,24 @@
...
 
-envoy_cc_library(
+cc_library(
+    name = "jni_import_lib",
+    srcs = [
+        "jni_import.h"
+    ]
+    deps = select({
+        "//bazel:android_os": [],
+        "//conditions:default": [ "//bazel:jni"],
```
keith, line 17: might want to restrict the visibility of this so no one accidentally uses it

keith (approved):

library/common/jni/jni_import.h
snowp: The way it works is that we still need a `#include <jni.h>` to bring it in from the NDK, so we expect this file to be included in both the android and JVM build. In the Android build we include the NDK JNI which uses `LIBNATIVEHELPER_INCLUDE_JNI_JNI_H_` as its header guard, so we check for `_JAVASOFT_JNI_H_` which is the header guard used by the JVM build to fail the build if we included the wrong jni.h for the android build

library/common/jni/BUILD
snowp: How would I set the visibility so that just this file can import jni.h?

keith: the best you can do is the package level so `"visibility" = ["//library/common/jni:__pkg__"]` which maybe isn't restrictive enough to be worth it

mattklein123: We could have a quick format check? Should be easy?

snowp: @keith Would there be an easy way to bazel query to block only allow jni_import to have a direct dependency?

keith: Maybe instead of doing this you can move this rule somewhere else that would be in a more limited BUILD file so the visibility would be enough?

mattklein123: > Maybe instead of doing this you can move this rule somewhere else that would be in a more limited BUILD file so the visibility would be enough?

That wouldn't prevent someone from depending on //bazel:jni though, right?

snowp: We could put jni_import.h in its own package, then limit bazel:jni to only be visible to that package. jni_import would have public visibility so others could depend on that, but not jni.h directly

snowp: This wouldn't prevent someone from writing another target in bazel/BUILD that pulls in jni.h, but it seems fairly robust beyond that

snowp:
fyi @jingwei99 in case you're curious, this tries to solve the underlying issue of what we've been dealing with

library/common/jni/BUILD
mattklein123: > This wouldn't prevent someone from writing another target in bazel/BUILD that pulls in jni.h, but it seems fairly robust beyond that

That's why I was suggesting a format check via buildifier, deps query or something else. Seems not that hard.

snowp:
@keith Got this green now if you want to review

keith (approved):

PR #1924, opened 2021-11-05
Use Explicit Flow Control for Cronvoy.

carloseltuerto:
This is an almost complete rewrite of the Cronvoy main logic (CronetUrlRequest and CronetUploadDataStream). More cleanup will be done with next PRs, and much more documentation will also be added - the logic is quite subtile.

Description: Reimplement Cronvoy to take advantage of the Explicit Flow Control
Risk Level: None: only Cronvoy is changed
Testing: CI - original tests are passing, and also all related `@Ignore`
Docs Changes: N/A
Release Notes: N/A


library/common/http/client.cc
```diff
@@ -80,6 +80,16 @@ void Client::DirectStreamCallbacks::encodeData(Buffer::Instance& data, bool end_
     closeStream();
   }
 
+  if (data.length() != 0 && explicit_flow_control_ && !response_data_) {
```
alyssawilk, line 83: can we comment why this move is necessary? Sadly I don't think I grok the race condition here

library/common/http/client.cc
```diff
@@ -80,6 +80,25 @@ void Client::DirectStreamCallbacks::encodeData(Buffer::Instance& data, bool end_
     closeStream();
   }
 
+  // The response_data_ is systematically assigned here because resumeData can
+  // incur an asynchronous callback to sendDataToBridge. It might look like
+  // safe to only set the response_data_ after invoking sendDataToBridge in this
```
alyssawilk, line 85: Thanks for explaining this one to me.  I think you can cut this down to just the top two lines as that's sufficient for me at least :-)

alyssawilk (commented):
LGTM pending comment update
thanks again for both catching and debugging this one!

alyssawilk (approved):
Woot!  I'm largely trusting you on the Java changes.  Would it make sense for @colibie to review those as well or you think this is good to go?

carloseltuerto:
> Woot! I'm largely trusting you on the Java changes. Would it make sense for @colibie to review those as well or you think this is good to go?

The sad reality is this PR is very hard to review without hours of explanations, and I am sure that there are subtile bugs.  This is code is still "work in progress". There are many of rough edges left, and I have to add a lot of documentation. This is based on CAS logic - efficient but quite tricky.

PR #1925, opened 2021-11-06
bazel: change .bazelversion to be symlink from envoy

keith:
These should likely always match since envoy's build configs could
depend on it.

### DIFF
--- a/.bazelversion
+++ b/.bazelversion
@@ -1,1 +1,1 @@
-4.2.1
+envoy/.bazelversion


goaway:
Thanks @keith! This sounds good to me. We have historically had a couple of stretches where we needed to build on a different version of bazel from the version on Envoy main, but this seems good to have as a default and it's not like we can't change it if we're forced to down the road.

Issue #1520, opened 2021-06-15
Cronvoy: RequestFinishInfo -> Metrics


carloseltuerto:
When a request concludes, the C++ Cronet Engine invokes a "native" Java method to expose some metrics. To implement this, I guess that adding a bunch of "x-envoy-metric-xyz" in the response header could be considered, as long as there is a response header in all circumstances (onSucceeded, onError, onCancel).

As far as I can tell, previous redirect metrics are discarded (otherwise, for example, ```dnsStartMs``` and ```dnsEndMs``` would not make sense.) 

```
    /**
     * Called by the native code on the network thread to report metrics. Happens before
     * onSucceeded, onError and onCanceled.
     */
    @CalledByNative
    private void onMetricsCollected(long requestStartMs, long dnsStartMs, long dnsEndMs,
            long connectStartMs, long connectEndMs, long sslStartMs, long sslEndMs,
            long sendingStartMs, long sendingEndMs, long pushStartMs, long pushEndMs,
            long responseStartMs, long requestEndMs, boolean socketReused, long sentByteCount,
            long receivedByteCount) {
        synchronized (mUrlRequestAdapterLock) {
            if (mMetrics != null) {
                throw new IllegalStateException("Metrics collection should only happen once.");
            }
            mMetrics = new CronetMetrics(requestStartMs, dnsStartMs, dnsEndMs, connectStartMs,
                    connectEndMs, sslStartMs, sslEndMs, sendingStartMs, sendingEndMs, pushStartMs,
                    pushEndMs, responseStartMs, requestEndMs, socketReused, sentByteCount,
                    receivedByteCount);
        }
        // Metrics are reported to RequestFinishedListener when the final UrlRequest.Callback has
        // been invoked.
    }
```

And as a reference, here is the C++ header file: https://source.chromium.org/chromium/chromium/src/+/master:net/base/load_timing_info.h And here is what every metric looks like:
```
  /**
   * Metrics collected for a single request. Most of these metrics are timestamps for events during
   * the lifetime of the request, which can be used to build a detailed timeline for investigating
   * performance.
   *
   * <p>Events happen in this order:
   *
   * <ol>
   *   <li>{@link #getRequestStart request start}
   *   <li>{@link #getDnsStart DNS start}
   *   <li>{@link #getDnsEnd DNS end}
   *   <li>{@link #getConnectStart connect start}
   *   <li>{@link #getSslStart SSL start}
[...]

PR #1937, opened 2021-11-11
First pass at APIs for final stream metrics

alyssawilk:
This pass only sends up metrics onComplete.  TODO onError and onCancel

Description:
Risk Level:
Testing:
Part of #1520


junr03 (commented):
At a high level this looks good, and in line with what would be useful to do some pretty comprehensive perf testing.

My main concern is about documenting the three different terms at play here:
stream info, stream intel, and stream metrics. I am afraid that future developers will be confused where to stick new information. 

As I see it, stream info is generally used in the context of Envoy's StreamInfo, StreamIntel is where we put information that is going to be useful on every callback, and stream metrics is where we put information that is only passed on stream termination. 

I know this is documented in the c_types and elsewhere but I wonder if we should be more explicit about it?

alyssawilk:
Happy to be more explicit but do you have a suggestion of where we document
it? I'm not super familar with what goes where yet :-)

goaway:
First of all, I think it's great to finally be exposing more of this via the API.

I had originally envisioned all of this simply going on StreamIntel. Totaled metrics would simply be in their final state on the last callback. Is a separate callback really necessary?

alyssawilk:
ok, per discussion with @goaway I'm going to make this FinalStreamIntel and pass up onError / onComplete / onCancel

goaway:
Thanks @alyssawilk this overall looks good. Thinking about it more though, final stream intel definitely will need the fields from the original stream intel one way or another - or we'll need to pass both structs on the terminal callbacks. This is because the final intel may be the only callback received for a stream (e.g. immediate cancellation or error).

Sorry to just now be thinking of this. :(

alyssawilk:
ok, @goaway I think this is ready for another pass.

@carloseltuerto I'm going to ask your help debugging the early return I added.  I think it's a lifetime issue, where I made the upstreamInfo a shared pointer, so we didn't have to copy all the data all the time, but by the last up call to onComplete, the stats store is gone (which  IMO shouldn't go away until the envoy engine does?) so it gets cranky if you reference any stats.

Worst case we can copy all the time but I _think_ it's a test bug that the store is going away before the final completion call, I just don't understand lifetime issues in that test to be sure.

goaway:
@alyssawilk feel free to copy/merge this in to fix kotlin formatting:

https://github.com/envoyproxy/envoy-mobile/pull/1952

There's a bazel target (`build kotlin_format_fix`) that _should_ work to fix kotlin formatting as well; unsure if it requires additional setup, but it runs locally just fine for me.

goaway (changes requested):
This looks good to me, modulu formatting issues. @alyssawilk do you mind opening a tracking issue to update the iOS side of things for parity? I don't think it needs to be an immediate blocker, however.

goaway (approved):
Thanks @alyssawilk, looks great.

PR #1942, opened 2021-11-15
cc: run tsan/asan and fix memory leaks

snowp:
Ensures that the envoy_http_callbacks used by the CC engine cleans up the provided data, verifying this by enabling ASAN/TSAN checks (which uncovered the issue to begin with).

Risk Level: Medium
Testing: Existing integration tests
Docs Changes: n/a
Release Notes: n/a


snowp:
This is blocked on #1609 since we are now building cc/ with clang, which is what rejects the designated initializers as C++2a extensions

crockeo (approved):
this LGTM. still building up context on envoy-mobile again, though. the good thing is this section of code isn't being used anywhere in production (afaik) so it's not dangerous either way 🙂

library/cc/stream_callbacks.cc
```diff
@@ -24,9 +24,9 @@ void* c_on_headers(envoy_headers headers, bool end_stream, envoy_stream_intel in
     }
     auto on_headers = stream_callbacks->on_headers.value();
     on_headers(builder.build(), end_stream, intel);
-  } else {
-    release_envoy_headers(headers);
   }
+
+  release_envoy_headers(headers);
```
Augustyniak, line 29: I think that the convention in here is that `on_headers` should do the release hence we used to have an `else` statement in here

library/python/bytes_view.cc
```diff
@@ -28,7 +28,6 @@ envoy_data pyBytesAsEnvoyData(py::bytes bytes) {
 // we just copy it over to python-land.
 py::bytes envoyDataAsPyBytes(envoy_data data) {
   auto bytes = py::bytes((const char*)data.bytes, data.length);
-  release_envoy_data(data);
```
Augustyniak, line 31: again, that's kind of a convention that if your method receives envoy raw struct it is responsible for releasing it - @goaway told me about this convention. In his words:


>the convention is always as follows:
>1. if you are passed or returned a bridge struct you must assume the passer/returner is >relying on you to call release
>2. if you pass or return a bridge struct you assume the passee/returnee will cause release to >be called on it at some point exactly once
>
>in most cases that means if you receive a bridge struct you should call release on it when >you’re done. however, an exception is if you receive it and then pass it to something else (e.g. in delegation). in line with the two rules above, now you should assume whoever you >passed it to will be calling release at some point, thus absolving you of that responsibility.
>one more special case deserves mention - when you want to pass or return a bridge struct, >but continue to use it after doing so. in this case you have two options:
>1. pass a deep copy.
>2. pass a shallow copy with release replaced with a no-op, so you don’t have to worry about >when the passee/returnee calls it.

library/cc/stream_callbacks.cc
snowp: on_headers doesn't receive the envoy_header struct, it gets the ResponseHeader type which doesn't allow it to free it. This means that it's on c_on_headers to free it regardless of the path it takes

Augustyniak (approved):

PR #1947, opened 2021-11-19
Engine Builder: add API (only Android) for fallback nameservers

junr03:
Description: this PR allows the EngineBuilder to configure fallback DNS servers.
Risk Level: low -- new API
Testing: added tests.


buildbreaker (approved):
Thanks @junr03 !

PR #1949, opened 2021-11-29
remove external folder

Reflejo:
This doesn't seem to be used at all and is affecting some source mapping while debugging the example app


buildbreaker (approved):

buildbreaker (approved):

goaway (approved):
SGTM

PR #1950, opened 2021-11-29
jni: fix support for non-direct byte buffers

goaway:
Description: #1906 introduced path whereby the check for whether a buffer is direct was skipped, then treated the -1 value returned by the capacity check as an unsigned size for allocation. This fix ensures the check always happens.
Risk Level: Low
Testing: Kotlin integration tests need to be re-enabled.

### DIFF
--- a/library/common/jni/jni_utility.cc
+++ b/library/common/jni/jni_utility.cc
@@ -113,11 +113,28 @@ jobject native_map_to_map(JNIEnv* env, envoy_map map) {
 }
 
 envoy_data buffer_to_native_data(JNIEnv* env, jobject j_data) {
-  size_t data_length = static_cast<size_t>(env->GetDirectBufferCapacity(j_data));
-  return buffer_to_native_data(env, j_data, data_length);
+  // Returns -1 if the buffer is not a direct buffer.
+  jlong data_length = env->GetDirectBufferCapacity(j_data);
+
+  if (data_length < 0) {
+    jclass jcls_ByteBuffer = env->FindClass("java/nio/ByteBuffer");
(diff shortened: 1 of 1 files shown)


library/common/jni/jni_utility.cc
```diff
@@ -113,11 +113,28 @@ jobject native_map_to_map(JNIEnv* env, envoy_map map) {
 }
 
 envoy_data buffer_to_native_data(JNIEnv* env, jobject j_data) {
-  size_t data_length = static_cast<size_t>(env->GetDirectBufferCapacity(j_data));
-  return buffer_to_native_data(env, j_data, data_length);
+  // Returns -1 if the buffer is not a direct buffer.
+  jlong data_length = env->GetDirectBufferCapacity(j_data);
+
+  if (data_length < 0) {
```
carloseltuerto, line 119: I really missed a step with my "ultra safe refactoring"...

Thanks!

carloseltuerto (approved):

library/common/jni/jni_utility.cc
goaway: No worries - I actually think our integration tests would have caught it; they're just unfortunately turned off at the moment. I intend to fix them up and re-enable them in the near future. :)

junr03: The integration tests do currently run on linux: https://github.com/envoyproxy/envoy-mobile/blob/main/.github/workflows/android_tests.yml#L134

junr03 (approved):
I have not had the chance to test on the Lyft apps (where I originally encountered the problem). But given @carloseltuerto approval I feel comfortable approving.

PR #1953, opened 2021-11-30
android DNS: add ability to use fallback nameservers

junr03:
Description: this PR uses config templating to introduce fallback DNS servers for c-ares in Android. Subsequent PR will add EngineBuilder configurability.
Risk Level: low -- the current config is a no-op.
Testing: e2e testing with the Lyft App


library/common/config/config.cc
```diff
@@ -24,6 +24,15 @@ const char* native_filter_template = R"(
...
 )";
 
+const char* dns_template = R"(
+                typed_dns_resolver_config:
+                  name: envoy.network.dns_resolver.cares
+                  typed_config:
+                    "@type": type.googleapis.com/envoy.extensions.network.dns_resolver.cares.v3.CaresDnsResolverConfig
+                    resolvers: {{ dns_resolvers }}
+                    use_resolvers_as_fallback: {{ dns_use_resolvers_as_fallback }}
+)";
```
goaway, line 34: Is it possible to specify a no-op default and override so we can avoid introducing new interpolation?

library/common/config/config.cc
```diff
@@ -267,6 +268,9 @@ R"(
                   base_interval: *dns_fail_base_interval
                   max_interval: *dns_fail_max_interval
                 dns_query_timeout: *dns_query_timeout
+                typed_dns_resolver_config:
+                  name: *dns_resolver_name
```
goaway, line 272: Need a default for this. I think it makes sense to just specify c-ares as the default above (along with `dns_resolver_config`, and iOS can still always override it with the apple dns resolves.

goaway:
Looks good overall, left one comment. Thanks for updating!

library/common/config/config.cc
junr03: @goaway I was wondering if this was the case. I make overrides for it on both the java and objc definitions so I saw no point in putting in a default in the common template given there is no common default.

goaway: Right now the default (header + config) with no overrides is a valid config. I'd like to keep it that way.

One reason is that we can start to use it in tests, instead of the canned configs some tests use (that have tendency to drift).

junr03: That's fair. Will update.

goaway: Thank you!

goaway (approved):
LGTM. Thanks @junr03!

library/cc/engine_builder.cc
```diff
@@ -100,6 +100,10 @@ std::string EngineBuilder::generateConfigStr() {
       {"dns_preresolve_hostnames", this->dns_preresolve_hostnames_},
       {"dns_refresh_rate", fmt::format("{}s", this->dns_refresh_seconds_)},
       {"dns_query_timeout", fmt::format("{}s", this->dns_query_timeout_seconds_)},
+      {"dns_resolver_name", "envoy.network.dns_resolver.cares"},
+      {"dns_resolver_config",
+       "{\"@type\":\"type.googleapis.com/"
```
goaway, line 105: nit: I'd suggest a string quoting approach that doesn't require a bunch of inline escaping

goaway (approved):
Approved, just one minor nit.

goaway:
Thanks @junr03!

PR #1954, opened 2021-11-30
Update Objective-C on error/cancel to include final_stream_intel

jpsim:
Description: This is a followup to https://github.com/envoyproxy/envoy-mobile/pull/1937.
Risk Level: I believe these extra parameters are safe to ignore if we don't need to propagate them to the Objective-C interface, but would like someone else to confirm that.
Testing: Example apps still work, and downstream (closed source) consumers now compile again.
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/library/objective-c/EnvoyHTTPStreamImpl.m
+++ b/library/objective-c/EnvoyHTTPStreamImpl.m
@@ -150,9 +150,11 @@ static void *ios_on_error(envoy_error error, envoy_stream_intel stream_intel,
   atomic_store(context->closed, NO);
 
   // Create native callbacks
-  envoy_http_callbacks native_callbacks = {ios_on_headers,  ios_on_data,                  ios_on_metadata,
-                                           ios_on_trailers, ios_on_error,                 ios_on_complete,
-                                           ios_on_cancel,   ios_on_send_window_available, context};
+  envoy_http_callbacks native_callbacks = {
+      ios_on_headers, ios_on_data,     ios_on_metadata, ios_on_trailers,
+      ios_on_error,   ios_on_complete, ios_on_cancel,   ios_on_send_window_available,
+      context
+  };
   _nativeCallbacks = native_callbacks;
 
   // We need create the native-held strong ref on this stream before we call start_stream because
--- a/library/objective-c/EnvoyHTTPStreamImpl.m
+++ b/library/objective-c/EnvoyHTTPStreamImpl.m
@@ -153,8 +153,7 @@ static void *ios_on_error(envoy_error error, envoy_stream_intel stream_intel,
   envoy_http_callbacks native_callbacks = {
       ios_on_headers, ios_on_data,     ios_on_metadata, ios_on_trailers,
       ios_on_error,   ios_on_complete, ios_on_cancel,   ios_on_send_window_available,
-      context
-  };
+      context};
   _nativeCallbacks = native_callbacks;
 
   // We need create the native-held strong ref on this stream before we call start_stream because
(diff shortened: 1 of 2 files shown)


library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -74,14 +75,15 @@
   return NULL;
 }
 
-static void *ios_on_cancel(envoy_stream_intel stream_intel, void *context) {
+static void *ios_on_send_window_available(envoy_stream_intel stream_intel, void *context) {
```
goaway, line 78: Looks like this got switched up with `ios_on_cancel`.

jpsim: Renamed to `ios_on_cancel_impl` in 7af1f6e

goaway:
Thanks for jumping on this @jpsim!

library/objective-c/EnvoyHTTPStreamImpl.m
goaway: Instead, could update `on_cancel` and just ignore the new arg? And then add a no-op on_send_window_available stub.

goaway (changes requested):
Just a small request with respect to handling the on_cancel/send_window callbacks (see above).

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -143,9 +150,9 @@ - (instancetype)initWithHandle:(envoy_stream_t)handle
 
   // Create native callbacks
   // TODO(goaway) fix this up to call ios_on_send_window_available
-  envoy_http_callbacks native_callbacks = {ios_on_headers,  ios_on_data,   ios_on_metadata,
-                                           ios_on_trailers, ios_on_error,  ios_on_complete,
-                                           ios_on_cancel,   ios_on_cancel, context};
```
goaway, line 148: This was a (awful) hack, because we needed a function with a valid signature. It would be better to use an explicit no-op callback for on_send_window_available until we have time to wire it through.

library/objective-c/EnvoyHTTPStreamImpl.m
jpsim: I was under the impression that `on_send_window_available` was calling `ios_on_cancel` on purpose.

If that can be a no-op for now, then I'll do that. I didn't want to change behavior in this PR without that context.

library/objective-c/EnvoyHTTPStreamImpl.m
jpsim: Will do.

jpsim: Done in d5867ef

library/objective-c/EnvoyHTTPStreamImpl.m
goaway: nit - since you're not using the parameters, you can skip naming them and just include the type.

goaway (approved):
Thanks @jpsim!

Left one minor nit, not a big deal though.

library/objective-c/EnvoyHTTPStreamImpl.m
jpsim: This is Objective-C though, not C++/Objective-C++, so I get this error if I omit the parameter name:

```
external/envoy_mobile/library/objective-c/EnvoyHTTPStreamImpl.m:66:54: warning: omitting the parameter name in a function definition is a C2x extension [-Wc2x-extensions]
                             envoy_final_stream_intel, void *context) {
                                                     ^
```

jpsim:
If there's nothing further, @goaway could you please merge when you get a chance so we can update our downstream consumers?

PR #1958, opened 2021-12-01
[EnvoyBridgeUtility] Fix NSString to envoy_data conversion

jpsim:
Description: Fix `NSString` to `envoy_data` conversion for non-ascii strings such as '台灣大哥大'. `-length` is the number of UTF-16 code units in the receiver whereas `-lengthOfBytesUsingEncoding:` is the number of UTF8 bytes. Using the latter is necessary since we're using this value to determine how many bytes to copy from the string.
Risk Level: Low.
Testing: Added an Objective-C unit test, and enabled running on CI.
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.github/workflows/ios_tests.yml
+++ b/.github/workflows/ios_tests.yml
@@ -32,3 +32,28 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: ./bazelw test --test_output=all --config=ios --build_tests_only --config=remote-ci-macos --remote_header="Authorization=Bearer $GITHUB_TOKEN" //test/swift/...
+  objctests:
+    name: objc_tests
+    runs-on: macos-11
+    timeout-minutes: 120
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          submodules: true
+      - id: check_context
+        name: 'Check whether to run'
+        run: |
+          if git rev-parse --abbrev-ref HEAD | grep -q ^main$ || git diff --name-only origin/main | grep -qe common/ -e objective-c/ -e swift/ -e bazel/ -e ^\.bazelrc$ -e ^envoy$ -e ^WORKSPACE$ -e ^.github/workflows/ios_tests.yml$ ; then
+            echo "Tests will run."
+            echo "::set-output name=run_tests::true"
+          else
+            echo "Skipping tests."
+            echo "::set-output name=run_tests::false"
+          fi
+      - name: 'Install dependencies'
+        run: ./ci/mac_ci_setup.sh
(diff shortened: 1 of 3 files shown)


library/objective-c/EnvoyBridgeUtility.h
```diff
@@ -24,7 +24,7 @@ static inline envoy_data *toNativeDataPtr(NSData *data) {
 }
 
 static inline envoy_data toManagedNativeString(NSString *s) {
-  size_t length = s.length;
+  size_t length = [s lengthOfBytesUsingEncoding:NSUTF8StringEncoding];
```
Reflejo, line 27: I believe you need +1 for the `'\0'`

jpsim: Indeed.

Reflejo (approved):

test/objective-c/EnvoyBridgeUtilityTest.m
```diff
@@ -18,4 +18,11 @@ - (void)testToNativeData {
   XCTAssertEqual(memcmp(nativeData.bytes, testData.bytes, 3), 0);
 }
 
+- (void)testToManagedNativeString {
```
junr03, line 21: nit: more descriptive test name.
```suggestion
- (void)testToManagedNativeStringUsingUTF8Chars {
```

library/objective-c/EnvoyBridgeUtility.h
jpsim: Turns out not. Added `testToManagedNativeStringUsingUTF8Chars` to validate.

jpsim:
@junr03 this is ready for review now.

Reflejo (approved):

.github/workflows/ios_tests.yml
```diff
@@ -32,3 +32,28 @@ jobs:
...
+            echo "Skipping tests."
+            echo "::set-output name=run_tests::false"
+          fi
+      - name: 'Install dependencies'
+        run: ./ci/mac_ci_setup.sh
+      - name: 'Run Objective-C library tests'
+        if: steps.check_context.outputs.run_tests == 'true'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: ./bazelw test --test_output=all --config=ios --build_tests_only --config=remote-ci-macos --remote_header="Authorization=Bearer $GITHUB_TOKEN" //test/objective-c/...
```
junr03, line 59: thanks for adding this. Did not realize this was missing.

junr03 (approved):

PR #1963, opened 2021-12-06
Adapt Cronvoy to use the onComplete callback to get the EnvoyFinalStreamIntel.

carloseltuerto:
- Initial integration on the FinalStreamIntel for Cronvoy
- Java tests added for Envoy-Mobile and Cronvoy

Description: Adapt Cronvoy to onComplete callBack 
Risk Level: Small
Testing: Tests were added for Envoy-Mobile and Cronvoy
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/library/common/http/client.cc
+++ b/library/common/http/client.cc
@@ -440,7 +440,7 @@ void Client::sendHeaders(envoy_stream_t stream, envoy_headers headers, bool end_
   // TODO: handle potential race condition with cancellation or failure get a stream in the
   // first place. Additionally it is possible to get a nullptr due to bogus envoy_stream_t
   // from the caller.
-  // https://github.com/lyft/envoy-mobile/issues/301
+  // https://github.com/envoyproxy/envoy-mobile/issues/301
   if (direct_stream) {
     ScopeTrackerScopeState scope(direct_stream.get(), scopeTracker());
     RequestHeaderMapPtr internal_headers = Utility::toRequestHeaders(headers);
@@ -487,7 +487,7 @@ void Client::sendData(envoy_stream_t stream, envoy_data data, bool end_stream) {
   // TODO: handle potential race condition with cancellation or failure get a stream in the
   // first place. Additionally it is possible to get a nullptr due to bogus envoy_stream_t
   // from the caller.
-  // https://github.com/lyft/envoy-mobile/issues/301
+  // https://github.com/envoyproxy/envoy-mobile/issues/301
   if (direct_stream) {
     ScopeTrackerScopeState scope(direct_stream.get(), scopeTracker());
     // The buffer is moved internally, in a synchronous fashion, so we don't need the lifetime
@@ -525,7 +525,7 @@ void Client::sendTrailers(envoy_stream_t stream, envoy_headers trailers) {
   // TODO: handle potential race condition with cancellation or failure get a stream in the
   // first place. Additionally it is possible to get a nullptr due to bogus envoy_stream_t
   // from the caller.
-  // https://github.com/lyft/envoy-mobile/issues/301
+  // https://github.com/envoyproxy/envoy-mobile/issues/301
   if (direct_stream) {
     ScopeTrackerScopeState scope(direct_stream.get(), scopeTracker());
     RequestTrailerMapPtr internal_trailers = Utility::toRequestTrailers(trailers);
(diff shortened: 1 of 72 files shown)


alyssawilk (approved):

alyssawilk (commented):
oh wait, this removes setFinalStreamIntel, so it's not copying any data up.

alyssawilk (approved):

library/common/http/client.cc
```diff
@@ -346,10 +352,8 @@ void Client::DirectStream::saveFinalStreamIntel() {
   setFromOptional(latency_info_.dns_end_ms,
                   request_decoder_->streamInfo().downstreamTiming().getValue(
                       "envoy.dynamic_forward_proxy.dns_end_ms"));
-  // TODO(alyssawilk) sort out why upstream info is problematic for cronvoy tests.
-  return;
   if (info.upstreamInfo().has_value()) {
-    latency_info_.upstream_info_ = request_decoder_->streamInfo().upstreamInfo();
+    callbacks_->setFinalStreamIntel(request_decoder_->streamInfo().upstreamInfo().get());
```
alyssawilk, line 356: So my goal here was to avoid saving all the many intel fields on every call to saveLatestStreamIntel (which is called often) but instead to latch the shared pointer, and only do all the work once at the end of the stream.
If that simply doesn't work because of the cronvoy lifetime (which I'm still perplexed and dismayed about) I think it's redundant to setFinalStreamIntel here, and before we pass the intel up, and we can remove those call sites and latency_info_.

carloseltuerto: Cronet updates the latest `sent_byte_count` on every callback, which is wha the code is doing here.

On the other hand, that value is only available in FinalStreamIntel, which obviously defeats the purpose of updating ip over each callback.

I will have to take another look at this logic.

library/common/types/c_types.h
```diff
@@ -153,6 +153,8 @@ typedef struct {
   int64_t connection_id;
   // The number of internal attempts to carry out a request/operation. 0 if not present.
   uint64_t attempt_count;
+  // The number of bytes sent upstream.
+  uint64_t sent_byte_count;
```
alyssawilk, line 157: I don't think adding this has any effect unless you add it to 
https://github.com/envoyproxy/envoy-mobile/blob/main/library/common/jni/jni_utility.cc#L93
https://github.com/envoyproxy/envoy-mobile/blob/main/library/swift/StreamIntel.swift#L25
etc.

alyssawilk (commented):
Nice!  C/C++ LGTM but if you want to go for completeness I think you want to handle the new bytecount in the bridge layers.  WDYT?

alyssawilk (approved):
LGTM given I'd actually prefer to handle the byte additions in a follow-up

library/common/http/client.cc
```diff
@@ -331,29 +305,45 @@ void Client::DirectStream::saveLatestStreamIntel() {
...
-  latency_info_.request_start_ms = std::chrono::duration_cast<std::chrono::milliseconds>(
-                                       info.startTimeMonotonic().time_since_epoch())
-                                       .count();
-  latency_info_.sent_byte_count = info.bytesSent();
-  latency_info_.received_byte_count = info.bytesReceived();
-  StreamInfo::TimingUtility timing(info);
-  setFromOptional(latency_info_.request_end_ms, timing.lastDownstreamRxByteReceived(),
-                  latency_info_.request_start_ms);
-  setFromOptional(latency_info_.dns_start_ms,
+  if (!request_decoder_) {
```
alyssawilk, line 311: I don't think the request decoder gets nulled out.  I think the canonical way to see if a stream is closed is to check
parent_.getStream(stream_handle_, ALLOW_ONLY_FOR_OPEN_STREAMS)

library/common/http/client.cc
```diff
@@ -101,6 +95,7 @@ void Client::DirectStreamCallbacks::encodeData(Buffer::Instance& data, bool end_
                                 GetStreamFilters::ALLOW_FOR_ALL_STREAMS));
   direct_stream_.saveLatestStreamIntel();
   if (end_stream) {
+    direct_stream_.saveFinalStreamIntel();
```
alyssawilk, line 98: how about putting these in closeStream()?

carloseltuerto: That's was my first idea - it crashes. Sigh.

PR #1964, opened 2021-12-08
stream intel: wire iOS FinalStreamIntel for parity with jvm/android

goaway:
Description: I've changed up my approach here. Previously, I mirrored Android/core APIs by having a separate struct for FinalStreamIntel. Upon further consideration, I feel it's a significant improvement to the API's usability and future flexibility to have FinalStreamIntel be a true superset (and derivative) of StreamIntel. With this change the terminal callbacks - onComplete (previously not exposed in the public API), onCancel, and onError contain a FinalStreamIntel object, and other callbacks contain the prior StreamIntel subset.
Risk Level: Moderate
Testing: Unit, Integration

### DIFF
--- a/examples/objective-c/hello_world/ViewController.m
+++ b/examples/objective-c/hello_world/ViewController.m
@@ -103,7 +103,7 @@ NSString *_REQUEST_SCHEME = @"https";
 
     [weakSelf addResponseMessage:message headerMessage:headerMessage error:nil];
   }];
-  [prototype setOnErrorWithClosure:^(EnvoyError *error, StreamIntel *ignored) {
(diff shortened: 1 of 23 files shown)


buildbreaker (dismissed):

goaway:
Note, there will be a forthcoming Android change which will align the two APIs.

goaway:
Thanks @buildbreaker!

junr03:
@goaway Isn't exposing onComplete somewhat disruptive? Given before the filter could expect nothing further to be fired after onHeaders/onData with endstream true -- or onTrailers?

PR #1965, opened 2021-12-09
adding alpn as a response header

alyssawilk:
part of https://github.com/envoyproxy/envoy-mobile/issues/1546

### DIFF
--- a/library/common/http/header_utility.cc
+++ b/library/common/http/header_utility.cc
@@ -43,9 +43,10 @@ RequestTrailerMapPtr toRequestTrailers(envoy_headers trailers) {
   return transformed_trailers;
 }
 
-envoy_headers toBridgeHeaders(const HeaderMap& header_map) {
-  envoy_map_entry* headers =
-      static_cast<envoy_map_entry*>(safe_malloc(sizeof(envoy_map_entry) * header_map.size()));
+envoy_headers toBridgeHeaders(const HeaderMap& header_map, absl::string_view alpn) {
+  int alpn_entry = alpn.empty() ? 0 : 1;
+  envoy_map_entry* headers = static_cast<envoy_map_entry*>(
+      safe_malloc(sizeof(envoy_map_entry) * (header_map.size() + alpn_entry)));
   envoy_headers transformed_headers;
   transformed_headers.length = 0;
   transformed_headers.entries = headers;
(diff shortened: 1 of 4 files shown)


library/common/http/header_utility.cc
```diff
@@ -65,6 +66,12 @@ envoy_headers toBridgeHeaders(const HeaderMap& header_map) {
 
         return HeaderMap::Iterate::Continue;
       });
+  if (!alpn.empty()) {
+    envoy_data key = Data::Utility::copyToBridgeData("x-envoy-upstream-alpn");
```
carloseltuerto, line 70: Naïve question: is copying the String really needed?

carloseltuerto (approved):

library/common/http/header_utility.cc
alyssawilk: I think that @goaway had plans for perf improvements but AFIK this is currently the way it's done.

junr03 (commented):
@alyssawilk can we add `x-envoy-upstream-alpn` to the const time header keys available in the ResponseHeaders? 

Additionally could we add the header to the HeaderMap and then convert to envoy_headers? Seems better to me to not special case this header and pollute our bridge utilities with a special header.

Lastly, if we think this response header might be beneficial to upstream envoy users, is this a response header that could be added in the Envoy router?

alyssawilk:
So I can add the header key and move where we add the header, but the problem is the code site where we call toBridgeHeaders is DirectStreamCallbacks::encodeHeaders, which has a const ResponseHeaderMap& headers

so we can either implement this in the copy utility, or we can do a full copy of all response headers, add the keyed header, and then do another full copy of all strings into the bridge.  I'd prefer to avoid the double copy but I can move it.

For the router, I don't think there's any other use case for adding ALPN - mostly folks would want it in access logs, not shipped downstream.  Even for envoy mobile, we think this is transient until we replace it with a proper API, so I don't think there's any changes we want to make to upstream.

junr03:
@alyssawilk makes sense. 🤔 I would like to avoid adding arbitrary headers to the bridge conversion utilities. I wonder if this value is something we want to ship out in the stream_intel struct instead? 

What is the intended consumption?

Out of curiosity what would the longer term API for this be?

alyssawilk:
" I wonder if this value is something we want to ship out in the stream_intel struct instead?"
I'm sending it in the headers per mike's advice at the mobile sync a week back.   StreamIntel is currently int only and he had some optimizations he wanted to make to avoid many copies before allowing strings.

library/common/http/header_utility.cc
goaway: Rather than augmenting the transformation call, I think it might be preferable to simply add the new header to Envoy's header map prior to transforming it to the bridge format.

goaway: I suppose that's mostly a nit though - if you'd prefer to move forward, that's fine, too.

goaway (approved):
Approved, but see suggestion above.

library/common/http/header_utility.cc
alyssawilk: yeah I prefer to move forward becasue as said if we add it to the header map we have to do a fully copy of the headers (the header map is const) before doing another full copy (to the shared memory struct)

PR #1968, opened 2021-12-14
Bumping Envoy to 247deb39110eca961af1ee8147a1cbb373c6cced247d

alyssawilk:
https://github.com/envoyproxy/envoy/compare/23e5fc2939e1782416155509edfca323150e8a59...eb39110eca961af1ee8147a1cbb373c6cced247d

### DIFF
--- a/library/common/http/client.cc
+++ b/library/common/http/client.cc
@@ -7,6 +7,7 @@
 #include "source/common/http/header_map_impl.h"
 #include "source/common/http/headers.h"
 #include "source/common/http/utility.h"
+#include "source/common/stream_info/utility.h"
 
 #include "library/common/bridge/utility.h"
 #include "library/common/buffer/bridge_fragment.h"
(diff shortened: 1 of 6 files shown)


library/common/http/client.cc
```diff
@@ -325,7 +326,9 @@ envoy_stream_intel Client::DirectStreamCallbacks::streamIntel() {
 
 void Client::DirectStream::saveLatestStreamIntel() {
   const auto& info = request_decoder_->streamInfo();
-  stream_intel_.connection_id = info.upstreamConnectionId().value_or(-1);
+  if (info.upstreamInfo()) {
```
jpsim, line 329: Do we need to set `connection_id` to `-1` in the case this fails to match previous behavior?

jpsim (approved):
+1 with one question

library/common/http/client.cc
alyssawilk: yeah I added an initializer in the .h file to -1 :-)

junr03 (approved):
thanks!

PR #1969, opened 2021-12-14
Various fixes to final stream intel to send in all cases

alyssawilk:
coauthered-by: carloseltuerto 
Most of https://github.com/envoyproxy/envoy-mobile/issues/1520

### DIFF
--- a/library/common/http/client.cc
+++ b/library/common/http/client.cc
@@ -199,6 +199,7 @@ void Client::DirectStreamCallbacks::closeStream() {
     size_t erased = client.streams_.erase(direct_stream_.stream_handle_);
     ASSERT(erased == 1, "closeStream should always remove one entry from the streams map");
   }
+  direct_stream_.request_decoder_ = nullptr;
 }
 
 void Client::DirectStreamCallbacks::onComplete() {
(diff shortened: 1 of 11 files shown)


carloseltuerto (approved):

library/common/http/client.cc
```diff
@@ -257,9 +213,7 @@ void Client::DirectStreamCallbacks::onComplete() {
     http_client_.stats().stream_failure_.inc();
   }
 
-  envoy_final_stream_intel final_intel;
-  setFinalStreamIntel(final_intel);
-  bridge_callbacks_.on_complete(streamIntel(), final_intel, bridge_callbacks_.context);
+  bridge_callbacks_.on_complete(streamIntel(), finalStreamIntel(), bridge_callbacks_.context);
```
junr03, line 216: We don't latch here because closeStream latches, correct?

library/common/http/client.h
```diff
@@ -37,17 +37,6 @@ struct HttpClientStats {
   ALL_HTTP_CLIENT_STATS(GENERATE_COUNTER_STRUCT)
 };
 
-struct LatencyInfo {
```
junr03, line 40: nice, one more struct consolidated :)

library/common/http/client.cc
```diff
@@ -286,9 +240,8 @@ void Client::DirectStreamCallbacks::onError() {
             direct_stream_.stream_handle_);
   http_client_.stats().stream_failure_.inc();
 
-  envoy_final_stream_intel final_intel;
-  setFinalStreamIntel(final_intel);
-  bridge_callbacks_.on_error(error_.value(), streamIntel(), final_intel, bridge_callbacks_.context);
+  bridge_callbacks_.on_error(error_.value(), streamIntel(), finalStreamIntel(),
```
junr03, line 243: same Q as onComplete

junr03 (commented):
lgtm, just a q

library/common/http/client.cc
alyssawilk: correct - the stream may be gone at the point onComplete is called

library/common/http/client.cc
alyssawilk: yeah we latch on reset which is the only thing that calls onerror

junr03 (approved):

PR #1976, opened 2021-12-17
common: make makeEnvoyMap more generic

snowp:
Changes makeEnvoyMap to work with not only a vector of pairs, but any
map-like container type.

Risk Level: Low
Testing: Existing tests
Docs Changes: n/a
Release Notes: n/a


buildbreaker (approved):
Thanks @snowp

PR #1987, opened 2021-12-24
buffer size: expand response buffer limit to 1Mb

junr03:
Description: expands the buffer size of the DirectStream from 65kb to 1mb. 
Risk Level: low - previously responses reaching the limit size would have resulted in an error.
Testing: tested with the example apps and this Envoy diff
```
diff --git a/source/common/http/filter_manager.cc b/source/common/http/filter_manager.cc
index 2864da7758..ec102fa515 100644
--- a/source/common/http/filter_manager.cc
+++ b/source/common/http/filter_manager.cc
@@ -1531,6 +1531,7 @@ Buffer::InstancePtr ActiveStreamEncoderFilter::createBuffer() {
       [this]() -> void { this->responseDataDrained(); },
       [this]() -> void { this->responseDataTooLarge(); },
       []() -> void { /* TODO(adisuissa): Handle overflow watermark */ });
+  ENVOY_LOG_MISC(error, "setting buffer limit {}", parent_.buffer_limit_);
   buffer->setWatermarks(parent_.buffer_limit_);
   return buffer;
 }
```

### DIFF
--- a/library/common/http/client.h
+++ b/library/common/http/client.h
@@ -228,7 +228,10 @@ private:
     // It only has an effect in explicit flow control mode, where when all buffers are drained,
     // on_send_window_available callbacks are called.
     void readDisable(bool disable) override;
-    uint32_t bufferLimit() override { return 65000; }
+    uint32_t bufferLimit() override {
+      // 1Mb
+      return 1024000;
+    }
     // Not applicable
     void setAccount(Buffer::BufferMemoryAccountSharedPtr) override {
       // Acounting became default in https://github.com/envoyproxy/envoy/pull/17702 but is a no=op.


junr03:
@alyssawilk you were totally right about https://github.com/envoyproxy/envoy-mobile/pull/1934#discussion_r752259640. The true origin of that buffer limit is the DirectStream in Envoy Mobile. This PR increases the limit to 1mb for now, I have created https://github.com/envoyproxy/envoy-mobile/issues/1988 to make the setting configurable.

I followed the `response_payload_too_large` message to the filter manager where the local response is sent, and from there to where `responseDataTooLarge()` is used. And then used the log above to see the value.

buildbreaker (approved):
Nice

PR #1995, opened 2022-01-05
Decompress even when `no-transform` is specified

jpsim:
Commit Message: Decompress even when `no-transform` is specified
Additional Description: This leverages the flag added in https://github.com/envoyproxy/envoy/pull/19399 to always decompress responses, regardless of the presence of a `no-transform` cache control header, matching the more lenient behavior of other client-side networking libraries such as OkHttp or URLSession, especially in cases where the remote server is not under the client developer's control.
Risk Level: Low, responses that would previously fail to decode will now succeed when a compressed response comes with a `no-transform` cache control header.
Testing: Added unit tests in Envoy and manually confirmed in an Android app that gzipped responses with `no-transform` that previously failed now succeed.
Docs Changes: Not needed.
Release Notes: Added.
Platform Specific Features: This is configured to apply regardless of the platform, so both iOS & Android will get this new behavior.

### DIFF
--- a/test/common/integration/BUILD
+++ b/test/common/integration/BUILD
@@ -8,6 +8,10 @@ envoy_cc_test(
     name = "client_integration_test",
     srcs = ["client_integration_test.cc"],
     repository = "@envoy",
+    # TODO(jpsim): Fix remote execution for these tests
+    tags = [
+        "no-remote-exec",
+    ],
     deps = [
         "//library/common/extensions/filters/http/local_error:config",
         "//library/common/extensions/filters/http/local_error:filter_cc_proto",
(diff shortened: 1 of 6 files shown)


test/common/integration/BUILD
```diff
@@ -8,6 +8,10 @@ envoy_cc_test(
     name = "client_integration_test",
     srcs = ["client_integration_test.cc"],
     repository = "@envoy",
+    # TODO(jpsim): Fix remote execution for these tests
```
jpsim, line 11: A change in Envoy has caused the integration tests to fail when executed remotely. I'm looking into it here: https://github.com/envoyproxy/envoy-mobile/pull/1996

I'd rather not block this PR on that investigation though.

jpsim: Bisecting has identified that https://github.com/envoyproxy/envoy/pull/19221 is responsible.

@alyssawilk considering you marked that PR as having a high risk level, do you think this is surfacing a real issue? Should we revert that change until we can better understand the reason these tests broke when executed remotely?

buildbreaker (approved):

PR #1997, opened 2022-01-06
[CI] Update Xcode to 13.2.1

jpsim:
Release notes: https://developer.apple.com/documentation/xcode-release-notes/xcode-13_2_1-release-notes

Description: Update the Xcode version used in build configuration files to use the latest stable Xcode release: Xcode 13.2.1.
Risk Level: Medium, this updates the versions of llvm, clang, swift as well as the Apple frameworks we link against.
Testing: Existing tests
Docs Changes: Updated.
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


jpsim:
@lfpino is there a documented process for updating EngFlow to install new Xcode versions on its remote execution workers?

lfpino:
Hi JP! We don't have an automatic way to do it, but we can try to generate new MacOS images with the new version. Let me check with our team and get back to you.

When would you like for this to happen?

CC @ulfjack @Yannic

jpsim:
Thanks for your quick reply, Luis!

> When would you like for this to happen?

It's not urgent, but I was expecting to be able to handle the changes myself fairly quickly, so it came as a surprise that there's infrastructure that I can't configure that's involved here.

In fact, by temporarily disabling EngFlow in the GitHub Actions scripts, everything works with Xcode 13.2.1 by executing jobs directly on the GitHub Actions agents. Of course, this takes much longer to compile so I'd like to wait for Xcode 13.2.1 support in EngFlow before merging this.

I think whatever comes of this I'll want to document the process for updating Xcode versions so that moving forward we know what to expect, and can give you a proper heads up, etc.

lfpino:
Yeah, that makes a lot of sense :-)

I can file an internal feature request for us to track this. IIRC we just haven't had this use case before.

And 100% agree that having good documentation on how to handle this is critical, so maybe we can add some instructions in the Envoy as part of the upgrade.

WDYT?

jpsim:
> And 100% agree that having good documentation on how to handle this is critical, so maybe we can add some instructions in the Envoy as part of the upgrade.
> 
> WDYT?

Yes, that'd be great. Ideally we'd have better documentation about how CI works overall, for both Envoy and Envoy Mobile. Maybe we can collaborate to write that up?

Based on my understanding, which likely is incorrect in some ways:

* Envoy primarily uses Azure Pipelines for its CI, with some remote execution happening via remotebuildexecution.googleapis.com
* Envoy Mobile primarily uses GitHub Actions for its CI, with remote execution happening via EngFlow
* For both projects, environment dependencies need to be mirrored with matching versions in both the CI host (Azure Pipelines & GH Actions) as well as its remote execution service (Google RBE & EngFlow)

Am I getting that right?

jpsim:
> I can file an internal feature request for us to track this. IIRC we just haven't had this use case before.

I don't mean to rush you with this question, I'm just trying to understand: how much lead time should we expect to request a new Xcode version on EngFlow executors? Are we talking on the order of days, weeks or months?

lfpino:
I'm super familiar on how Envoy mobile CI works but not so much about Envoy. I can definitely help writing something :-)

I believe the three points you mention are mostly correct, except that for Envoy Mobile we use Engflow and the GH runner doesn't necessarily need to match the Engflow worker configuration but we do use the same docker image for both (i.e. running the CI job and running the actual Bazel actions).

W.r.t timing, once Apple has made XCode available it's usually a matter of days (even hours) to upgrade, so I'd say that asking us  ~1-2 weeks in advance is enough. We also have an oncall that can take care of this quicker for urgent situations. So it's just a matter of asking and it may not even be worth automating this fully since there's only a few versions per year, but we can always discuss it.

jpsim:
Ok thanks for sharing. We’d like to request support for Xcode 13.2.1 in EngFlow whenever possible, and we’ll merge this as soon as that’s available. There’s no big rush but we’re ready when you are.

Yannic:
It's unfortunately not super straight forward to upgrade Xcode on our side. Bazel only sends the Xcode version to RE, so we need to have Xcode installed on the executors (because Apple's license says so). The way we do that today is by installing Xcode when building our release images, which means that we need to cut a release to update Xcode. Additionally, Xcode needs a lot of space on the images, so we can only fit a certain number of versions into the image. Plus, we need to coordinate changes there between everyone using the image which takes time.

I have something on my todo list to improve the situation and at least allow us to install more than 2-3 Xcode versions at a time in an image, but there's a lot of other things on my todo list as well.

We'll start the process of adding Xcode 13.2.1 to our images now, but it'll probably be next week until it's rolled out TBH.

jpsim:
If I'm understanding you correctly, you can only have a single macOS image rolled out at any given time across all your customers?

Thanks for kicking off the process. Please keep us posted of your progress.

jpsim:
Hi @Yannic & @lfpino, how's the process going to add this version of Xcode to the EngFlow macOS image going?

lfpino:
Hey @jpsim, Yannic has been making some changes to our images to add the new XCode, should be ready in the next few days :)

jpsim:
Hi @Yannic & @lfpino, how this going? We're keeping this older version of Xcode provisioned in our CI fleet just for this project now, so it'd be good to move to Xcode 13.2.1 soon. Thanks!

lfpino:
Hey @jpsim, we're very close, @Yannic is sending some internal changes to make this process simpler, should be done soon :)

jpsim:
ok thanks for the update!

lfpino:
You should be able to try now, we rolled out new images that contain the new XCode. Let us know if you see any problems.

jpsim:
Thanks @Yannic & @lfpino, CI passed with remote exec. This is ready for review now.

Yannic (approved):

keith (approved):

PR #1999, opened 2022-01-13
bump boringssl to Envoy's version

alyssawilk:
Also fixes a build error for those of us with pedantic compilers


alyssawilk:
ping!

snowp (approved):
Thanks!

PR #2004, opened 2022-01-16
Enable the "received byte count" feature.

carloseltuerto:
After a loot of `cout`, it turns out that the number of bytes received for the headers only account for the "trimmed" headers. What is amusing is both EM and Cronet are doing the same here: they exclude the status line, and perform trimming.

Description: Report stream_intel_.received_byte_count to the Java Layer.
Risk Level: Small
Testing: Tests added for C++, tests enabled in Java, CI
Docs Changes: N/A
Release Notes: N/A
Fixes: 1426


library/common/http/client.cc
```diff
@@ -298,6 +298,9 @@ void Client::DirectStream::saveLatestStreamIntel() {
   }
   stream_intel_.stream_id = static_cast<uint64_t>(stream_handle_);
   stream_intel_.attempt_count = info.attemptCount().value_or(0);
+  if (info.getUpstreamBytesMeter()) {
+    stream_intel_.received_byte_count = info.getUpstreamBytesMeter()->wireBytesReceived();
```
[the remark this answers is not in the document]
carloseltuerto: The crux of the problem is on the [encodeHeaders](https://github.com/envoyproxy/envoy-mobile/blob/main/library/common/http/client.cc#L36) invocation, determining how many bytes are left unprocessed is not obvious - there is no Buffer to interrogate.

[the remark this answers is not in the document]
alyssawilk: but on encodeHeaders we pass it all up, no?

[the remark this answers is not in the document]
carloseltuerto: Unfortunately, no. The received byte count already includes some coming from the response body. How can we subtract those?

library/common/types/c_types.h
```diff
@@ -153,6 +153,8 @@ typedef struct {
   int64_t connection_id;
   // The number of internal attempts to carry out a request/operation. 0 if not present.
   uint64_t attempt_count;
+  // The number of bytes received from upstream.
```
alyssawilk, line 156: So this is close but not 100% accurate right?
Generally we'll include all the body bytes read, except on the call to pass headers up at which point we only send up header bytes. Do you think clarifying is going to cause more confusion than it solves?

library/java/org/chromium/net/impl/CronetUrlRequest.java
```diff
@@ -631,6 +653,19 @@ private boolean streamEnded() {
     return cronvoyCallbacks != null && cronvoyCallbacks.mEndStream;
   }
 
+  private void recordEnvoyFinalStreamIntel(EnvoyFinalStreamIntel envoyFinalStreamIntel) {
+    mEnvoyFinalStreamIntel = envoyFinalStreamIntel;
+    if (mUrlResponseInfo != null) { // Null if cancelled before receiving a Response.
+      mUrlResponseInfo.setReceivedByteCount(envoyFinalStreamIntel.getReceivedByteCount() +
```
alyssawilk, line 659: so I thought cronvoy had the following behavior
if we'd read 100 header bytes and 5000 body bytes from the network but reset the stream when only 2500 bytes had been sent up to the java layer then reset the stream, that we'd want to only report the 100+2500 bytes the java layer had seen?  I assumed you'd be doing the byte accounting of "headers + body bytes cronet had seen" in the java layers but it looks like we're reporting all bytes including buffered ones?  Is my understanding off?

alyssawilk (commented):
Sweet!  Great to see progress on this one

library/java/org/chromium/net/impl/CronetUrlRequest.java
carloseltuerto: The `mEnvoyFinalStreamIntel` is expected to be null if the request is cancelled before receiving the ResponseHeader

alyssawilk: right, but if java gets the headers, and say half the body bytes which were read off the wire, what would cronet expect to report for bytes read: bytes read off the wire, or bytes handed up to java?

library/common/types/c_types.h
carloseltuerto: Concocted something - done.

alyssawilk (approved):

library/java/org/chromium/net/impl/CronetUrlRequest.java
alyssawilk: still curious about this one

alyssawilk:
@goaway any API concerns? If not I'll merge tomorrow

library/java/org/chromium/net/impl/CronetUrlRequest.java
carloseltuerto: That's a good question. There are no tests for that. Here is what I found after many hours of "cout": Cronet is not consistent when reporting the received byte count. There is one [test with an OK response](https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/android/test/javatests/src/org/chromium/net/CronetUrlRequestTest.java;l=171) where all the bytes are counted, i.e. == `streamInfo().getUpstreamBytesMeter()->wireBytesReceived()` == 86, and there is another [test with a "NOT FOUND"](https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/android/test/javatests/src/org/chromium/net/CronetUrlRequestTest.java;l=735) where the status line is excluded from received byte count, and the expected received byte count is 120 (should be 142)
 
I "cout"ed the Buffer in codec_impl.cc This is the content of the test with "an OK response":
```
=========
HTTP/1.1 200 OK^M
Connection: close^M
Content-Length: 3^M
Content-Type: text/plain^M
^M
GET
=========== size: 86
```
The size is perfectly matching the value in the original Cronet test.

This is the content of the test with a "NOT FOUND":
```
=========
HTTP/1.1 404 Not Found^M
Content-Length: 96^M
^M
<!DOCTYPE html>
<html>
<head>
<title>Not found</title>
<p>Test page loaded.</p>
</head>
</html>

=========== size: 142
```
But this one the original Cronet test expects 120 bytes: this looks like the size without the Status Line.

I'm not yet sure how to reconcile this.

goaway:
My general concern with bytes sent/received type measurements is that they're usually ambiguous about what precisely they're measuring. Encoding/framing at each network layer changes the number of bytes "seen" by other layers. For example, HTTP/2+ framing adds some bytes (probably not accounted for here), but OTOH header compression can substantially reduce what ultimately gets sent at L4 (also probably not accounted for here) and below. Something I've proposed in the past is an accounting based on the network layer doing the measuring.

(I think @alyssawilk also alludes to this above in her question.)

Is the goal here to measure the number of bytes received at L7 _as if_ it were an unmodified HTTP/1 response?

I think it would be great if this field could be named and documented clarify what in fact is being measured.

goaway:
Sorry, to clarify, I think `receivedBytes` is a somewhat ambiguous name that will lead to different people making different assumptions about what the field represents.

carloseltuerto:
> My general concern with bytes sent/received type measurements is that they're usually ambiguous about what precisely they're measuring. Encoding/framing at each network layer changes the number of bytes "seen" by other layers. For example, HTTP/2+ framing adds some bytes (probably not accounted for here), but OTOH header compression can substantially reduce what ultimately gets sent at L4 (also probably not accounted for here) and below. Something I've proposed in the past is an accounting based on the network layer doing the measuring.
> 
> (I think @alyssawilk also alludes to this above in her question.)
> 
> Is the goal here to measure the number of bytes received at L7 _as if_ it were an unmodified HTTP/1 response?
> 
> I think it would be great if this field could be named and documented clarify what in fact is being measured.

@alyssawilk @goaway @RyanTheOptimist 
Mike, this is a nightmare. I still have not found the exact recipe to mimic Cronet. I don't know what Cronet is counting. Now I start suspecting that it does not matter much. The only thing I'm sure of is the count sent by the onHeaders callback: it reflects only the trimmed headers, without the Status Line - that's for HTTP1 only - my guess this one is the same for other protocols/compression schemes.

goaway:
@carloseltuerto How would you like to proceed? If we want to commit an arbitrary definition, my main concern would be that it be named unambiguously and documented as to what it counts.

If we were to choose what we expose, I'd propose we report a layer 4 accounting of what gets read and written to the socket for a given stream - though exposing this may require some additional work in Envoy.

carloseltuerto:
@goaway @alyssawilk @RyanTheOptimist I have been working for more than a day to have the an environment where I can perform some tests with the original Chromium::Cronet. I was able to observe more details. It looks like that Cronet has two behaviors:

- When the response is in a single packet: onHeaders includes all the bytes received up to the Body, and onSucceeded returns all the bytes received (Status line, untrimmed Headers, Body)
- When the response is more than one packet: onHeaders only includes the trimmed Headers (excludes Status line), and onSucceeded returns the sum of the body size and trimmed Headers.

I suspect that when the CronetUrlRequestTest were written, the authors did not consider any of this. They rather let the test fail the first time, and then plugged the real number they saw.

I don't want to mimic this logic. So here is a suggestion:
- onHeader always returns the trimmed Headers size with the stream_intel_
- onData  trimmed Headers size + previous and current Buffer passed to onData with the stream_intel
- onError, onCancel, onSucceed return the real number of bytes: `streamInfo().getUpstreamBytesMeter()->wireBytesReceived()` with the final_stream_intel_

This logic has this flaw: the sum of onHeader and onData call will never be the one returned in `final_stream_intel_.received_byte_count`. If this is plain data, the latter will be bigger, and if the data is compressed, then the latter will likely be smaller. However, I believe that it is important to reflect the correct number of bytes received on the final callbacks. What is being returned by the non-final callbacks does not matter much. That will be an improvement when compared with the original Cronet.

carloseltuerto:
@goaway @alyssawilk @RyanTheOptimist The suggestion has been implemented. Please let me know what you think - thanks!

library/java/io/envoyproxy/envoymobile/engine/types/EnvoyStreamIntel.java
```diff
@@ -18,4 +18,8 @@
    * The number of internal attempts to carry out a request/operation.
    */
   public long getAttemptCount();
+  /*
+   * The number of bytes received from upstream.
```
alyssawilk, line 22: update commments?

alyssawilk (commented):
Nice!  I think this is much cleaner.  One comment nit and you're good to go from my end

library/java/io/envoyproxy/envoymobile/engine/types/EnvoyStreamIntel.java
carloseltuerto: Comments updates, and also using the same names everywhere. Thanks.

alyssawilk (approved):

goaway (approved):
Thanks @carloseltuerto! This makes sense to me.

PR #2005, opened 2022-01-18
stream_intel: replacing a lost memset

alyssawilk:
Fixing a bug where final stream intel might not be updated correctly


carloseltuerto (commented):
-1 is the default value for timestamps in Cronet.

I would much prefer this in client.h

    `envoy_final_stream_intel envoy_final_stream_intel_{-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,-1,0,0,0};`

This requires a small change in extra_stream_info.cc

alyssawilk:
All the types are uint64s.  I think we'd have to rework all the APIs if we want to allow -1?

PR #2006, opened 2022-01-18
Default timestamp to -1 for envoy_final_stream_intel

carloseltuerto:
Description: Default timestamp to -1 for envoy_final_stream_intel
Risk Level: small
Testing: CI
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/library/common/http/client.h
+++ b/library/common/http/client.h
@@ -280,7 +280,8 @@ private:
     bool explicit_flow_control_ = false;
     // Latest intel data retrieved from the StreamInfo.
     envoy_stream_intel stream_intel_{-1, -1, 0};
-    envoy_final_stream_intel envoy_final_stream_intel_;
+    envoy_final_stream_intel envoy_final_stream_intel_{-1, -1, -1, -1, -1, -1, -1,
+                                                       -1, -1, -1, -1, 0,  0,  0};
     StreamInfo::BytesMeterSharedPtr bytes_meter_;
   };
 
--- a/library/common/types/c_types.h
+++ b/library/common/types/c_types.h
@@ -157,34 +157,36 @@ typedef struct {
 
 /**
  * Contains internal HTTP stream metrics which sent at stream end.
+ *
+ * Note: for the signed fields, -1 means not present.
  */
 typedef struct {
   // The time the request started, in ms since the epoch.
-  uint64_t request_start_ms;
+  int64_t request_start_ms;
   // The time the DNS resolution for this request started, in ms since the epoch.
-  uint64_t dns_start_ms;
+  int64_t dns_start_ms;
   // The time the DNS resolution for this request completed, in ms since the epoch.
-  uint64_t dns_end_ms;
+  int64_t dns_end_ms;
   // The time the upstream connection started, in ms since the epoch.
   // This may not be set if socket_reused is false.
-  uint64_t connect_start_ms;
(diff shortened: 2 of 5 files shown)


library/common/http/client.h
```diff
@@ -280,7 +280,8 @@ class Client : public Logger::Loggable<Logger::Id::http> {
     bool explicit_flow_control_ = false;
     // Latest intel data retrieved from the StreamInfo.
     envoy_stream_intel stream_intel_{-1, -1, 0};
-    envoy_final_stream_intel envoy_final_stream_intel_;
```
alyssawilk, line 283: also update library/common/extensions/filters/http/platform_bridge/filter.cc

library/common/types/c_types.h
```diff
@@ -157,34 +157,36 @@ typedef struct {
...
-  uint64_t sending_start_ms;
+  int64_t sending_start_ms;
   // The time the last byte of the request was sent upstream, in ms since the epoch.
-  uint64_t sending_end_ms;
+  int64_t sending_end_ms;
   // The time the first byte of the response was received, in ms since the epoch.
-  uint64_t response_start_ms;
+  int64_t response_start_ms;
   // The time the last byte of the request was received, in ms since the epoch.
-  uint64_t request_end_ms;
```
alyssawilk, line 187: I believe you also need to change every other API layer if you're going to do this.  For example, the swift also uses unsigned ints: library/swift/FinalStreamIntel.swift
Also we should probably have type consistency between stream intel and final stream intel, so if you move this to unsigned you should move those.

alyssawilk (changes requested):

library/common/types/c_types.h
carloseltuerto: For Java, it turns out that the language only supports signed Long. Kotlin also used signed Long in its current implementation. But you are right, I missed Swift - now adapted. Thanks!

library/swift/FinalStreamIntel.swift
```diff
@@ -2,30 +2,31 @@
 import Foundation
 
 /// Exposes one time HTTP stream metrics, context, and other details.
+/// Note: -1 means "not present" for the fields of type Int64.
```
jpsim, line 5: Why wouldn't we use an `Optional` here, i.e. the language's built-in mechanism to indicate that a value is "not present"? That would prevent some classes of misuse errors for these fields.

library/common/http/client.h
carloseltuerto: Done - thanks.

alyssawilk (approved):
I think this is fine but definitely needs Lyft sign-off

library/swift/FinalStreamIntel.swift
carloseltuerto: Well, this PR goal does not really change the situation. In reality, there was an important comment that was missing previously:
`/// Note: 0 means "not present" for any field where its name ends with Ms`
And I agree that this is a suboptimal API. But it is on par with the Kotlin and Java versions - same pattern.
My guess is dedicated PR for all 3 languages would be appropriate.

goaway (approved):
I'm okay proceeding with this as it follows the existing convention. That said, I would generally prefer if we used real optionals for public fields that might not be present, both because it aligns better with the Swift and Kotlin ecosystems, and because it's a better representation for "lack of a value" in general.

PR #2007, opened 2022-01-19
Solve CI flakiness for test/java/org/chromium/net/urlconnection:urlconnection_test 

carloseltuerto:
Reduce the number of concurrent requests for CronetUrlRequestTest.testManyRequests to 1000.

This solves this crash:
```
[2022-01-23 17:01:31.159][9613895][critical][assert] [external/envoy/source/common/network/connection_impl.cc:604] assert failure: !connecting_.
```

This ran on a 2017 MacBookPro without errors:

```
bazel test --define=signal_trace=disabled --runs_per_test=1000 test/java/org/chromium/net:cronet_url_request_test
```

However, this PR does not solve the root cause, it only mitigates it. **It should not be possible for this ASSERT to fail**.

Description: EM has a 1024 concurrent request limit that is now respected by Cronet tests
Risk Level: N/A
Testing: Solves CI "android_tests / java_tests_mac" failure
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/bazel/envoy_mobile_dependencies.bzl
+++ b/bazel/envoy_mobile_dependencies.bzl
@@ -71,8 +71,8 @@ def kotlin_dependencies():
             # Test artifacts
             "org.assertj:assertj-core:3.12.0",
             "junit:junit:4.12",
-            "org.mockito:mockito-inline:2.28.2",
-            "org.mockito:mockito-core:2.28.2",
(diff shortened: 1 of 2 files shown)


goaway (approved):

goaway:
Thanks @carloseltuerto! For what it's worth, the limit is fairly arbitrary and I'm not opposed to increasing it.

PR #2014, opened 2022-01-24
tooling: first pass at oncall tooling

alyssawilk:
copy/paste/edited from
https://github.com/envoyproxy/envoy/blob/main/.github/workflows/pr_notifier.yml

General functionality:
send daily reminders to folks listed as reviewers. Lyft folk: please let me know who else if any should be added
send on-call channel updates for PRs which are out-SLO (haven't been reviewed in 2 days, and aren't tagged as waiting)

Major differences from the Envoy tooling:
removed all the first-pass review and api review cruft
no nags for unassigned PRs - I think we'll want this at some point but I think most repo contributors have been assigning their own when the PR is ready, WDYT?


.github/actions/pr_notifier/pr_notifier.py
```diff
@@ -0,0 +1,171 @@
...
+import github
+from slack_sdk import WebClient
+from slack_sdk.errors import SlackApiError
+
+REVIEWERS = {
+    'alyssawilk': 'U78RP48V9',
+    'junr03': 'U79K0Q431',
+    'RyanTheOptimist': 'U01SW3JC8GP',
+    'goaway': 'U7TDPD3L2',
+    'snowp': 'U93KTPQP6',
```
goaway, line 29: I think it would make sense to add buildbreaker, jpsim, and Augustyniak here - this is just for review reminders, right?

goaway (dismissed):
Thanks @alyssawilk! This sounds great. Suggested a few more names, but this sounds good as an initial setup to me either way.

goaway (approved):

.github/actions/pr_notifier/pr_notifier.py
alyssawilk: cc @buildbreaker @jpsim @Augustyniak so when you eventually get slack pings from the "PR reminders" app, you know where they came from :-)

PR #2017, opened 2022-01-25
envoy: 72287840959d

junr03:


junr03:
I bet https://github.com/envoyproxy/envoy/pull/19564/files is making the expectation in the IdleTimeoutTests swift test fail

jpsim:
What course of action do you recommend to address that timeout test failure?

PR #2022, opened 2022-01-25
Fixed network policy of client_integration_test

### LINKED ISSUE #2020: [CI] Run integration tests on EngFlow
In https://github.com/envoyproxy/envoy-mobile/pull/1995 I disabled remote execution for the integration test CI jobs because of test failures that were only occurring when using remote execution but ran fine on GitHub Actions workers directly.

Here's the kind of failure we were seeing when running these tests with remote execution: https://github.com/envoyproxy/envoy-mobile/runs/4739746396?check_suite_focus=true

And the same tests running successfully without remote execution: https://github.com/envoyproxy/envoy-mobile/runs/4740616118?check_suite_focus=true

WillEngFlow:
//test/common/integration:client_integration_test requires networking
which is not allowed by default. This allows the test to use the
localhost network.

Signed-off-by: Will <will@engflow.com>

Description: //test/common/integration:client_integration_test requires networking which is not allowed by default. This allows the  test to use the localhost network.
Risk Level: Low
Testing: See  unit_tests  workflow
Docs Changes: N/A
Release Notes: N/A
Fixes #2020

### DIFF
--- a/test/common/integration/BUILD
+++ b/test/common/integration/BUILD
@@ -7,11 +7,11 @@ envoy_package()
 envoy_cc_test(
     name = "client_integration_test",
     srcs = ["client_integration_test.cc"],
-    repository = "@envoy",
     exec_properties = {
         # TODO(willengflow): Remove this once the sandboxNetwork=off works for ipv4 localhost addresses.
         "sandboxNetwork": "standard",
     },
+    repository = "@envoy",
     deps = [
         "//library/common/extensions/filters/http/local_error:config",
         "//library/common/extensions/filters/http/local_error:filter_cc_proto",
--- a/test/common/integration/BUILD
+++ b/test/common/integration/BUILD
@@ -8,10 +8,10 @@ envoy_cc_test(
     name = "client_integration_test",
     srcs = ["client_integration_test.cc"],
     repository = "@envoy",
-    # TODO(jpsim): Fix remote execution for these tests
-    tags = [
-        "no-remote-exec",
-    ],
+    exec_properties = {
+        # TODO(willengflow): Remove this once the sandboxNetwork=off works for ipv4 localhost addresses.
+        "sandboxNetwork": "standard",
+    },
     deps = [
         "//library/common/extensions/filters/http/local_error:config",
         "//library/common/extensions/filters/http/local_error:filter_cc_proto",


WillEngFlow:
@jpsim @goaway @lfpino

goaway (dismissed):

goaway:
Thanks @WillEngFlow!

WillEngFlow:
@jpsim It looks like this is failing because the `exec_properties` attribute does not exist on `envoy_cc_test` upstream. Do you mind adding it for me and piping it into the wrapped `cc_test`?

`envoy_cc_test` defined here: https://github.com/envoyproxy/envoy/blob/519774f742bb2500cf5f6a00933c9662337acbb2/bazel/envoy_test.bzl#L143

An example of piping `exec_properties` through via 'envoy_mobile_jni_kt_test`: https://github.com/envoyproxy/envoy-mobile/blob/cef4ec0a29d3adf09a0df68da3d2aa8c988a7335/bazel/kotlin_test.bzl#L4

jpsim:
Sure I'll plumb that through and see if it helps. I'll do the same for the Swift integration test that similarly had remote exec disabled.

jpsim:
[[bazel] allow specifying `exec_properties` in `envoy_cc_test`](https://github.com/envoyproxy/envoy/pull/19711)

jpsim:
I've validated that it works: https://github.com/envoyproxy/envoy-mobile/runs/4959253557

My upstream Envoy PR (https://github.com/envoyproxy/envoy/pull/19711) was approved and is being auto-merged. Once that's in, I'll bump the Envoy submodule commit to point to it and you'll be able to rebase this PR.

jpsim:
https://github.com/envoyproxy/envoy-mobile/pull/2027

WillEngFlow:
> #2027

Thanks for all the work @jpsim! I'm tracking the submodule update.

jpsim:
I just merged #2027 so if you update, you should be able to pass `exec_properties` in `envoy_cc_test`.

WillEngFlow:
> I just merged #2027 so if you update, you should be able to pass `exec_properties` in `envoy_cc_test`.

Thanks @jpsim! I just pushed the most recent fix. I did get some linter errors on the WORKSPACE file after rebasing (in the pre-push check) but no changes were made to WORKSPACE and my changes checked out so I pushed with `--no-verify`

jpsim:
@WillEngFlow i also often get errors in the pre-push hook, but `pre-commit —all-files` works. I need to look into fixing that at some point.

in this case, I think the BUILD file you edited needs to move the attribute to match the definition.

WillEngFlow:
> Thanks for seeing this through!

Thanks for all the help for a first time contributor!

PR #2027, opened 2022-01-27
envoy: 71248e512

jpsim:
Changes: https://github.com/envoyproxy/envoy/compare/519774f742bb2500cf5f6a00933c9662337acbb2...71248e51257e7f9f6fbe9cfb2ead834cd8dff550


bazel/envoy_mobile_repositories.bzl
```diff
@@ -144,7 +144,7 @@ def android_repos():
     http_archive(
         name = "build_bazel_rules_android",
         urls = ["https://github.com/bazelbuild/rules_android/archive/v0.1.1.zip"],
-        sha256 = "cd06d15dd8bb59926e4d65f9003bfc20f9da4b2519985c27e190cddc8b7a7806",
+        sha256 = "7cb2b2e11aafe08b2ccb6eeb11f21f52c45e1ad1cdd0132ddf7bde8462bd9e20",
```
jpsim, line 147: https://github.com/bazelbuild/rules_android/issues/43

ahumesky: The change in the zip files was the result of a change on Github's side. They've rolled back the change that they believe caused the change in zip files:
https://twitter.com/tgummerer/status/1488493440103030787

So the hash should go back to what it was before.

jpsim: Thanks for linking to that twitter thread. I'll switch back once GitHub starts serving the archive with the previous checksum again (still serving 7cb2b2e right now).

jpsim: An alternate URL (https://github.com/bazelbuild/rules_android/archive/refs/tags/v0.1.1.zip) produces the expected checksum.

goaway (approved):
Looks good, thanks @jpsim.

PR #2028, opened 2022-01-27
bazel: remove java-related .bazelrc flags that have no effect

benjaminp:
Description: These flags have to do with toolchain resolution, which is not used in Bazel 4.x for Java rules.
Risk Level: low
Testing: run CI
Docs Changes: N/A

Signed-off-by: Benjamin Peterson <benjamin@engflow.com>


lfpino:
@jpsim or @goaway could you please approve this so that Benjamin can run the GH actions as well? It's his first contribution.

jpsim:
> could you please approve this so that Benjamin can run the GH actions as well? It's his first contribution.

I actually don't have permissions to trigger CI jobs either, but it looks like they ran and passed now.

jpsim (dismissed):
+1 as long as this was tested locally and worked too.

snowp:
Is this still relevant?

PR #2029, opened 2022-01-27
build: avoid signal handling for android envoy mobile builds

buildbreaker:
Signed-off-by: Alan Chiu achiu@lyft.com

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -2,7 +2,6 @@
 try-import ./envoy/.bazelrc
 
 # Common flags for all builds
-build --define signal_trace=disabled
 build --define=google_grpc=disabled
 build --define=hot_restart=disabled
 build --define=tcmalloc=disabled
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -2,6 +2,7 @@
 try-import ./envoy/.bazelrc
 
 # Common flags for all builds
+build --define signal_trace=disabled
```
vikrum, line 5: Should there be an '=' after define like the below lines?

jpsim (approved):

jpsim (approved):
Thanks

goaway:
I'd suggest we hold off on merging this at least until we can observe the effects in the wild. Even then, I'm not certain if it makes sense to commit this as the library default. I've opened a [tracking issue](https://github.com/envoyproxy/envoy-mobile/issues/2033) with instructions for disabling handling.

PR #2034, opened 2022-02-02
Migrate to Bazel 5.0.0

jpsim:
* Update Envoy: https://github.com/envoyproxy/envoy/compare/71248e51257e7f9f6fbe9cfb2ead834cd8dff550...bbcd487bd178ef5254545941ac369d944da41913
* Remove various `remotejdk_11` configurations in .bazelrc which we now inherit from Envoy
* Work around https://github.com/bazelbuild/bazel/issues/14260 androidndk fetching issue


keith (approved):
seems good unless we can come up with another workaround

.bazelrc
```diff
@@ -16,10 +16,6 @@ build --spawn_strategy=local
 build --verbose_failures
 build --workspace_status_command=envoy/bazel/get_workspace_status
 build --xcode_version=13.2.1
-build --host_javabase=@bazel_tools//tools/jdk:remote_jdk11
```
Bencodes, line 19: There is a new way to provide these values on Bazel 5.0 if you want to preserve this. At minimum you might want to provide the `java_language_version` and `tool_java_language_version` just to make sure that it's consistent moving forward.

Ex:
```
build \
  --java_language_version=8 \
  --tool_java_language_version=8 \
  --java_runtime_version=lyft_jdk11 \
  --tool_java_runtime_version=lyft_jdk11
```

jpsim: These are actually defined in Envoy now, and we import that `.bazelrc` file: https://github.com/envoyproxy/envoy/blob/708742e9e8ec44e634a02d16eee0dce826130298/.bazelrc#L19-L20

keith (approved):

PR #2037, opened 2022-02-02
Migrate EngFlow configurations from rbe_autoconfig to rbe_configs_gen

jpsim:
Following the instructions in https://github.com/bazelbuild/bazel-toolchains

Specifically:

    $ ./rbe_configs_gen \
            --bazel_version="4.2.1" \
            --toolchain_container="docker.io/envoyproxy/envoy-build-ubuntu:514e2f7bc36c1f0495a523b16aab9168a4aa13b6" \
            --exec_os="linux" \
            --target_os="linux" \
            --output_src_root="/path/to/envoy-mobile" \
            --output_config_path="third_party/rbe_configs" \
            --cpp_env_json=clang.json \
            --generate_java_configs=False
    $ cat clang.json
    {"CC": "/opt/llvm/bin/clang", "CXX": "/opt/llvm/bin/clang++"}

This will be necessary to migrate to Bazel 5.0.0.


keith:
@Yannic can you review? is this how this should work now?

Yannic:
@lfpino PTAL

lfpino:
I'll take a look @keith, can you please add me as a reviewer?

keith:
I cannot, maybe you have to be part of the org to be added as a reviewer?

jpsim:
The GitHub UI also won't let me.

image

lfpino:
I guess I'll review nonetheless :')

third_party/rbe_configs/cc/armeabi_cc_toolchain_config.bzl
```diff
@@ -0,0 +1,82 @@
...
+    artifact_name_patterns = []
+    make_variables = []
+
+    tool_paths = [
+        tool_path(name = "ar", path = "/bin/false"),
+        tool_path(name = "compat-ld", path = "/bin/false"),
+        tool_path(name = "cpp", path = "/bin/false"),
+        tool_path(name = "dwp", path = "/bin/false"),
+        tool_path(name = "gcc", path = "/bin/false"),
+        tool_path(name = "gcov", path = "/bin/false"),
```
lfpino, line 50: This will break coverage. You need to force run coverage and make sure you add all of the env variables

third_party/rbe_configs/config/BUILD
```diff
@@ -0,0 +1,52 @@
...
+        "@bazel_tools//platforms:x86_64",
+        "@bazel_tools//tools/cpp:clang",
+    ],
+    exec_properties = {
+        "container-image": "docker://envoyproxy/envoy-build-ubuntu@sha256:4965eb16ba40bccf8dbf1cb03a544c8671e5a239396a90fd7159a24195181b2a",
+        "OSFamily": "Linux",
+        "Pool": "linux",
+        # Necessary to workaround https://github.com/google/sanitizers/issues/916, otherwise, dangling threads in the
+        # docker container fail tests on teardown (example: https://github.com/envoyproxy/envoy-mobile/runs/3443649963)
+        "dockerAddCapabilities": "SYS_PTRACE",
```
lfpino, line 50: This capability is only needed for asan/tsan, I think you should have a separate platform here.

lfpino (changes requested):
The change is almost there but the current toolchain would break coverage so please fix that one first.

third_party/rbe_configs/cc/armeabi_cc_toolchain_config.bzl
jpsim: Fixed in 155ed54bf106da69b705346d0bd18ffe7453149b

third_party/rbe_configs/config/BUILD
jpsim: Addressed in 2d37c209a508911db89f01cb7e08b226e5e4a66d.

jpsim:
Thanks for the review, @lfpino.

* Coverage was indeed broken, that's fixed now. The CI job was effectively being skipped because this PR didn't touch any source or test files
* I've split out the asan platform into its own configuration so that other executions don't have the `SYS_PTRACE` Docker capability since it's generally not needed

Can you please take another look?

lfpino (approved):
LGTM!

keith (approved):

PR #2042, opened 2022-02-06
Implement RequestFinishInfo -> Metrics

### LINKED ISSUE #1520: Cronvoy: RequestFinishInfo -> Metrics

carloseltuerto:
Fixes the Epoch timestamp issue: now compatible with the Cronet logic.

__WARNING__: changes the meaning of `end_request`. Will be renamed in a follow-up PR

Description: Implement ExperimentalCronetEngine#addRequestFinishedListener()
Risk Level: Small
Testing: CI, C++ test, ignored tests due to issue #1520 are not ignored anymore!
Docs Changes: n/a
Release Notes: n/a
Fixes #1520 


library/common/stream_info/extra_stream_info.cc
```diff
@@ -6,50 +6,60 @@ namespace Envoy {
...
   CONSTRUCT_ON_FIRST_USE(std::string, "envoy_mobile.extra_stream_info");
 }
 
-void setFinalStreamIntel(StreamInfo& stream_info, envoy_final_stream_intel& final_intel) {
+void setFinalStreamIntel(StreamInfo& stream_info, TimeSource& time_source,
+                         envoy_final_stream_intel& final_intel) {
+  final_intel.request_start_ms = std::chrono::duration_cast<std::chrono::milliseconds>(
+                                     stream_info.startTime().time_since_epoch())
+                                     .count();
+  int64_t offset_ms =
```
alyssawilk, line 29: I think it's not clear what we're doing with this offset - mind commenting?

alyssawilk (commented):
yay!  So exciting to see the test finally working :-)

library/common/stream_info/extra_stream_info.cc
carloseltuerto: Added a little comment.

Done.

goaway (approved):
Looks fine to me. 👍

PR #2043, opened 2022-02-08
Rename request_end to stream_end, and adjust documentation.

carloseltuerto:
Pure refactoring.

For consistency, request_start has also been renamed to stream_start.

Description: Rename request_end to stream_end, and adjust documentation.
Risk Level: None
Testing: CI
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/library/java/io/envoyproxy/envoymobile/engine/types/EnvoyFinalStreamIntel.java
+++ b/library/java/io/envoyproxy/envoymobile/engine/types/EnvoyFinalStreamIntel.java
@@ -7,9 +7,9 @@ package io.envoyproxy.envoymobile.engine.types;
  */
 public interface EnvoyFinalStreamIntel {
   /*
-   * The time the request started, in ms since the epoch.
+   * The time the stream started (a.k.a request started), in ms since the epoch.
    */
-  public long getRequestStartMs();
+  public long getStreamStartMs();
   /*
    * The time the DNS resolution for this request started, in ms since the epoch.
    */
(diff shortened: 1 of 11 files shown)


library/java/io/envoyproxy/envoymobile/engine/types/EnvoyFinalStreamIntel.java
```diff
@@ -51,9 +51,9 @@
    */
   public long getResponseStartMs();
   /*
-   * The time the last byte of the request was received, in ms since the epoch.
+   * The time when the stream reached a final state (Error, Cancel, Success), in ms since the epoch.
```
alyssawilk, line 54: while we're in here, did your last PR correct to unix epoch or are all epochs here the local clock epoch?  If the later would you be willing to comment at the top of each intel block what epoch is because I default to unix.

alyssawilk (approved):
LGTM, with optional semi-related improvement.  Please get cross-company sign off on this one though!

library/java/io/envoyproxy/envoymobile/engine/types/EnvoyFinalStreamIntel.java
carloseltuerto: There is only one definition of Epoch: 1970:01:01?00:00:00GMT.

The "Tick Clock" of a computer, what you named "local clock", is almost never based on Epoch. Typically, the Tick Clock starts at zero when a computer boots up. Trying to "extract" an Epoch timestamp from a Tick Clock is non-sensical. The sad reality is we do have code like this: `monotonicTime().time_since_epoch()`. This makes no sense, but that's used as a trick to retrieve the "long" value representing the number of Ticks.

RyanTheOptimist (approved):
Thanks for doing this! It looks great. Though the format and coverage checks are failing. Do you happen to know why?

carloseltuerto:
The "Format" error was of my making, but all other errors are CI flakiness.

I rerun CI until all passes - the trick I use here it is close the PR and re-open it just after. That triggers a new CI run...

goaway (approved):
Stream seems like a better name to me, thanks!

PR #2046, opened 2022-02-09
ios: add explicit flow control onSendWindowAvailable to public interface

goaway:
Description: Adds the previously Kotlin-only feature to the iOS public interface.
Risk Level: Low
Testing: Manual
Release Notes: Added onSendWindowAvailable callback on iOS for when explicit flow control is enabled. This brings it in line with the Kotlin API.

### DIFF
--- a/library/objective-c/EnvoyHTTPStreamImpl.m
+++ b/library/objective-c/EnvoyHTTPStreamImpl.m
@@ -62,6 +62,17 @@ static void *ios_on_trailers(envoy_headers trailers, envoy_stream_intel stream_i
   return NULL;
 }
 
+static void *ios_on_send_window_available(envoy_stream_intel stream_intel, void *context) {
+  ios_context *c = (ios_context *)context;
+  EnvoyHTTPCallbacks *callbacks = c->callbacks;
+  dispatch_async(callbacks.dispatchQueue, ^{
+    if (callbacks.onSendWindowAvailable) {
+      callbacks.onSendWindowAvailable(stream_intel);
+    }
+  });
+  return NULL;
+}
+
 static void *ios_on_complete(envoy_stream_intel stream_intel,
                              envoy_final_stream_intel final_stream_intel, void *context) {
   ios_context *c = (ios_context *)context;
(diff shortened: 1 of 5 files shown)


jpsim:
In my opinion, we should bump SwiftLint's [line length limit](https://github.com/envoyproxy/envoy-mobile/blob/da9bc5496e384a6de8606233f13254d9016f1556/.swiftlint.yml#L111-L112) from 100 to 120. 100 is quite low considering modern screen resolutions.

jpsim:
Although if we do that, we should probably keep it in sync with the [kotlin linter too](https://github.com/envoyproxy/envoy-mobile/blob/da9bc5496e384a6de8606233f13254d9016f1556/.kotlinlint.yml#L22-L23).

library/objective-c/EnvoyHTTPStreamImpl.m
```diff
@@ -62,6 +62,17 @@
   return NULL;
 }
 
+static void *ios_on_send_window_available(envoy_stream_intel stream_intel, void *context) {
```
jpsim, line 65: Why can't we use types in this interface and avoid the cast below?

```suggestion
static void *ios_on_send_window_available(envoy_stream_intel stream_intel, ios_context *context) {
```

library/swift/StreamPrototype.swift
```diff
@@ -103,6 +103,22 @@ public class StreamPrototype: NSObject {
     return self
   }
 
+  /// Specify a callback for when additional send window becomes available.
```
jpsim, line 106:
```suggestion
  /// Specify a callback to be invoked when an additional send window becomes available.
```

library/objective-c/EnvoyHTTPStreamImpl.m
goaway: Because the address of this function is ultimately assigned to a C function pointer in the bridge layer which is agnostic of what actually goes in here. It's how Java (and C++/Python, and Rust) can all pass through something different here to handle their respective callbacks with their own types and state.

goaway:
> In my opinion, we should bump SwiftLint's [line length limit](https://github.com/envoyproxy/envoy-mobile/blob/da9bc5496e384a6de8606233f13254d9016f1556/.swiftlint.yml#L111-L112) from 100 to 120. 100 is quite low considering modern screen resolutions.

No strong feelings, but I think we chose 100 because it's what the parent project uses.

jpsim (approved):
Capturing this in the version history would be good too.

PR #2047, opened 2022-02-09
bazel: Add custom rule to produce android debug info

keith:
When building the jni dylib for android, we previously stripped all
debug info to decrease the artifact size. With this change we now
produce the same stripped binary as before, but before stripping it we
create a dump of the debug info suitable for crash reporting.

This is made overly difficult for a few reasons:

1. Bazel doesn't support fission for Android https://github.com/bazelbuild/bazel/pull/14765
2. Extra outputs from rules are not propagated up the dependency tree,
   so just building `android_dist` at the top level, isn't enough to get
   the extra outputs built as well
3. Building the library manually alongside the android artifact on the
   command line results in 2 separate builds, one for android as a
   transitive dependency of `android_dist` and one for the host
   platform

This change avoids #1 fission for now, but the same approach could be used
once that change makes its way to a bazel release.

This change fixes #2 by using a separate output group that can be depended
on by the genrule that writes to dist while avoiding #3 because the custom
rule producing these uses the android transition.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -69,7 +69,7 @@ build:tsan-dev --test_env="TSAN_OPTIONS=report_atomic_races=0"
 
 # Exclude debug info from the release binary since it makes it too large to fit
 # into a zip file. This shouldn't affect crash reports.
-build:release-common --define=no_debug_info=1
+build:release-ios --define=no_debug_info=1
 
 # Flags for release builds targeting iOS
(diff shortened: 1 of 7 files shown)


keith:
I'm thinking I can do something to avoid having to pass the extra target, looking at that now

snowp (approved):
Thanks!

Bencodes (approved):

PR #2049, opened 2022-02-10
size: optimizations for dynamic library

Reflejo:
These are the different low hanging fruit fixes to reduce the dynamic library size. Note wins were calculated in isolation against a fixed target for just one architecture so they are not necesarily additive.

- remove api v2 dependency (Mostly done here: https://github.com/envoyproxy/envoy/pull/18817)
    - Estamated wins: -1.8Mi

- compile optimizing for size
    - [commit c2d7084] Estimated wins: -0.66Mi

- set default visibility to hidden
    - [commit ed2975e] Estimated wins: -5.66Mi

- override pgv validations with NOP functions
    - [commit 255eb5a] Estimated wins: -3.7Mi

**For the measured case (a few months back) the size reduction accounted for about -46.2%**

This PR is better reviewed commit by commit.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,5 +1,16 @@
 workspace(name = "envoy_mobile")
 
+# TODO[fz]: Remove once envoy updates PGV
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+http_archive(
+    name = "bazel_gazelle",
+    sha256 = "de69a09dc70417580aabf20a28619bb3ef60d038470c7cf8442fafcf627c21cb",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/bazel-gazelle/releases/download/v0.24.0/bazel-gazelle-v0.24.0.tar.gz",
+        "https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.24.0/bazel-gazelle-v0.24.0.tar.gz",
+    ],
+)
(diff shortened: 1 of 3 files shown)


buildbreaker (dismissed):

WORKSPACE
```diff
@@ -1,5 +1,36 @@
...
+    version = "v0.6.0",
+)
+
+git_repository(
+    name = "com_envoyproxy_protoc_gen_validate",
+    commit = "79071f0f8b04188b297a0517a6e55b2d3641ab5a",
+    remote = "https://github.com/envoyproxy/protoc-gen-validate"
+)
+
+# End of PGV overrides
```
buildbreaker, line 32: Would this be better put in: https://github.com/envoyproxy/envoy-mobile/blob/main/bazel/envoy_mobile_repositories.bzl?

Adding it here wouldn't be able to be used if envoy mobile is a submodule (e.g like a private envoy mobile for a given org/company)

buildbreaker (dismissed):

WORKSPACE
```diff
@@ -1,5 +1,16 @@
 workspace(name = "envoy_mobile")
 
+# TODO[fz]: Remove once envoy updates PGV
```
jpsim, line 3: Specifically, any release including this PR: https://github.com/envoyproxy/protoc-gen-validate/pull/533

PR #2052, opened 2022-02-11
support building swift/obj-c targets on macos

snowp:
Right now the swift/obj-c targets cannot build on their own, relying on being wrapped in a ios_test or similar in order to
build with the right libraries. This makes some few surgical changes in order to let the targets build on their own. This opens up for running swift_binaries etc. that consume EM, with a few of the iOS specific bits disabled.

Risk Level: Low, macos only
Testing: Verified that we can build swift_binaries that target 
```
        "@envoy_mobile//library/swift:ios_lib",
        "@envoy_mobile//library/objective-c:envoy_engine_objc_lib",
```
on macos
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/library/objective-c/BUILD
+++ b/library/objective-c/BUILD
@@ -20,9 +20,14 @@ objc_library(
         "EnvoyHTTPStreamImpl.m",
         "EnvoyLogger.m",
         "EnvoyNativeFilterConfig.m",
-        "EnvoyNetworkMonitor.m",
         "EnvoyStringAccessor.m",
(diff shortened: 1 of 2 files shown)


keith (dismissed):

keith (approved):

jpsim:
This broke reachability on iOS too. Fix here is to enable it for both iOS & macOS: https://github.com/envoyproxy/envoy-mobile/pull/2114

PR #2053, opened 2022-02-11
symbols: conditionally create symbols map

Reflejo:


BUILD
```diff
@@ -59,7 +59,7 @@ genrule(
     set -- $(SRCS)
     chmod 755 $$1
     chmod 755 $$2
-    chmod 755 dist/*.objdump
+    chmod 755 -f dist/*.objdump || true
```
keith, line 62: @snowp changed this internally to something else, we should copy that instead

library/kotlin/io/envoyproxy/envoymobile/BUILD
```diff
@@ -10,9 +10,10 @@ android_artifacts(
     android_library = ":envoy_lib",
     archive_name = "envoy",
     manifest = "EnvoyManifest.xml",
-    native_deps = [
-        "//library/common/jni:libenvoy_jni.so.debug_info",
-    ],
+    native_deps = select({
+        "//bazel:is_symbols_map_enabled": ["//library/common/jni:libenvoy_jni.so.debug_info"],
```
keith, line 14: instead of a new conditional could we just use `opt_build`?

BUILD
Reflejo: copied

BUILD
```diff
@@ -59,11 +59,16 @@ genrule(
...
     chmod 755 $$1
     chmod 755 $$2
-    chmod 755 dist/*.objdump
     cp $$1 dist/envoy.aar
     cp $$2 dist/envoy-pom.xml
     shift 2
-    cp $$@ dist/
+
+    mkdir -p dist/symbols
+    if [[ -f "$$@" ]]; then
```
keith, line 67: for multiple archs this is multiple files so this part is invalid bash I think? `if [[ -f "foo.symbols bar.symbols" ]]`

keith (approved):

PR #2057, opened 2022-02-14
pom: Update developers in pom xml file

jpsim:
To keep in sync with OWNERS.


jpsim:
ptal @buildbreaker

buildbreaker (approved):

PR #2058, opened 2022-02-15
Have QuicTestServer to support Cronet request

carloseltuerto:
Attempts to mimic this: https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/testing/test_server/data/quic_data/simple.txt
The original QuicTestServer reads that file and caches it while booting up, and precomputes the full response out of it.

Description: provide a canned response to a specific request.
Risk Level: N/A - only changes test infrastructure
Testing: CI
Docs Changes: N/A
Release Notes: N/A


carloseltuerto:
I wish. But then I would need to configure Cronvoy in such a way that it would be able to establish an H3 connection to a an H3 test server.
Today Cronvoy uses the default EM Client Configuration. And then, I would need a way to tell the EM Client to "fake" the certificate verification.
I don't know yet how to do that.

alyssawilk (approved):

Issue #332, opened 2019-08-15
core: add multiple engine support


goaway:
The skeleton for this is place, but it needs to be implemented.

rebello95:
Note that when we have support for this, we should consolidate some of the iOS integration tests into one file/case to exercise this functionality: https://github.com/envoyproxy/envoy-mobile/pull/1298

junr03:
Moving context from #2003: 

Per @carloseltuerto: 

> Today, due to some EnvoyProxy limitations, Envoy-Mobile can not run more than one Engine at a given moment. For the YouTube App, only one Engine is not viable - it needs at least two concurrent Engines.

> TODO: upstream tests with tear down of A first, tear down of B first, different logging paths.

junr03:
Per @alyssawilk's [comment](https://github.com/envoyproxy/envoy-mobile/issues/2003#issuecomment-1055772742):

> Ok @goaway @junr03 the upstream part is "done enough" insofar as if you bump main it should work (and if not, it definitely works with envoy.restart_features.no_runtime_singleton = true, as that's what is regression tested)

junr03:
This PR https://github.com/envoyproxy/envoy-mobile/pull/2085 pulls in all of @alyssawilk's upstream changes.

alyssawilk:
Oh nice, this one has the details of what needs to be done.  Thanks for  filing it @goaway and digging it up, @junr03

PR #2059, opened 2022-02-15
Fix iOS termination crash in `ProvisionalDispatcher`

jpsim:
When iOS is terminating a process, it may release some static memory, which could cause the `event_dispatcher_` variable to point to invalid memory, leading to a crash.

A crash that occurs when an app is already in the process of being terminated isn't too bad, considering it's already shutting down, but this matters to some extent, considering that crash reporters will report this as an app crash, and the OS may choose to penalize your app by not proactively suggesting it.

This should fix a crash with a backtrace such as this one:

```
EXC_BAD_ACCESS: Attempted to dereference garbage pointer 0x8000000000000010.

0 Envoy::Event::ProvisionalDispatcher::post(std::__1::function<void ()>)
1 Envoy::EngineHandle::runOnEngineDispatcher(long, std::__1::function<void (Envoy::Engine&)>)
2 start_stream
3 -[EnvoyHTTPStreamImpl initWithHandle:callbacks:explicitFlowControl:]
4 -[EnvoyEngineImpl startStreamWithCallbacks:explicitFlowControl:]
5 StreamPrototype.start(queue:)
```

This is a follow up to https://github.com/envoyproxy/envoy-mobile/pull/2056 after discussion with Snow.


jpsim:
Same caveat as #2056 applies here: I haven't been able to reproduce this crash, nor validate that this would actually fix it. I'm just making an educated guess based on the backtrace at the time of the crash.

goaway:
Thanks @jpsim!

A couple thoughts:
This looks like it's probably not threadsafe, and races seem likely in these sorts of circumstances.
Would this state make sense to be contained within the Dispatcher instead of the Engine? After all, we return the result of the `post` which could be `false` if things are shut down. (This is partly why `post` returns a value.)

jpsim:
> Would this state make sense to be contained within the Dispatcher instead of the Engine?

Does the dispatcher have any way to detect if we've terminated the engine?

goaway:
> > Would this state make sense to be contained within the Dispatcher instead of the Engine?
> 
> Does the dispatcher have any way to detect if we've terminated the engine?

It could be explicitly informed. Let me know if you'd like to chat about this change synchronously.

jpsim:
Something like this is needed for multi-engine support (#332).

library/common/engine.cc
```diff
@@ -175,6 +175,8 @@ envoy_status_t Engine::terminate() {
     main_thread_.join();
   }
 
+  dispatcher_->terminate();
```
snowp, line 178: Should this be before the join? I'm not that familiar with this but isn't the dispatcher running on main_thread_, so if we're calling it here then the dispatcher is already not running?

library/common/event/provisional_dispatcher.h
```diff
@@ -69,6 +74,7 @@ class ProvisionalDispatcher : public ScopeTracker {
   std::list<Event::PostCb> init_queue_ GUARDED_BY(state_lock_);
   Event::Dispatcher* event_dispatcher_{};
   Thread::ThreadSynchronizer synchronizer_;
+  bool terminated_ GUARDED_BY(state_lock_){};
```
snowp, line 77: Doesn't matter a ton but if you made this atomic it could be checked outside of the lock

snowp (commented):
Few questions but conceptually this makes sense I think

I would think that we could add a simple test that verifies that once we call terminate we no longer delegate to the inner dispatcher

library/common/event/provisional_dispatcher.cc
```diff
@@ -12,6 +12,12 @@ void ProvisionalDispatcher::drain(Event::Dispatcher& event_dispatcher) {
   // because of behavioral oddities in Event::Dispatcher: event_dispatcher_->isThreadSafe() will
   // crash.
   Thread::LockGuard lock(state_lock_);
+
+  // Don't perform any work on the dispatcher if marked as terminated.
+  if (terminated_) {
+    return;
+  }
```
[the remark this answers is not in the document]
jpsim: I'm not sure, but it seems safest to check here considering engine termination is exposed in the public API so a user could invoke it at any time.

We've also already acquired the lock here, which is the expensive part, the boolean check should add effectively no overhead.

jpsim:
> I would think that we could add a simple test that verifies that once we call terminate we no longer delegate to the inner dispatcher

I'll take a stab at writing a test that does this, but for what it's worth the existing tests fail with https://github.com/envoyproxy/envoy-mobile/pull/2129 without this, so it's soon to be implicitly covered by our existing tests.

library/common/event/provisional_dispatcher.h
jpsim: Considering that this only needs to be checked in places where we've already acquired the lock, it seems we should leverage that.

library/common/engine.cc
jpsim: I don't know. I can move it and see if the tests in https://github.com/envoyproxy/envoy-mobile/pull/2129 still pass.

jpsim: Actually, this causes a data race, caught by the TSan tests: https://github.com/envoyproxy/envoy-mobile/runs/6026097993?check_suite_focus=true

```
WARNING: ThreadSanitizer: data race (pid=22)
  Read of size 8 at 0x7b5400021280 by main thread:
    #0 memcpy ??:? (main_interface_test+0x19060ae)
    #1 Envoy::Thread::ThreadId::isEmpty() const ??:? (main_interface_test+0x1b9d8ee)
    #2 Envoy::Event::DispatcherImpl::runFatalActionsOnTrackedObject(std::__cxx11::list<std::unique_ptr<Envoy::Server::Configuration::FatalAction, std::default_delete<Envoy::Server::Configuration::FatalAction> >, std::allocator<std::unique_ptr<Envoy::Server::Configuration::FatalAction, std::default_delete<Envoy::Server::Configuration::FatalAction> > > > const&) const ??:? (main_interface_test+0x3eac80c)
```

Reverting.

snowp (approved):
I'm not convinced this solves it 100% (since we set this after we call `event_dispatcher_.exit()`, it feels racy), but will definitely make things better so this works for me

jpsim:
I’m also not convinced the termination crash will be completely fixed but I do think this will help.

PR #2062, opened 2022-02-16
Fix `bump_support_rotation` GitHub Action

jpsim:
Failed with the following error when I ran it manually a few minutes ago:

```
fatal: The current branch support-bump-1854030781 has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin support-bump-1854030781
```

https://github.com/envoyproxy/envoy-mobile/runs/5219745723?check_suite_focus=true


jpsim:
Still not enough. Will troubleshoot in https://github.com/envoyproxy/envoy-mobile/pull/2063 before opening a PR to fix.

PR #2065, opened 2022-02-18
Fix bump_support_rotation GitHub Action

jpsim:
Failed with the following error when I ran it manually a few days ago:

```
fatal: The current branch support-bump-1854030781 has no upstream branch.
To push the current branch and set the remote as upstream, use

    git push --set-upstream origin support-bump-1854030781

https://github.com/envoyproxy/envoy-mobile/runs/5219745723?check_suite_focus=true

Validated [here](https://github.com/jpsim/gh_action_test/blob/main/.github/workflows/pr_test.yml).


jpsim:
Sadly something's still broken: https://github.com/envoyproxy/envoy-mobile/runs/5254292662?check_suite_focus=true

image

PR #2068, opened 2022-02-18
happy eyeballs: fix missing runtime configuration

goaway:
Description: In addition to needing the DNS lookup family to be set to ALL, it turns out Happy Eyeballs also requires a runtime flag (`envoy.reloadable_features.allow_multiple_dns_addresses`) to be enabled to function.
Risk Level: Moderate
Testing: Local & CI

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -39,6 +39,7 @@ const std::string config_header = R"(
 - &dns_fail_max_interval 10s
 - &dns_query_timeout 25s
 - &dns_lookup_family V4_PREFERRED
+- &dns_multiple_addresses false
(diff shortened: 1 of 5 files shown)


jpsim (approved):
+1 as long as there's no reason we'd want to be able to enable the `dns_multiple_addresses` feature independently of happy eyeballs. We can add that later if we eventually want that anyway.

goaway:
> +1 as long as there's no reason we'd want to be able to enable the `dns_multiple_addresses` feature independently of happy eyeballs. We can add that later if we eventually want that anyway.

Yep, we can just change this later if we discover we want to configure it separately.

goaway:
Thanks @jpsim!

PR #2075, opened 2022-02-28
ci: create baseline and experimental test app pipelines

goaway:
Description: Creates distinct test app profiles for end-to-end testing of baseline and new/experimental features in CI on Swift and Kotlin. Previously only a baseline configuration was run in CI which resulted in blindspots, especially with respect to upstream changes.
Risk Level: Low
Testing: Local & CI


Augustyniak (commented):
I like the general idea of us testing how the engine behaves when all/some of its features are enabled.

So just to make sure that you know where I am coming from. As I understood that when we were talking about these changes (back when you proposed them originally) the idea behind them was to "continue to make sure that the engine continues to perform well even if all of its feature are enabled". 

Some of the thoughts after a brief look at the PR:

1. It adds a lot of complexity/maintenance burden as we introduce another copy(ies) of filters and some other components and some of the upstream changes will require us to update all of these copies.
2. It may add to the confusion for people who want to contribute to Envoy Mobile. We have 3(?) example apps per platform now as opposed to one and that may lead to a lot of questions. If we decide to stick with this approach we should add some documents that would explain why we need all of these example apps (and to make people update them as they keep adding features?)
3. Couldn't we just add a few integration tests (one with all feature disabled/one with all features enabled) that would start the engine and perform a requests (or a few of them) to see whether EM works well? Wouldn't that give us exactly the same level of confidence as all of these newly introduced example apps do? At the same time, with integration tests approach we could cut significantly reduce the amount of added code (and using integration tests to verify things seems to be a more common practice than using custom example apps).


1) and 2) are the issues that I see with the currently proposed approach. 3) is a question regarding an alternative that I can see.

Augustyniak:
After talking with @goaway offline:

- @goaway is going to open a GH issue that will track the future work for simplifying (removing) code from new CI example apps. We are going to remove some http filters, simplify apps' ViewControllers etc.
- we are going to add some documentation that would explain what's the intent behind these new CI example apps and how they should be updated when somebody add a new feature to Envoy Mobile.
- new CI example apps are going to be moved out of `examples` directory

Augustyniak (approved):
Let's make sure that we do simplify newly added soon. I will advocate for removing these if we see that they add too much of maintenance burden.

PR #2076, opened 2022-03-01
envoy: 3267361cedf936

junr03:
Signed-off-by: José Ulises Niño Rivera <junr03@users.noreply.github.com>

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -119,6 +119,11 @@ build:remote-ci-common --spawn_strategy=remote,sandboxed,local
 # sends a keepalive if we don't receive any messages (server updates every 60s).
 build:remote-ci-common --grpc_keepalive_time=61s
 #############################################################################
+# remote-ci-local-java8: Use a local Java 8 (JDK)
+#############################################################################
+build:remote-ci-local-java8 --java_runtime_version=local_jdk
+build:remote-ci-local-java8 --java_language_version=8
+#############################################################################
 # remote-ci-linux: These options are linux-only using GCC by default
 #############################################################################
 # Common Engflow flags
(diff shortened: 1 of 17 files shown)


goaway (approved):

junr03:
@alyssawilk I am seeing the swift tests fail with 

```
Error: 3-02 00:07:17.506][129825][error][envoy_bug] [external/envoy/source/common/runtime/runtime_features.cc:85] envoy bug failure: false. Details: Unable to find runtime feature envoy.restart_features.use_apple_api_for_dns_lookups
```

I think it is related to https://github.com/envoyproxy/envoy/pull/19880 trying to understand if something should change in order to succeed. But if I understand your PR correctly, it should have been a no-op from the user perspective

alyssawilk:
oh nooooo.  Ironically I'd almost started out with a map of flag name to flag, but found I could avoid that with the reflection code.  Worst case I can add and maintain that map for a bit, but if it's Ok let me check in with the absl team and see if we can sort out what's broken here.

alyssawilk:
absl team has awesome turnaround!
"Yes. Flag literals are compiled out by default on mobile platforms.  
This can be changed with build flag.
-DABSL_FLAGS_STRIP_NAMES=0"

junr03:
@alyssawilk amazing! That is exactly what it looked like. Let me update the build to reflect that.

alyssawilk:
Thank you, and sorry about the breakage!  Even 15 years at Google that was a new one to me :-/

.bazelrc
```diff
@@ -24,6 +24,9 @@ build --define cxxopts=-std=c++17
 build --features=-per_object_debug_info
 # Suppress deprecated declaration warnings due to extensive transitive noise from protobuf.
 build --copt -Wno-deprecated-declarations
+# Abseil Flag literals are compiled out by default on mobile platforms. These flags are needed for
+# Envoy's runtime system. This option prevents them from being compiled out.
+build --copt -DABSL_FLAGS_STRIP_NAMES=0
```
jpsim, line 29: Should this be defined upstream instead? Fine here for now.

jpsim: Yes, I think it makes most sense to land here first so we can unblock this update, then remove it once it's available upstream.

library/common/extensions/retry/options/network_configuration/predicate.cc
```diff
@@ -32,12 +32,19 @@ NetworkConfigurationRetryOptionsPredicate::updateOptions(
     return Upstream::RetryOptionsPredicate::UpdateOptionsReturn{absl::nullopt};
   }
 
-  auto& extra_stream_info =
+  StreamInfo::ExtraStreamInfo* extra_stream_info =
       filter_state->getDataMutable<StreamInfo::ExtraStreamInfo>(StreamInfo::ExtraStreamInfo::key());
 
+  if (extra_stream_info == nullptr) {
```
jpsim, line 38: Nit:

```suggestion
  if (!extra_stream_info) {
```

jpsim (approved):
I didn't look at the upstream changes. I'm assuming you have.

library/common/extensions/retry/options/network_configuration/predicate.cc
junr03: I did this to kick CI 😂 while we were having all those issues yesterday. I'll leave it in because I dont want to run CI again. But will update in a subsequent PR

jpsim: Not a problem to leave as is.

PR #2084, opened 2022-03-03
requirements: move iOS requirement to 12.0+

junr03:
Signed-off-by: José Ulises Niño Rivera <junr03@users.noreply.github.com>


jpsim:
We should check with our known active users to make sure they don't still support iOS 11.

If we do this, we can clean up some availability checks like this one: https://github.com/envoyproxy/envoy-mobile/blob/1c4128c2a7aa1c890cfeb71aafa466975b569565/library/objective-c/EnvoyEngineImpl.m#L445

junr03:
@jpsim sounds good. I added to next week's community call agenda. Will wait to move on this until then.

jpsim:
Other places to change: https://cs.github.com/envoyproxy/envoy-mobile?q=11.0

jpsim (approved):
I think we can merge this now. I made the remaining changes I called out (hope you don't mind) and no one has objected in the [mailing list](https://groups.google.com/g/envoy-mobile-announce/c/k5-71EGWMKs), [Slack post](https://envoyproxy.slack.com/archives/CKQ2LK23G/p1646941463278129) or community call.

junr03:
@jpsim ok! Sounds good.

PR #2087, opened 2022-03-07
routing/cluster changes: move cluster selection to Envoy's routing table

junr03:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -46,7 +46,7 @@ const std::string config_header = R"(
 - &enable_interface_binding false
 - &h2_connection_keepalive_idle_interval 100000s
 - &h2_connection_keepalive_timeout 10s
-- &h2_hostnames []
+- &h2_hostnames ["api.foo.bar"]
 - &metadata {}
 - &stats_domain 127.0.0.1
 - &stats_flush_interval 60s
@@ -239,13 +239,15 @@ static_resources:
               routes:
 #{custom_routes}
               - match:
+                  prefix: "/"
                   headers:
                   - name: ":scheme"
-                    value: http
+                    string_match:
+                      exact: http
                 request_headers_to_remove:
                 - x-forwarded-proto
                 route:
-                  cluster: base
+                  cluster: base_clear
                   timeout: 0s
                   retry_policy:
                     per_try_idle_timeout: *per_try_idle_timeout
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -167,7 +167,6 @@ const char* config_template = R"(
                 direct_response: { status: 404, body: { inline_string: "not found" } }
                 request_headers_to_remove:
                 - x-forwarded-proto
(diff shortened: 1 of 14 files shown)


library/common/config/config.cc
```diff
@@ -75,7 +76,7 @@ const std::string config_header = R"(
       address:
         socket_address: { address: *statsd_host, port_value: *statsd_port }
 
-!ignore protocol_defs: &http1_protocol_options_defs
+!ignore http1_protocol_options_defs: &http1_protocol_options
```
junr03, line 79: @goaway how come this is in the header but `base_protocol_options_defs` is in the config_template?

library/common/http/client.cc
```diff
@@ -585,53 +584,8 @@ void Client::removeStream(envoy_stream_t stream_handle) {
 }
 
 namespace {
-
-const LowerCaseString ClusterHeader{"x-envoy-mobile-cluster"};
 const LowerCaseString H2UpstreamHeader{"x-envoy-mobile-upstream-protocol"};
```
junr03, line 587: @goaway on Friday we discussed having:
1. :scheme dictate if the request is routed to non-tls http1.1
2. route specifically to h2 if the hostname is included.
3. route everything else to alpn.

Looking at the code that uses `x-envoy-mobile-upstream-protocol` it seems that we have the gRPC platform bindings force routing to h2. I can preserve that functionality. But do we want to allow for _any_ request to force routing via `x-envoy-mobile-upstream-protocol`?

library/objective-c/EnvoyConfiguration.m
```diff
@@ -100,6 +102,18 @@ - (nullable NSString *)resolveTemplate:(NSString *)templateYAML {
         appendString:[[NSString alloc] initWithUTF8String:route_cache_reset_filter_insert]];
   }
 
+  BOOL hasH2Hostnames = self.h2Hostnames.count > 0;
+  if (hasH2Hostnames) {
```
junr03, line 106: TODO: actually build this

junr03:
@goaway Alternatively, we could start with a PR that leaves all the `x-envoy-mobile-cluster` and `x-envoy-mobile-upstream-protocol` logic alone, and only adds the forced h2 cluster routing.

library/common/config/config.cc
goaway: The header defines defaults. Overrides are injected between the header and the template, so any objects that contain internal values that you want to be overridable must be defined in the template.

library/common/http/client.cc
goaway: I think I might remove the protocol-based mapping (since it's incomplete anyways) and just leave in the cluster override header and use that for the gRPC case.

junr03:
Closed in favor of https://github.com/envoyproxy/envoy-mobile/pull/2088

PR #2098, opened 2022-03-14
[Android] Replace uses of `StringJoiner` with `StringBuilder`

jpsim:
Description: [StringJoiner](https://developer.android.com/reference/java/util/StringJoiner) was added in Android API level 24 but we still support back to API level 21, so this results in `NoClassDefFoundError` errors on devices running older Android versions.

Risk Level: Medium, this code is simple and unit tested.
Testing: Existing unit tests
Docs Changes: None
Release Notes: Added

### DIFF
--- a/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
+++ b/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
@@ -4,7 +4,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 import java.util.regex.Matcher;
-import java.util.StringBuilder;
+import java.lang.StringBuilder;
 import javax.annotation.Nullable;
 
 import io.envoyproxy.envoymobile.engine.types.EnvoyHTTPFilterFactory;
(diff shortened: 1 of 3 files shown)


library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -160,20 +160,28 @@ String resolveTemplate(final String templateYAML, final String platformFilterTem
 
     String dnsFallbackNameserversAsString = "[]";
     if (!dnsFallbackNameservers.isEmpty()) {
-      StringJoiner sj = new StringJoiner(",", "[", "]");
+      StringBuilder sb = new StringBuilder("[");
```
Reflejo, line 163: Can we use `TextUtils` instead? https://developer.android.com/reference/android/text/TextUtils#join(java.lang.CharSequence,%20java.lang.Object[])

jpsim: This package doesn't currently depend on android packages:

```
library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java:3: error: package android.text does not exist
import android.text.TextUtils;
                   ^
```

Augustyniak (approved):
I can see that we have unit tests that verify that the config looks good when we have ONE h2 raw domain or one DNS fallback domain but it would probably be good to add tests for cases when we have multiple domains too. Consider doing it as part of this PR.

Augustyniak (approved):
Great, thank you.

PR #2103, opened 2022-03-15
[iOS/macOS] Remove `EngineBuilder.enableNetworkPathMonitor(_:)`

jpsim:
**Description**

We ran some experiments in the Lyft apps and found no regressions with this enabled, so we believe it's safe to remove the legacy `SCNetworkReachability` implementation and always use the `NWPathMonitor` code path.

Also enable path monitor reachability state observation on macOS while I'm in there.

It was disabled in https://github.com/envoyproxy/envoy-mobile/pull/2052 because we didn't have a minimum macOS version set but this API is available as of 10.14.

**Risk Level:** Medium. It's possible the previous `SCNetworkReachability` implementation performs better under some circumstances but if that's the case we weren't able to find them.
**Testing:** Ran pre-release and release experiments in Lyft's production environments
**Docs Changes:** Updated
**Release Notes:** Updated

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -9,6 +9,7 @@ build --features=debug_prefix_map_pwd_is_dot
 build --features=swift.cacheable_swiftmodules
 build --features=swift.debug_prefix_map
 build --host_force_python=PY3
+build --macos_minimum_os=10.14
 build --ios_minimum_os=12.0
 build --ios_simulator_device="iPhone 13"
 build --ios_simulator_version=15.2
--- a/docs/root/api/starting_envoy.rst
+++ b/docs/root/api/starting_envoy.rst
@@ -369,25 +369,6 @@ Specify a closure to be called by Envoy to access arbitrary strings from Platfor
   // Swift
   builder.addStringAccessor(name: "demo-accessor", accessor: { return "PlatformString" })
 
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-``enableNetworkPathMonitor``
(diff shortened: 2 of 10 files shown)


Augustyniak (commented):
@jpsim Thank you for the follow up. I would prefer if we postponed this merge for the time being. 

As you mentioned in your PR description we have not found any circumstances under which the old API `SCNetworkReachability` performs better than `NWPathMonitor` one does but since we are about to start experimenting with interface binding feature which relies heavily on values returned by interface selection logic I would like to derisk ourselves here and keep the method for the time being (in case we decide to run interface binding with `SCNetworkReachability` API enabled - that's unlikely at this point I think but still). 

We should know more about the state of interface binding within the next few weeks - if all looks good in there then we can remove the `enableNetworkPathMonitor` method. What do you think? I do not see any big downsides of keeping this method alive for another few weeks.

jpsim:
> We should know more about the state of interface binding within the next few weeks - if all looks good in there then we can remove the `enableNetworkPathMonitor` method. What do you think? I do not see any big downsides of keeping this method alive for another few weeks.

Yes, I'm in no rush to merge this.

jpsim:
As was pointed out in https://github.com/envoyproxy/envoy-mobile/pull/2114 & https://github.com/envoyproxy/envoy-mobile/pull/2116, this functionality was broken since February 11, so the experiment we ran was invalid and we will need to run it again.

Closing as we need to keep this around until that concludes.

PR #2104, opened 2022-03-15
ios: add support for toggling trust chain verification

goaway:
Description: Adds feature parity to the Android/JVM engine builder, and allows TLS trust chain verification to be bypassed.
Risk Level: Low
Testing: Unit & CI

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -110,6 +110,9 @@ trailing_comma:
   mandatory_comma: true
 line_length:
   - 100
+type_body_length:
+  - 300 # warning
+  - 400 # error
 file_length:
   warning: 300
   ignore_comment_only_lines: true
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -111,8 +111,8 @@ trailing_comma:
 line_length:
   - 100
 type_body_length:
-  - 300 # warning
-  - 400 # error
+  warning: 300
+  error: 400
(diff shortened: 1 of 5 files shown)


goaway:
Note this currently differs slightly from the Android/Java API, which exposes a binary enumeration to represent trust verification status (which matches Envoy internal configuration). I think this public API is simpler, and if folks agree, propose updating the Java side to this model as well.

jpsim:
> I think this public API is simpler, and if folks agree, propose updating the Java side to this model as well.

I agree, unless there's any possibility of adding a 3rd option in the future, at which point an enum could be both more descriptive and more extensible.

goaway:
> This looks mostly good to me, but I wonder if we should have any guardrails to prevent our users from accidentally turning this on?
It defaults to verified+enforced, which I think should be safe enough right now. I've considered bucketing some of these features under an "Advanced/Experimental" mechanism previously, and if we make that distinction, then this could go there.

> I agree, unless there's any possibility of adding a 3rd option in the future, at which point an enum could be both more descriptive and more extensible.
My intuition is it's unlikely we'll expose a third option here even if one is added internally (some configuration is advanced enough that it probably warrants manual configuration).

carloseltuerto:
Thanks Mike!

FYI: For Java, I picked an Enum for readability purposes, even if I don't foresee that there would be more values.

PR #2105, opened 2022-03-15
update envoy to 82b1f66

snowp:
Signed-off-by: Snow Pettersen <snowp@lyft.com>

Description:
Risk Level: Medium
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

PR #2107, opened 2022-03-16
envoy: c911818721586

junr03:
Includes https://github.com/envoyproxy/envoy/pull/19758 so we now need to set the config option.

Will wait until @snowp resolves the upstream break in https://github.com/envoyproxy/envoy-mobile/pull/2105 to land this

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -326,6 +326,7 @@ R"(
         dns_cache_config: *dns_cache_config
     transport_socket: *base_tls_socket
     upstream_connection_options: &upstream_opts
+      set_local_interface_name_on_upstream_connections: true
       tcp_keepalive:
         keepalive_interval: 5
         keepalive_probes: 1


library/common/config/config.cc
```diff
@@ -326,6 +326,7 @@ R"(
         dns_cache_config: *dns_cache_config
     transport_socket: *base_tls_socket
     upstream_connection_options: &upstream_opts
+      set_local_interface_name_on_upstream_connections: true
```
jpsim, line 329: How much of a perf impact does this have for EM use cases compared to turning it off? Would it be worth offering as a config option or turning off when log level is set to some values?

junr03: https://github.com/envoyproxy/envoy-mobile/issues/2111 good point. Alternatively there is further work in upstream envoy that can make this functionality more performant. 

For this PR though I'm biased to leave it on by default as that has been the case for all experiments we've run in the last quarter.

jpsim: Sure, I'm ok with that too, but was curious if you had measured the impact on us.

PR #2114, opened 2022-03-18
[iOS/macOS] Fix reachability

jpsim:
Enable path monitor reachability state observation on both iOS and macOS.

It was disabled in https://github.com/envoyproxy/envoy-mobile/pull/2052 because we didn't have a minimum macOS version set but this API is available as of 10.14.

It seems the `@platforms//os:macos` condition meant that the host was macOS, not that the compilation target was macOS.

Risk Level: Low, re-enabling something that was enabled until fairly recently.
Testing: Validated that preferred network changes are logged when toggling WiFi on/off in the Swift example apps on a physical iPhone. Also confirmed that the library still builds for macOS with `./bazelw build //library/objective-c:envoy_engine_objc_lib --config=macos`.
Release Notes: Added.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -9,6 +9,7 @@ build --features=debug_prefix_map_pwd_is_dot
 build --features=swift.cacheable_swiftmodules
 build --features=swift.debug_prefix_map
 build --host_force_python=PY3
+build --macos_minimum_os=10.14
 build --ios_minimum_os=12.0
 build --ios_simulator_device="iPhone 13"
 build --ios_simulator_version=15.2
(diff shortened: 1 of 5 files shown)


Augustyniak (approved):

jpsim:
Even looking at the [logs for the iOS example apps in CI](https://github.com/envoyproxy/envoy-mobile/runs/5606340890?check_suite_focus=true), you see the expected reachability log:

```
2022-03-18 20:36:42.614775+0000 app[15355:44507] [Envoy] setting preferred network to WLAN
```

Issue #2077, opened 2022-03-01
Envoy prefs-based key-value store


alyssawilk:
both https://github.com/envoyproxy/envoy-mobile/issues/1587 and https://github.com/envoyproxy/envoy-mobile/issues/1587 have limited utility while they only cache to the in-memory key value store.

Upstream we have a file based key-value store (https://github.com/envoyproxy/envoy/blob/main/source/extensions/key_value/file_based/config.cc) but for android we need 
1) an API to write strings to and read strings from the android prefs file and
2) a key value store which uses that API to load and flush from prefs

We've discussed this one at the weekly meeting, but there was no tracking bug so filing one now :-)

PR #2120, opened 2022-03-21
key_value: structure for prefs based key value store

alyssawilk:
Part of https://github.com/envoyproxy/envoy-mobile/issues/2077


goaway:
Thanks for creating this @alyssawilk. Instead of callling it "PrefsBased", I'd suggest the name "PlatformKeyValueStore" - this is what we've named other components that have a platform-level implementation that they delegate to.

This will need a configurable string identifier for the platform API - this is how we discover platform APIs at runtime.

goaway:
Thanks for creating this @alyssawilk. Instead of callling it "PrefsBased", I'd suggest the name "PlatformKeyValueStore" - this is what we've named other components that have a platform-level implementation that they delegate to.

This will need a configurable string identifier for the platform API - this is how we discover platform APIs at runtime.

alyssawilk:
renamed.  can you tell me a bit more about the string identifier, or point to examples?

alyssawilk:
ping!

goaway:
PlatformBridgeFilters are maybe the best (if also somewhat complicated) example of this.

See:
https://github.com/envoyproxy/envoy-mobile/blob/a1f5c57a7ee7a4eb3be8592963c1cb1218436c2b/library/common/extensions/filters/http/platform_bridge/filter.cc#L73

An C API that represents the implementation is defined and and the actual delegating calls are registered (as part of engine initialization) in a global map called the Api::Registry. Lookup of the implementing calls occurs using a unique name for the filter implementation (there are obviously some runtime shortcomings here, but this can all be improved).

Filters are complicated, because a platform-level object actually gets instantiated each time the Envoy filter chain does. For a durable store, this is probably unnecessary and the same implementation can be re-used for all calls.

goaway:
At the end of the day, the important thing is a token that can be specified in config, which can then be used both to register the API during runtime initialization, and to look up the API in compiled extension code.

Having thought about it more, I suppose we could use a single static token for this, but the parallel that struck me with filters is that you might actually have or want multiple implementations of this (e.g. preferences vs. secure storage).

alyssawilk:
IMO it'd be really helpful to have a single static token for now, since the factory that creates the k-v store doesn't have access to the hooks for up calls.  or we could add that string to the prefs proto, if folks want to customize.  I think that'd be part of your part of the implementation though?

goaway:
@alyssawilk All it would need though is a single string field in its config proto though, right?

library/common/extensions/key_value/platform/config.h
```diff
@@ -0,0 +1,62 @@
...
+#include "library/common/extensions/key_value/platform/platform.pb.validate.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace KeyValue {
+
+class PlatformInterface {
+public:
+  virtual ~PlatformInterface() {}
+  virtual void flush(const std::string& key, const std::string& contents) PURE;
```
goaway, line 15: nit: I would suggest one of `store`, `save`, `write` or `put` as opposed to `flush` probably.

library/common/extensions/key_value/platform/config.cc
```diff
@@ -0,0 +1,52 @@
...
+                                          ProtobufMessage::ValidationVisitor& validation_visitor,
+                                          Event::Dispatcher& dispatcher, Filesystem::Instance&) {
+  const auto& typed_config = MessageUtil::downcastAndValidate<
+      const ::envoy::config::common::key_value::v3::KeyValueStoreConfig&>(config,
+                                                                          validation_visitor);
+  const auto file_config = MessageUtil::anyConvertAndValidate<
+      envoymobile::extensions::key_value::platform::PlatformKeyValueStoreConfig>(
+      typed_config.config().typed_config(), validation_visitor);
+  auto milliseconds =
+      std::chrono::milliseconds(DurationUtil::durationToMilliseconds(file_config.flush_interval()));
```
goaway, line 43: While I understand the value in supporting periodic flushes, would it make sense for the base case to be synchronous/immediate?

alyssawilk: I don't think we want to for any of the current use cases - flushing on every dns resolution or every response headers seems expensive.  @RyanTheOptimist for thoughts

RyanTheOptimist: Yeah for a per-request or per-DNS resolution store, immediate sync would be too expensive. Chrome's analogous system uses periodic writes, fwiw. I could imagine that we might have other use cases for the key-value store for which immediate write might be desirable so maybe there is value in having that as an option. But for the uses we currently have, periodic flushes make sense to me.

goaway: Okay that's fine - you all are the primary users of this right now anyways. I just thought it might make for easier testing and/or open up other simple use cases that aren't as perf sensitive.

goaway (approved):
Thanks @alyssawilk!

PR #2122, opened 2022-03-21
prevent dns refresh if network if is unchanged

Reflejo:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


library/java/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
```diff
@@ -105,11 +106,13 @@ public void onReceive(Context context, Intent intent) {
 
   private void handleNetworkChange() {
     NetworkInfo networkInfo = connectivityManager.getActiveNetworkInfo();
-    if (networkInfo == null) {
-      AndroidJniLibrary.setPreferredNetwork(ENVOY_NET_GENERIC);
-      return;
+    int networkType = networkInfo == null ? -1 : networkInfo.getType();
+    if (networkType == previousNetworkType) {
```
goaway, line 110: Should we have this check on iOS as well?

Reflejo: We can. The new network monitor path is easy. The other one is a bit more painful given how it is set up. I'll add it to the new path (done [here](https://github.com/envoyproxy/envoy-mobile/pull/2122/commits/5e3fdc7420c64089b188282be25d26f8030d2d36)). 

Having said that, this is a bigger problem on android than iOS given how often we call `handleNetworkChange`

jpsim (approved):
+1 strongly considering Rafal's suggestion to keep the log more user-friendly.

goaway (approved):
Rerunning CI. Suggest we move forward, and we can improve the log in a followup.

PR #2123, opened 2022-03-21
api: new dns and connection configuration options 

goaway:
Description: Add EngineBuilder options for configuring minimum DNS refresh rate and per-host max connections.
Risk Level: Low
Testing: Unit & CI

### DIFF
--- a/library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
+++ b/library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
@@ -34,15 +34,17 @@ open class EngineBuilder(
   private var dnsRefreshSeconds = 60
   private var dnsFailureRefreshSecondsBase = 2
   private var dnsFailureRefreshSecondsMax = 10
-  private var dnsQueryTimeoutSeconds = 25
-  private var dnsPreresolveHostnames = "[]"
   private var dnsFallbackNameservers = listOf<String>()
   private var dnsFilterUnroutableFamilies = false
+  private var dnsQueryTimeoutSeconds = 25
+  private var dnsMinRefreshSeconds = 60
+  private var dnsPreresolveHostnames = "[]"
   private var enableHappyEyeballs = false
   private var enableInterfaceBinding = false
   private var h2ConnectionKeepaliveIdleIntervalMilliseconds = 100000000
   private var h2ConnectionKeepaliveTimeoutSeconds = 10
   private var h2RawDomains = listOf<String>()
+  private var maxConnectionsPerHost = 7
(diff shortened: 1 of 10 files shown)


library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -244,6 +259,18 @@ open class EngineBuilder(
     return this
   }
 
+  /**
+   * Set the maximum number of connections to open to a single host.
```
Augustyniak, line 263: Can we mention in the docs what the default max is (7)?

library/swift/EngineBuilder.swift
```diff
@@ -211,6 +225,17 @@ open class EngineBuilder: NSObject {
     return self
   }
 
+  /// Set the maximum number of connections to open to a single host.
```
Augustyniak, line 228: same as above - can we add a comment which would explain what the default maximum is (7)?

Augustyniak, line 228: same as above - can we add a comment which would explain what the default maximum is (7)?

Augustyniak (dismissed):
with 2 nits

library/swift/EngineBuilder.swift
goaway: There's not really a good reason - 7 is just in the range of normal defaults for this kind of thing. I did want to get in the habit of mentioning defaults in the docstrings though, which was why I mentioned it here.

Augustyniak (approved):

Augustyniak (approved):

library/swift/EngineBuilder.swift
```diff
@@ -400,6 +425,7 @@ open class EngineBuilder: NSObject {
       dnsFailureRefreshSecondsBase: self.dnsFailureRefreshSecondsBase,
       dnsFailureRefreshSecondsMax: self.dnsFailureRefreshSecondsMax,
       dnsQueryTimeoutSeconds: self.dnsQueryTimeoutSeconds,
+      dnsMinRefreshSeconds: self.dnsMinRefreshSeconds,
```
Augustyniak, line 428: I just realized that we have `dnsFailureRefreshSecondsMax` which puts word `max` as suffix 🤷 may be worth following up to make naming consistent

goaway: For this PR, the name corresponds exactly with the internal setting. I have thought about going through and cleaning up  some of the inconsistencies in our config API. Perhaps if we create the advanced config split, that would be a good opportunity to also perform some of this cleanup.

PR #2128, opened 2022-03-25
iOS: Add `includeUnroutableDNSResults` to engine builder

jpsim:
If set, the resolver will avoid the system's heuristics to only return IPv4 or IPv6 addresses that it considers to be "routable", instead returning all possible IPv4 or IPv6 addresses.

This setting is ignored if the DNS lookup family is set to v4-only or v6-only.

This may be a useful setting to specify if the addresses considered unroutable by the system's heuristics may in practice be routable.

We may want to enable this in combination with happy eyeballs.

---

Description: iOS: Add `includeUnroutableFamilies` to engine builder
Risk Level: Low, Apple-only, off by default
Testing: I've confirmed both branches of the code paths are hit when the flag is on and off on a physical iPhone. I will validate the behavioral changes next before merging. Unit tests in progress...
Docs Changes: In progress...
Release Notes: In progress...

---

To Do:

- [ ] Validate Behavior
- [ ] Unit / Integration Tests
- [ ] Docs
- [ ] Release Notes


library/objective-c/EnvoyConfiguration.m
```diff
@@ -16,6 +16,7 @@ - (instancetype)initWithAdminInterfaceEnabled:(BOOL)adminInterfaceEnabled
                               enableHappyEyeballs:(BOOL)enableHappyEyeballs
                            enableInterfaceBinding:(BOOL)enableInterfaceBinding
                     enforceTrustChainVerification:(BOOL)enforceTrustChainVerification
+                enableDNSFilterUnroutableFamilies:(BOOL)enableDNSFilterUnroutableFamilies
```
goaway, line 19: nit: how about `filterUnroutableDNSResults` (or perhaps, `filterUnroutableDNSResultIPs`)?

goaway: `enableAppleDNSFiltering`

library/objective-c/EnvoyConfiguration.m
```diff
@@ -145,11 +147,12 @@ - (nullable NSString *)resolveTemplate:(NSString *)templateYAML {
...
   [definitions appendFormat:@"- &dns_refresh_rate %lus\n", (unsigned long)self.dnsRefreshSeconds];
   [definitions appendFormat:@"- &dns_resolver_name envoy.network.dns_resolver.apple\n"];
-  // No additional values are currently needed for Apple-based DNS resolver.
   [definitions
       appendFormat:@"- &dns_resolver_config "
                    @"{\"@type\":\"type.googleapis.com/"
-                   @"envoy.extensions.network.dns_resolver.apple.v3.AppleDnsResolverConfig\"}\n"];
+                   @"envoy.extensions.network.dns_resolver.apple.v3.AppleDnsResolverConfig\", "
+                   @"\"include_unroutable_families\": %@}\n",
+                   self.enableDNSFilterUnroutableFamilies ? @"false" : @"true"];
```
goaway, line 155: LGTM

library/objective-c/EnvoyConfiguration.m
jpsim: Filter is ambiguous if we're keeping or rejecting unroutable results.

Perhaps `excludeUnroutableDNSResults` since we can share that setting across platforms since we also have the option to do that on the c-ares side.

jpsim:
Thanks for the review, @goaway although the most important part of this to review is the Envoy side: https://github.com/envoyproxy/envoy/pull/20527

jpsim:
Closing as stale now that we've migrated to the Envoy repo: https://github.com/envoyproxy/envoy/blob/main/mobile/README.md

PR #2129, opened 2022-03-26
engine: return new handles from `EngineHandle::initEngine`

jpsim:
Until now, `EngineHandle` has always had a single value of `1`. In other words, the Envoy engine has always been a singleton with no way to run multiple independent engines.

In order to support multiple concurrent engines (https://github.com/envoyproxy/envoy-mobile/issues/332), we first need to make `EngineHandle::initEngine` create a new engine and rather than have functions that use the engine do a singleton lookup, we can use the `envoy_engine_t` as an opaque handle type to get back the `Envoy::Engine` instance.

Several places were using a hardcoded value of `1`, so we need to plumb through the engine handles.

With this change although you can now create multiple engine instances, it is not safe to do so because there is still some static state that will need to be untangled. Also note that like the singleton we have had until now, new engine instances that are created will leak, as there is no way to fully release the engine through the public API. This is intentional in order to reduce the risk associated with this change.

Mike and I made these changes while pairing. The bulk of the ideas here are really from him.

Description:
Risk Level: High.
Testing: Existing tests & Lyft apps validation.
Docs Changes: None.
Release Notes: Added.

Co-authored-by: Alyssa Wilk <alyssar@chromium.org>


jpsim:
Interestingly I don't see the asan leaks locally, only on CI:

```
$ ./bazelw test //test/common/... --config=macos --config=asan-dev
//test/common:engine_common_test                                (cached) PASSED in 44.2s
//test/common:engine_test                                       (cached) PASSED in 38.2s
//test/common:main_interface_test                               (cached) PASSED in 38.1s
//test/common/bridge:utility_test                               (cached) PASSED in 17.8s
//test/common/buffer:bridge_fragment_test                       (cached) PASSED in 15.1s
//test/common/common:lambda_logger_delegate_test                (cached) PASSED in 13.0s
//test/common/data:utility_test                                 (cached) PASSED in 13.7s
//test/common/extensions/filters/http/assertion:assertion_filter_test (cached) PASSED in 43.6s
//test/common/extensions/filters/http/platform_bridge:platform_bridge_filter_integration_test (cached) PASSED in 32.3s
//test/common/extensions/filters/http/platform_bridge:platform_bridge_filter_test (cached) PASSED in 42.9s
//test/common/extensions/retry/options/network_configuration:network_configuration_retry_options_test (cached) PASSED in 43.2s
//test/common/http:client_test                                  (cached) PASSED in 49.9s
//test/common/http:header_utility_test                          (cached) PASSED in 12.8s
//test/common/integration:client_integration_test               (cached) PASSED in 69.9s
//test/common/network:configurator_test                         (cached) PASSED in 19.2s
//test/common/network:src_addr_socket_option_impl_test          (cached) PASSED in 33.9s
//test/common/network:synthetic_address_impl_test               (cached) PASSED in 15.6s
//test/common/stats:utility_test                                (cached) PASSED in 41.2s
//test/common/stream_info:extra_stream_info_test                (cached) PASSED in 13.8s
//test/common/thread:lock_guard_test                            (cached) PASSED in 11.9s
```

test/java/org/chromium/net/CronetUrlRequestContextTest.java
```diff
@@ -1117,7 +1117,10 @@ public void testInitEngineAndStartRequest() {
   @Test
   @SmallTest
   @Feature({"Cronet"})
-  public void testInitEngineStartTwoRequests() throws Exception {
+  @Ignore(
```
jpsim, line 1120: @carloseltuerto could you help me take a look at this test failure? It's unclear to me why my changes would cause this to start failing. Is fixing this a blocker to removing the engine singleton?

carloseltuerto: This test is straightforward: a single engine where each request is sent sequentially.

To make sure there is no test pollution, commenting out all the other tests could help. (I don't know how to ask a single java test to run with bazel)

jpsim: You're right, this test does pass if I run it in isolation, disabling all other tests in this file. Thanks for the pointer, I can look into what's not being properly cleaned up.

jpsim:
Merging main into this branch surfaced a new assertion failure: https://github.com/envoyproxy/envoy/blob/abd0f8646d7f148fb174268afb7ca9299122f15f/source/common/common/logger.cc#L57

Looking...

jpsim:
Posting some findings that Alyssa shared with me to address the remaining Envoy assertion failures:

* We'll need to set `envoy.disallow_global_stats` to `true`, as is done in Envoy's multi-engine tests: https://github.com/envoyproxy/envoy/blob/79f4d0d70f5d62e9a496e6cd6461424fc523f989/test/integration/multi_envoy_test.cc#L9
* `engine.cc` sets up logging delegates, and they used to be global and now aren't, so that mechanism will need to be updated

alyssawilk (dismissed):
optional: I'd comment inline if we can (just so folks don't have to git blame to figure out what's going on

jpsim:
This isn't quite ready for review yet, namely the logging delegates still needs to be reworked in order to outlive the engine.

> `engine.cc` sets up logging delegates, and they used to be global and now aren't, so that mechanism will need to be updated

jpsim:
Oops, I meant to add `Co-authored-by: Mike Schore <mike.schore@gmail.com>` to that last commit, thanks @goaway!

Augustyniak:
Should we update the description of the PR? It contains some WIP notes.

Augustyniak:
Let's mention https://github.com/envoyproxy/envoy-mobile/issues/332 in the description of the PR too.

jpsim:
@Augustyniak I've updated the PR description to make more sense as a commit message when merging this.

docs/root/intro/version_history.rst
```diff
@@ -15,6 +15,9 @@ Breaking changes:
 - iOS: enable usage of ``NWPathMonitor`` by default (:issue:`#2329 <2329>`)
 - iOS: replace ``enableNetworkPathMonitor`` with a new ``setNetworkMonitoringMode`` API to allow disabling monitoring (:issue:`#2345 <2345>`)
 - iOS: release artifacts no longer embed bitcode
+- api: engines are no longer a singleton, you may need to update your code to only create engines once and hold on to them.
+  You also cannot assume that an `envoy_engine_t` value of `1` will return the default engine.
```
goaway, line 19: nit: Not sure this strictly needs to be in release notes, since technically this is not a public type/interface.

goaway (approved):
Nice work @jpsim!

PR #2130, opened 2022-03-28
stats: enable more metrics

Augustyniak:

Description: Modify config to enable more metrics. Mainly focused on downstream requests, DNS updates and h2 ping timeouts. The changes add around 44 new metrics (a move from around 250 to around 290).
Risk Level: Low
Testing: N/A
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -389,6 +389,12 @@ stats_config:
             regex: '^cluster\.[\w]+?\.upstream_rq_[\w]+'
         - safe_regex:
             google_re2: {}
+            regex: '^cluster\.[\w]+?\.update_(attempt|success|failure)'
+        - safe_regex:
+            google_re2: {}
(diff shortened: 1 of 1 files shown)


jpsim (approved):

library/common/config/config.cc
```diff
@@ -387,6 +387,12 @@ stats_sinks: *stats_sinks
         - safe_regex:
             google_re2: {}
             regex: '^cluster\.[\w]+?\.upstream_rq_[\w]+'
+        - safe_regex:
+            google_re2: {}
```
jpsim, line 391: Is there ever a situation we'd want to use re2?

Augustyniak: Frankly speaking - I do not know. I was an ignorant and followed the established pattern + I'm just more familiar with standard regexes so felt more comfortable doing that.

PR #2133, opened 2022-03-28
WIP build: add flatbuffers

snowp:

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/bazel/envoy_mobile_dependencies.bzl
+++ b/bazel/envoy_mobile_dependencies.bzl
@@ -1,6 +1,7 @@
 load("@build_bazel_rules_swift//swift:repositories.bzl", "swift_rules_dependencies")
 load("@build_bazel_rules_apple//apple:repositories.bzl", "apple_rules_dependencies")
 load("@build_bazel_apple_support//lib:repositories.bzl", "apple_support_dependencies")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 load("@rules_jvm_external//:defs.bzl", "maven_install")
 load("@rules_detekt//detekt:dependencies.bzl", "rules_detekt_dependencies")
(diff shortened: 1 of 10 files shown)


keith (approved):

bazel/flatbuffers.bzl
```diff
@@ -0,0 +1,33 @@
...
+        name = "{}_fb_kt_lib".format(name),
+        srcs = [":{}_fb_kt_srcs".format(name)],
+        deps = ["@maven//:com_google_flatbuffers_flatbuffers_java"],
+    )
+
+    flatbuffer_library_public(
+        name = "{}_fb_swift_srcs".format(name),
+        srcs = ["test.fbs"],
+        outs = [
+            "test_generated.swift",
```
jpsim, line 30: Nit: why `test_generated.swift`? Can this be `$(name).generated.swift`?

```suggestion
            "{}.generated.swift".format(name),
```

jpsim (approved):
It'd be good to have some tests validating that a buffer can be read and written across the generated language bindings, but not necessary right now. If we start using flatbuffers in EM directly we can add tests for that specifically.

The java version is 2.0.3 but the C++ and Swift versions are 2.0.0. Are those versions compatible with each other?

bazel/flatbuffers.bzl
snowp: Ah yeah the swift rules here aren't generic, just hard coded to the test.fbs input, let me update

snowp:
I would expect them to be compatible per semver, also from looking into the format a bit I believe the underlying binary format is very stable, most changes to the library are around updating the per language bindings. For example, this seems like one of the patch changes: https://github.com/google/flatbuffers/pull/6785

keith (approved):

PR #2134, opened 2022-03-29
android: add support for registering a platform KV store

goaway:
Description: This PR allows platform implementations of a generic (and potentially persistent) key-value store to be registered with the Engine. Internal components can look up these implementations by name. The internal implementation and wiring in this PR is platform agnostic, but this PR adds a public API for Java/Kotlin only, initially. iOS and other public APIs to follow in upcoming PRs.

Part of #2077.

Risk Level: Low
Testing: New Unit + Integration Coverage


library/common/extensions/filters/http/test_kv_store/config.cc
```diff
@@ -0,0 +1,29 @@
+#include "library/common/extensions/filters/http/test_kv_store/config.h"
+
+#include "library/common/extensions/filters/http/test_kv_store/filter.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace TestKeyValueStore {
```
RyanTheOptimist, line 8: I would have expected the name of this to be something like PlatformKeyValueStore. Is this filter intended to be used in production? Maybe it's just test-only?

library/java/io/envoyproxy/envoymobile/engine/types/EnvoyKeyValueStore.java
```diff
@@ -0,0 +1,27 @@
...
+   * @param key,     key identifying the value to be returned.
+   * @return String, value mapped to the key, or null if not present.
+   */
+  String read(String key);
+
+  /**
+   * Remove a value from the key value store implementation.
+   *
+   * @param key,     key identifying the value to be removed.
+   * @return String, value mapped to the key, or null if not present.
```
RyanTheOptimist, line 16: nit: Looks like this returns void?

library/java/io/envoyproxy/envoymobile/engine/types/EnvoyKeyValueStore.java
```diff
@@ -0,0 +1,27 @@
...
+   */
+  void remove(String key);
+
+  /**
+   * Save a value to the key value store implementation.
+   *
+   * @param key,     key identifying the value to be saved.
+   * @param value,   the value to be saved.
```
RyanTheOptimist, line 24: nit: Looks like this returns void?

library/common/extensions/filters/http/test_kv_store/filter.h
```diff
@@ -0,0 +1,52 @@
...
+
+private:
+  const envoy_kv_store* kv_store_;
+  const std::string test_key_;
+  const std::string test_value_;
+};
+
+using TestKeyValueStoreFilterConfigSharedPtr = std::shared_ptr<TestKeyValueStoreFilterConfig>;
+
+class TestKeyValueStoreFilter final : public ::Envoy::Http::PassThroughFilter {
```
RyanTheOptimist, line 34: nit: Please add a comment which describes what this filter does.

library/common/extensions/key_value/platform/config.cc
```diff
@@ -4,10 +4,37 @@
...
+    envoy_data bridged_value = bridged_store_.read(bridged_key, bridged_store_.context);
+    return Data::Utility::copyToString(bridged_value);
+  }
+
+  void save(const std::string& key, const std::string& contents) override {
+    envoy_data bridged_key = Data::Utility::copyToBridgeData(key);
+    envoy_data bridged_value = Data::Utility::copyToBridgeData(contents);
+    bridged_store_.save(bridged_key, bridged_value, bridged_store_.context);
+  }
+
```
RyanTheOptimist, line 33: I would have expected a remove() method. Is that not required?

library/common/jni/jni_interface.cc
```diff
@@ -787,6 +788,58 @@ static void* jvm_on_send_window_available(envoy_stream_intel stream_intel, void*
...
 
+// JvmKeyValueStoreContext
+static envoy_data jvm_kv_store_read(envoy_data key, const void* context) {
+  jni_log("[Envoy]", "jvm_kv_store_read");
+  JNIEnv* env = get_env();
+
+  jobject j_context = static_cast<jobject>(const_cast<void*>(context));
+
+  jclass jcls_JvmKeyValueStoreContext = env->GetObjectClass(j_context);
+  jmethodID jmid_read = env->GetMethodID(jcls_JvmKeyValueStoreContext, "read", "([B)[B");
```
RyanTheOptimist, line 799: It's too bad there is no magic for auto-generating these method signature strings, but c'est la vie.

RyanTheOptimist (commented):
Nice! A few small comments. I think it would also be nice to flesh out the PR description to explain a bit more the details of what's going on.

library/common/jni/jni_interface.cc
goaway: In fact, there is! ...but I never figured out how it might work with Bazel. But `javac` itself has support for generating jni headers based on `native` method signatures in class definitions in java.

library/common/extensions/key_value/platform/config.cc
goaway: @alyssawilk shared that you all didn't have a use case for it right now (which I sort of found surprising, but then again, if you supply the KV store implementation, you can always remove stuff from it elsewhere). I went ahead and wired `remove` at the other layers anyways, simply because I do expect we will have a use for it.

library/common/extensions/filters/http/test_kv_store/config.cc
goaway: Your presumption is correct - this is a test-only filter that exist alongside some other test-only filters in an entirely unintuitive location. Ideally, these should be moved elsewhere. I can certainly add a clarifying comment.

goaway:
> Nice! A few small comments. I think it would also be nice to flesh out the PR description to explain a bit more the details of what's going on.

👍

library/java/io/envoyproxy/envoymobile/engine/types/EnvoyKeyValueStore.java
goaway: Sorry I may be being dense - what's the fix here? I fixed the `@return` for remove, but this method doesn't have one.

library/common/extensions/filters/http/test_kv_store/BUILD
```diff
@@ -0,0 +1,45 @@
...
+    ],
+)
+
+envoy_cc_extension(
+    name = "config",
+    srcs = ["config.cc"],
+    hdrs = ["config.h"],
+    repository = "@envoy",
+    deps = [
+        ":test_kv_store_filter_lib",
```
jpsim, line 42: Maybe worth a comment explaining why this isn't ideal but necessary right now to allow testing the interface?

goaway: I did add this comment here:
https://github.com/envoyproxy/envoy-mobile/pull/2134/files#diff-f38766704cb1bb34da15e2496c577b3bd09a101137f0f2f684c6dc869821fe60R18

Or did were you thinking of something else? This is mostly a holdover from how things were set up historically and a desire to limit the scope of this PR. :)

jpsim: No need to add another comment, that one's sufficient. Thanks!

library/common/extensions/key_value/platform/c_types.h
```diff
@@ -0,0 +1,35 @@
...
+// NOLINT(namespace-envoy)
+
+#ifdef __cplusplus
+extern "C" { // function pointers
+#endif
+
+/**
+ * Function signature for reading value from implementation.
+ */
+typedef envoy_data (*envoy_kv_store_read_f)(envoy_data key, const void* context);
```
jpsim, line 14: Is there a reason you went with a synchronous approach here? Is the idea that consumers would take care of moving this to a separate thread if they don't want to block on reading/writing?

library/common/extensions/key_value/platform/c_types.h
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Function signature for saving value to implementation.
+ */
+typedef void (*envoy_kv_store_save_f)(envoy_data key, envoy_data value, const void* context);
```
jpsim, line 19: Can this document what's expected as the `context` pointer?

library/common/jni/jni_interface.cc
jpsim: Yeah, this is bound to bite us at some point. `cxx.rs` automatically generates these which helps avoid typos. If we start introducing Rust at this layer it would be worth considering using it for JNI codegen.

library/java/io/envoyproxy/envoymobile/engine/types/EnvoyKeyValueStore.java
RyanTheOptimist: Oh, shoot. You're right. I somehow totally misread this. :(

RyanTheOptimist (dismissed):

library/common/jni/jni_interface.cc
```diff
@@ -787,6 +788,58 @@ static void* jvm_on_send_window_available(envoy_stream_intel stream_intel, void*
...
+
+  env->DeleteLocalRef(j_value);
+  env->DeleteLocalRef(j_key);
+  env->DeleteLocalRef(jcls_JvmKeyValueStoreContext);
+
+  return native_data;
+}
+
+static void jvm_kv_store_remove(envoy_data key, const void* context) {
+  jni_log("[Envoy]", "jvm_store_remove");
```
jpsim, line 812: Nit:

```suggestion
  jni_log("[Envoy]", "jvm_kv_store_remove");
```

library/common/extensions/key_value/platform/c_types.h
goaway: Well, for `read` asynchronous would introduce a whole new calling paradigm ("onRead"?). Honestly, I'd think an implementation should probably do caching/pre-caching/memoization if read performance is likely to be a bottleneck somewhere. For `write`/`remove `, yes, I think that's also better off as a feature of the actual provided platform implementation.

library/common/extensions/key_value/platform/c_types.h
goaway: Sure.

library/common/jni/jni_interface.cc
goaway: Ah, good catch.

library/proguard.txt
```diff
@@ -39,6 +39,10 @@
    <methods>;
 }
 
+-keep, includedescriptorclasses class io.envoyproxy.envoymobile.engine.JvmKeyValueStoreContext {
```
jpsim, line 42: How is this file updated?

library/common/extensions/key_value/platform/c_types.h
goaway: I added a more general description to the root c_types.h, since we use these everywhere.

jpsim:
Can you please add a changelog entry for the user-facing change?

library/common/extensions/key_value/platform/c_types.h
jpsim: ok. Another thing to consider is whether or not errors should be propagated between the platform and native layers.

library/proguard.txt
goaway: Manually, when a new JNI-facing class is added.

library/common/types/c_types.h
```diff
@@ -7,6 +7,16 @@
 
 // NOLINT(namespace-envoy)
 
+/**
+ * Throughout this file one may note that most callbacks take a void* context parameter, and most
```
jpsim, line 11: 👌 excellent comment

jpsim (approved):
Nice work. Looking forward to seeing the default persistent kv store too.

library/java/io/envoyproxy/envoymobile/engine/types/EnvoyKeyValueStore.java
goaway: All good - thanks for all the comments. :)

PR #2141, opened 2022-03-31
bazel: Add flatbuffers Swift hack

keith:
The generated code couldn't be used with private_deps because it
imported FlatBuffers normally and created a bunch of public types.
Ideally this would be an option to the generator instead of doing this.

This will have to be updated to a loop to support multiple input files


snowp (approved):
Thanks!

jpsim (approved):
nice

keith:
> nice

you use this loosely I see

jpsim:
The end result is nice. The hack is not that bad imho in that it is localized and should be relatively easily upstream-able as a configuration as you said.

PR #2149, opened 2022-04-05
internal: pass engine handles throughout API

jpsim:
In preparation for removing the singleton. This is split off into its own independent change to minimize the scope of the main singleton removal work.

We were already doing this in many places, this just adds more instances of passing the engine handle so we do so comprehensively.

Description: Pass engine handles throughout API
Risk Level: Medium, changing some JNI interfaces, removing some static state
Testing: Unit tests and sample apps
Docs Changes: N/A internal only
Release Notes: N/A internal only
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/library/java/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
+++ b/library/java/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
@@ -34,7 +34,7 @@ public class AndroidNetworkMonitor extends BroadcastReceiver {
   private static volatile AndroidNetworkMonitor instance = null;
 
   private int previousNetworkType = ConnectivityManager.TYPE_DUMMY;
-  private long engineHandle;
+  private EnvoyEngine envoyEngine;
   private ConnectivityManager connectivityManager;
   private NetworkCallback networkCallback;
   private NetworkRequest networkRequest;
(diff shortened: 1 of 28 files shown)


library/java/io/envoyproxy/envoymobile/engine/AndroidNetworkMonitor.java
```diff
@@ -34,6 +34,7 @@ public class AndroidNetworkMonitor extends BroadcastReceiver {
   private static volatile AndroidNetworkMonitor instance = null;
 
   private int previousNetworkType = ConnectivityManager.TYPE_DUMMY;
+  private long engineHandle;
```
goaway, line 37: How about we store a reference to the engine, and let the engine make the JNI call, so we don't have to expose the handle?

jpsim: Giving that a shot in [22148bc](https://github.com/envoyproxy/envoy-mobile/pull/2149/commits/22148bcf9678f861fe0fc0c451c969f27fc638b6)

goaway (approved):
Nice work on this @jpsim. Finally had a chance to look through it all. Looks good.

PR #2160, opened 2022-04-13
bazel: Allow multiple definitions for armeabi android links

keith:
When working on rust support we hit this issue
https://github.com/rust-lang/compiler-builtins/issues/353 which seems to
still be persistent even in the case you have the nightly version
mentioned there. I haven't been able to reproduce the issue in a
separate project, but the core seems to be that both libgcc and rust's
compiler builtins module vendor some float math symbols and both are
included. This change is as scoped as possible so that no real issues
could sneak in with this disabled, which should mostly be covered by
only passing this for the single arch.

### DIFF
--- a/library/common/jni/BUILD
+++ b/library/common/jni/BUILD
@@ -51,6 +51,14 @@ cc_library(
     alwayslink = True,
 )
 
+config_setting(
+    name = "android_armeabi",
+    constraint_values = [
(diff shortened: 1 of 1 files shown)


snowp (approved):
Multiple definitions is a bit scary, but I imagine we'd uncover something that foundational in the emulator tests

Issue #1642, opened 2021-07-29
Cronvoy: Bidirectional Stream


carloseltuerto:
Cronet supports [bidirectional stream](https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/android/api/src/org/chromium/net/BidirectionalStream.java), Envoy Mobile, not yet.

Those are the two peculiar parameters:
```
        /**
         * Sets priority of the stream which should be one of the
         * {@link #STREAM_PRIORITY_IDLE STREAM_PRIORITY_*} values.
         * The stream is given {@link #STREAM_PRIORITY_MEDIUM} priority if
         * this method is not called.
         *
         * @param priority priority of the stream which should be one of the
         *         {@link #STREAM_PRIORITY_IDLE STREAM_PRIORITY_*} values.
         * @return the builder to facilitate chaining.
         */
        public abstract Builder setPriority(int priority);

        /**
         * Delays sending request headers until {@link BidirectionalStream#flush()}
         * is called. This flag is currently only respected when QUIC is negotiated.
         * When true, QUIC will send request header frame along with data frame(s)
         * as a single packet when possible.
         *
         * @param delayRequestHeadersUntilFirstFlush if true, sending request headers will
         *         be delayed until flush() is called.
         * @return the builder to facilitate chaining.
         */
        public abstract Builder delayRequestHeadersUntilFirstFlush(
                boolean delayRequestHeadersUntilFirstFlush);
```

This is the base of gRPC, so its reimplementation is mandatory.

PR #2164, opened 2022-04-18
CronetBidirectionalStream 

carloseltuerto:
This is the first PR to address #1642. Most of the Cronet tests have been ported. However:

- The QUIC mode where the request header is [**guaranteed**](https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/android/api/src/org/chromium/net/BidirectionalStream.java;l=79-91) to be sent with the first flushed ByteBuffers needs to be implemented in EM first.
- There are some `@Ignore` tests:
  - Testing specific error codes
  - Causing EM to crash
  - FinalIntel are flaky (not always set)
  - FinalIntel not available at all when cancelling before receiving the response headers

The implementation does not use `synchronized`. The logic is 100% Compare-And-Swap based.

Description: Initial CronetBidirectionalStream implementation
Risk Level: none: only touches Cronvoy
Testing: CI, and multi concurrent runs
Docs Changes: N/A
Release Notes: N/A


library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
```diff
@@ -0,0 +1,242 @@
...
+ * operations allowed with the EnvoyHTTPStream - subsequent operations will be ignored silently.
+ * However, in the event that that one or more EnvoyHTTPStream operations are currently being
+ * executed, the "cancel" operation gets postponed: the last concurrent operation will invoke
+ * "cancel" at the end.
+ *
+ * <p>Instance of this class start with a state of "BUSY_STARTING". This ensure that if a cancel
+ * is invoked while the stream is being created, that cancel will be executed only once the stream
+ * is completely initialized. Doing otherwise leads to unpredictable outcomes.
+ */
+final class CancelProofEnvoyStream {
```
StefanoDuo, line 27: Will this also be used for plain url requests?

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
```diff
@@ -25,157 +36,93 @@
...
-          if (mEndOfStream) {
-            mReadState = State.READING_DONE;
-            maybeOnSucceeded = (mWriteState == State.WRITING_DONE);
-          } else {
-            mReadState = State.WAITING_FOR_READ;
-          }
+        @NextAction
+        int nextAction =
+            mState.nextAction(mEndOfStream ? Event.LAST_READ_COMPLETED : Event.READ_COMPLETED);
+        if (nextAction == NextAction.INVOKE_ON_READ_COMPLETED_CALLBACK) {
```
StefanoDuo, line 120: Commenting here but this is more of a general question: is `mState.nextAction(@Event)` good enough to really know the next action or do you also need to know the context within it is being called (i.e., does an action have different meanings in different contexts)?

I'm trying to understand if it would be possible for me to reason about this as a simple `hashmap<@Action, functionExecutinAction>` or if there are more nuances (I would love to be able to reason about the next action logic and the action to function mapping on their own).

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
```diff
@@ -790,43 +651,279 @@ private void onCallbackException(Exception e) {
...
+    }
+    if (value == null) {
+      throw new NullPointerException("Invalid header value.");
+    }
+    if (!isValidHeaderName(header) || value.contains("\r\n")) {
+      throw new IllegalArgumentException("Invalid header " + header + "=" + value);
+    }
+  }
+
+  private static boolean isValidHeaderName(String header) {
```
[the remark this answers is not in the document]
carloseltuerto: https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/android/java/src/org/chromium/net/impl/JavaUrlRequest.java;l=259

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
```diff
@@ -790,43 +651,279 @@ private void onCallbackException(Exception e) {
...
+  }
+
+  private static void validateHeader(String header, String value) {
+    if (header == null) {
+      throw new NullPointerException("Invalid header name.");
+    }
+    if (value == null) {
+      throw new NullPointerException("Invalid header value.");
+    }
+    if (!isValidHeaderName(header) || value.contains("\r\n")) {
```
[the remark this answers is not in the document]
carloseltuerto: https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/android/java/src/org/chromium/net/impl/JavaUrlRequest.java;l=250

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
```diff
@@ -233,163 +176,160 @@ public void run() {
...
+    }
+  }
+
+  /**
+   * Returns, potentially, an exception to report through the "onError" callback, even though no
+   * stream has been created yet. This awkward error reporting solely exists to mimic Cronet.
+   */
+  @Nullable
+  private static CronetException engineSimulatedError(Map<String, List<String>> requestHeaders) {
+    if (requestHeaders.get(":scheme").get(0).equals("http")) {
```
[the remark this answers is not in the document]
carloseltuerto: That's implemented in C++, I have not yet paid serious attention to this. There is a test that trickles this error. As usual, it is not because a test is passing that we can imply that this is the correct implementation. Far from it. I will add to TODO.

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
```diff
@@ -233,163 +176,160 @@ public void run() {
...
+    }
+    mEnvoyRequestHeaders =
+        buildEnvoyRequestHeaders(mMethod, mRequestHeaders, mUserAgent, mInitialUrl);
+    // Cronet C++ layer exposes reported errors here with an onError callback. EM does not.
+    @Nullable CronetException startUpException = engineSimulatedError(mEnvoyRequestHeaders);
+    @Event int startingEvent;
+    if (startUpException == null) {
+      startingEvent = Event.USER_START;
+      if (!mDelayRequestHeadersUntilFirstFlush) {
+        startingEvent |= Event.USER_START_WITH_HEADERS;
```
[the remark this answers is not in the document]
carloseltuerto: I'll change that - way too hacky.

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
```diff
@@ -790,43 +651,279 @@ private void onCallbackException(Exception e) {
...
+    switch (mState.nextAction(Event.ERROR)) {
+    case NextAction.CANCEL:
+      mStream.cancel();
+      break;
+    case NextAction.PROCESS_ERROR:
+      failWithException();
+      break;
+    default:
+      Log.e(CronetUrlRequestContext.LOG_TAG,
+            "An exception has already been previously recorded. This one is ignored.", exception);
```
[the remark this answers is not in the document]
carloseltuerto: There is a third possible Action: TAKE_NO_MORE_ACTIONS - occurs when terminating (includes DONE)

Here is a scenario:

```
thread1: mException.compareAndSet(null, exception);
thread2: mState.nextAction(Event.USER_CANCEL);
thread1: switch (mState.nextAction(Event.ERROR)) {
```

The `switch` will return `NextAction.TAKE_NO_MORE_ACTIONS` in this case. The `mException` is set, but that exception will be ignored: a "user cancel" preempted it. mState is the sole master.

I'm not sure that I really answered you question, but hopefully those details are useful.

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
```diff
@@ -25,157 +36,93 @@
...
-            return;
-          }
-          if (mEndOfStream) {
-            mReadState = State.READING_DONE;
-            maybeOnSucceeded = (mWriteState == State.WRITING_DONE);
-          } else {
-            mReadState = State.WAITING_FOR_READ;
-          }
+        @NextAction
+        int nextAction =
```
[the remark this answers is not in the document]
carloseltuerto: If the nextAction is not `INVOKE_ON_READ_COMPLETED_CALLBACK`, it means that we are in a terminating state. Event `READY_TO_FINISH` is only applicable for the success path. When there is cancel/error in progress, the Stream will rather end with one of these two user callbacks: `mCallback.onFailed()`, `mCallback.onCanceled()`.

From a correctness standpoint, we could execute `mCallback.onReadCompleted()` regardless of an ongoing  terminating state. Keep in mind that a cancel/error can occur just after `nextAction == INVOKE_ON_READ_COMPLETED_CALLBACK`. So `mCallback.onReadCompleted()` will get executed even though we are in terminating state anyway. We are just making the occurrence of such situation less likely.

You will notice that Cronet has the very same behaviour: 

https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/android/java/src/org/chromium/net/impl/CronetBidirectionalStream.java;l=200

If a cancel occurs on line 199.5, then `mCallback.onReadCompleted()` will be executed anyway.

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
carloseltuerto: Yes. The same logic is already is already present in CronetUrlRequest, but spread around.  I need to clean that out. In fact, CronetUrlRequest needs a big cleanup.

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
carloseltuerto: Many NextAction are found in more than one place in `CronetBidirectionalStream`. What is a little bit better are the Events: they usually present in only one method (there are exception that I can fix).

The Mapping is a bit more complicated: 

```
Map<Pair<Event, State>, NextAction>
```

As far as I can tell, there are only two action that are generic: CARRY_ON and TAKE_NO_MORE_ACTIONS. So it is not impossible to rather return a Function, but it will require some extra consideration for these two generic NextAction.

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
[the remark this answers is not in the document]
StefanoDuo: Oh, there is this bit of code at the bottom:
```
if (event == Event.ON_CANCEL || event == Event.ON_ERROR || event == Event.ON_COMPLETE) {
    // If this assert triggers it means that the C++ EnvoyMobile contract has been breached.
    assert (originalState & State.DONE) == 0; // Or there is a blatant bug.
} else if ((originalState & State.TERMINATING_STATES) != 0) {
    // Unfortunately, this check can not occur at the beginning of the loop: any encountered
    // IllegalStateException/IllegalArgumentException have precedence.
    return NextAction.TAKE_NO_MORE_ACTIONS; // No need to loop - this is irreversible.
}
```

`return NextAction.TAKE_NO_MORE_ACTIONS` bypasses the `final nextAction` contract hence my confusion. Would it be possible to rewrite it without bypassing that? If it complicates the code unnecessarily we can leave as is but maybe add a comment at the top about this?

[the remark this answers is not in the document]
carloseltuerto: I've reworked this part. Hopefully this is in a better shape now.

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
```diff
@@ -233,259 +174,237 @@ public void run() {
...
-
-      assert !mPendingData.isEmpty() || !mFlushData.isEmpty();
-
-      // Move buffers from mPendingData to the flushing queue.
-      if (!mPendingData.isEmpty()) {
-        mFlushData.addAll(mPendingData);
-        mPendingData.clear();
+      if (mState.nextAction(Event.USER_FLUSH) == NextAction.FLUSH_HEADERS) {
+        mStream.sendHeaders(mEnvoyRequestHeaders, /* endStream= */ mReadOnly);
       }
```
StefanoDuo, line 306: This will always be executed at most once, right?

If so, can you move this out of the loop?

StefanoDuo (commented):
I haven't looked at `CancelProofEnvoyStream`, `CronetUrlRequest*` and the tests since they are orthogonal to the bidi stream implementation.

library/java/org/chromium/net/impl/CronetBidirectionalState.java
```diff
@@ -0,0 +1,415 @@
...
+    int SEND_DATA = 3;                // Send one ByteBuffer on the wire, if any.
+    int READ = 4;                     // Start reading the next chunk of the response body.
+    int INVOKE_ON_READ_COMPLETED = 5; // Initiate the completion of a read operation.
+    int INVOKE_ON_ERROR_RECEIVED = 6; // Initiate the completion of a network Error.
+    int CANCEL = 7;                   // Tell EM to cancel. Can be an user induced, or due to error.
+    int INVOKE_ON_WRITE_COMPLETED_CALLBACK = 8; // Tell the User that a write operation completed.
+    int INVOKE_ON_READ_COMPLETED_CALLBACK = 9;  // Tell the User that a read operation completed.
+    int FINISH_UP = 10;     // Tell the User that the stream is done and was completed successfully.
+    int PROCESS_ERROR = 11; // Tell the User the stream completed in error.
+    int PROCESS_CANCEL = 12;       // Tell the User the stream completed in a cancelled state.
```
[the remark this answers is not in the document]
carloseltuerto: Good suggestion - thanks.

library/java/org/chromium/net/impl/CronetBidirectionalState.java
```diff
@@ -0,0 +1,415 @@
...
+        assert (originalState & State.END_STREAM_READ) == 0;
+        nextState |= State.END_STREAM_READ;
+        // FOLLOW THROUGH
+      case Event.ON_DATA:
+        assert (originalState & State.WAITING_FOR_READ) == 0;
+        nextAction = NextAction.INVOKE_ON_READ_COMPLETED;
+        break;
+
+      case Event.ON_COMPLETE:
+        assert (originalState & State.ON_COMPLETE_RECEIVED) == 0;
```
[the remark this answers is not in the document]
carloseltuerto: Ditto

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
```diff
@@ -233,259 +174,237 @@ public void run() {
...
-          if (!doesMethodAllowWriteData(mInitialMethod) && mRequestHeadersSent) {
-            mWriteState = State.WRITING_DONE;
-          } else {
-            mWriteState = State.WAITING_FOR_FLUSH;
-          }
-        }
-
         try {
-          mCallback.onStreamReady(CronetBidirectionalStream.this);
+          if (!mState.isTerminating()) {
```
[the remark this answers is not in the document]
carloseltuerto: With Cronet, yes. For example, if you send and "http" request (without the "s"), then this ends up with this callback. I had to mimic that, because EM does not behave like this.

library/java/org/chromium/net/impl/CronetBidirectionalState.java
```diff
@@ -0,0 +1,415 @@
...
+
+      case Event.USER_CANCEL:
+        if ((originalState & State.STARTED) == 0) {
+          nextAction = NextAction.CARRY_ON; // Cancel came too soon - no effect.
+        } else if ((originalState & State.ON_COMPLETE_RECEIVED) != 0) {
+          nextState |= State.USER_CANCELLED | State.DONE;
+          nextAction = NextAction.PROCESS_CANCEL;
+        } else {
+          nextState |= State.USER_CANCELLED | State.CANCELLING;
+          nextAction = NextAction.CANCEL;
```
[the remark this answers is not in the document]
carloseltuerto: Done - thanks.

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
```diff
@@ -233,259 +174,237 @@ public void run() {
...
-      mPendingData.add(buffer);
-      if (endOfStream) {
-        mEndOfStreamWritten = true;
-      }
+    Preconditions.checkDirect(buffer);
+    if (!buffer.hasRemaining() && !endOfStream) {
+      throw new IllegalArgumentException("Empty buffer before end of stream.");
+    }
+    if (mState.nextAction(endOfStream ? Event.USER_LAST_WRITE : Event.USER_WRITE) ==
+        NextAction.WRITE) {
```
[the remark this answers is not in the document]
carloseltuerto: The outcome can be to do nothing:

https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/android/java/src/org/chromium/net/impl/CronetBidirectionalStream.java;l=346

But still you are correct that we could be exhaustive here, not just the switch cases.

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
```diff
@@ -233,259 +174,237 @@ public void run() {
...
+      // Another Thread is already copying pending buffers - can't be done concurrently.
+      // However, the thread which started with a zero count will loop until this count goes back
+      // to zero. For all intent and purposes, this has a similar outcome as using synchronized {}
+      return;
+    }
+    do {
+      WriteBuffer pendingBuffer;
+      while ((pendingBuffer = mPendingData.poll()) != null) {
+        mFlushData.add(pendingBuffer);
       }
```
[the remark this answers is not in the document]
carloseltuerto: I tried to write up something - hopefully this is not too confusing...

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
carloseltuerto: Yes - this statement can only be true no more than once. Changed - thanks!

library/java/org/chromium/net/impl/CronetUrlRequestContext.java
```diff
@@ -327,7 +336,7 @@ private static void postObservationTaskToExecutor(Executor executor, Runnable ta
     try {
       executor.execute(task);
     } catch (RejectedExecutionException failException) {
-      // TODO(carloseltuerto): use Envoy-Mobile logs - this is a hack.
+      // TODO: use Envoy-Mobile logs - this is a hack.
```
RyanTheOptimist, line 339: OOC, what would be required to use an EM log? Would you mind filing an issue about this and linking to it here?

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
```diff
@@ -0,0 +1,114 @@
...
+ * <p>This contraption ensures that once a "cancel" operation is invoked, there will be no further
+ * operations allowed with the EnvoyHTTPStream - subsequent operations will be ignored silently.
+ * However, in the event that that one or more EnvoyHTTPStream operations are currently being
+ * executed, the "cancel" operation gets postponed: the last concurrent operation will invoke
+ * "cancel" at the end.
+ *
+ * <p>Instance of this class start with a state of "Busy Starting". This ensure that if a cancel
+ * is invoked while the stream is being created, that cancel will be executed only once the stream
+ * is completely initialized. Doing otherwise leads to unpredictable outcomes.
+ */
```
RyanTheOptimist, line 22: Would you mind commenting on the thread safety of this class? I suspect it's entirely thread-safe but it would be good to confirm.

library/java/org/chromium/net/impl/CronetBidirectionalState.java
```diff
@@ -0,0 +1,431 @@
...
+ *  {@code AtomicInteger.compareAndSet()}.
+ *
+ *  <p>All methods in this class are Thread Safe.
+ */
+final class CronetBidirectionalState {
+
+  /**
+   * Enum of the events altering the global state. There are 3 types of events: User induced
+   * (prefixed with USER_), EM Callbacks (prefixed with ON_), and internal events (the remaining
+   * ones).
```
RyanTheOptimist, line 25: Your call, but given the description in this comment, I expected to see all the USER_ values first, then the ON_ values and finally "everything else".

library/java/org/chromium/net/impl/CronetBidirectionalState.java
```diff
@@ -0,0 +1,431 @@
...
+    int INVOKE_ON_FAILED = 12;     // Tell the User the stream completed in an error state.
+    int INVOKE_ON_CANCELED = 13;   // Tell the User the stream completed in a cancelled state.
+    int TAKE_NO_MORE_ACTIONS = 14; // The stream is already in error state - don't do anything else.
+  }
+
+  /**
+   * Bitmap used to express the global state of the BIDI Stream. Each bit represent one element of
+   * the global state.
+   */
+  @IntDef(flag = true, // Not used as an Enum, and this is not used as the argument of a "switch".
```
RyanTheOptimist, line 116: nit: "This is not used as an Enum nor as the argument of a switch statement."

library/java/org/chromium/net/impl/CronetBidirectionalState.java
```diff
@@ -0,0 +1,431 @@
...
+    return RequestFinishedInfo.SUCCEEDED;
+  }
+
+  /**
+   * Establishes what is the next action by taking in account the current global state, and the
+   * provided {@link Event}. This method has one important side effect: the resulting global state
+   * is saved through an Atomic operation. For few cases, this method will throw when the state is
+   * not compatible with the event.
+   */
+  @NextAction int nextAction(@Event final int event) { // "final" just to avoid dumb mistakes.
```
RyanTheOptimist, line 187: nit: I think I would remove these comments about "final"

library/java/org/chromium/net/impl/CronetBidirectionalState.java
```diff
@@ -0,0 +1,431 @@
...
+    while (true) {
```
RyanTheOptimist, line 188: I don't think I understand why this is in a loop. The method is quite big so maybe I'm not able to hold it all in memory. Would it be possible to extract the body of this `while (true) {` out into a stand-alone method which does a single iteration of the loop?

library/java/org/chromium/net/impl/CronetBidirectionalState.java
```diff
@@ -0,0 +1,431 @@
...
+        }
+        break;
+
+      default:
+        throw new AssertionError("switch is exhaustive");
+      }
+
+      if (mState.compareAndSet(originalState, nextState)) {
+        return nextAction;
+      }
```
RyanTheOptimist, line 428: There is a LOT of logic in this loop, not surprisingly. I'm not quite sure how to verify that all the transitions are correct. Is there a Cronet state transition diagram somewhere?

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
```diff
@@ -1,182 +1,141 @@
...
-    /* Writing to the remote, {@code onWritevCompleted()} callback will be called when done. */
-    int WRITING = 9;
-    /* There is no more data to write and stream is half-closed by the local side. */
-    int WRITING_DONE = 10;
-  }
+public final class CronetBidirectionalStream
+    extends ExperimentalBidirectionalStream implements EnvoyHTTPCallbacks {
+
+  private static final String X_ENVOY = "x-envoy";
+  private static final String X_ENVOY_SELECTED_TRANSPORT = "x-envoy-upstream-alpn";
```
RyanTheOptimist, line 48: X_ENVOY_UPSTREAM_ALPN, perhaps?

RyanTheOptimist (commented):
High-level comment. There are thousands of lines of code added in the PR. Are there any ways we could consider breaking it up into smaller pieces?

In any case, this is awesome!

carloseltuerto:
> High-level comment. There are thousands of lines of code added in the PR. Are there any ways we could consider breaking it up into smaller pieces?
> 
> In any case, this is awesome!

This PR is a deal with the devil: you have code avoiding as much as possible "task switching"/"synchronisation", however, an implementation with blocking semantics (java "synchronized") would probably be easier (still tricky).

And there is a class that could be in an isolated PR: CronetBidirectionalState (and the test: CronetBidirectionalStateTest). However, my appreciation of the situation is this would be a pointless exercise. By just looking at that class and its test, it is impossible to infer what how it is used: you will the caller too to get a sense of what is going on.

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
```diff
@@ -0,0 +1,242 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.IntUnaryOperator;
+import io.envoyproxy.envoymobile.engine.EnvoyHTTPStream;
+
+/**
+ * Consistency layer above the {@link EnvoyHTTPStream} preventing unwarranted Stream operations
```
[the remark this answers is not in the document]
carloseltuerto: Done, thanks

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
```diff
@@ -0,0 +1,242 @@
...
+ * after a "cancel" operation. There are no "synchronized" - this is CAS based logic.
+ *
+ * <p>This contraption ensures that once a "cancel" operation is invoked, there will be no further
+ * operations allowed with the EnvoyHTTPStream - subsequent operations will be ignored silently.
+ * However, in the event that that one or more EnvoyHTTPStream operations are currently being
+ * executed, the "cancel" operation gets postponed: the last concurrent operation will invoke
+ * "cancel" at the end.
+ *
+ * <p>Instance of this class start with a state of "BUSY_STARTING". This ensure that if a cancel
```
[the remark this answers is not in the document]
carloseltuerto: You commented an outdated file - this was removed.

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
```diff
@@ -0,0 +1,114 @@
...
+  private static final int CANCEL_BIT = 0x8000;
+  private final AtomicInteger mState = new AtomicInteger(1); // Busy starting.
+  private volatile EnvoyHTTPStream mStream;                  // Cancel can come from any Thread.
+
+  /** Sets the stream. Can only be invoked once. */
+  void setStream(EnvoyHTTPStream stream) {
+    assert mStream == null;
+    mStream = stream;
+    if (cancelNeedsToBeInvoked()) {
+      mStream.cancel(); // Cancel was called meanwhile, so now this is honored.
```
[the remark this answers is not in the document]
carloseltuerto: Changed to "awaitingCancel".

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
```diff
@@ -0,0 +1,114 @@
...
+   */
+  void cancel() {
+    while (true) {
+      int count = mState.get();
+      if ((count & CANCEL_BIT) != 0) {
+        return; // Cancel already invoked.
+      }
+      if (mState.compareAndSet(count, count | CANCEL_BIT)) {
+        if (count == 0) {
+          mStream.cancel(); // Was not busy with other EM operations - cancel right now.
```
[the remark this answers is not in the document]
carloseltuerto: That's CAS magic: https://en.wikipedia.org/wiki/Compare-and-swap

This technique is extensively used in Operating Systems and low level telecom logic. The whole PR is based on this.

The contract is the mutation succeeds only if the original value matches the expected one - this is atomic at the assembly language level: most CPUs have dedicated mnemonics for this operation - extremely efficient.

And this might look like an infinite loop. It is infinite only if many Threads are **eternally** attempting to concurrently change the value. In fact, CAS is pretty bad under heavy contention - in that case it is probably better to go with "synchronized" blocks. In our case, there is none or very little contention. What matters is correctness.

This PR requires CAS knowledge. I am not too sure how disseminate that knowledge.

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
```diff
@@ -0,0 +1,114 @@
...
+
+  private static final int CANCEL_BIT = 0x8000;
+  private final AtomicInteger mState = new AtomicInteger(1); // Busy starting.
+  private volatile EnvoyHTTPStream mStream;                  // Cancel can come from any Thread.
+
+  /** Sets the stream. Can only be invoked once. */
+  void setStream(EnvoyHTTPStream stream) {
+    assert mStream == null;
+    mStream = stream;
+    if (cancelNeedsToBeInvoked()) {
```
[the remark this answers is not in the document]
carloseltuerto: Changed to: `decreaseConcurrentlyRunningStreamOperationsAndReturnTrueIfAwaitingCancel`

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
```diff
@@ -0,0 +1,114 @@
...
+      mStream.cancel(); // Cancel was called meanwhile, so now this is honored.
+    }
+  }
+
+  /** Initiates the sending of one chunk of the request body if the state permits. */
+  void sendData(ByteBuffer buffer, boolean finalChunk) {
+    if (cancelHasAlreadyBeenInvoked()) {
+      return; // Already Cancelled - to late to send something.
+    }
+    // The Envoy Mobile library only cares about the capacity - must use the correct ByteBuffer
```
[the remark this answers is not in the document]
carloseltuerto: Yes, EnvoyMobile does not care about buffer.position.

https://github.com/envoyproxy/envoy-mobile/blob/main/library/common/jni/jni_utility.cc#L167

I added the "length" feature - only capacity was considered before...

library/java/org/chromium/net/impl/CronetUrlRequest.java
```diff
@@ -94,11 +94,10 @@ public final class CronetUrlRequest extends UrlRequestBase {
   }
 
   private static final String X_ENVOY = "x-envoy";
-  private static final String X_ENVOY_SELECTED_TRANSPORT = "x-android-selected-transport";
```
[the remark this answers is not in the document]
carloseltuerto: Done, thanks.

library/java/org/chromium/net/impl/CronetUrlRequestContext.java
carloseltuerto: Well, I believe that the Java Logger is controlled by the C++ layer with Cronet. In fact, I am not sue at all how logging works with Cronet, and how it should work with EM.

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
carloseltuerto: Done, thanks.

library/java/org/chromium/net/impl/CronetBidirectionalState.java
carloseltuerto: Sure - done.

library/java/org/chromium/net/impl/CronetBidirectionalState.java
carloseltuerto: Done.

library/java/org/chromium/net/impl/CronetBidirectionalState.java
carloseltuerto: The complexity lies in the fact that EM and Cronet offer a different API, and because with CAS incurs some internal operations. The exact diagram is quite big, unwieldy. The best short term is to explain to you the general lines.

At this point, the only think that can give you a little not of confidence are the Cronet tests: they are the same as the original ones. Furthermore, I chased for weeks all the race conditions. I ran the test in batches of 500, many times. Good news, the tests are quite good at finding issues under heavy contention (all my M1 CPUs were running hot...)

library/java/org/chromium/net/impl/CronetBidirectionalStream.java
carloseltuerto: Done, thanks.

library/java/org/chromium/net/impl/CronetBidirectionalState.java
carloseltuerto: Thanks! Done.

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
```diff
@@ -0,0 +1,114 @@
...
+ * executed, the "cancel" operation gets postponed: the last concurrent operation will invoke
+ * "cancel" at the end.
+ *
+ * <p>Instance of this class start with a state of "Busy Starting". This ensure that if a cancel
+ * is invoked while the stream is being created, that cancel will be executed only once the stream
+ * is completely initialized. Doing otherwise leads to unpredictable outcomes.
+ */
+final class CancelProofEnvoyStream {
+
+  private static final int CANCEL_BIT = 0x8000;
```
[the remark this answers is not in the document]
carloseltuerto: Done, thanks.

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
```diff
@@ -7,48 +7,58 @@
...
+ * <p>Instances of this class start with a state of "Busy Starting". This ensure that if a cancel
  * is invoked while the stream is being created, that cancel will be executed only once the stream
  * is completely initialized. Doing otherwise leads to unpredictable outcomes.
  */
 final class CancelProofEnvoyStream {
 
   private static final int CANCEL_BIT = 0x8000;
+  /**
+   * mState mainly maintains a counter of how many Stream operations are currently in-flight.
+   * However, when 15 is flipped (0x8000), it indicates that cancel operation hase been requested.
```
RyanTheOptimist, line 29: nit: "However when bit 15 (0x80000) is set, it indicates that the cancel operation has been requested"

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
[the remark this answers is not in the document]
RyanTheOptimist: Thanks for the rename! They're super verbose but very clear.

Though this comment was requesting "previously" instead of "meanwhile" in the comment.

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
```diff
@@ -7,48 +7,58 @@
...
+   */
   private final AtomicInteger mState = new AtomicInteger(1); // Busy starting.
   private volatile EnvoyHTTPStream mStream;                  // Cancel can come from any Thread.
 
   /** Sets the stream. Can only be invoked once. */
   void setStream(EnvoyHTTPStream stream) {
     assert mStream == null;
     mStream = stream;
-    if (cancelNeedsToBeInvoked()) {
+    if (decreaseConcurrentlyRunningStreamOperationsAndReturnTrueIfAwaitingCancel()) {
```
RyanTheOptimist, line 43: It's not obvious to me why this decreases the running count. Else where in the class we do:
```
incrementCount();
doThing()
decrementCount();
```
But here we're just decrementing the count. Is this the 2nd half of some other operation? If so, can you add a comment which explains what the other half of this operation is?

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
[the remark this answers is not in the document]
RyanTheOptimist: I see. This still sounds super scary. Can you file an issue about this confusion between what the Java layer thinks is in the buffer and what the C layer thinks is in the buffer? You clearly don't need to fix this but it would be good to have this written down and we can sort it out later.

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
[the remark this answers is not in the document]
RyanTheOptimist: Yeah, I google'd "compare and set" and I think I understand that primitive (after swapping back in some CS knowledge from years ago :>) and how it is used in this PR. 

But CAS is merely a primitive. The reason it's OK to call cancel() here is because once the CANCEL_BIT is set, mState in unable to be incremented. It can only be decremented. So once mState = CANCEL_BIT it will never change. (Because that's the contract that the other methods agree to). Is that assessment correct?

As for infinite looping... This method will busy loop until any in-flight operations finish, right? As such if any of those operations need access to the current thread, they will deadlock, I think? Presumably, something prevents that from happening. Can you expand on what that is? Also it looks like setStream() finished an operation that gets started elsewhere. Is it possible that cancel() is called before setStream() but for some reason setStream() is blocked on the same thread that cancel() is called from? I think you've thought about this enough that this is all probably true, but it would be good to have this all written down.

library/java/org/chromium/net/impl/CronetBidirectionalState.java
RyanTheOptimist: Since Even only has a few values, we could encode both nextAction and nextState into a single long (and probably even an int) and just return that, right? That being said, thanks for extracting out the isAlreadyFinalState() method. That helps!

library/java/org/chromium/net/impl/CronetBidirectionalState.java
```diff
@@ -0,0 +1,454 @@
...
+  @NextAction
+  int nextAction(@Event final int event) {
+    while (true) {
+      @State final int originalState = mState.get();
+
+      if (isAlreadyFinalState(event, originalState)) {
+        return NextAction.TAKE_NO_MORE_ACTIONS; // No need to loop - this is irreversible.
+      }
+
+      @NextAction final int nextAction; // "final" guarantees that it is assigned exactly once.
```
RyanTheOptimist, line 199: This comment simply describes the java definition of final, right? I don't think that deserves a comment.

library/java/org/chromium/net/impl/CronetBidirectionalState.java
```diff
@@ -0,0 +1,454 @@
...
+        throw new AssertionError("switch is exhaustive");
+      }
+
+      // With "Compare And Swap", the contract is the mutation succeeds only if the original value
+      // matches the expected one - this is atomic at the assembly language level: most CPUs have
+      // dedicated mnemonics for this operation - extremely efficient. And this might look like an
+      // infinite loop. It is infinite only if many Threads are eternally attempting to concurrently
+      // change the value. In fact, "Compare And Swap" is pretty bad under heavy contention - in
+      // that case it is probably better to go with "synchronized" blocks. In our case, there is
+      // none or very little contention. What matters is correctness and efficiency.
```
RyanTheOptimist, line 403: This comment talks about infinite looping but it's in front of an if statement. I think if we want to keep this comment, it belongs elsewhere.

library/java/org/chromium/net/impl/CronetBidirectionalState.java
RyanTheOptimist: As you say, I suspect that this is "correct" and that the testing you did gives good confidence for that. But I have no idea how to look at this code and understand if it is correct. For example, this case:
```
      case Event.READY_TO_FINISH:
        if ((originalState & State.ON_COMPLETE_RECEIVED) != 0 &&
            (originalState & State.READ_DONE) != 0 && (originalState & State.WRITE_DONE) != 0) {
          nextState |= State.DONE;
          nextAction = NextAction.INVOKE_ON_SUCCEEDED;
        } else {
          nextAction = NextAction.CARRY_ON;
        }
        break;
```

It seems like this suggests that we're done when we've finished reading AND we've finished writing (and we get ON_COMPLETE_RECEIVED). That sounds plausible. But are there other things that should have finished too? Probably not :) Are there perhaps two different state machines that we could consider: a Read state machine and a Write state machine? (This matches some of the state machines in Chrome net code and matches the QUIC Stream state machine in RFC 9000 so maybe it makes sense here). If changes to the Read state machine do not affect the Write state machine, this might be a way to simplify the complexity. But maybe this is not the right mental model?

library/java/org/chromium/net/impl/CronetUrlRequestContext.java
RyanTheOptimist: Please file an issue about this and link to it here.

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
```diff
@@ -7,48 +7,58 @@
...
  * However, in the event that that one or more EnvoyHTTPStream operations are currently being
  * executed, the "cancel" operation gets postponed: the last concurrent operation will invoke
  * "cancel" at the end.
  *
- * <p>Instance of this class start with a state of "Busy Starting". This ensure that if a cancel
+ * <p>Instances of this class start with a state of "Busy Starting". This ensure that if a cancel
  * is invoked while the stream is being created, that cancel will be executed only once the stream
  * is completely initialized. Doing otherwise leads to unpredictable outcomes.
  */
 final class CancelProofEnvoyStream {
```
RyanTheOptimist, line 24: Are there unit tests of this class? I don't think I see them.

RyanTheOptimist (commented):
Thanks for the thoughtful response! It looks like CancelProofEnvoyStream could be extracted from the PR as a stand-alone change (even if it wouldn't be used yet). I would definitely consider doing that.

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
carloseltuerto: Done, thanks.

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
carloseltuerto: The other half is done by the Constructor: the counter starts with a value of "1' (not "0" as someone would expect.

Comment added, thanks.

carloseltuerto:
Ryan wrote:

[I see. This still sounds super scary. Can you file an issue about this confusion between what the Java layer thinks is in the buffer and what the C layer thinks is in the buffer? You clearly don't need to fix this but it would be good to have this written down and we can sort it out later.](https://github.com/envoyproxy/envoy-mobile/pull/2164#discussion_r865090104)

I created an "issue" for this: #2247

library/java/org/chromium/net/impl/CronetBidirectionalState.java
carloseltuerto: Yeah - lately Java coders tend to remove the "final" because it is not rally needed. In this case, it is important to keep it for maintenance purposes.

Removed.

library/java/org/chromium/net/impl/CronetBidirectionalState.java
carloseltuerto: Moved just above the `while (true)`

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
carloseltuerto: Added come tests.

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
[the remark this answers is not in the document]
carloseltuerto: Discussed in person, and comments were added.

library/java/org/chromium/net/impl/CronetBidirectionalState.java
carloseltuerto: Discussed in person. Conclusion: it is doable to to have a state for "read" and another "state" for write. However, there is some added trickiness by following that path: can foster some race conditions. Nothing that can't be circumvented, but this is not trivial. Status quo for the moment.

library/java/org/chromium/net/impl/CancelProofEnvoyStream.java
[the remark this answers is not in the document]
carloseltuerto: Issue #2247 filed.

library/java/org/chromium/net/impl/CronetBidirectionalState.java
carloseltuerto: Discussed in person. Status quo for the moment.

RyanTheOptimist (approved):
Thanks for doing all this, Charles. I'm slightly apprehensive about the maintainability / complexity of the state machine, but we and always address that later if it turns out to be problematic. LGTM

@alyssawilk care to take a look?

library/java/org/chromium/net/impl/CronetUrlRequestContext.java
carloseltuerto: Issue filed: #2262

RyanTheOptimist (approved):

RyanTheOptimist:
> @RyanTheOptimist I strongly suspect my pass isn't going to add much value to the review you've already done. I'd be inclined to merge as-is and as you say we can revisit complexity later if we need to. SGTY or you actively prefer I do a pass?

Works for me! I'll go ahead and merge now.

PR #2165, opened 2022-04-18
http client: add cancel log and limit callback to open streams

goaway:
Description: Adds a debug log statement to help more clearly indicate when the a stream has been cancelled via the programmatic API. Additionally, noticed that the cancellation callback could theoretically be called for a manual stream that was not currently open, and moved the call into the guard.
Risk Level: Low
Testing: CI

### DIFF
--- a/library/common/http/client.cc
+++ b/library/common/http/client.cc
@@ -517,16 +517,15 @@ void Client::cancelStream(envoy_stream_t stream) {
     ScopeTrackerScopeState scope(direct_stream.get(), scopeTracker());
     removeStream(direct_stream->stream_handle_);
 
+    ENVOY_LOG(debug, "[S{}] application cancelled stream", stream);
+    direct_stream->callbacks_->onCancel();
+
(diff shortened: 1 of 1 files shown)


library/common/http/client.cc
```diff
@@ -517,14 +517,16 @@ void Client::cancelStream(envoy_stream_t stream) {
...
-
     // Since https://github.com/envoyproxy/envoy/pull/13052, the connection manager expects that
     // response code details are set on all possible paths for streams.
     direct_stream->setResponseDetails(getCancelDetails());
 
     // Only run the reset callback if the stream is still open.
     if (stream_was_open) {
+      ENVOY_LOG(debug, "[S{}] application cancelled stream", stream);
+
+      direct_stream->callbacks_->onCancel();
```
goaway, line 528: This distinction with the callback only actually matters when the stream is being manually controlled; in the default case if the stream not being present in the map, it's equivalent to it not being open.

Augustyniak (approved):

PR #2171, opened 2022-04-21
Add documentation on renewing GPG keys

jpsim:
These had expired, and rather than have to hunt around the Internet for how to renew them when this happens again in 2024, document the steps here.


goaway (approved):
Sounds like we need to have a discussion about where to store the private key?

jpsim:
> Sounds like we need to have a discussion about where to store the private key?

I've been made aware of a shared Envoy LastPass folder, so I've uploaded it there as an item titled "Sonatype EnvoyMobile".

goaway:
Great, thanks @jpsim!

PR #2175, opened 2022-04-21
Update podspec

jpsim:
* Fix summary string.
* Automatically set version based on the contents of the `VERSION` file.
* Add `APPLICATION_EXTENSION_API_ONLY` so Envoy Mobile can be used in app extension targets if desired.
* Advertise support for Swift 5.6
* Add `Gemfile` to pin what version of CocoaPods we use.
* Add `release_validation` GitHub Action to run on PRs and `main` commits to validate the podspec. Other release validation checks can be added here in the future.

Risk Level: Low, we haven't pushed CocoaPods releases in over two years
Testing: CI job
Docs Changes: N/A
Release Notes: N/A


.github/workflows/release_validation.yml
```diff
@@ -0,0 +1,21 @@
...
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  cocoapods_lib_lint:
+    name: cocoapods_lib_lint
+    runs-on: macos-11
+    timeout-minutes: 120
```
Augustyniak, line 13: Does the timeout has to be so high? Just curious

EnvoyMobile.podspec
```diff
@@ -1,17 +1,18 @@
 Pod::Spec.new do |s|
     s.name = 'EnvoyMobile'
-    s.version = '0.4.5'
+    s.version = `cat VERSION`.chomp
```
Augustyniak, line 3: nice. love it

Augustyniak (approved):
Nice! Great to see this being improved. Thank you for working on this

.github/workflows/release_validation.yml
jpsim: No, just copy pasted from a different action. Reduced to 20. The last run took 5 minutes, so 4x headroom should be plenty.

Augustyniak (approved):

PR #2177, opened 2022-04-21
build: set 'local' tag on android_dist builds

goaway:
Description: a061f3dcdd2cdeec50a171973eabf0660bc9a8c6 caused build failures when attempting to build the android_dist target locally. This ensures the target can write the to the `dist/` directory when built.
Risk Level: Low
Testing: Local & CI

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -72,6 +72,9 @@ genrule(
     touch $@
     """,
     stamp = True,
+    # This action writes to a non-hermetic output location, so running it
+    # remotely isn't currently possible.
+    tags = ["local"],
 )
 
 define_kt_toolchain(
--- a/BUILD
+++ b/BUILD
@@ -17,8 +17,8 @@ unzip -o $< -d dist/
 touch $@
 """,
     stamp = True,
-    # This action writes to a non-hermetic output location, so running it
-    # remotely isn't currently possible.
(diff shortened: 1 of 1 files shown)


.bazelrc
```diff
@@ -16,6 +16,7 @@ build --macos_minimum_os=10.14
 build --ios_minimum_os=12.0
 build --ios_simulator_device="iPhone 13"
 build --ios_simulator_version=15.2
+build --spawn_strategy=local
```
jpsim, line 19: Can you please add a comment with context here? Or likely even better, a link to a GitHub issue tracking re-adding sandboxing?

goaway: Well, I expect sandboxing doesn't work because we actually try to output to local `dist/` directory. I can add a comment.

jpsim (dismissed):

BUILD
```diff
@@ -72,6 +72,9 @@ genrule(
     touch $@
     """,
     stamp = True,
+    # This action writes to a non-hermetic output location, so running it
+    # remotely isn't currently possible.
```
keith, line 76: this isn't about running it remotely, just running it at all since it's write files outside of bazel's execroot

goaway: Sure, I just copied the comment, but you're right of course - I'll update it.

jpsim (approved):

Issue #2101, opened 2022-03-15
arm64 darwin binaries of bazelisk fail to link Swift tests


jpsim:
Since merging arm64 support for bazelisk yesterday in https://github.com/envoyproxy/envoy-mobile/pull/2096, running the Swift tests fails at the linking stage with the following error on my M1 Max MacBook Pro:

<details>

<summary>Logs</summary>

```console
./bazelw test //test/swift:test --config=ios
Starting local Bazel server and connecting to it...
DEBUG: /private/var/tmp/_bazel_jsimard/7140f038a5c13b01a171c93151a000da/external/rules_proto_grpc/repositories.bzl:302:18: Dependency 'bazel_skylib' has already been declared with a different version. Found urls=("https://github.com/bazelbuild/bazel-skylib/releases/download/1.1.1/bazel-skylib-1.1.1.tar.gz",), expected ("https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/archive/be3b1fc838386bdbea39d9750ea4411294870575.tar.gz", "https://github.com/bazelbuild/bazel-skylib/archive/be3b1fc838386bdbea39d9750ea4411294870575.tar.gz")
DEBUG: /private/var/tmp/_bazel_jsimard/7140f038a5c13b01a171c93151a000da/external/rules_proto_grpc/repositories.bzl:302:18: Dependency 'com_google_protobuf' has already been declared with a different version. Found urls=("https://github.com/protocolbuffers/protobuf/releases/download/v3.16.0/protobuf-all-3.16.0.tar.gz",), expected ("https://mirror.bazel.build/github.com/protocolbuffers/protobuf/archive/19ca196e4dcc8ff03136c950cf1bc298d5e1ce1e.tar.gz", "https://github.com/protocolbuffers/protobuf/archive/19ca196e4dcc8ff03136c950cf1bc298d5e1ce1e.tar.gz")
DEBUG: /private/var/tmp/_bazel_jsimard/7140f038a5c13b01a171c93151a000da/external/rules_proto_grpc/repositories.bzl:311:18: Dependency 'zlib' has already been declared with a different rule kind. Found bind, expected http_archive
[...]

PR #2185, opened 2022-04-22
[bazelw] Use arm64 binary on darwin-arm64

jpsim:
Now that we've pulled in https://github.com/envoyproxy/envoy/pull/20852, I'm no longer seeing linker errors when running Swift tests for iOS as reported in https://github.com/envoyproxy/envoy-mobile/issues/2101.

Risk Level: Low to the libraries, Moderate to local development on an M1 Mac.
Testing: `./bazelw test //test/swift:test --config=ios` passes
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/bazelw
+++ b/bazelw
@@ -20,9 +20,7 @@ fi
 bazel_platform="$bazel_os-$bazel_arch"
 case "$bazel_platform" in
   darwin-arm64)
-    # TODO: Support M1 once this is resolved: https://github.com/envoyproxy/envoy-mobile/issues/2101
-    bazel_platform="darwin-amd64"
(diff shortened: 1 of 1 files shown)


keith (approved):

jpsim:
Verified with a few common build+test targets: `ios_dist`, `android_dist`, `/test/cc/...`, `//test/swift:test`

The only thing I know to be broken is `./bazelw test //test/swift:test --config=ios --ios_multi_cpus=x86_64` due to https://github.com/google/xctestrunner/pull/40, but that's not something I've ever needed in my local development.

Something else may be broken, but I haven't seen it.

PR #2188, opened 2022-04-22
bazel: Include debug info in Android release binary

Augustyniak:
Description: Include debug info in the Android release binary. This allows us to have an objdump file with all of the relevant symbols. The change is needed as we moved from symbol mapping table files to symbol mapping files in https://github.com/envoyproxy/envoy-mobile/commit/f955257e940b7bf3a9f4463fd2f061d3eea6f6ea
Risk Level: low, tooling only
Testing: N/A
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -76,10 +76,6 @@ build:tsan-dev --copt -DEVENT__DISABLE_DEBUG_MODE
 # https://github.com/google/sanitizers/issues/953
 build:tsan-dev --test_env="TSAN_OPTIONS=report_atomic_races=0"
 
-# Exclude debug info from the release binary since it makes it too large to fit
-# into a zip file. This shouldn't affect crash reports.
-build:release-common --define=no_debug_info=1
-
 # Compile releases optimizing for size (eg -Os, etc).
 build:release-common --config=sizeopt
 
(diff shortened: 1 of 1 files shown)


Augustyniak:
this is increasing the size of objdump files 4x - from around 250MB to around 1GB.

.bazelrc
```diff
@@ -76,10 +76,6 @@ build:tsan-dev --copt -DEVENT__DISABLE_DEBUG_MODE
 # https://github.com/google/sanitizers/issues/953
 build:tsan-dev --test_env="TSAN_OPTIONS=report_atomic_races=0"
 
-# Exclude debug info from the release binary since it makes it too large to fit
-# into a zip file. This shouldn't affect crash reports.
-build:release-common --define=no_debug_info=1
```
keith, line 81: need to change to use for ios but not both instead, because the download size is an issue otherwise

Augustyniak: OK, that makes sense

keith (approved):

jpsim:
FYI we won't be able to deploy this to our apps until we merge https://github.com/envoyproxy/envoy-mobile/pull/2186, or a different fix for that SIGABRT crash.

Augustyniak:
@keith mind re-approving? I needed to fix some linter issues and force push to sign-off on my changes.

keith (approved):

PR #2191, opened 2022-04-23
Fix race condition when posting CronetMetrics

### LINKED ISSUE #2136: unexpected java_tests_mac failure
https://github.com/envoyproxy/envoy-mobile/runs/5744148090?check_suite_focus=true

WARNING: /Users/runner/work/envoy-mobile/envoy-mobile/test/common/jni/BUILD:17:10: in cc_binary rule //test/common/jni:libquic_test_server_jni.so: target '//test/common/jni:libquic_test_server_jni.so' depends on deprecated target '@local_jdk//:jni_header': Don't depend on targets in the JDK workspace; use @bazel_tools//tools/jdk:current_java_runtime instead (see https://github.com/bazelbuild/bazel/issues/5594)

carloseltuerto:
At the end of a request, mRequestFinishedListener is used to post the CronetMetrics. Before doing so, two events must have occurred first: the "final user callback" and the "final Network callback". The Thread involved with the last of these two events is in charge of the posting the CronetMetrics - this is intrinsically racy. This PR ensures a proper behaviour.

Description: Solves CI test flakiness: unexpected java_tests_mac failure: test/java/org/chromium/net:net_tests 
Risk Level: N/A (Cronet is still in development)
Testing: CI, and local multi runs
Docs Changes: N/A
Release Notes: N/A
Fixes #2136

### DIFF
--- a/library/java/org/chromium/net/impl/AtomicCombinatoryState.java
+++ b/library/java/org/chromium/net/impl/AtomicCombinatoryState.java
@@ -1,0 +1,38 @@
+package org.chromium.net.impl;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * CAS logic based class providing a mean to ensure that the last awaited "stateEvent" will be
+ * identified as so. Typically a "stateEvent" is a single bit flip.
+ *
+ * <p>This class is Thread Safe.
+ */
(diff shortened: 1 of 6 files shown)


alyssawilk (approved):
wow, thanks for tracking this one down!

PR #2200, opened 2022-04-25
Update releasing.rst

jpsim:
The version in the podspec no longer needs to be manually bumped as of #2175.


Augustyniak (approved):
I wonder whether we should not start requiring folks to update `version_history.rst` file every time they make a change in the repo. That's how it work in the main Envoy repo I think. What do you think?

jpsim:
Yes we should try to remember to keep it up to date as we go.

PR #2201, opened 2022-04-26
Release 0.4.6

jpsim:
Adding changes since 0.4.5 to release notes


jpsim:
I manually went through the changes in https://github.com/envoyproxy/envoy-mobile/compare/v0.4.5...main and added the most relevant ones.

jpsim:
Because the CocoaPods podspec points to the precompiled release asset, I'll need to think of a good way to make that work when we change versions. Punting on that until after the next release.

Augustyniak (approved):

jpsim:
Changed version to 0.4.6 after discussing with @Augustyniak because we want to clean up some builder default values before bumping the minor version number.

Augustyniak (approved):

PR #2210, opened 2022-04-26
ci: Re-enable release_validation job

jpsim:
This doesn't handle version bumps so I had to temporarily disable it to release 0.4.6.


jpsim:
Actually I don't think we should re-enable this until it works when bumping versions.

PR #2211, opened 2022-04-26
bazel: remove rules_cc references

keith:
I'm removing rules_cc from
envoy as well https://github.com/envoyproxy/envoy/pull/19760 so this
mirrors that.


snowp (approved):
Thanks!

keith:
I'm leaving rules_java now since we're relying on a newer version of that repo's toolchain support now

snowp (approved):

PR #2214, opened 2022-04-26
build: simplify jnilib copy

snowp:
The genrule doesn't work well with sandboxing on Linux since we're modifying files we're not supposed to touch:

```
chmod: changing permissions of 'bazel-out/aarch64-dbg/bin/loop/test/kt/io/envoyproxy/loop/logging/libloop_test_lib.so': Read-only file system
```

This simplifies the rule to just copy the .so into a .jnilib.

Risk Level: Low, build only
Testing: Existing tests
Docs Changes: n/a
Release Notes: n/a


keith:
i don't understand what this was trying to do before, but maybe we can use `copy_file` from skylib to simplify?

keith (approved):

PR #2220, opened 2022-04-27
cleanup: remove the unused stats sink metrics_service

abeyad:
This was discovered because running tools/vscode_compdb.sh was failing
due to build failures in metrics_service.  A discussion on the
[envoy-mobile slack channel](https://envoyproxy.slack.com/archives/CKQ2LK23G/p1650986208103949)
mentioned that the metrics_service sink is no longer used and can be removed.

Risk Level: Low
Testing: Relying on CI
Docs Changes: N/A
Release Notes: N/A


alyssawilk:
thanks for the clean up!
alas repokitteh is an envoy-repo thing so the slash actions don't work on the envoy mobile repo.  I'll assign to one of the folks form Lyft and he can reassign as makes sense :-)

jpsim:
Thanks for the PR! This does indeed appear unused.

PR #2221, opened 2022-04-27
Fix Envoy Mobile bug where writing prevents the read loop from running,

### LINKED ISSUE #2213: onResponseHeaders not called on continuous sendData
cc: @alyssawilk @goaway 

Pushing more ByteBuffers always has priority over performing the onResponseHeaders callback. You could have received the ResponseHeaders one hour ago - still no callback indicating so if you keep pushing more data.

This problem is systematic on H2 with explicit flow control when using the EM Network Thread to perform the callback logic.

There is a specific test for this in Cronet: https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/android/test/javatests/src/org/chromium/net/BidirectionalStreamTest.java;l=465
This test is being ported, but is currently failing due to this issue.

And there is a PR with a simple test demonstrating the failure: #2212

RyanTheOptimist:
Fix Envoy Mobile bug where writing prevents the read loop from running,
by scheduling repeated onSendWindowAvailable calls in the next
dispatcher iteration.

Adds the regression test Charles cooked up in #2212

Fixes #2213 

Risk Level: Low
Testing: New regression test
Docs Changes: N/A
Release Notes: N/A


RyanTheOptimist:
@alyssawilk please take a look. I'm not entirely sure if this is the right approach.

RyanTheOptimist:
FYI: This PR is not complete; I'm looking for guidance if this approach seems reasonable.

library/common/http/client.cc
```diff
@@ -471,9 +471,11 @@ void Client::sendData(envoy_stream_t stream, envoy_data data, bool end_stream) {
     if (direct_stream->explicit_flow_control_ && !end_stream) {
       if (direct_stream->read_disable_count_ == 0) {
         // If there is still buffer space after the write, notify the sender
-        // that send window is available.
+        // that send window is available, on the next dispatcher iteration.
         direct_stream->wants_write_notification_ = false;
-        direct_stream->callbacks_->onSendWindowAvailable();
+        scheduled_callback_ = dispatcher_.createSchedulableCallback(
```
alyssawilk, line 476: can't remember offhand: do we need to create a new one each time, or only the one time?

alyssawilk (commented):
yeah I think this is the right away to go about it, thanks for tackling it! I'd be inclined to unit test as well as you get it ready.

library/common/http/client.cc
```diff
@@ -471,9 +471,11 @@ void Client::sendData(envoy_stream_t stream, envoy_data data, bool end_stream) {
     if (direct_stream->explicit_flow_control_ && !end_stream) {
       if (direct_stream->read_disable_count_ == 0) {
         // If there is still buffer space after the write, notify the sender
-        // that send window is available.
+        // that send window is available, on the next dispatcher iteration.
```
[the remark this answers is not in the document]
RyanTheOptimist: Sure. Done.

library/common/http/client.cc
RyanTheOptimist: Oh nice. That seems to work. Done!

alyssawilk (approved):
LBTM :-)

RyanTheOptimist (commented):
Unit test coverage added via the changes to the existing tests.

RyanTheOptimist:
@danzh2010 FYI

RyanTheOptimist:
@jpsim Would you mind taking a look at this? Or redirect if someone else would be a better choice?

RyanTheOptimist:
@goaway as per our discussion in the community meeting, what do you think of this change?

goaway (dismissed):
Based on our discussion this past week, and given the fact that this is all effectively guarded by the explicit flow control flag (which is off by default), I think it's reasonable to move forward with this.

Thanks for the discussion and investigation @RyanTheOptimist and @carloseltuerto.

goaway:
Rerunning the failed CI job.

library/common/http/client.cc
alyssawilk: apparently it didn't work - can we call out why we always need to create in a comment?

RyanTheOptimist: Good point. Done.

alyssawilk (approved):
ugh, I want to prematurely optimize that callback creation so badly :-P

goaway (approved):

carloseltuerto:
When removing this line,

[`@Ignore("https://github.com/envoyproxy/envoy-mobile/issues/2213")`](https://github.com/envoyproxy/envoy-mobile/blob/main/test/java/org/chromium/net/BidirectionalStreamTest.java#L453)

This unfortunately, this ends up with crash more than 50% of the time:
```
#
#
#  SIGSEGV (0xb) at pc=0x0000000105cb1b00, pid=77193, tid=32123
#
# JRE version: OpenJDK Runtime Environment Zulu11.50+19-CA (11.0.12+7) (build 11.0.12+7-LTS)
# Java VM: OpenJDK 64-Bit Server VM Zulu11.50+19-CA (11.0.12+7-LTS, mixed mode, tiered, compressed oops, g1 gc, bsd-aarch64)
# Problematic frame:
# May 10, 2022 7:06:33 PM io.netty.handler.logging.LoggingHandler channelInactive
INFO: [id: 0xc19fa4fa, L:/0:0:0:0:0:0:0:0:61521] INACTIVE
May 10, 2022 7:06:33 PM io.netty.handler.logging.LoggingHandler channelUnregistered
INFO: [id: 0xc19fa4fa, L:/0:0:0:0:0:0:0:0:61521] UNREGISTERED
V  [libjvm.dylib+0x379b00]  jni_GetObjectClass+0x120
#
# No core dump will be written. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
```

PR #2226, opened 2022-04-28
bazel: Test with `6.0.0-pre.20220414.2`

jpsim:
To get ahead of any issues with https://github.com/envoyproxy/envoy/pull/21070


keith:
needs https://github.com/envoyproxy/envoy-mobile/pull/2227

PR #2228, opened 2022-04-28
use 64 bit emulators for test

snowp:
This is much more common than 32 bit, so it makes sense to test for this over x86

Risk Level: Low
Testing: Existing tests
Docs Changes: n/a
Release Notes: n/a


snowp:
@keith @jpsim

jpsim:
Was there an issue on 32-bit archs you were investigating?

jpsim (approved):

snowp:
Yep ran into a hard to debug JNI issue that only happened on x86 specifically, didn't feel necessary to debug that given x86 is not really a thing for actual mobile devices

PR #2233, opened 2022-05-02
swift: add DrString tool & fix documentation comment issues

jpsim:
This will help ensure that Swift docstrings are in the correct format and that all documented interfaces are complete.


Augustyniak (approved):
This is great 💯

PR #2237, opened 2022-05-02
network: handle no network connectivity state

Augustyniak:

Description: The original thought behind this PR was for Envoy Mobile to be able to differentiate between "no network connectivity" and wi-fi connectivity cases. This is what dictated changes in `_reachability_callback` method. During testing I noticed that for every transition from "no connection" to "wi-fi connection" I was seeing two updates: first saying that we have wi-fi connectivity and the second one saying that there is not connectivity.

It turned out that this was always the case and my change to remove `if (flags == 0)` from the implementation of `_reachability_callback` just surfaced this issue. After some googling I found https://github.com/ashleymills/Reachability.swift/issues/95#issuecomment-229401474 which seems to provide an explanation for as to why I saw 2 consecutive updates. The issue have been caused by us passing a fake domain to the `SCNetworkReachabilityCreateWithName` method which is why I decided to move to the `SCNetworkReachabilityCreateWithAddress` method instead. That's something that AFNetworking does for cases when it does not have access to a domain - see https://github.com/AFNetworking/AFNetworking/blob/master/AFNetworking/AFNetworkReachabilityManager.m#L150-L153.

Risk Level: medium
Testing: Manual
Docs Changes:
Release Notes:

### DIFF
--- a/library/objective-c/EnvoyNetworkMonitor.m
+++ b/library/objective-c/EnvoyNetworkMonitor.m
@@ -7,7 +7,7 @@
 #import <SystemConfiguration/SystemConfiguration.h>
 #import <netinet/in.h>
 
-static NSString * network_type_description(envoy_network_t network_type) {
+static NSString *network_type_description(envoy_network_t network_type) {
   if (network_type == ENVOY_NET_GENERIC) {
     return @"Generic";
   } else if (network_type == ENVOY_NET_WLAN) {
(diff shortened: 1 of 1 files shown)


Augustyniak:
I am still not sure as to whether the proposed solution is the right one as:
1. I do not see an initial update of the network state (after the app is launched). This can be fixed.
2. The update that informs us about wi-fi -> no connection transition is fired only after we re-enabled the wi-fi 🤔 (and is followed by the next update which tells us that wi-fi connection is back).

PR #2240, opened 2022-05-03
network: perform post-DNS connection drain on a per-host basis

goaway:
Description: Limits connection draining to be based on specific hosts as each host's DNS resolution occurs following a triggered DNS refresh.
Risk Level: Moderate
Testing: Additional Unit Coverage

### DIFF
--- a/library/common/network/configurator.cc
+++ b/library/common/network/configurator.cc
@@ -181,11 +181,12 @@ void Configurator::reportNetworkUsage(envoy_netconf_t configuration_key, bool ne
 }
 
 void Configurator::onDnsResolutionComplete(
-    const std::string& host, const Extensions::Common::DynamicForwardProxy::DnsHostInfoSharedPtr&,
+    const std::string& resolved_host,
+    const Extensions::Common::DynamicForwardProxy::DnsHostInfoSharedPtr&,
     Network::DnsResolver::ResolutionStatus) {
   if (enable_drain_post_dns_refresh_) {
     // Check if the set of hosts pending drain contains the current resolved host.
-    if (hosts_to_drain_.erase(host) == 0) {
+    if (hosts_to_drain_.erase(resolved_host) == 0) {
       return;
     }
 
@@ -194,9 +195,11 @@ void Configurator::onDnsResolutionComplete(
     // drain connections. It may be possible to refine this logic in the future.
     // TODO(goaway): check the set of cached hosts from the last triggered DNS refresh for this
     // host, and if present, remove it and trigger connection drain for this host specifically.
-    ENVOY_LOG_EVENT(debug, "netconf_post_dns_drain_cx", host);
+    ENVOY_LOG_EVENT(debug, "netconf_post_dns_drain_cx", resolved_host);
+
+    // Pass predicate to only drain connections to the resolved host (for any cluster).
     cluster_manager_.drainConnections(
-        [host](const Upstream::Host& host) { return host.hostname() == host; });
+        [resolved_host](const Upstream::Host& host) { return host.hostname() == resolved_host; });
   }
 }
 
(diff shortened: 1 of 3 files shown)


library/common/network/configurator.cc
```diff
@@ -181,25 +181,32 @@ void Configurator::reportNetworkUsage(envoy_netconf_t configuration_key, bool ne
...
 void Configurator::onDnsResolutionComplete(
-    const std::string& host, const Extensions::Common::DynamicForwardProxy::DnsHostInfoSharedPtr&,
+    const std::string& resolved_host,
+    const Extensions::Common::DynamicForwardProxy::DnsHostInfoSharedPtr&,
     Network::DnsResolver::ResolutionStatus) {
-  if (enable_drain_post_dns_refresh_ && pending_drain_) {
-    pending_drain_ = false;
+  if (enable_drain_post_dns_refresh_) {
+    // Check if the set of hosts pending drain contains the current resolved host.
+    if (hosts_to_drain_.erase(resolved_host) == 0) {
```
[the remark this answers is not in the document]
goaway: You're correct that this would be wrong if this was a vector, but `hosts_to_drain_` is an `absl::flat_hash_set`.

Here is what is being called:
https://github.com/abseil/abseil-cpp/blob/master/absl/container/flat_hash_set.h#L232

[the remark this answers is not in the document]
goaway: Another way to read this is "return if nothing was erased".

library/common/network/configurator.cc
```diff
@@ -229,9 +236,15 @@ void Configurator::refreshDns(envoy_netconf_t configuration_key, bool drain_conn
...
     ENVOY_LOG_EVENT(debug, "netconf_refresh_dns", std::to_string(configuration_key));
-    pending_drain_ = drain_connections && enable_drain_post_dns_refresh_;
-    // TODO(goaway): capture the list of currently cached hosts here in a set, to be checked
-    // for draining when DNS resolutions occur.
+
+    if (drain_connections && enable_drain_post_dns_refresh_) {
+      dns_cache->iterateHostMap(
+          [&](absl::string_view host,
+              const Extensions::Common::DynamicForwardProxy::DnsHostInfoSharedPtr&) {
+            hosts_to_drain_.emplace(host);
```
Augustyniak, line 244: should we explicitly empty `hosts_to_drain_` before we start adding hosts here? I think that we cannot end up in a state where we have dups in here but what if we had a host in the past that does not exist anymore (not sure whether this can happen)? 


Imagine we had DNS refresh 1) that included host x and y and DNS refresh 2) that included y host only. We want `hosts_to_drain_` to contain only `y` after the second DNS refresh is initialized

Augustyniak, line 244: should we explicitly empty `hosts_to_drain_` before we start adding hosts here? I think that we cannot end up in a state where we have dups in here but what if we had a host in the past that does not exist anymore (not sure whether this can happen)? 


Imagine we had DNS refresh 1) that included host x and y and DNS refresh 2) that included y host only. We want `hosts_to_drain_` to contain only `y` after the second DNS refresh is initialized

Augustyniak (approved):
with a comment/question

library/common/network/configurator.cc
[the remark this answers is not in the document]
jpsim: ok thanks for sharing that. I still haven't come around to appreciating this overall C++ pattern of returning iterators from functions, but this isn't the place to debate that.

jpsim (approved):

library/common/network/configurator.cc
goaway: I'm not actually sure if hosts can ever cease to exist in Envoy's cache implementation, since at the end of their TTL, I think they're simply refreshed. That said, previous re-resolutions should still exist even if this is triggered again before prior resolutions complete - in other words, I think the race where a host might persist exists irrespective of whether we call `clear` here.

In the case you describe:
The resolutions from 1) should still exist and be outstanding even if 2) is triggered, and thus we should still observe resolutions for x and y.

Augustyniak: Is it possible that 1) DNS refresh is cancelled and we never receives completion callback associated with it?

goaway: My understanding is the callback we're listening for should fire regardless of the result state of the resolution. @mattklein123 added it because other callbacks didn't necessarily.

goaway: `hosts_to_drain_` is a set, and should not contain dups. I don't think emptying the set in here or above actually closes the window on the small window for a race where a host ceases to exist between us enumerating hosts and then triggering the refresh.

I think putting a time bound on outstanding resolutions would, but that adds additional complexity, and given the narrow window, I'm not sure if it's really needed.

library/common/network/configurator.cc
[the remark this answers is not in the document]
goaway: I agree working with iterators is a bit of a mental adjustment. That said, in this case, the function is simply returning the number of elements that were deleted as a `size_type`. :)

Issue #2239, opened 2022-05-02
Fix `android_tests` CI jobs with bazel 6.x


jpsim:
When updating to bazel 6.x, the `android_tests` CI job [fails](https://github.com/envoyproxy/envoy-mobile/runs/6260254234?check_suite_focus=true) with the following error:

> Error: Exception in thread "main" java.lang.UnsupportedClassVersionError: com/google/testing/junit/runner/BazelTestRunner has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0

Issue #2238, opened 2022-05-02
Fix `android_build` CI jobs with bazel 6.x


jpsim:
When updating to bazel 6.x, the `android_build` CI job [fails](https://github.com/envoyproxy/envoy-mobile/runs/6260253760?check_suite_focus=true) with the following error:

> ERROR: /private/var/tmp/_bazel_runner/b225883c33850f3f191a570e7b70ec29/external/bazel_tools/tools/jdk/BUILD:29:19: While resolving toolchains for target @bazel_tools//tools/jdk:current_java_runtime: no matching toolchains found for types @bazel_tools//tools/jdk:runtime_toolchain_type

PR #2243, opened 2022-05-03
bazel: test with version `6.0.0-pre.20220421.3`

jpsim:
To see if #2238 or #2239 are resolved.


jpsim:
Some CI issues with workers running out of disk space ([Slack](https://envoyproxy.slack.com/archives/C02QMNG92A3/p1651605933888989)).

jpsim:
Disk space aside, this does appear to fix #2239 but not #2238.

PR #2250, opened 2022-05-05
CancelProofEnvoyStream

carloseltuerto:
Description: Wrapper for the EnvoyStream guaranteeing that a another stream method can't be called after a "cancel"
Risk Level: None - Cronvoy only
Testing: CI
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/test/java/org/chromium/net/impl/CancelProofEnvoyStreamTest.java
+++ b/test/java/org/chromium/net/impl/CancelProofEnvoyStreamTest.java
@@ -17,7 +17,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicReference;
 
 import io.envoyproxy.envoymobile.engine.EnvoyHTTPStream;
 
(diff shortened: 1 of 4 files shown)


test/java/org/chromium/net/impl/CancelProofEnvoyStreamTest.java
```diff
@@ -0,0 +1,379 @@
...
+    mSendDataBlock.open();
+    mReadDataBlock.open();
+    // Wait for all the Threads to die.
+    for (Thread thread : threads) {
+      thread.join();
+    }
+    assertThat(mCancelInvocationCount.get()).isOne();
+  }
+
+  private class MockedStream extends EnvoyHTTPStream {
```
RyanTheOptimist, line 347: Can you add a comment which explains the various Block/Latch behaviors of this class?

test/java/org/chromium/net/impl/CancelProofEnvoyStreamTest.java
```diff
@@ -0,0 +1,379 @@
...
+
+    assertThat(mCancelInvocationCount.get()).isOne();
+  }
+
+  @Test
+  public void cancel_withoutStreamSet() {
+    // Does nothing because the "Concurrently Running Stream Operations Count" starts with "1".
+    // This is the desired outcome - the cancel is postponed. Once the stream is set, the next
+    // Stream Operation will invoke "cancel".
+    cancelProofEnvoyStream.cancel();
```
RyanTheOptimist, line 254: Should this check that the cancel invocation count is 0?

RyanTheOptimist (dismissed):
Looks great. Two tiny comments.

test/java/org/chromium/net/impl/CancelProofEnvoyStreamTest.java
carloseltuerto: It is the `MockedStream` that can increment the cancel invocation count, and this test does not invoke `setStream`: it is pointless to verify it.

Added a comment.

test/java/org/chromium/net/impl/CancelProofEnvoyStreamTest.java
carloseltuerto: Added a comment. Short story: a `ConditionVariable` is in reality a `CountDownLatch` where the initial count is "1".

RyanTheOptimist (approved):

Issue #1575, opened 2021-07-07
Cronvoy: support Android TLS mechanism


carloseltuerto:
Android must control the available root certificates, as users can choose to revoke certificates on their Android devices. Hence, Android provides a Java API to verify a Server certificate. As usual, Cronet has a Java layer allowing JNI interactions with related Android system libraries. That Cronet layer is fairly complex:

- [```AndroidNetworkLibrary.verifyServerCertificates```](https://source.chromium.org/chromium/chromium/src/+/main:net/android/java/src/org/chromium/net/AndroidNetworkLibrary.java;l=108): that's the main method.
- [```AndroidCertVerifyResult```](https://source.chromium.org/chromium/chromium/src/+/main:net/android/java/src/org/chromium/net/AndroidCertVerifyResult.java): this is the response to the above method. Requires some JNI invocations from the C++ layer to extirpate the data.
- [```X509Util```](https://source.chromium.org/chromium/chromium/src/+/main:net/android/java/src/org/chromium/net/X509Util.java): that's where the subtle details reside. These are more JNI invocations in there.

Also, to help with testing, there are two methods dealing with root certificates:
- [```AndroidNetworkLibrary.addTestRootCertificate```](https://source.chromium.org/chromium/chromium/src/+/main:net/android/java/src/org/chromium/net/AndroidNetworkLibrary.java;l=132)
- [```AndroidNetworkLibrary.clearTestRootCertificates```](https://source.chromium.org/chromium/chromium/src/+/main:net/android/java/src/org/chromium/net/AndroidNetworkLibrary.java;l=142)

PR #2251, opened 2022-05-05
Android's certificate verification JNI layer

StefanoDuo:
# Cronvoy: preparation to unittest certificate verification JNI
Properly testing Android certificate verification logic requires calling
into framework code. That cannot be done through Robolectric (framework
APIs are faked). Hence, actual instrumentation tests are needed.

Having said that, after some exploration on the state of Bazel/Android
instrumentation tests, I've realized that the current state of things is
non-optimal. As a stopgap solution (until the infrastructure improves),
we've decided to test the JNI layer through unittests which fake
X509Util. This should be a reasonable middle ground, as X509Util is mostly a
compatibility layer on top of Android's API.

This CL introduces FakeX509Util and adds allows tests to swap that in by
adding a test-only API to AndroidNetworkLibrary.

FakeX509Util is not particularly clever: from its perspective a
certificate is just a string and for a verification to succeeds, all
certificates in a chain must be root certificates (weird, but we don't
really care about its logic).

# Add utility functions to translate between C++ and Java std constructs
Add utility functions to translate between:
- C++ strings and Java strings
- C++ vector of bytes and Java array of bytes

These utility functions will be used in the next commit to implement the
JNI layer for certificate verifications (and its unit tests).

# Implement JNI layer for Android's certificate verification
Add native hooks for AndroidNetworkLibrary Java APIs.

Write unittests that mimic calls from native to Java for these new
hooks, by making use of a fake X509Util and additional test-only Java
-> C++ JNI calls.


StefanoDuo:
@RyanTheOptimist I'm not sure who's going to review this code, sending over to you for the time being, feel free to redirect.

Note: this is still rough around the edges (I still need to look into the lifetime of JNI objects) but I wanted to touch base to make sure we don't have are on the same page. My idea is that once these JNI calls are in place, the async C++ cert verifier (yet to be built) will simply synchronously call into this.

Main type of feedback I'm looking for is: general shape of things and where we want the code to live. The PR is composed by 3 commits which you should be able to look at independently (I can't create child PR afaik).

library/common/jni/jni_interface.cc
```diff
@@ -1031,3 +1032,121 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_setPreferredNetwork(JNIEnv* env
   return set_preferred_network(static_cast<envoy_engine_t>(engine),
                                static_cast<envoy_network_t>(network));
 }
+
+bool Java_AndroidCertVerifyResult_isIssuedByKnownRoot(JNIEnv* env, jobject result) {
```
RyanTheOptimist, line 1036: Can you help me understand how these methods or called? I guess there is some sort of mapping from the c++ function name to some sort of Java class name? Or maybe not?

library/common/types/c_types.h
```diff
@@ -488,3 +488,27 @@ typedef struct {
...
+  // Certificate is not trusted because it has expired.
+  CERT_VERIFY_STATUS_ANDROID_EXPIRED = -3,
+  // Certificate is not trusted because it is not valid yet.
+  CERT_VERIFY_STATUS_ANDROID_NOT_YET_VALID = -4,
+  // Certificate is not trusted because it could not be parsed.
+  CERT_VERIFY_STATUS_ANDROID_UNABLE_TO_PARSE = -5,
+  // Certificate is not trusted because it has an extendedKeyUsage field, but
+  // its value is not correct for a web server.
+  CERT_VERIFY_STATUS_ANDROID_INCORRECT_KEY_USAGE = -6,
+} envoy_cert_verify_status_t;
```
RyanTheOptimist, line 514: Should this perhaps be named `envoy_cert_verify_status_android_t` to match the java name?

library/java/org/chromium/net/FakeX509Util.java
```diff
@@ -0,0 +1,40 @@
+package org.chromium.net;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Fake utility functions for verifying X.509 certificates.
```
RyanTheOptimist, line 9: It would be nice to mention the behavior of this class. For example, if certs A, B, C are passed in to verifyServerCertificate() what will be required for this to return OK vs failure?

library/java/org/chromium/net/FakeX509Util.java
```diff
@@ -0,0 +1,40 @@
...
+                                                                 String authType, String host) {
+    if (certChain == null || certChain.length == 0 || certChain[0] == null) {
+      throw new IllegalArgumentException(
+          "Expected non-null and non-empty certificate "
+          + "chain passed as |certChain|. |certChain|=" + Arrays.deepToString(certChain));
+    }
+
+    for (byte[] cert : certChain) {
+      String fakeCert = new String(cert);
+      if (!validFakeCerts.contains(fakeCert)) {
```
RyanTheOptimist, line 33: Could/should we also be able to verify that the subjectAltName in the certificate matches the specified hostname?

test/java/org/chromium/net/CertificateVerificationTest.java
```diff
@@ -0,0 +1,72 @@
...
+  @Test
+  public void testVerifyCertificateCall() throws Exception {
+    String[] fakeCertChain1 = new String[] {"fake cert"};
+    byte[][] certChain1 = new byte[][] {fakeCertChain1[0].getBytes()};
+
+    String[] fakeCertChain2 = new String[] {"fake cert", "another fake cert"};
+    byte[][] certChain2 = new byte[][] {fakeCertChain2[0].getBytes(), fakeCertChain2[1].getBytes()};
+
+    AndroidCertVerifyResult result =
+        (AndroidCertVerifyResult)JniLibrary.callCertificateVerificationFromNative(certChain1, "",
```
RyanTheOptimist, line 44: nit: here and below, instead of "" can we create (final?) local variables for authType and host and pass those in?

test/java/org/chromium/net/CertificateVerificationTest.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @After
+  public void tearDown() throws Exception {
+    JniLibrary.callClearTestRootCertificateFromNative();
+    AndroidNetworkLibrary.setFakeCertificateVerificationForTesting(true);
+  }
+
+  @Test
+  public void testVerifyCertificateCall() throws Exception {
```
RyanTheOptimist, line 36: nit: there are a number of different cases being tested in this method. (for example no trusted roots are provide, cases where the certs match vs mismatch.) It would be nice to break this into a number of different tests so that if there are failures, the name of the failed test explains the incorrect behavior.

RyanTheOptimist (commented):
Exciting! This approach looks great. A few comments... In particular, up to you but I would consider sending out the changes to jni_utility.h/cc in a distinct PR as they are so self-contained.

library/common/jni/jni_utility.cc
```diff
@@ -255,3 +257,61 @@ envoy_map to_native_map(JNIEnv* env, jobjectArray entries) {
...
+}
+
+void ConvertJavaStringToUTF8(JNIEnv* env, jstring str, std::string* result) {
+  const jsize length = env->GetStringLength(str);
+  const jchar* chars = env->GetStringChars(str, NULL);
+  std::string utf8 = std::wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t>{}.to_bytes(
+      reinterpret_cast<const char16_t*>(chars));
+  *result = utf8;
+  env->ReleaseStringChars(str, chars);
+}
```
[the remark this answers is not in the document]
StefanoDuo: Yes, that is a good idea. The current tests confirms that the transformations are reversible (A -> X -> A), but not that the intermediate state X is valid/sound.

Do you know if there are any C++ tests already setup for the other utilities? I've tried looking around but I didn't find anything.

StefanoDuo (commented):
> In particular, up to you but I would consider sending out the changes to jni_utility.h/cc in a distinct PR as they are so self-contained.

That is/was my intention, but AFAIK there is no way to create child PR on github and I wanted to parallelize the review process (ideally the 3 initial commit would have been 3 different CLs).

P.S. What is the accepted way to {D}CHECK inside EM?

Also, I've noticed that in the jni code we're not checking for Java exceptions?

library/java/org/chromium/net/FakeX509Util.java
RyanTheOptimist: Sure, that sounds fine. As we do more with FakeX509Util, we can fine-tune any behavior we need.

test/java/org/chromium/net/CertificateVerificationTest.java
```diff
@@ -33,40 +36,89 @@ public void tearDown() throws Exception {
...
-    String[] fakeCertChain1 = new String[] {"fake cert"};
-    byte[][] certChain1 = new byte[][] {fakeCertChain1[0].getBytes()};
-
-    String[] fakeCertChain2 = new String[] {"fake cert", "another fake cert"};
-    byte[][] certChain2 = new byte[][] {fakeCertChain2[0].getBytes(), fakeCertChain2[1].getBytes()};
+  public void testChainWithNonRootCertificate() throws Exception {
+    final String[] fakeCertChain = new String[] {"fake cert"};
+    final byte[][] certChain = new byte[][] {fakeCertChain[0].getBytes()};
+    final String authType = "";
+    final String host = "";
```
RyanTheOptimist, line 43: Your call, but I would be inclined to turn these 4 variables into members of the test class to avoid repeating them each time. But totally optional.

library/java/org/chromium/net/FakeX509Util.java
```diff
@@ -20,6 +24,10 @@ public static void addTestRootCertificate(byte[] rootCertBytes) {
 
   public static void clearTestRootCertificates() { validFakeCerts.clear(); }
 
+  /* Fake certificate chain verification. Returns CertVerifyStatusAndroid.OK only if all
```
RyanTheOptimist, line 27: nit: It would be nice to make the first part of this comment a complete sentence. Perhaps something like "Performs fake certificate chain verification."? Also, should the comment start with /**?

library/common/jni/jni_utility.cc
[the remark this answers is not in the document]
RyanTheOptimist: Hm, no I don't think so. Maybe ask on slack? Where are the A -> X -> A tests? I think I missed them?

library/common/jni/jni_utility.cc
```diff
@@ -279,39 +279,49 @@ jobjectArray ToJavaArrayOfByteArray(JNIEnv* env, const std::vector<std::string>&
...
                                         std::vector<std::string>* out) {
   size_t len = env->GetArrayLength(array);
   out->resize(len);
+
   for (size_t i = 0; i < len; ++i) {
     jbyteArray bytes_array = static_cast<jbyteArray>(env->GetObjectArrayElement(array, i));
     jsize bytes_len = env->GetArrayLength(bytes_array);
-    jbyte* bytes = env->GetByteArrayElements(bytes_array, nullptr);
+    // We don't care whether the returned array is a copy or not as we're simply
+    // copying it into C++ owned memory.
```
RyanTheOptimist, line 296: I get what this comment is saying and it makes sense. But it threw me for a loop because it talks about copy twice, first to say we don't care and secondly to say we're copying. Of course, the first time is talking about GetByteArrayElements() and the second is talking about this method, avaArrayOfByteArrayToStringVector. So how about a slight tweak:

 // It doesn't matter if the array returned by GetByteArrayElements is a copy or not,
 // as the data will be simply be copied into C++ owned memory below.

library/common/jni/jni_interface.cc
```diff
@@ -1031,3 +1032,121 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_setPreferredNetwork(JNIEnv* env
...
+jobjectArray jvm_cert_get_certificate_chain_encoded(JNIEnv* env, jobject result) {
+  jclass jcls_AndroidCertVerifyResult = env->FindClass("org/chromium/net/AndroidCertVerifyResult");
+  jmethodID jmid_getCertificateChainEncoded =
+      env->GetMethodID(jcls_AndroidCertVerifyResult, "getCertificateChainEncoded", "()[[B");
+
+  return static_cast<jobjectArray>(
+      env->CallObjectMethod(jcls_AndroidCertVerifyResult, jmid_getCertificateChainEncoded, result));
+}
+
+void ExtractCertVerifyResult(JNIEnv* env, jobject result,
```
RyanTheOptimist, line 1060: This method doesn't match the Java_ or jvm_ naming convention. it looks like it's called exclusively from jvm_verify_x509_cert_chain(). Should it be modified to fit one of them or does it have a different purpose?

library/common/jni/jni_interface.cc
```diff
@@ -1031,3 +1032,121 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_setPreferredNetwork(JNIEnv* env
...
+                             std::vector<std::string>* verified_chain) {
+  *status = jvm_cert_get_status(env, result);
+
+  *is_issued_by_known_root = jvm_cert_is_issued_by_known_root(env, result);
+
+  jobjectArray chain_byte_array = jvm_cert_get_certificate_chain_encoded(env, result);
+  JavaArrayOfByteArrayToStringVector(env, chain_byte_array, verified_chain);
+}
+
+static jobject call_jvm_verify_x509_cert_chain(const std::vector<std::string>& cert_chain,
```
RyanTheOptimist, line 1072: This one also starts with call_. Is that to reflect a different purpose?

RyanTheOptimist (commented):
I think this looks great. Just a few comments.

library/common/jni/jni_utility.cc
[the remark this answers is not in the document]
StefanoDuo: > Hm, no I don't think so. Maybe ask on slack?

I'll do that on Monday.

> Where are the A -> X -> A tests? I think I missed them?

They are sort of disguised, every CertificateVerificationTest test is basically doing A -> X -> A, in fact they:
1. Start with a Java string
2. Call into C++ through `Java_io_envoyproxy_envoymobile_engine_JniLibrary_callCertificateVerificationFromNative`, which
    * Converts the Java objects into C++ objects
        * `jobjectArray certChain` into `std::vector<std::string> cert_chain`
        * `jstring jhost` into `std::string host`
    * Calls into `call_jvm_verify_x509_cert_chain`, which
        * Reverses the transformation
        * Calls back into Java where we compare the obtained strings against the originals

P.S. Now that I look at it again I'm not sure if it makes sense for the certificate chain to be a `std::vector<std::string>` as we're basically manipulating it as a `std::vector<uint8_t*>`, we can discuss this later on though.

library/java/org/chromium/net/FakeX509Util.java
StefanoDuo: Done, thanks!

library/java/org/chromium/net/FakeX509Util.java
StefanoDuo: I'll add the host/authType checks on Monday, I've replied to the other things in the meantime.

library/common/jni/jni_interface.cc
RyanTheOptimist: Sure! Adding a comment which explains the state and points are a potential TODO sounds great.

library/common/jni/jni_interface.cc
RyanTheOptimist: Filing an issue and leaving as-is sounds perfect. Thanks!

library/common/jni/jni_utility.cc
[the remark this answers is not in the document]
RyanTheOptimist: Ah, char vs uint8_t is one of the most obnoxious "features" of C++. I'd say pick whatever is most convenient.

StefanoDuo (commented):
@RyanTheOptimist: at this point, if the only bits worth discussing are the utilities function, I can split them in a different CL if it's going to make your life easier. Let me know what you prefer.

library/common/strings/string_conversions.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <string>
+
+std::string UTF16ToUTF8(const char16_t* utf16, size_t len);
```
RyanTheOptimist, line 5: nit: Comments for these methods, pelase.

library/common/strings/string_conversions.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <string>
+
```
RyanTheOptimist, line 4: I think these methods should probably be in a namespace, right?

RyanTheOptimist (commented):
Looks great! Just a couple of comments. Feel free to leave the new utility functions in, or break them out. Either way.

library/common/strings/string_conversions.h
StefanoDuo: Done (sorry about that).

library/common/strings/string_conversions.h
StefanoDuo: Done.

test/common/strings/string_conversions_test.cc
```diff
@@ -0,0 +1,11 @@
+#include <string>
+
+#include "gtest/gtest.h"
+#include "library/common/strings/string_conversions.h"
+
+TEST(StringConversions, SameEncoding) {
+  std::string utf8 = u8"z\u00df\u6c34\U0001f34c";
+  std::u16string utf16 = u"z\u00df\u6c34\U0001f34c";
+
+  EXPECT_EQ(utf16, UTF8ToUTF16(utf8.data(), utf8.size()));
```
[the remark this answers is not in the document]
StefanoDuo: I've refreshed my memory about Unicode, UTF8 and UTF16. I haven't found any character which has the same UTF8 and UTF16 encoding so I've renamed the tests and also added some sanity checking on the strings being tested.

goaway:
It's late for me right now, but I'll do a full pass on this tomorrow during my a.m.

RyanTheOptimist (approved):

library/common/jni/jni_interface.cc
```diff
@@ -1031,3 +1032,127 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_setPreferredNetwork(JNIEnv* env
   return set_preferred_network(static_cast<envoy_engine_t>(engine),
                                static_cast<envoy_network_t>(network));
 }
+
+bool jvm_cert_is_issued_by_known_root(JNIEnv* env, jobject result) {
+  jclass jcls_AndroidCertVerifyResult = env->FindClass("org/chromium/net/AndroidCertVerifyResult");
```
goaway, line 1037: This call produces a `LocalRef` that must be deleted or it will leak.

library/common/jni/jni_interface.cc
```diff
@@ -1031,3 +1032,127 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_setPreferredNetwork(JNIEnv* env
...
+  jmethodID jmid_isIssuedByKnownRoot =
+      env->GetMethodID(jcls_AndroidCertVerifyResult, "isIssuedByKnownRoot", "()Z");
+  return env->CallBooleanMethod(jcls_AndroidCertVerifyResult, jmid_isIssuedByKnownRoot, result);
+}
+
+envoy_cert_verify_status_android_t jvm_cert_get_status(JNIEnv* env, jobject result) {
+  jclass jcls_AndroidCertVerifyResult = env->FindClass("org/chromium/net/AndroidCertVerifyResult");
```
goaway, line 1044: This call produces a `LocalRef` that must be deleted or it will leak.

library/common/jni/jni_interface.cc
```diff
@@ -1031,3 +1032,127 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_setPreferredNetwork(JNIEnv* env
...
+  jmethodID jmid_getStatus = env->GetMethodID(jcls_AndroidCertVerifyResult, "getStatus", "()I");
+
+  return static_cast<envoy_cert_verify_status_android_t>(
+      env->CallIntMethod(jcls_AndroidCertVerifyResult, jmid_getStatus, result));
+}
+
+jobjectArray jvm_cert_get_certificate_chain_encoded(JNIEnv* env, jobject result) {
+  jclass jcls_AndroidCertVerifyResult = env->FindClass("org/chromium/net/AndroidCertVerifyResult");
```
goaway, line 1052: This call produces a `LocalRef` that must be deleted or it will leak.

library/common/jni/jni_interface.cc
```diff
@@ -1031,3 +1032,127 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_setPreferredNetwork(JNIEnv* env
...
+
+  jobjectArray chain_byte_array = jvm_cert_get_certificate_chain_encoded(env, result);
+  JavaArrayOfByteArrayToStringVector(env, chain_byte_array, verified_chain);
+}
+
+static jobject call_jvm_verify_x509_cert_chain(const std::vector<std::string>& cert_chain,
+                                               std::string auth_type, std::string host) {
+  jni_log("[Envoy]", "jvm_verify_x509_cert_chain");
+  JNIEnv* env = get_env();
+  jclass jcls_AndroidNetworkLibrary = env->FindClass("org/chromium/net/AndroidNetworkLibrary");
```
goaway, line 1082: This call produces a `LocalRef` that must be deleted or it will leak.

library/common/jni/jni_utility.cc
```diff
@@ -259,8 +258,9 @@ envoy_map to_native_map(JNIEnv* env, jobjectArray entries) {
 }
 
 jstring ConvertUTF8ToJavaString(JNIEnv* env, const std::string& str) {
-  std::u16string utf16 =
-      std::wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t>{}.from_bytes(str);
+  // JNI's NewStringUTF expects "modified" UTF8 so instead convert the string to UTF16 and pass it
```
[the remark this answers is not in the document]
goaway: To avoid surprises due to the weirdness of Modified UTF-8, the convention we've adopted a this layer is to pass true UTF-8 byte arrays across the JNI and then encode/decode them to java.lang.Strings on the Java side. For simplicity's sake (and perhaps to avoid the risk of extra logic manipulating buffers) it would be nice to maintain that convention, if possible, rather than introduce Modified UTF-8 handling in C++.

library/common/types/c_types.h
```diff
@@ -488,3 +488,27 @@ typedef struct {
...
+  // Certificate is not trusted because it has expired.
+  CERT_VERIFY_STATUS_ANDROID_EXPIRED = -3,
+  // Certificate is not trusted because it is not valid yet.
+  CERT_VERIFY_STATUS_ANDROID_NOT_YET_VALID = -4,
+  // Certificate is not trusted because it could not be parsed.
+  CERT_VERIFY_STATUS_ANDROID_UNABLE_TO_PARSE = -5,
+  // Certificate is not trusted because it has an extendedKeyUsage field, but
+  // its value is not correct for a web server.
+  CERT_VERIFY_STATUS_ANDROID_INCORRECT_KEY_USAGE = -6,
+} envoy_cert_verify_status_android_t;
```
goaway, line 514: Can this enum be made generic such that it could be utilized by other platforms (currently iOS and C++)?

goaway (changes requested):
Thanks for the PR @StefanoDuo! Sorry it took me a while to catch it on my radar, and a few more days for me to actually do the full review.

I left some comments above.

Am I correct in my assumption that this code is intended to eventually support a platform-implementable cert validator? If so, I'd propose packaging the API into a struct with a matching Java/Kotlin interface (it looks like FakeX509Util might be the implementation of that interface). This is similar to what we've done with other platform-implementable extensions. Then a trivial Test implementation (FakeX509Util) could drive your tests. A similar approach can be seen in my currently open PR for registering a platform-implementable key-value store:
https://github.com/envoyproxy/envoy-mobile/pull/2134

(Note there are substantial improvements that can be made to how we manage calling into these externally-implemented extensions; the current mechanism is a minimally-viable placeholder.)

library/common/jni/jni_utility.cc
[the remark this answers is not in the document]
goaway: This is a simple example that shows the JNI side (this supports an implementation of a simple extension to retrieve runtime-defined platform Strings):

https://github.com/envoyproxy/envoy-mobile/blob/aa0beeea7079dbc6c9021f9bd6fd4801dbb1381a/library/common/jni/jni_interface.cc#L819

And here is the Java side:

https://github.com/envoyproxy/envoy-mobile/blob/aa0beeea7079dbc6c9021f9bd6fd4801dbb1381a/library/java/io/envoyproxy/envoymobile/engine/JvmStringAccessorContext.java

library/common/jni/jni_interface.cc
goaway: To clarify, the above is true if the stack originates in the JVM. If the stack originates on a non-JVM-managed thread (e.g. the Envoy Mobile main thread), no allocated references will be automatically cleaned up.

StefanoDuo: Whoa, that should be explicitly stated in the docs (I'm guessing it is and I failed to find it)! Thanks a lot for expanding on this.

For posterity's sake, [Android's JNI tips page](https://developer.android.com/training/articles/perf-jni#local-and-global-references) explicitly mentions this:
> One unusual case deserves separate mention. If you attach a native thread with AttachCurrentThread, the code you are running will never automatically free local references until the thread detaches. Any local references you create will have to be deleted manually.

library/common/jni/jni_utility.cc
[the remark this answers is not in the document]
StefanoDuo: I've updated the code to pass true UTF-8 byte arrays across the JNI interface as you suggested.

I don't think I need to wrap stuff in an accessor for my use case (the strings are only needed for the duration of the JNI call). Let me know if this looks okay to you!

goaway (approved):
Looks good. Thanks for making the updates @StefanoDuo!

StefanoDuo:
(I forgot to mention [the associated bug](https://github.com/envoyproxy/envoy-mobile/issues/1575))

PR #2267, opened 2022-05-11
configuration: filter unroutable addresses on Android by default

Augustyniak:
Description: After experimenting with it in Lyft apps we saw a positive impact of enabling filtering. It's enabled to all Lyft users who have EM enabled.
Risk Level: Low, experimented with.
Testing: N/A
Docs Changes:  Done
Release Notes: Done


Augustyniak:
@alyssawilk Updating this so that it matches what we use internally at Lyft :)

goaway (approved):

jpsim (approved):
+1

I have some work in progress in #2128 to add this on iOS and flip the wording to better reflect the default, but this is good to merge in the meantime.

Thanks for adding a changelog entry!

PR #2269, opened 2022-05-12
tests: add -Xcheck:jni to kotlin integration tests by default

goaway:


jpsim (approved):
Fantastic. Based on my reading of [the docs](https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/clopts002.html) for the option there isn't much of a downside to enabling this for tests, so this works for me.

jpsim:
I do wonder if this would have caught this return value mismatch: https://github.com/envoyproxy/envoy-mobile/pull/2252

goaway:
Thanks @jpsim!

PR #2270, opened 2022-05-12
configuration: enable h2 ping by default

Augustyniak:
Description: Enable h2 ping by default. Lyft currently has this functionality enabled to all of users who have Envoy Mobile running. It turned out to be effective when it comes to getting rid of stale connections.
Risk Level: Low, enables well tested functionality by default.
Testing: N/A
Docs Changes: Done
Release Notes: Done


library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -256,7 +256,8 @@ open class EngineBuilder(
 
   /**
    * Add a rate at which to ping h2 connections on new stream creation if the connection has
-   * sat idle.
+   * sat idle. Defaults to 1 millisecond which effectively enables h2 ping functionlity
```
jpsim, line 259:
```suggestion
   * sat idle. Defaults to 1 millisecond which effectively enables h2 ping functionality
```

library/swift/EngineBuilder.swift
```diff
@@ -203,7 +203,8 @@ open class EngineBuilder: NSObject {
   }
 
   /// Add a rate at which to ping h2 connections on new stream creation if the connection has
-  /// sat idle.
+  /// sat idle. Defaults to 1 millisecond which effectively enables h2 ping functionlity
```
jpsim, line 206:
```suggestion
  /// sat idle. Defaults to 1 millisecond which effectively enables h2 ping functionality
```

jpsim (changes requested):
Typos. We should also document how to disable now that it's not the default.

Thanks for adding release notes!

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
Augustyniak: 🤦

Augustyniak:
> Typos. We should also document how to disable now that it's not the default.

Great point. Update the doc. Not sure whether it's OK now - let me know. This API for enabling/disabling H2 ping should be eventually updated to access true/false as opposed to this time interval I think. That's a separate thing tho.

jpsim (approved):
This is fine as it's documenting existing behavior but we should make this field an optional to make it more obvious how to disable it.

PR #2271, opened 2022-05-12
Add version history entries for user-facing changes

jpsim:
I found these while looking at the changeset since 0.4.6.


Augustyniak (approved):

RyanTheOptimist:
Oh sorry! I didn't realize this was a thing to be done. Thanks!

jpsim:
No problem, it is in the PR template [here](https://github.com/envoyproxy/envoy-mobile/blob/512d5246e8c16481bded48ee65470fbfec56ca52/PULL_REQUEST_TEMPLATE.md?plain=1#L8) but we (reviewers/maintainers) should do a better job at actually checking for that.

PR #2272, opened 2022-05-12
net: enable happy eyeballs by default

goaway:


jpsim (commented):
Can you please reflect this in the changelog especially since it's a breaking change?

Docs CI job might fail without https://github.com/envoyproxy/envoy-mobile/pull/2274

Augustyniak (approved):
My preference would be to add an information about what the default is to method's docs but leaving up to you to decide. At some point it would be good to have an API that allows a user to check what the current engine(builder) settings are - right now folks need to look into the source code to see how a given functionality is configured (unless a given method is documented).

Augustyniak (approved):

jpsim (approved):

jpsim:
Going to merge this so I can cut a release.

PR #2274, opened 2022-05-12
docs: fix sphinx reference mismatch warning

jpsim:
I'm not very familiar with reST so I'm not sure if this is correct, but it does remove the following warning:

> WARNING: Mismatch: both interpreted text role prefix and reference suffix


jpsim:
I'm not sure why this wasn't caught in the CI runs for https://github.com/envoyproxy/envoy-mobile/pull/2263 🤔

goaway (approved):

PR #2275, opened 2022-05-12
api: disallow setting 'host' header directly

### LINKED ISSUE #2244: when 'host` and `:authority` are both set on headers, it would fail DNS resolution
Steps:

- make a RequestHeadersBuilder with authority be (“[api.pinterest.com](api.pinterest.com)”, *)
- add a `test` header: add(“host”, “[api.pinterest.com](api.pinterest.com)”)
- send request
- then will see dns resolution error. In my testing, my host appears to be “[api.pins.com](http://api.pins.com/), [api.pins.com](http://api.pins.com/)” in tracing a DNS error.

image

Not sure if this is a known issue. I had to remove `host` to work around this.

goaway:
Description: The ':authority' header is used to direct the connection internally and this is set when initializing the builder.

Fixes #2244.

Risk Level: Low
Testing: Unit


Augustyniak (commented):
Nice! Thank you for doing this Mike - can you update the version history file? Looks good otherwise I think

Augustyniak (approved):

Augustyniak (approved):

PR #2276, opened 2022-05-13
iOS: fix termination crash in `ProvisionalDispatcher` (again)

jpsim:
I believe https://github.com/envoyproxy/envoy-mobile/pull/2059 helped reduce these occurrences, but we're still seeing some termination crashes where a different thread is waiting on `std::thread::join()`.

So moving this to before the `main_thread_.join()` may help.

Risk Level: Moderate, there's a possibility this will introduce a data race, but even if that's the case, it'll be when the engine is being terminated.
Release Notes: Added


library/common/engine.cc
```diff
@@ -161,22 +161,22 @@ envoy_status_t Engine::terminate() {
...
     ASSERT(event_dispatcher_);
+    ASSERT(dispatcher_);
 
     // Exit the event loop and finish up in Engine::run(...)
     if (std::this_thread::get_id() == main_thread_.get_id()) {
       // TODO(goaway): figure out some way to support this.
       PANIC("Terminating the engine from its own main thread is currently unsupported.");
     } else {
       event_dispatcher_->exit();
+      dispatcher_->terminate();
```
snowp, line 172: Is the provisional dispatcher valid after event_dispatcher_->exit() is called? Should the order here be flipped?

snowp (commented):
What data race are you concerned about? Seems like this is all done under a lock

I wonder if it would be helpful to set up a integration test which keeps issuing calls against the dispatcher while shutting down, then using bazels "run this test 1000 times" feature to see if we're still hitting crashes

library/common/engine.cc
goaway: I'd actually suggest moving the `exit()` call inside of `ProvisionalDispatcher::terminate()` to make this less delicate.

goaway:
> What data race are you concerned about? Seems like this is all done under a lock
> 
> I wonder if it would be helpful to set up a integration test which keeps issuing calls against the dispatcher while shutting down, then using bazels "run this test 1000 times" feature to see if we're still hitting crashes

If you look at the logic in `envoy_status_t Engine::terminate()`, the lock is released (necessarily) before attempting to join the main thread. This sequence of events in this PR is the one I alluded to earlier that I thought would be more robust.

Thanks @jpsim!

goaway (changes requested):
See minor suggestion above.

library/common/engine.cc
jpsim: Done.

jpsim: I've reverted it to what it was before. Moving the exit into `terminate()` caused the java tests to crash:

```
#
#
#  SIGSEGV (0xb) at pc=0x000000014156e5d9, pid=28881, tid=29959
#
# JRE version: OpenJDK Runtime Environment Zulu11.50+19-CA (11.0.12+7) (build 11.0.12+7-LTS)
# Java VM: OpenJDK 64-Bit Server VM Zulu11.50+19-CA (11.0.12+7-LTS, mixed mode, tiered, compressed oops, g1 gc, bsd-amd64)
# Problematic frame:
# C  [libndk_envoy_jni.jnilib+0x1fee5d9]  Envoy::Event::ProvisionalDispatcher::terminate()+0x29
```

goaway (approved):
Thanks @jpsim!

PR #2277, opened 2022-05-14
tools: Enable the VSCode completion db to use bazelisk if available

abeyad:
Running tools/vscode_compdb.py to generate the code completion symbols for VSCode produces an error because the .bazelversion in Envoy may not be available in certain environments. The recommendation is to use bazelisk, but the compilation database script defaults to using bazel. This change enables the VSCode compilation database generation script to use bazelisk if it is available on the system; if not, it will use bazel.

Signed-off-by: Ali Beyad abeyad@google.com

Risk Level: low
Testing: local
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/tools/vscode_compdb.sh
+++ b/tools/vscode_compdb.sh
@@ -4,8 +4,12 @@
 # This is a modification of evnoy/tools/vscode/refresh_compdb.sh which hits
 # the correct envoy-mobile Bazel targets.
 
+# Use bazelisk if present, if not use bazel.
+bazel_or_isk=bazelisk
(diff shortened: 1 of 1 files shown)


goaway (approved):
Thank you!

jpsim:
Sorry I was slow to review, but we have a wrapper around bazel that we use to have a consistent experience no matter what the user has installed and available in their environment. Updating to use that in https://github.com/envoyproxy/envoy-mobile/pull/2284

abeyad:
Oh nice, didn't know about `bazelw`, thanks JP!

PR #2281, opened 2022-05-16
add support for http/2 prior knowledge over plaintext

snowp:
This adds support for using HTTP/2 over plaintext by setting "h2c" as the protocol, which matches the mechanism used by other HTTP clients/servers. This allows using http/2 prior knowledge with plaintext, which was previously only possible when using TLS.

Risk Level: Medium
Testing: New UT
Docs Changes: n/a
Release Notes: Added

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -386,6 +386,17 @@ R"(
     upstream_connection_options: *upstream_opts
     circuit_breakers: *circuit_breakers_settings
     typed_extension_protocol_options: *h1_protocol_options
+  - name: base_h2_clear
+    connect_timeout: *connect_timeout
+    lb_policy: CLUSTER_PROVIDED
+    cluster_type: *base_cluster_type
+    transport_socket:
+      name: envoy.transport_sockets.raw_buffer
+      typed_config:
+        "@type": type.googleapis.com/envoy.extensions.transport_sockets.raw_buffer.v3.RawBuffer
+    upstream_connection_options: *upstream_opts
+    circuit_breakers: *circuit_breakers_settings
+    typed_extension_protocol_options: *h2_protocol_options
   - name: base_h2
     http2_protocol_options: {}
     connect_timeout: *connect_timeout
--- a/library/common/http/client.cc
+++ b/library/common/http/client.cc
@@ -599,6 +599,7 @@ const char* BaseCluster = "base";
 const char* H2Cluster = "base_h2";
 const char* H3Cluster = "base_h3";
 const char* ClearTextCluster = "base_clear";
+const char* ClearTextH2Cluster = "base_h2_clear";
 
 } // namespace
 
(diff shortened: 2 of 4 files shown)


library/common/http/client.cc
```diff
@@ -599,18 +599,23 @@ const char* BaseCluster = "base";
...
   // Determine upstream cluster:
   // - Use TLS with ALPN by default.
   // - Use http/2 or ALPN if requested explicitly via x-envoy-mobile-upstream-protocol.
-  // - Force http/1.1 if request scheme is http (cleartext).
+  // - Force http/1.1 if request scheme is http (cleartext) unless the protocol is set to h2c.
   const char* cluster{};
   auto protocol_header = headers.get(ProtocolHeader);
   if (headers.getSchemeValue() == Headers::get().SchemeValues.Http) {
-    cluster = ClearTextCluster;
+    if (!protocol_header.empty() && protocol_header[0]->value().getStringView() == "h2c") {
```
goaway, line 614: Just want to call out that while any value for the header can be theoretically interpreted here, the iOS and Android public interfaces actually restrict setting this header, instead exposing a RequestBuilder option to set the HTTP protocol.

If the intent is to call this from iOS/Android you'll likely want to add an option to the respective RequestBuilders.

goaway (approved):
Thanks @snowp! This seems fine to me, but see the comment above if the plan is to leverage this from iOS/Android application contexts.

library/common/config/config.cc
```diff
@@ -386,6 +386,17 @@ R"(
     upstream_connection_options: *upstream_opts
     circuit_breakers: *circuit_breakers_settings
     typed_extension_protocol_options: *h1_protocol_options
+  - name: base_h2_clear
```
alyssawilk, line 389: If we're not including HTTP/3 by default, I think we should also make this something which is configured off but configurable on.  h2c can be really problematic on the internet given middelboxes.

goaway: This won't get used in any context by default, and will require setting a special header, or more likely (per my previous comment) setting the override via the RequestBuilder. The cluster will exist, however, but be unused with the setup in this PR - is that a concern?

alyssawilk: Happy to get a second opinion from @RyanTheOptimist but I think if most people don't want it and it adds to config complexity it'd be better to have it not present by default.

RyanTheOptimist: I agree with @alyssawilk. h2c has the potential to leak secrets onto the wire. Since that is super dangerous and h2c is quite uncommon I would strongly suggest that the bar for using it in Envoy Mobile be has high as possible. (In particular, I think the bar for using h3 should not be higher than the bar for using h2c)

goaway: If we want to raise the bar, one option would be to require it to be enabled via the EngineBuilder before the cluster is injected. EngineBuilder documentation could come with a prominent warning about usage being risky.

alyssawilk: worth noting, if I recall correctly h2c frequently doesn't work on the internet - because it's in the clear, all the routing infra that spies on HTTP tries to parse it, and often just kills off connections.  that may have changed since last folks tried but I wouldn't bet on it actually working well in the wild

PR #2282, opened 2022-05-16
ci: add scheduled job to update Envoy weekly

jpsim:
We're currently 2.5 weeks behind, and there are a significant number of breaking changes to account for.

This would be easier to manage if we kept the updates smaller.

### DIFF
--- a/.github/workflows/submodule_update.yml
+++ b/.github/workflows/submodule_update.yml
@@ -1,0 +1,41 @@
+name: submodule_update
+
+on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


Augustyniak (approved):
One idea that comes to my mind is that we could in theory assign it to a current EM on-call person (we have a rotation) but that could overcomplicate things and I think that having this job as an automated thing is a big improvements on its own. Thank you for working on this.

jpsim:
> I'll do that as a follow-up.

As promised: #2288

PR #2290, opened 2022-05-18
extensions: register `TestAccessor` filter

jpsim:
We use this in Lyft's integration with Envoy Mobile. It should be harmless to add even though we don't currently need it, and it will be helpful to keep the Envoy Mobile configuration close to the Lyft one considering that we're running Envoy Mobile in production and can help identify issues at scale.


goaway:
I think I'm lacking some context - let's chat today.

jpsim:
I'm checking to see if this is actually needed in our Lyft-side tests and will just close the PR if it's not needed.

jpsim:
We're already configuring this extension in https://github.com/envoyproxy/envoy-mobile/blob/1eb14c9d3026dd956b46873b2fe0287d456294af/bazel/envoy_mobile_test_extensions.bzl#L6

PR #2293, opened 2022-05-18
mobile: moving the c++ integration test to use default config

alyssawilk:
Risk Level: n/a
Testing: test still passes
Docs Changes: n/a (test only)
Release Notes: n/a (test only)

### DIFF
--- a/test/common/integration/client_integration_test.cc
+++ b/test/common/integration/client_integration_test.cc
@@ -171,9 +171,6 @@ public:
   // There are some config modifiers below which do result in deltas.
   static std::string defaultConfig() {
     Platform::EngineBuilder builder;
-
-    ProtobufMessage::StrictValidationVisitorImpl visitor;
-    envoy::config::bootstrap::v3::Bootstrap bootstrap;
     std::string config_str = absl::StrCat(config_header, builder.generateConfigStr());
     return config_str;
   }
--- a/test/common/integration/client_integration_test.cc
+++ b/test/common/integration/client_integration_test.cc
@@ -10,6 +10,8 @@
 
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
+#include "library/cc/engine_builder.h"
+#include "library/common/config/internal.h"
 #include "library/common/data/utility.h"
 #include "library/common/http/client.h"
 #include "library/common/http/header_utility.h"
@@ -46,9 +48,8 @@ typedef struct {
 } callbacks_called;
 
 void validateStreamIntel(const envoy_final_stream_intel& final_intel) {
-  // This test doesn't do DNS lookup
-  EXPECT_EQ(-1, final_intel.dns_start_ms);
-  EXPECT_EQ(-1, final_intel.dns_end_ms);
+  EXPECT_NE(-1, final_intel.dns_start_ms);
+  EXPECT_NE(-1, final_intel.dns_end_ms);
   // This test doesn't do TLS.
   EXPECT_EQ(-1, final_intel.ssl_start_ms);
   EXPECT_EQ(-1, final_intel.ssl_end_ms);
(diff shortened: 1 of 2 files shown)


alyssawilk:
Also question for Lyft, it took me ages to figure out that my generated config was invalid because I wasn't prepending the config header.  I was going to add comments in engine.h about that, but I notice it's not exposed via engine.h only internal.h which implies it generally shouldn't be used.

Can y'all think of a clarifying comment I could add to the Engine's generateConfigStr which would make it more clear what one has to do to make it valid?

test/common/integration/client_integration_test.cc
```diff
@@ -73,11 +74,26 @@ void validateStreamIntel(const envoy_final_stream_intel& final_intel) {
...
                               public testing::TestWithParam<Network::Address::IpVersion> {
 public:
-  ClientIntegrationTest() : BaseIntegrationTest(GetParam(), bootstrap_config()) {
+  ClientIntegrationTest() : BaseIntegrationTest(GetParam(), defaultConfig()) {
     use_lds_ = false;
     autonomous_upstream_ = true;
     defer_listener_finalization_ = true;
     HttpTestUtility::addDefaultHeaders(default_request_headers_);
+    config_helper_.addConfigModifier([](envoy::config::bootstrap::v3::Bootstrap& bootstrap) {
+      // The default stats config has overenthusiastic filters.
```
jpsim, line 83: What filters cause issues exactly? Should this be adjusted?

jpsim:
> it took me ages to figure out that my generated config was invalid because I wasn't prepending the config header

I don't have the historical context here, but if the config header is necessary, why not always include it? Seems like we should always try to produce valid configs when going through our APIs.

goaway:
Thanks for doing this @alyssawilk! It's actually an item I'd had on my TODO list for a while.

> Also question for Lyft, it took me ages to figure out that my generated config was invalid because I wasn't prepending the config header. I was going to add comments in engine.h about that, but I notice it's not exposed via engine.h only internal.h which implies it generally shouldn't be used.
> 
> Can y'all think of a clarifying comment I could add to the Engine's generateConfigStr which would make it more clear what one has to do to make it valid?

It's been a while since I've looked at the cc builder, but in general, the builders aren't meant to output a complete valid Envoy config. Rather they're meant to output formatted overrides to values initially defined in the `config_header`, all of which will eventually be prepended to the `config_template`. They still currently have access to the template, because there's a few cases where anchors/overrides don't yet suffice, and so some limited interpolation is still done.

The only thing that should need to be done to produce a complete valid config is concatenating the `config_header` with the `config_template`. (This bypasses the means by which any EngineBuilder configuration is applied.) The header is not exposed because it's actually very large, and in normal use one should never actually need to touch it. This test case, however, is perhaps a bit special, since you're trying to independently produce a complete valid config. 

> I don't have the historical context here, but if the config header is necessary, why not always include it? Seems like we should always try to produce valid configs when going through our APIs.

The `config_header` is added as part of the startup process when preparing to run the main thread. It's a last step because the different pieces of config can be very large at that point and we want to avoid unnecessary data copying/manipulation. But in the sense of running EM in the typical fashion, it is always included.

My suggestion for this would be to not use the cc `Platform::EngineBuilder` at all, and instead simply concatenate the `config_header` and `config_template`.

alyssawilk:
@goaway would love to touch base with you today but I think I want to use the builder because as I understand it it's the place we'd mimic java and kotlin APIs for changing config?  I want to be able to do things like enable http/3 not through explicit config modifiers in the test, but through the same mechanisms we use elsewhere and I thought builder was the way to do that (though I could easily misunderstand)

alyssawilk:
for example one of the upcoming tasks I've got is to add optional brotli support.  I think I want the builder for that.

alyssawilk:
ok, filed https://github.com/envoyproxy/envoy-mobile/issues/2303 and I think this is otherwise good to go?

goaway (approved):

PR #2295, opened 2022-05-18
config: cleaning up deprecated configs

alyssawilk:
Switching off of deprecated http2 and watchdog configs. 
_Should_ be a functional no-op.

Risk Level:
Testing: no test changes
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -203,7 +203,11 @@ const char* config_template = R"(
     wait_for_warm_on_init: false
     connect_timeout: *connect_timeout
     dns_refresh_rate: *dns_refresh_rate
-    http2_protocol_options: {}
+    typed_extension_protocol_options:
+      envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
+        "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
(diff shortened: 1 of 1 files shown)


alyssawilk:
I'm  not sure that we need both main and worker watchdogs but I believe this is functionally equivalent to the prior config, which seems the safest change.

library/common/config/config.cc
```diff
@@ -389,7 +393,11 @@ R"(
     circuit_breakers: *circuit_breakers_settings
     typed_extension_protocol_options: *h1_protocol_options
   - name: base_h2
-    http2_protocol_options: {}
+    typed_extension_protocol_options:
+      envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
+        "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
+        explicit_http_config:
+          http2_protocol_options: {}
```
goaway, line 400: Just to confirm - this will still configure this cluster to use h2 without protocol negotiation, correct?

alyssawilk: yeah, auto_config does alpn, explicit_http does whatever protocol you explicitly configure.

goaway (approved):

PR #2300, opened 2022-05-19
instrumentation: add timers and warnings to platform callbacks

goaway:
Description: Adds Envoy histograms (timers) to most platform-provided callbacks in the Http::Client and in PlatformBridge Filters. Beyond a hardcoded threshold (1s), a warn-level LOG_EVENT will be emitted for slow-to-execute callbacks.
Risk Level: Low
Testing: Manual & Existing Coverage

### DIFF
--- a/library/common/extensions/filters/http/platform_bridge/filter.cc
+++ b/library/common/extensions/filters/http/platform_bridge/filter.cc
@@ -67,14 +67,17 @@ static void envoy_filter_reset_idle(const void* context) {
 }
 
 PlatformBridgeFilterConfig::PlatformBridgeFilterConfig(
+    Server::Configuration::FactoryContext& context,
     const envoymobile::extensions::filters::http::platform_bridge::PlatformBridge& proto_config)
-    : filter_name_(proto_config.platform_filter_name()),
+    : root_scope_(context.scope()),
+      stats_(generateStats("", root_scope_)),
+      filter_name_(proto_config.platform_filter_name()),
       platform_filter_(static_cast<envoy_http_filter*>(
           Api::External::retrieveApi(proto_config.platform_filter_name()))) {}
 
 PlatformBridgeFilter::PlatformBridgeFilter(PlatformBridgeFilterConfigSharedPtr config,
                                            Event::Dispatcher& dispatcher)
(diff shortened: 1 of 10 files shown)


library/common/extensions/filters/http/platform_bridge/filter.cc
```diff
@@ -23,6 +23,9 @@ namespace HttpFilters {
 namespace PlatformBridge {
 
 namespace {
+
+constexpr auto SlowCallbackWarningThreshold = std::chrono::seconds(1);
```
jpsim, line 27: +1 to starting at 1 second here. Ideally we roll this out at scale and see that it's rarely hit, if ever, and we can reduce it accordingly. Ideally pt2 is that we find a good way to make this configurable.

jpsim:
Since this is a user-facing change, can you please add an entry to the version history doc?

library/common/extensions/filters/http/platform_bridge/filter.h
```diff
@@ -14,17 +17,51 @@ namespace Extensions {
...
   PlatformBridgeFilterConfig(
+      Server::Configuration::FactoryContext& context,
       const envoymobile::extensions::filters::http::platform_bridge::PlatformBridge& proto_config);
 
+  PlatformBridgeFilterStats& stats() const { return stats_; }
   const std::string& filter_name() { return filter_name_; }
   const envoy_http_filter* platform_filter() const { return platform_filter_; }
 
 private:
+  static PlatformBridgeFilterStats generateStats(const std::string& stat_prefix,
```
jpsim, line 56: Why have the stat prefix if we only set it to the empty string?

goaway: Ah, good question, it's
1) because I Considered making it the filter name and think that still might be worth considering, but I held off because I'm not sure how consistently/well those are formatted for use in stats,

and
2) because this is basically the signature used everywhere in Envoy for this pattern.

jpsim (dismissed):
Approving now so we can observe the stats over the weekend, but I'll review in more detail next week.

jpsim (approved):

PR #2305, opened 2022-05-20
ci: automate weekly releases

jpsim:
Our [documentation](https://github.com/envoyproxy/envoy-mobile/blob/v0.4.6.20220513-4/docs/root/development/releasing/releasing.rst#pre-release-versioning) states that we cut weekly pre-releases, but that hasn't been the case.

This PR automates pushing a new tag on a weekly basis on Mondays at 8am EST. Once the tag is pushed, the `release.yml` GitHub Actions kick off the remainder of the release process.

It's also possible to manually trigger the workflow on other days to cut a new pre-release version.

Note that for simplicity, there can only be one pre-release cut per day. If a manual pre-release is needed after this automation has already run, a new tag can be pushed manually.


goaway (approved):

goaway:
Nice work @jpsim!

jpsim:
I'll merge this and trigger the workflow to confirm it works.

jpsim:
Ran it here: https://github.com/envoyproxy/envoy-mobile/runs/6529542994?check_suite_focus=true

It successfully created and pushed the tag: https://github.com/envoyproxy/envoy-mobile/releases/tag/v0.4.6.20220520

image

However, the release automation failed because of an invalid key:

> The workflow is not valid. .github/workflows/release.yml (Line: 166, Col: 9): Unexpected value 'prerelease'

Fixing that next.

jpsim:
Fix: https://github.com/envoyproxy/envoy-mobile/pull/2306

PR #2307, opened 2022-05-20
Revert "android: use local addresses as opposed to prefix"

jpsim:
Reverts envoyproxy/envoy-mobile#2081

This broke detection of ipv6 interfaces.


jpsim:
No changelog entry for this since the bug was only ever shipped in pre-release versions.

Augustyniak (approved):

PR #2315, opened 2022-05-25
test: making C++ integration test more authentic

alyssawilk:
Removing the modifier which cleared stats, now that the stats clusters work for IPv6.

Risk Level: n/a (test only)
Docs Changes: n/a
Release Notes: n/a


RyanTheOptimist (approved):
Sweet

Issue #1572, opened 2021-07-06
Cronvoy: check if sending clear text is permitted


carloseltuerto:
With [Marshmallow (V6.0 / API level 23)](https://source.android.com/setup/start/build-numbers#platform-code-names-versions-api-levels-and-ndk-releases), Android library exposes two system methods indicating if sending a request in clear text (without TLS) is acceptable:

- [public boolean isCleartextTrafficPermitted()](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/security/NetworkSecurityPolicy.java#49)
- [public boolean isCleartextTrafficPermitted(String hostname)](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/security/NetworkSecurityPolicy.java#73)

Cronet seems to [honor](https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/cronet_url_request_context.cc;l=357) this policy systematically for Android M+.

And this is how this is [implemented](https://source.chromium.org/chromium/chromium/src/+/main:net/android/java/src/org/chromium/net/AndroidNetworkLibrary.java;l=368) in Java, and how the [C++ calls it](https://source.chromium.org/chromium/chromium/src/+/main:net/url_request/url_request_http_job.cc;l=214).

The cheap solution here is to implement this within the Cronvoy shim, but I suspect that when Envoy Mobile will be used directly on Android (without the Cronvoy shim), then it should also respect this policy.

alyssawilk:
@goaway @junr03 thoughts on this one?

PR #2318, opened 2022-05-25
android: attempting to respect Android's cleartext policies

alyssawilk:
Rejecting requests which don't have an https scheme if Android disallows cleartext

Risk Level: High 
Testing: NEED HELP HERE
Docs Changes: n/a
Release Notes: inline
Fixes https://github.com/envoyproxy/envoy-mobile/issues/1572


library/common/jni/BUILD
```diff
@@ -199,3 +181,24 @@ cc_library(
...
+        "jni_support.h",
+        "jni_utility.h",
+        "jni_version.h",
+    ],
+    deps = [
+        "//library/common/jni/import:jni_import_lib",
+        "//library/common/types:c_types_lib",
+        "@envoy//source/common/common:assert_lib",
+    ] + select({
+        "@envoy//bazel:android_logger": [":ndk_jni_support"],
```
alyssawilk, line 201: @keith this is probably not the right select for using android - do you know if there's a better opton?

keith: We have this internally https://github.com/envoyproxy/envoy/pull/21459 seems useful upstream at this point, wdyt

library/common/http/client.cc
```diff
@@ -479,6 +480,16 @@ void Client::sendHeaders(envoy_stream_t stream, envoy_headers headers, bool end_
   if (direct_stream) {
     ScopeTrackerScopeState scope(direct_stream.get(), scopeTracker());
     RequestHeaderMapPtr internal_headers = Utility::toRequestHeaders(headers);
+
+    // This is largely a check for the android platform: is_cleartext_permitted
+    // is a no-op for other platforms.
+    if (internal_headers->getSchemeValue() != "https" &&
+        !is_cleartext_permitted(internal_headers->getHostValue())) {
+      direct_stream->request_decoder_->sendLocalReply(
+          Http::Code::BadRequest, "Cleartext is not permitted", nullptr, absl::nullopt, "");
```
goaway, line 489: Rather than issuing the direct reply, could we simply call the on_error callback here?

goaway:
Thanks @alyssawilk! Overall, I think the setup here looks good, but let me know what you think about the comment above.

library/common/http/client.cc
goaway: I guess the current setup will result in the filter chain being created, whereas firing the callback directly wouldn't(?). I don't know if that's desirable or not. Maybe it is, if someone is listening for errors in a filter?

But I suppose I'm fine either way, as long as we're consistent about we handle this sort of thing for any other cases moving forward.

goaway (approved):
Thanks @alyssawilk!

Like we talked about previously, I will work on an integration test for this once it's merged.

PR #2321, opened 2022-05-26
builder: making compressor configurable

alyssawilk:
As a precursor to adding brotli, making the decompressor configurable.

TODO: wire up the APIs once I have Mike's approval of the general design

Description:
Risk Level: low
Testing: new unit tests
Docs Changes: n/a
Release Notes: NEED TO DO


library/cc/engine_builder.cc
```diff
@@ -92,6 +92,35 @@ EngineBuilder& EngineBuilder::setDeviceOs(const std::string& device_os) {
   return *this;
 }
 
+EngineBuilder& EngineBuilder::setDecompressor(bool decompressor_on) {
+  this->decompressor_filter_ = decompressor_on;
+  return *this;
+}
+
+std::string EngineBuilder::generateFiltersStr() {
```
alyssawilk, line 100: I think with the alias reworking, this is pretty minimal YAML for adjusting the filter chain, WDYT?

library/cc/engine_builder.h
```diff
@@ -33,6 +33,7 @@ class EngineBuilder {
   EngineBuilder& setAppVersion(const std::string& app_version);
   EngineBuilder& setAppId(const std::string& app_id);
   EngineBuilder& setDeviceOs(const std::string& app_id);
+  EngineBuilder& setDecompressor(bool decompressor_on);
```
alyssawilk, line 36: I could imagine int he long term this being a yaml config string but I'd like to keep it simple for now :-)

goaway: This is fine for now. I'd actually like to move away from passing in YAML strings into the builder altogether (and have 
a general alternative approach in mind).

library/common/config/config.cc
```diff
@@ -70,6 +89,34 @@ const std::string config_header = R"(
...
+    host_ttl: 86400s
+    dns_min_refresh_rate: *dns_min_refresh_rate
+    dns_refresh_rate: *dns_refresh_rate
+    dns_failure_refresh_rate:
+      base_interval: *dns_fail_base_interval
+      max_interval: *dns_fail_max_interval
+    dns_query_timeout: *dns_query_timeout
+    typed_dns_resolver_config:
+      name: *dns_resolver_name
+      typed_config: *dns_resolver_config
```
goaway, line 115: I like the cleanup here, but any of these that contain aliases need to be in `config_template` string rather than `config_header` string in order to maintain their ability to be overridden via EngineBuilders. Should be sufficient to simply relocate this block.

goaway (changes requested):
Thanks @alyssawilk - this is indeed what I meant (you can also see another case this is used in EnvoyConfiguration.java).

The cleanup looks good as well - I just see one issue with the setup that should be straightforward to fix.

alyssawilk:
That was actually horribly painful!  Either this was easier than prior changes or I'm finally getting the hang of this.  Good for another pass :-)

library/java/org/chromium/net/impl/NativeCronetEngineBuilderImpl.java
```diff
@@ -127,7 +127,7 @@ private EnvoyConfiguration createEnvoyConfiguration() {
         mDnsRefreshSeconds, mDnsFailureRefreshSecondsBase, mDnsFailureRefreshSecondsMax,
         mDnsQueryTimeoutSeconds, mDnsMinRefreshSeconds, mDnsPreresolveHostnames,
         mDnsFallbackNameservers, mEnableDnsFilterUnroutableFamilies, mEnableDrainPostDnsRefresh,
-        mEnableHttp3, mEnableHappyEyeballs, mEnableInterfaceBinding,
+        mEnableHttp3, true, mEnableHappyEyeballs, mEnableInterfaceBinding,
```
goaway, line 130: nit: The pattern of passing all named values here is sort of convenient for remembering which parameter is which.

goaway (approved):
Just a couple nits - feel free to update as you feel warranted and otherwise merge.

Thanks @alyssawilk!

alyssawilk:
With my almost approval I started on the brotli filter to find it also uses the decompressor filter, so renaming everything to gzip.

goaway (approved):

PR #2324, opened 2022-05-26
Compress xcframework release asset

jpsim:
bazel uses zip as a container with no compression.

Reduces the size of the zip from 1.0GB to 352MB.

Risk Level: Low
Testing: Ran the commands locally
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -126,8 +126,11 @@ jobs:
             --config=remote-ci-macos \
             --remote_header="Authorization=Bearer $GITHUB_TOKEN" \
             //:ios_xcframework
-      - name: 'Move Envoy.xcframework.zip'
-        run: mv bazel-bin/library/swift/Envoy.xcframework.zip .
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -126,8 +126,11 @@ jobs:
             --config=remote-ci-macos \
             --remote_header="Authorization=Bearer $GITHUB_TOKEN" \
             //:ios_xcframework
-      - name: 'Move Envoy.xcframework.zip'
-        run: mv bazel-bin/library/swift/Envoy.xcframework.zip .
+      - name: 'Recompress Envoy.xcframework.zip' # Recompress as bazel uses zip as a container with no compression
```
keith, line 129: really? seems like a bug we should fix upstream since that zip is always meant for distribution

keith (approved):

.github/workflows/release.yml
jpsim: Yes, really:

```
$ du -h bazel-bin/library/swift/Envoy.xcframework.zip
1.0G	bazel-bin/library/swift/Envoy.xcframework.zip
```

PR #2325, opened 2022-05-26
Remove (broken) CocoaPods support

jpsim:
And update documentation to reflect that SwiftPM is supported.

CocoaPods support is blocked on https://github.com/CocoaPods/CocoaPods/issues/11372.

Risk Level: Low
Testing: N/A
Docs Changes: Added
Release Notes: Added


jpsim:
Until CocoaPods support is fixed, there's no point in keeping the CI jobs around since it just adds to CI time without providing any value.

keith (approved):

jpsim:
Posted about the possibility of adding CocoaPods support via dynamic xcframeworks: https://envoyproxy.slack.com/archives/CKQ2LK23G/p1653591518897199

in case possible users find that particularly helpful.

Still merging this PR because doing so would require further work to the release pipeline, and I can pull that in from here if needed.

PR #2332, opened 2022-05-31
stats: include new metrics in config

goaway:

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -459,7 +459,7 @@ stats_config:
             regex: '^http.hcm.downstream_rq_[\w]+'
         - safe_regex:
             google_re2: {}
-            regex: '.*pbf_filter.*'
+            regex: '^pbf_filter.*'
         - safe_regex:
             google_re2: {}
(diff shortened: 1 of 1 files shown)


library/common/config/config.cc
```diff
@@ -454,6 +457,9 @@ stats_sinks: *stats_sinks
         - safe_regex:
             google_re2: {}
             regex: '^http.hcm.downstream_rq_[\w]+'
+        - safe_regex:
+            google_re2: {}
+            regex: '.*pbf_filter.*'
```
Augustyniak, line 462: can we make this regex more specific? Can we specify the prefix as opposed to using `.*` matching?

goaway:
Incidentally, something I considered here, and that we may want to consider for a future iteration, is including the platform filter name in the stat itself. I held off because I wanted to check if the string is sanitized (it is, but I only checked that after the initial implementation), and also because I'm always concerned about variable stat components that could blow out the number of stats emitted. Still, it may be worth considering as some sort of configurable option.

Issue #2100, opened 2022-03-15
[CI] Use Xcode 13.3


jpsim:
This is now the latest stable Xcode release and has some important fixes to Swift Concurrency back-deployment that could be beneficial to us.

* Version: 13.3 (13E113)
* Release Notes: https://developer.apple.com/documentation/xcode-release-notes/xcode-13_3-release-notes
* Download: https://developer.apple.com/services-account/download?path=/Developer_Tools/Xcode_13.3/Xcode_13.3.xip

Will require EngFlow to update their executors just as was done for https://github.com/envoyproxy/envoy-mobile/pull/1997

cc @lfpino @Yannic

lfpino:
Ack @jpsim

jpsim:
@lfpino any progress on this?

In #1997 you mentioned that you were making changes to make this process simpler, so I was hoping it'd be faster to update this time around.

lfpino:
Hi @jpsim, sorry for the delay, yeah we've done some changes to make this faster but still not fully automated. I'll create an internal issue to move forward with this.

jpsim:
If you're just starting on this now, I suggest we look at having Xcode 13.3.1 (13E500a) which is now the latest stable release.

Yannic:
Working on it now, ETA later this week

Yannic:
Unfortunately, this will take longer than expected: Xcode 13.3.1 requires macOS 12, but we're still running macOS 11 on the worker instances 😞

jpsim:
I understand.

I don't know what your mac infra automation looks like, but you may want to look at [tart](https://github.com/cirruslabs/tart) which allows virtualizing macOS based on image snapshots pulled from an OCI-compatible container registry, similar to how Docker works.

jpsim:
I updated the ticket to reflect that we should just jump straight to 13.4 (13F17a), which is now the latest stable release, and similarly also requires macOS 12.

lfpino:
Thanks for the tips @jpsim, we have an internal issue to upgrade to MacOS 12 and will keep you posted.

jpsim:
Thanks for the update. Looking forward to it and I hope you find a good solution to make the process smoother moving forward. I'm happy to workshop some ideas with you.

PR #2335, opened 2022-06-02
swift: generate API docs as a DocC archive

jpsim:
And include in GitHub release artifacts

Updates rules_swift to pull in https://github.com/bazelbuild/rules_swift/pull/838

When users open the archive, they'll have rich docs they can explore using Xcode's documentation navigator.

When we update to Xcode 13.4, we'll be able to generate HTML docs with this approach and publish to the documentation website. Here's a demo of what that static site looks like: https://envoy-mobile-docc-demo.netlify.app/documentation/envoy/

https://user-images.githubusercontent.com/474794/171704142-993072a3-fbc2-4db6-9c70-65e9e1a47aac.mp4

Risk Level: Low
Testing: Local, CI
Docs Changes: Added
Release Notes: Added

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -132,7 +132,7 @@ jobs:
         with:
           name: ios_framework
           path: ./Envoy.xcframework.zip
-      - name: 'Build Envoy.doccarchive'
+      - name: 'Build Envoy.doccarchive.zip'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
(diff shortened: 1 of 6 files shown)


goaway (approved):
This is fantastic, thank you @jpsim!

Ultimately, it would be great if we could find a way to also publish these to EM's official documentation tree, but this is a great first step.

jpsim:
> Ultimately, it would be great if we could find a way to also publish these to EM's official documentation tree, but this is a great first step.

I completely agree, and this will be possible when we update to Xcode 13.4, tracked in https://github.com/envoyproxy/envoy-mobile/issues/2100.

jpsim:
> this will be possible when we update to Xcode 13.4, tracked in #2100.

For more context, this is the Xcode 13.4 feature we need for static website hosting: https://github.com/apple/swift-docc/pull/44

PR #2345, opened 2022-06-08
iOS: add a way to disable network monitoring

jpsim:
We may be interested in experimenting with network monitoring & interface switching completely disabled to assess the impact of this feature.

Disabling network switching is not something we recommend generally.

Risk Level: Low, adds the ability to disable network switching on an opt-in basis
Testing: Updated unit tests
Docs Changes: Added
Release Notes: Added


goaway (approved):
Thanks @jpsim

PR #2347, opened 2022-06-08
iOS: stop embedding bitcode in releases

jpsim:
Bitcode is all but dead at this point.

Here's a deprecation note from the Xcode 14 beta 1 release notes:

https://developer.apple.com/documentation/Xcode-Release-Notes/xcode-14-release-notes

> Starting with Xcode 14, bitcode is no longer required for watchOS and
> tvOS applications, and the App Store no longer accepts bitcode
> submissions from Xcode 14.

> Xcode no longer builds bitcode by default and generates a warning
> message if a project explicitly enables bitcode: “Building with
> bitcode is deprecated. Please update your project and/or target
> settings to disable bitcode.” The capability to build with bitcode
> will be removed in a future Xcode release. IPAs that contain bitcode
> will have the bitcode stripped before being submitted to the App
> Store. Debug symbols for past bitcode submissions remain available
> for download. (86118779)

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -93,10 +93,8 @@ build:release-common --config=sizeopt
 build:release-common --copt=-fvisibility=hidden
 
 # Flags for release builds targeting iOS
-build:release-ios --apple_bitcode=embedded
 build:release-ios --config=ios
 build:release-ios --config=release-common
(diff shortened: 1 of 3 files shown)


keith (approved):
yay

keith (approved):

PR #2349, opened 2022-06-09
iOS: add configurable gzip and brotli decompression options

goaway:
Description: Previously, gzip decompression was always enabled in base configuration. This makes both gzip and brotli decompression configurable, with gzip defaulting to on (as it was previously) and brotli defaulting to off. No changes to usage of EngineBuilder need to be made to maintain existing behavior.
Risk Level: Moderate
Testing: Added config coverage.


jpsim (approved):
Assuming these aren’t mutually exclusive. Thanks for adding docs and release notes!

goaway:
cc @alyssawilk

PR #2351, opened 2022-06-09
cleanup: consistently use nullptr in cc contexts

goaway:


jpsim (approved):
Is there no clang-tidy configuration for this?

goaway:
> Is there no clang-tidy configuration for this?

Could be; probably worth looking into.

goaway:
Thanks @jpsim!

jpsim:
> Could be; probably worth looking into.

Looks like it: https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-nullptr.html

jpsim:
It not clear to me how to enable that in `.clang-format`

PR #2355, opened 2022-06-09
config: hopefully fixing C++ config default for apple

alyssawilk:
Replacing the default dns resolver in the config string to be correct for the OS in question.
Replaced overrides in the swift code now that the base config handles it.

Risk Level: low
Testing: existing tests should verify correctness
Docs Changes: n/a
Release Notes: did not add

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -69,9 +69,17 @@ const std::string config_header = R"(
 - &dns_multiple_addresses true
 - &dns_preresolve_hostnames []
 - &dns_refresh_rate 60s
-- &dns_resolver_name envoy.network.dns_resolver.cares
+)"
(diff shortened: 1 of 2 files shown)


jpsim:
Can you also remove this line? https://github.com/envoyproxy/envoy-mobile/blob/2ea19b7f8862f8d102750b1e9b95350969eb8c2e/library/objective-c/EnvoyConfiguration.m#L157

alyssawilk:
Good call, thanks!

jpsim (approved):
Thanks!

goaway (approved):
Makes sense to me. Longer-term, I can see this being potentially useful to expose through the builder, but not as anything that would need to block this change. Thanks @alyssawilk!

Issue #2303, opened 2022-05-19
move c++ integration tests over to fully using the builder APIs


alyssawilk:
ideally this should be a follow on to https://github.com/envoyproxy/envoy-mobile/pull/2293 so we're consistently testing the full builder, not just the config generation bits of it.

alyssawilk:
So I've made some headway on this but I think I'm getting lost in the APIs somewhere.

From existing C++ tests it looks like the C++ way of doing these things is 
engine_->streamClient()->newStreamPrototype();
and from there I can replace the existing callbacks of the form of 

-    bridge_callbacks_.on_headers = [](envoy_headers c_headers, bool, envoy_stream_intel intel,
-                                      void* context) -> void* {

with

 .setOnHeaders([this](Platform::ResponseHeadersSharedPtr headers, bool) {

but that doesn't have the intel.

While I'm game to migrate the tests to using the builders, I don't think I want to embark on the yak shaving mission which is porting them to raw C, or to changing all the C++ APIs to match the C APIs.  @goaway is there a simple way to just get the Http::Client the test is looking for or am I good to put this on the back burner for a bit?

PR #2362, opened 2022-06-09
test: got the client_integration_test moved over to the builder.

alyssawilk:
Risk Level: low
Testing: yep
Release Notes: inline
Fixes https://github.com/envoyproxy/envoy-mobile/issues/2303


alyssawilk:
@goaway I expect this to fail CI as I haven't gotten the RTDS test passing yet, but client integration test is good to go and there's a lot of moving parts here I'd like your thoughts on, especially the weirdness of taking config from the builder, using it to initialize the integration test config helper, then serializing the config helper back to the builder so we can test custom configs.
(not sure how else we do RTDS, happy to have your thoughts here as I can tweak in parallel with fixing up the rtds test)

thanks!

goaway:
Discussed a bit over Slack. Overall this seems fine to me, though at least on a case-by-case basis I think it's worth considering whether a given override might simply make sense as a Builder option. I'll also be thinking over possibilities for allowing for direct overrides - at least on C++ - rather than needing to do the serialization ahead of time and passing that back through.

library/cc/stream.h
```diff
@@ -18,7 +18,7 @@ class Stream {
   Stream(envoy_engine_t engine_handle, envoy_stream_t handle);
 
   Stream& sendHeaders(RequestHeadersSharedPtr headers, bool end_stream);
-  Stream& sendData(envoy_data data);
+  Stream& sendData(envoy_data data, bool end_stream);
```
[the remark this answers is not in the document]
alyssawilk: but doesn't sendHeaders with end stream implicitly close?  That's a pretty weird abstraction IMO but yeah this PR is not the place to change it.

alyssawilk:
ok as discussed, moved the default to "use builder config" but I don't think we want to block this PR on builder RTDS support.

library/cc/engine_builder.h
```diff
@@ -46,6 +47,9 @@ class EngineBuilder {
   // Filter): EngineBuilder& addNativeFilter(name: String = UUID.randomUUID().toString(),
   // typedConfig: String): EngineBuilder& addStringAccessor(name: String, accessor:
   // EnvoyStringAccessor): EngineBuilder {
+protected:
+  void setOverrideConfigForTests(std::string config) { config_override_for_tests_ = config; }
+  void setAdminAddressPathForTests(std::string admin) { admin_address_path_for_tests_ = admin; }
```
alyssawilk, line 52: alternately could publicly add key-value and hash map for generic command line override.

library/cc/request_headers_builder.cc
```diff
@@ -28,6 +28,11 @@ RequestHeadersBuilder::addUpstreamHttpProtocol(UpstreamHttpProtocol upstream_htt
   return *this;
 }
 
+RequestHeadersBuilder& RequestHeadersBuilder::addKeyValue(std::string key, std::string value) {
```
goaway, line 31: nit: I'd call this`set`, for alignment with the other builders: https://github.com/envoyproxy/envoy-mobile/blob/main/library/kotlin/io/envoyproxy/envoymobile/RequestHeadersBuilder.kt#L43

goaway (dismissed):

jpsim (dismissed):
Yay, one more up to date API!

PR #2363, opened 2022-06-10
ci: pin external GitHub Action

goaway:
Description: Improves security with respect to third-party actions. See:
https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions

I also read through the recent commits and updated to the latest release, since it contained a few bugfixes.

Risk: Low
Testing: CI


Augustyniak (approved):
In general the SHA seems to be a valid `peter-evans/create-pull-request` SHA. One thing that I want to point out in public - and that was briefly discussed by us offline - is that this PR moves us from `v3` to `v4.0.4`.

PR #2366, opened 2022-06-14
iOS: fix xcframework upload in release workflow

jpsim:


jpsim:
This morning's automated release failed because of this: https://github.com/envoyproxy/envoy-mobile/runs/6862017912?check_suite_focus=true

keith (approved):

PR #2367, opened 2022-06-14
stats: disable global stats in config

jpsim:
These are already being filtered out of our stats inclusion list, so this isn't a user-facing change, but this is needed to support the singleton removal work in https://github.com/envoyproxy/envoy-mobile/pull/2129.

Splitting this out into its own PR to reduce the scope of the singleton removal change.

Risk Level: Low
Testing: None
Docs Changes: N/A, not a user-facing change
Release Notes: N/A, not a user-facing change

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -508,6 +508,8 @@ layered_runtime:
     - name: static_layer_0
       static_layer:
         envoy:
+          # This disables envoy bug stats, which are filtered out of our stats inclusion list anyway
+          disallow_global_stats: true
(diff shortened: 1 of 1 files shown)


alyssawilk (approved):
LGTM though it might be worth explaining why (global stats not working for multi-envoy)

alyssawilk:
haha I meant in the code, but worst case folks can blame and look a the PR :-P

jpsim:
Added some more inline in the code here https://github.com/envoyproxy/envoy-mobile/pull/2371

PR #2368, opened 2022-06-15
api: return opaque engine handle in `onEngineRunning`

jpsim:
To support interop between APIs across different languages.

This is needed to support the engine singleton removal work in https://github.com/envoyproxy/envoy-mobile/pull/2129 because previously no matter the language, you could always assume that the engine handle would be `1`, but with the singleton removed, we need to expose a mechanism to pass the engine instance from the higher level language APIs to the C++ APIs.

Risk Level: Moderate
Testing: Unit tests, integration tests and in Lyft's downstream Envoy Mobile cross-language integration
Docs Changes: Updated
Release Notes: Added

To Do:

- [x] Update Swift
- [x] Update ObjC
- [x] Update Java
- [x] Update Kotlin
- [x] Update C++
- [x] Update Python (not exposing in the API for now)


jpsim:
I'll go with a different approach.

PR #2369, opened 2022-06-15
bazel: use `device_and_simulator` rule

jpsim:
To simplify running sample iOS apps for physical devices.

### DIFF
--- a/docs/root/development/debugging/ios_local.rst
+++ b/docs/root/development/debugging/ios_local.rst
@@ -55,21 +55,12 @@ compiled source files in C/C++/Objective-C/Objective-C++/Swift:
 Running on a real iPhone
 ------------------------
 
-Although building and running on a physical device is
-`not yet officially supported <https://github.com/buildbuddy-io/rules_xcodeproj/issues/285>`_
-by rules_xcodeproj, following these steps may work:
+You can build and run the example iOS apps on a physical iPhone by
+following these steps:
 
-1. Add ``build --ios_multi_cpus=arm64`` to ``.bazelrc``.
-2. Add a ``.mobileprovision`` provisioning profile capable of signing
-   the app you want to run next to its ``BUILD`` file.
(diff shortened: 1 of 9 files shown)


examples/BUILD
```diff
@@ -0,0 +1,26 @@
...
+    name = "ios_provisioning_profile",
+    managed_by_xcode = True,
+    provisioning_profile = ":xcode_managed_profile",
+    tags = ["manual"],
+    visibility = ["//visibility:public"],
+)
+
+local_provisioning_profile(
+    name = "xcode_managed_profile",
+    profile_name = "iOS Team Provisioning Profile: *",
```
brentleyjones, line 23: Since this will be different for some accounts, I think it would be worthwhile to extract this to `PROFILE_NAME` like so:

```
# If needed, change to the name shown by clicking the (i) next to
# "Xcode Managed Profile" in the "Signing and Capabilities" tab
PROFILE_NAME = "iOS Team Provisioning Profile: *"
```

docs/root/development/debugging/ios_local.rst
```diff
@@ -55,21 +55,12 @@ compiled source files in C/C++/Objective-C/Objective-C++/Swift:
...
-4. Remove all apps without a provisioning profile from the ``xcodeproj``
-   configuration in Envoy Mobile's root ``BUILD`` file.
-5. Follow the same steps as defined in the
+You can build and run the example iOS apps on a physical iPhone by
+following these steps:
+
+1. Set the ``TEAMID`` value in ``examples/BUILD`` to your Apple
+   Developer Team ID found in the
+   `Apple Developer Portal <https://developer.apple.com/account/#!/membership>`_.
+2. Follow the same steps as defined in the
```
brentleyjones, line 64: If you extract the `PROFILE_NAME` per my other comment, I would add a step between 1. and 2. here calling out potentially needing to change that as well.

brentleyjones (dismissed):

PR #2380, opened 2022-06-21
Bump Envoy sha to f49d4f2

Augustyniak:
Description: Bumping upstream Envoy sha so that it includes recent changes that allow us to force IPv6 use on Android. The Envoy sha that we use in this PR is NOT from Envoy main. The idea is to pick up only @mattklein123's change from https://github.com/envoyproxy/envoy/pull/21803 without any other changes. We are going to move back to Envoy main later this week.
Testing: Manual, passed
Docs Changes: N/A
Release Notes: N/A


goaway (approved):

Augustyniak:
I see some timeouts - I think that they are related to the changes from https://github.com/envoyproxy/envoy-mobile/pull/2184 that removed dependency of some of the CI checks on `androidbuild` and `iosbuild` jobs.

PR #2383, opened 2022-06-23
api: make iOS Headers and HeadersBuilder case-insensitive

Augustyniak:
Description: Make the lookup of headers in `HeadersBuilder` and `Headers` case-insensitive _and_ preserve the original casing of headers. Implemented on iOS only. Android side-changes are going to be implemented in a follow up PR. GH issue: https://github.com/envoyproxy/envoy-mobile/issues/2390.
Risk Level: Low
Testing: Unit Tests
Docs Changes: Updated
Release Notes: Updated


Augustyniak:
@goaway This is now ready for review. I incorporate your offline feedback with regard to making `Headers` case-insensitive too.

Augustyniak:
I would expect Dr String to catch the issue that I found manually in [6389322](https://github.com/envoyproxy/envoy-mobile/pull/2383/commits/6389322f5fb40484a56ea7ea4891e042900f7aa7). @jpsim any idea as to why this issue wasn't caught?

jpsim:
DrString fix: https://github.com/envoyproxy/envoy-mobile/pull/2389

jpsim:
Can you update this TODO?

https://github.com/envoyproxy/envoy-mobile/blob/b357bab025b0e7f10fdf8fe1a23a10dc0b41c290/examples/swift/async_await/ContentView.swift#L78-L80

library/swift/HeadersBuilder.swift
```diff
@@ -3,14 +3,17 @@ import Foundation
 private let kRestrictedPrefixes = [":", "x-envoy-mobile"]
 
 private func isRestrictedHeader(name: String) -> Bool {
-  return name == "host" || kRestrictedPrefixes.contains { name.hasPrefix($0) }
+  let lowercasedName = name.lowercased()
+  return lowercasedName == "host" || kRestrictedPrefixes.contains { lowercasedName.hasPrefix($0) }
```
jpsim, line 7: Eagerly converting each character to its lowercase form could be expensive.

Can you use Foundation's case-insensitive comparison APIs instead? E.g. https://developer.apple.com/documentation/foundation/nsstring/1416849-range

Augustyniak: good idea, Note - we will still need to lowercase inside of the implementation of `HeadersContainer`. We could probably avoid that if we decide to use some 'smart' container in there but not sure whether it's worth doing for now.

library/swift/Headers.swift
```diff
@@ -4,45 +4,58 @@ import Foundation
 /// To instantiate new instances, see `{Request|Response}HeadersBuilder`.
 @objcMembers
 public class Headers: NSObject {
```
jpsim, line 6: Can we merge the `Headers` type with the `HeadersContainer` type? I don't see any reason to add that additional layer, but maybe I'm missing a reason why it's necessary?

library/swift/HeadersBuilder.swift
Augustyniak: We could do some tweaks to make Swift dictionary case-insensitive (i.e., https://stackoverflow.com/a/47134176) but not sure what's cheaper - iterating over all keys and doing case insensitive comparisons or lowercasing the header name. I could do some benchmarking here but my light preference (I am biased) it to keep the lowercasing operations they are now. Let me know what you think.

library/swift/HeadersBuilder.swift
```diff
@@ -3,14 +3,19 @@ import Foundation
 private let kRestrictedPrefixes = [":", "x-envoy-mobile"]
 
 private func isRestrictedHeader(name: String) -> Bool {
-  return name == "host" || kRestrictedPrefixes.contains { name.hasPrefix($0) }
+  let isHostHeader = name.range(of: "host", options: [.caseInsensitive]) != nil
```
jpsim, line 6: This will match `FOOHOSTBAR`. This should work:

```suggestion
  let isHostHeader = name.caseInsensitiveCompare("host") == .orderedSame
```

Augustyniak: sorry, that's a pretty high level of sloppiness from my side. Updating.

library/swift/Headers.swift
```diff
@@ -4,45 +4,58 @@ import Foundation
...
-  let headers: [String: [String]]
+  let container: HeadersContainer
 
   /// Get the value for the provided header name.
   ///
-  /// - parameter name: Header name for which to get the current value.
+  /// - parameter name: The case-insensitive header name for which to
+  ///                   get the current value.
+  ///
```
jpsim, line 13: Nit: Extra line

```suggestion
```

library/swift/Headers.swift
```diff
@@ -4,45 +4,58 @@ import Foundation
 /// To instantiate new instances, see `{Request|Response}HeadersBuilder`.
 @objcMembers
 public class Headers: NSObject {
-  let headers: [String: [String]]
+  let container: HeadersContainer
 
   /// Get the value for the provided header name.
   ///
-  /// - parameter name: Header name for which to get the current value.
+  /// - parameter name: The case-insensitive header name for which to
```
jpsim, line 11: Nit: I'd revert this to the previous description since there's nothing about this value that's case-insensitive, it's the lookup operation that's case-insensitive. I'd add a `- note:` to that effect instead.

library/swift/Headers.swift
```diff
@@ -4,45 +4,58 @@ import Foundation
...
+  ///                   get the current value.
+  ///
   ///
   /// - returns: The current headers specified for the provided name.
   public func value(forName name: String) -> [String]? {
-    return self.headers[name]
+    return self.container.value(forName: name)
   }
 
   /// Accessor for all underlying headers as a map.
```
jpsim, line 20: I'd add a `- warning:` to discourage using this dictionary to do equality key-based lookups, possibly linking to http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2 for context.

library/swift/Headers.swift
```diff
@@ -4,45 +4,58 @@ import Foundation
...
   ///
   /// - returns: The underlying headers.
   public func allHeaders() -> [String: [String]] {
```
jpsim, line 23: Consider renaming this function to `caseSensitiveHeaders()` to raise awareness to the fact that users will likely need to perform case-insensitive lookups if they want to find values for given keys.

library/swift/HeadersBuilder.swift
```diff
@@ -3,14 +3,19 @@ import Foundation
 private let kRestrictedPrefixes = [":", "x-envoy-mobile"]
 
 private func isRestrictedHeader(name: String) -> Bool {
-  return name == "host" || kRestrictedPrefixes.contains { name.hasPrefix($0) }
+  let isHostHeader = name.caseInsensitiveCompare("host") == .orderedSame
+  let hasRestrictedPrefix = kRestrictedPrefixes
```
jpsim, line 7: Avoid eagerly computing this if the name is `host`

```suggestion
  lazy var hasRestrictedPrefix = kRestrictedPrefixes
```

library/swift/HeadersBuilder.swift
```diff
@@ -3,14 +3,19 @@ import Foundation
...
+    .contains { name.range(of: $0, options: [.caseInsensitive, .anchored]) != nil }
+  return isHostHeader || hasRestrictedPrefix
 }
 
 /// Base builder class used to construct `Headers` instances.
+/// It preserves the original casing of headers and enforces
+/// a case-insensitive look up and setting of headers.
```
jpsim, line 14: Nit: lookup is a noun in this case

```suggestion
/// a case-insensitive lookup and setting of headers.
```

library/swift/HeadersContainer.swift
```diff
@@ -0,0 +1,110 @@
+/// The container which manages the underlying headers map.
+/// It maintains the original casing of passed header names.
+/// It treats headers names as case-insensitive for the purpose
+/// of header lookups and header name conflict resolutions.
+struct HeadersContainer: Equatable {
+  private var headers: [String: Header]
+
+  // Represents a headers name together with all of its values.
+  // It preserves the original casing of the header name.
```
jpsim, line 9: Nit: this reads as a doctring

```suggestion
  /// Represents a headers name together with all of its values.
  /// It preserves the original casing of the header name.
```

library/swift/HeadersContainer.swift
```diff
@@ -0,0 +1,110 @@
...
+  /// Initialize a new instance of the receiver using the provided headers map.
+  ///
+  /// - parameter headers: The headers map.
+  init(headers: [String: [String]]) {
+    var underlyingHeaders = [String: Header]()
+    for (name, value) in headers {
+      let lowercasedName = name.lowercased()
+      /// Dictionaries in Swift are unordered collections. Process headers with names
+      /// that are the same when lowercased in an alphabetical order to avoid a situation
+      /// in which the result of the initialization is underministic i.e., we want
```
jpsim, line 37:
```suggestion
      /// in which the result of the initialization is non-derministic i.e., we want
```

jpsim (commented):
I'm wondering if the internal APIs that currently take header dictionaries should be updated to take a `Headers` instance instead.

Also, we should consider adding subscripting support to `Headers` for more ergonomic lookups.

library/swift/Headers.swift
Augustyniak: added warning

library/swift/Headers.swift
Augustyniak: added note

library/swift/Headers.swift
Augustyniak: updated the name

Augustyniak:
>I'm wondering if the internal APIs that currently take header dictionaries should be updated to take a Headers instance instead.

Created an issue to track this https://github.com/envoyproxy/envoy-mobile/issues/2394.

> Also, we should consider adding subscripting support to Headers for more ergonomic lookups.

Created an issue to track this https://github.com/envoyproxy/envoy-mobile/issues/2393

library/swift/Headers.swift
Augustyniak: One of my previous solutions was to merge `Headers` and `HeadersContainer` but I decided against it as:

- I wanted `HeadersContainers` to be a struct - not possible with `Headers` as it has to inherit from `NSObject`. There are some negative of using structs tho - more copying.
- `HeadersBuilder` depends on a lot of stuff that's currently in `HeadersContainer`. If we merge `Headers` and `HeadersContainer` then `HeadersBuilder `would either have to have duplicated logic between `Headers `and `HeadersBuilder` or make `HeadersBuilder` depend on `Headers`. I decided that having a separate `HeadersContainer` class was a better approach to ensure that all of the case-insensitive lookup logic lives in one place and is shared between `Headers` and `HeaderBuilder`.

jpsim (dismissed):

jpsim (approved):
🙌

examples/objective-c/hello_world/ViewController.m
```diff
@@ -90,9 +90,9 @@ - (void)performRequest {
     NSString *message = [NSString stringWithFormat:@"received headers with status %i", statusCode];
 
     NSMutableString *headerMessage = [NSMutableString new];
-    for (NSString *name in headers.allHeaders) {
+    for (NSString *name in headers.caseSensitiveHeaders()) {
```
goaway, line 93: I like this rename. 👍

goaway (dismissed):
I like this solution. Nice work @Augustyniak.

jpsim (approved):

PR #2387, opened 2022-06-27
ci: change submodule_update PR branch name to use a timestamp strategy

jpsim:
To prevent it from force-pushing to an existing branch like it did in https://github.com/envoyproxy/envoy-mobile/pull/2373


Augustyniak (approved):

Augustyniak:
Thank you for following up on this one.

PR #2348, opened 2022-06-09
docs: explain admin interface and stats

Augustyniak:
Description: Add documentation for stats and admin interface features.
Risk Level: None
Testing: N/A
Docs Changes: N/A
Release Notes: N/A

PR #2388, opened 2022-06-28
admin: disable admin html

jpsim:
This isn't something we want to expose when running Envoy Mobile.

For more details: https://github.com/envoyproxy/envoy/pull/21814

Risk Level: Low, will impact anyone who's used this feature before.
Testing: Verified that `admin_html.cc` is not compiled.
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -3,6 +3,7 @@ try-import ./envoy/.bazelrc
 
 # Common flags for all builds
 build --platform_mappings=envoy/bazel/platform_mappings
+build --define=admin_html=disabled
 build --define=google_grpc=disabled
 build --define=hot_restart=disabled
 build --define=tcmalloc=disabled


Augustyniak:
How does this work with `enableAdminInterface` method that we currently have on the builder itself? I assume that it will no longer be possible to visit an admin webpage after we call `enableAdminInterface` on the builder (and build the engine)?

jpsim:
> How does this work with `enableAdminInterface` method that we currently have on the builder itself?

I was only vaguely aware of this configuration point. I didn't know that the html interface was what we were explicitly exposing. Knowing that this is an explicitly enabled feature changes the discussion on this for sure.

I had started a discussion on Slack here: https://envoyproxy.slack.com/archives/C02F93EEJCE/p1656350561377359

> I assume that it will no longer be possible to visit an admin webpage after we call `enableAdminInterface` on the builder (and build the engine)?

Yes that's my understanding as well based on the description in https://github.com/envoyproxy/envoy/pull/21814.

Augustyniak (approved):

Augustyniak:
> I was only vaguely aware of this configuration point. I didn't know that the html interface was what we were explicitly exposing. Knowing that this is an explicitly enabled feature changes the discussion on this for sure.

The feature was about exposing `curl localhost:9901/stats` and not the admin (`/`) endpoint itself so I think that it should be fine to proceed here. I have a PR that explains how to use EM admin facilities in https://github.com/envoyproxy/envoy-mobile/pull/2348

snowp:
This should be good to merge no?

jpsim:
No, we decided to keep the feature enabled in our last community meeting, but to eventually replace it with safer programmatic APIs to provide the same kind of debugging features.

PR #2401, opened 2022-07-01
connectivity_manager: rename/refactor from Network::Configurator

goaway:
Description: This is intended to better capture the role and responsibilities of this class in Envoy Mobile. It still "configures connections", but more generally, it holds library/device connectivity state and provides mechanisms to leverage it.
Risk Level: Low
Testing: Updated tests.
Docs Changes: N/A
Release Notes: N/A


library/common/extensions/retry/options/network_configuration/predicate.cc
```diff
@@ -11,8 +11,10 @@ NetworkConfigurationRetryOptionsPredicate::NetworkConfigurationRetryOptionsPredi
     const envoymobile::extensions::retry::options::network_configuration::
         NetworkConfigurationOptionsPredicate&,
     Upstream::RetryExtensionFactoryContext& context) {
-  network_configurator_ = Network::ConfiguratorHandle{context.singletonManager()}.get();
-  RELEASE_ASSERT(network_configurator_ != nullptr, "unexpected nullptr network configurator");
+  network_connectivity_manager_ =
+      Network::ConnectivityManagerHandle{context.singletonManager()}.get();
+  RELEASE_ASSERT(network_connectivity_manager_ != nullptr,
+                 "unexpected nullptr network connectivity_manager");
```
Augustyniak, line 17:
```suggestion
                 "unexpected nullptr network_connectivity_manager");
```

Augustyniak (commented):
left one specific comment plus an overall comment: Not sure whether we need `network` prefix in variables/methods that works with `Network::ConnectivityManager` - "connectivity" part of the name itself seems to be implying "network" but it's really just me nit picking. Consider dropping it.

Augustyniak (approved):

PR #2415, opened 2022-07-12
iOS: fix retain cycles in `EnvoyNetworkMonitor`

jpsim:
And properly clean up reachability ref.

Risk Level: Low
Testing: Validated using the memory graph debugger in Xcode with both reachability and path monitor modes
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/library/objective-c/EnvoyNetworkMonitor.m
+++ b/library/objective-c/EnvoyNetworkMonitor.m
@@ -26,6 +26,11 @@
   if (_path_monitor) {
     nw_path_monitor_cancel(_path_monitor);
   }
+  if (_reachability_ref) {
+    SCNetworkReachabilitySetCallback(_reachability_ref, nil, nil);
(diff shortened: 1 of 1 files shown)


library/objective-c/EnvoyNetworkMonitor.m
```diff
@@ -38,6 +43,7 @@ - (void)startPathMonitor {
   nw_path_monitor_set_queue(_path_monitor, queue);
 
   __block envoy_network_t previousNetworkType = -1;
+  envoy_engine_t engineHandle = _engineHandle;
```
Augustyniak, line 46: Why was this needed?

Augustyniak (approved):

library/objective-c/EnvoyNetworkMonitor.m
jpsim: It was causing a retain cycle in the nw update block.

PR #2419, opened 2022-07-14
dns: allow using the `getaddrinfo`-based system DNS resolver

jpsim:
For non-Apple platforms like Linux and Android.

Updated the Kotlin, Java & C++ builders.

Added here: https://github.com/envoyproxy/envoy/pull/22080

Risk Level: Low to moderate, off by default but I had to refactor some stuff.
Testing: Added.
Docs Changes: Documentation comments.
Release Notes: Added.

### DIFF
--- a/docs/root/intro/version_history.rst
+++ b/docs/root/intro/version_history.rst
@@ -44,7 +44,7 @@ Features:
 - android: create simple persistent SharedPreferencesStore (:issue: `#2319 <2319>`)
 - iOS: A documentation archive is now included in the GitHub release artifact (:issue: `#2335 <2335>`)
 - api: improved C++ APIs compatibility with Java / Kotlin / Swift (:issue `#2362 <2362>`)
-- api: add option to use the a ``getaddr``-based system DNS resolver instead of c-ares (:issue: `#2419 <2419>`)
+- api: add option to use the a ``getaddrinfo``-based system DNS resolver instead of c-ares (:issue: `#2419 <2419>`)
 
 0.4.6 (April 26, 2022)
 ========================
--- a/docs/root/intro/version_history.rst
+++ b/docs/root/intro/version_history.rst
@@ -44,6 +44,7 @@ Features:
 - android: create simple persistent SharedPreferencesStore (:issue: `#2319 <2319>`)
 - iOS: A documentation archive is now included in the GitHub release artifact (:issue: `#2335 <2335>`)
 - api: improved C++ APIs compatibility with Java / Kotlin / Swift (:issue `#2362 <2362>`)
+- api: add option to use the a ``getaddr``-based system DNS resolver instead of c-ares (:issue: `#2419 <2419>`)
 
 0.4.6 (April 26, 2022)
 ========================
--- a/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
+++ b/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
@@ -36,6 +36,7 @@ public class EnvoyConfiguration {
   public final String dnsPreresolveHostnames;
   public final List<String> dnsFallbackNameservers;
   public final Boolean dnsFilterUnroutableFamilies;
+  public final Boolean dnsUseSystemResolver;
   public final Boolean enableDrainPostDnsRefresh;
   public final Boolean enableHttp3;
   public final Boolean enableGzip;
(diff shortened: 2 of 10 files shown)


docs/root/intro/version_history.rst
```diff
@@ -44,6 +44,7 @@ Features:
 - android: create simple persistent SharedPreferencesStore (:issue: `#2319 <2319>`)
 - iOS: A documentation archive is now included in the GitHub release artifact (:issue: `#2335 <2335>`)
 - api: improved C++ APIs compatibility with Java / Kotlin / Swift (:issue `#2362 <2362>`)
+- api: add option to use the a ``getaddr``-based system DNS resolver instead of c-ares (:issue: `#2419 <2419>`)
```
Augustyniak, line 47:
```suggestion
- api: add option to use the a ``getaddrinfo``-based system DNS resolver instead of c-ares (:issue: `#2419 <2419>`)
```

library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -228,10 +232,19 @@ String resolveTemplate(final String configTemplate, final String platformFilterT
...
+    if (dnsUseSystemResolver) {
+      dnsResolverName = "envoy.network.dns_resolver.getaddrinfo";
+      dnsResolverConfig =
+          "{\"@type\":\"type.googleapis.com/envoy.extensions.network.dns_resolver.getaddrinfo.v3.GetAddrInfoDnsResolverConfig\"}";
+    } else {
+      dnsResolverName = "envoy.network.dns_resolver.cares";
+      dnsResolverConfig = String.format(
+          "{\"@type\":\"type.googleapis.com/envoy.extensions.network.dns_resolver.cares.v3.CaresDnsResolverConfig\",\"resolvers\":%s,\"use_resolvers_as_fallback\": %s, \"filter_unroutable_families\": %s}",
+          dnsFallbackNameserversAsString, !dnsFallbackNameservers.isEmpty() ? "true" : "false",
+          dnsFilterUnroutableFamilies ? "true" : "false");
```
Augustyniak, line 246: no action needed - just want to point out that for our new resolver `dnsFilterUnroutableFamilies` is effectively always equal to `true` as I pointed out in https://github.com/envoyproxy/envoy/pull/22080/files#r919203157. 

Just something for future us - we will need to clean up all of the DNS flags/options that we expose.

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -218,6 +219,19 @@ open class EngineBuilder(
     return this
   }
 
+  /**
+   * Specify whether to use the getaddr-based system DNS resolver or the c-ares resolver.
```
Augustyniak, line 223: I start to think that it was not a typo that you wrote `getaddr` but not sure why we would not want to mention the whole name of the method.

```suggestion
   * Specify whether to use the getaddrinfo-based system DNS resolver or the c-ares resolver.
```

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -218,6 +219,19 @@ open class EngineBuilder(
...
+   * Defaults to false.
```
Augustyniak, line 224: Should we add a `note` in here that would explain that specifying that EM should use system resolves causes EM ignore some of the other options exposed by the builder - mainly min DNS refresh and filtering of unroutable families. We could mention that `enableDNSUseSystemResolver` means DNS refresh time equal to 60s and enabled filtering of unroutable families.

library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -77,6 +78,7 @@ public enum TrustChainVerification {
    * @param dnsPreresolveHostnames       hostnames to preresolve on Envoy Client construction.
    * @param dnsFallbackNameservers       addresses to use as DNS name server fallback.
    * @param dnsFilterUnroutableFamilies  whether to filter unroutable IP families or not.
+   * @param dnsUseSystemResolver         whether to use the getaddr-based system resolver or c-ares.
```
jpsim, line 81:
```suggestion
   * @param dnsUseSystemResolver         whether to use the getaddrinfo-based system resolver or c-ares.
```

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
jpsim: Yes.

Augustyniak (dismissed):

Augustyniak (approved):

PR #2421, opened 2022-07-14
Remove Tulsi configuration & docs

jpsim:
Now that we've integrated with rules_xcodeproj and that's working well.


Augustyniak (approved):

Augustyniak:
Nice!

PR #2342, opened 2022-06-07
Adding knobs to configure brotli

alyssawilk:
Adding the option to configure brotli in the bulder

Risk Level: low off by default
Testing: cc unit tests, updated java tests
Docs Changes: n/a
Release Notes: inline

PR #2423, opened 2022-07-14
Add a filter for applying Android socket tags

RyanTheOptimist:
Add a filter for applying Android socket tags

Testing:  Functionality verified as part of the run of the experimentation example app. Also filed #2494 to track improvements.
Docs Changes: Updated docs/root/api/starting_envoy.rst
Release Notes: Added


RyanTheOptimist:
@alyssawilk can you take a look? I know I need tests, but hopefully I can lean on whatever testing gets set up for the cleartext check.

library/common/config/config.cc
```diff
@@ -206,6 +209,8 @@ const char* config_template = R"(
     typed_config: *local_error_config
   - name: envoy.filters.http.dynamic_forward_proxy
     typed_config: *dfp_config
+  - name: envoy.filters.http.socket_tag
+    typed_config: *socket_tag_config
```
alyssawilk, line 213: If we're going to have this on by default, we should call out in the release notes

alyssawilk: Also I wonder if it should live in library/common/extensions/filters/http/network_configuration/filter.cc ?
That's where @Augustyniak is doing the proxy work I believe

library/common/extensions/filters/http/socket_tag/filter.cc
```diff
@@ -0,0 +1,41 @@
...
+namespace SocketTag {
+
+Http::FilterHeadersStatus SocketTagFilter::decodeHeaders(Http::RequestHeaderMap& request_headers,
+                                                         bool) {
+  auto socket_tag_header = Http::LowerCaseString("socket-tag");
+  if (request_headers.get(socket_tag_header).empty()) {
+    return Http::FilterHeadersStatus::Continue;
+  }
+
+  std::string tag_string(request_headers.get(socket_tag_header)[0]->value().getStringView());
```
alyssawilk, line 19: comments?

library/common/network/socket_tag_socket_option_impl.cc
```diff
@@ -0,0 +1,54 @@
...
+    : optname_(0, 0, "socket_tag"), uid_(uid), traffic_stats_tag_(traffic_stats_tag) {}
+
+bool SocketTagSocketOptionImpl::setOption(
+    Socket& socket, envoy::config::core::v3::SocketOption::SocketState state) const {
+  if (state != envoy::config::core::v3::SocketOption::STATE_PREBIND) {
+    return true;
+  }
+
+  if (!isSupported()) {
+    // ENVOY_LOG(warn, "Failed to set unsupported option on socket");
```
alyssawilk, line 23: ?

test/java/integration/AndroidEnvoyExplicitFlowTest.java
```diff
@@ -323,6 +323,7 @@ public MockResponse dispatch(RecordedRequest recordedRequest) {
                                           .setHttpMethod(RequestMethod.POST)
                                           .setUrl(mockWebServer.url("post/flowers").toString())
                                           .addHeader("content-length", "23")
+                                          .addHeader("socket-tag", "0,0")
```
alyssawilk, line 326: wait is the interface for adding these headers from Java?  I thought that it was something like the cleartext or proxy things where android would inform what a given tag should be?

alyssawilk (commented):
wohoo progress!
high level pass because I feel guilty for prioritizing my FINALLY PASSING factory PR.
let's sort out where the code should live filter-wise with Lyfties since I'm hoping we can remove some of the boilerplate

test/java/integration/AndroidEnvoyExplicitFlowTest.java
RyanTheOptimist: As I understand it, the application specifies the tag/uid and then calls:
https://developer.android.com/reference/android/net/TrafficStats#setThreadStatsUid(int)
https://developer.android.com/reference/android/net/TrafficStats#setThreadStatsTag(int)
So in EM, I would expect the application (at the java layer) to set the headers and then this filter will apply them to the socket. Does that sound right?

library/common/config/config.cc
RyanTheOptimist: Oh, interesting. So instead of a new filter, we'd just add this logic to the existing network_configuration filter? That's a reasonable idea. @Augustyniak, what do you think?

Augustyniak: @RyanTheOptimist  I do not have a strong opinion here but if I had to choose I think that your current approach is the one I would go with.

I think that in the case of my [proxy PR](https://github.com/envoyproxy/envoy-mobile/pull/2416) there is an extra argument to be made for keeping the proxy logic as part of the existing `network_configuration/filter` since my filter proxy logic retrieves the current proxy settings from the `connectivity_manager` to late apply it to the stream (`streamInfo()->filterState()->setData(...)`) and in general our `network_configuration/filter` and `connectivity_manager` seem to be somehow 'paired'. So if I add something to `connectivity_manager` and I need a corresponding filter logic I put that filter logic in `network_configuration/filter` filter.

Since your logic does not depend on the `connectivity_manager` at all I think that I would keep it separate from our connectivity filter but then again - it's not a strong opinion and I could be persuaded that either approach is the better one.

test/java/integration/AndroidEnvoyExplicitFlowTest.java
alyssawilk: is every application supposed to do this, or is it an opt-in sort of thing?  If everyone is expected to do it I'm not sure how we'd get tags if using the C++ APIS, and I think we have to call up at some point to get the ID.  If it's not used by everyone, I wonder if we should make it off by default and add a config knob to turn it on for cronvoy?

RyanTheOptimist: As I understand it, this is to be used when an app is doing networking on behalf of another app, or vice versa. So if an app is not doing that, then it would not need to use this Android API. Cronet exposes an API which makes use of this behavior via: https://chromium.googlesource.com/chromium/src/+/refs/heads/main/components/cronet/android/api/src/org/chromium/net/ExperimentalUrlRequest.java#52

So making this off by default and adding a knob that cronvoy turns on seems fine. That being said, I'm not sure there's much harm in making it always on, but happy to make it off by default. Do you have any pointers to how to make such a knob work?

alyssawilk: I'd be mildly inclined to not have it on by default, or built in by default.  It's already possible to add custom filters - for example I took the "custom" brotli filter out of the cronet tests when I added explicit configuration support for brotli here:
https://github.com/envoyproxy/envoy-mobile/pull/2342/files#diff-3a624bebe9faf2818e3aaeec574223c875803bdbcb6bc2e36c890ce944a3b337L32

I would say we could skip having an explicit knob, but expose default filter config via insert like this:
https://github.com/envoyproxy/envoy-mobile/pull/2342/files#diff-90bc04f1f5d0a531db783375205a81872a6bb7830871d24a01ea1678e6a8793bR161

and then turn it up that way once we do tests.  WDYT?

RyanTheOptimist: Done, I think!

docs/root/intro/version_history.rst
```diff
@@ -37,6 +37,7 @@ Features:
 - api: add option to control whether Envoy should drain connections after a soft DNS refresh completes. (:issue:`#2225 <2225>`, :issue:`#2242 <2242>`)
 - api: add option to disable the gzip decompressor. (:issue: `#2321 <2321>`) (:issue: `#2349 <2349>`)
 - api: add option to enable the brotli decompressor. (:issue `#2342 <2342>`) (:issue: `#2349 <2349>`)
+- api: add option to enable socket tagging. (:issue `#1512 <1521>`)
```
alyssawilk, line 40: Oh you went whole hog and added an explicit API.
I think we should call out what socket tagging is somewhere as well.  I'm not sure what the Envoy-Mobile docs requirements are, but in Envoy we'd usually explain in the proto file and link autogenerated docs.  not sure if E-M does the autogen docs but best case you can grep for a prior example, worst case you can doc up the proto and hard link to it (and Lyfties can correct i that's wrong)

library/common/extensions/filters/http/socket_tag/filter.cc
```diff
@@ -0,0 +1,44 @@
...
+
+  // The socket-tag header must contain a pair of number separated by a comma, e.g.:
+  // socket-tag: 123,456
+  // The first number contains the UID and the second contains the traffic stats tag.
+  std::string tag_string(request_headers.get(socket_tag_header)[0]->value().getStringView());
+  std::pair<std::string, std::string> data = absl::StrSplit(tag_string, ',');
+  uid_t uid;
+  uint32_t traffic_stats_tag;
+  if (!absl::SimpleAtoi(data.first, &uid) || !absl::SimpleAtoi(data.second, &traffic_stats_tag)) {
+    return Http::FilterHeadersStatus::StopIteration;
```
alyssawilk, line 27: StopIteration or sendLocalReply with error?  I'd think we wanted the latter?  do we have an e2e test for this?

library/common/extensions/filters/http/socket_tag/filter.cc
```diff
@@ -0,0 +1,44 @@
...
+
+  auto options = std::make_shared<Network::Socket::Options>();
+  options->push_back(std::make_shared<Network::SocketTagSocketOptionImpl>(uid, traffic_stats_tag));
+  callbacks_->addUpstreamSocketOptions(options);
+  request_headers.remove(socket_tag_header);
+  return Http::FilterHeadersStatus::Continue;
+}
+
+void SocketTagFilter::setDecoderFilterCallbacks(Http::StreamDecoderFilterCallbacks& callbacks) {
+  callbacks_ = &callbacks;
```
alyssawilk, line 38: AFIK base class saves as decoder_callbacks_

library/common/extensions/filters/http/socket_tag/filter.h
```diff
@@ -0,0 +1,33 @@
...
+#include "library/common/extensions/filters/http/socket_tag/filter.pb.h"
+#include "library/common/types/c_types.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace SocketTag {
+
+/**
+ * Filter to set upstream socket options based on network conditions.
```
alyssawilk, line 17: based on network conditions or application supplied headers?

library/common/extensions/filters/http/socket_tag/filter.proto
```diff
@@ -0,0 +1,6 @@
+syntax = "proto3";
+
+package envoymobile.extensions.filters.http.socket_tag;
+
+message SocketTag {
```
alyssawilk, line 5: let's comment somewhere (here?) about how socket tagging works and maybe why you want it (the bits you kindly already explained to me, but in writing =P)

library/common/network/socket_tag_socket_option_impl.cc
```diff
@@ -0,0 +1,49 @@
...
+SocketTagSocketOptionImpl::SocketTagSocketOptionImpl(uid_t uid, uint32_t traffic_stats_tag)
+    : optname_(0, 0, "socket_tag"), uid_(uid), traffic_stats_tag_(traffic_stats_tag) {}
+
+bool SocketTagSocketOptionImpl::setOption(
+    Socket& socket, envoy::config::core::v3::SocketOption::SocketState state) const {
+  if (state != envoy::config::core::v3::SocketOption::STATE_PREBIND) {
+    return true;
+  }
+
+  int fd = socket.ioHandle().fdDoNotUse();
```
alyssawilk, line 22: Can we comment here why this is OK for multiplexed protocols (where many streams may share the connection)
Also API comments on any limitations (not sure if it'll crash or fail if used on internal socket, or any other without fd)

library/common/network/socket_tag_socket_option_impl.cc
```diff
@@ -0,0 +1,49 @@
...
+absl::optional<Socket::Option::Details> SocketTagSocketOptionImpl::getOptionDetails(
+    const Socket&, envoy::config::core::v3::SocketOption::SocketState /*state*/) const {
+  if (!isSupported()) {
+    return absl::nullopt;
+  }
+
+  static std::string name = "socket_tag";
+  Socket::Option::Details details;
+  details.name_ = optname_;
+  //  details.value_ = tag_->dataForLogging();
```
alyssawilk, line 42: ?

library/common/network/socket_tag_socket_option_impl.h
```diff
@@ -0,0 +1,35 @@
...
+  absl::optional<Details>
+  getOptionDetails(const Network::Socket& socket,
+                   envoy::config::core::v3::SocketOption::SocketState state) const override;
+  bool isSupported() const override;
+
+private:
+  const Network::SocketOptionName optname_;
+
+  uid_t uid_;
+  uint32_t traffic_stats_tag_;
```
alyssawilk, line 31: comments

library/java/io/envoyproxy/envoymobile/engine/JniLibrary.java
```diff
@@ -315,6 +315,11 @@ protected static native int recordHistogramValue(long engine, String elements, b
    */
   public static native String brotliConfigInsert();
 
+  /**
+   * Provides a configuration insert that may be used to enable socket tagging.
+   */
+  public static native String socketTagConfigInsert();
```
alyssawilk, line 321: so you know I'd envisioned this being done as brotli was _before_ my brotli PR, and having cronet just add this as a custom filter.  I'm 100% with doing the extras but it comes with a higher API/docs burden
though really it's worth doccing anyway.  :-)

library/java/org/chromium/net/AndroidNetworkLibrary.java
```diff
@@ -83,4 +103,153 @@ public static void clearTestRootCertificates()
...
+    // Have ParcelFileDescriptor relinquish ownership of the file descriptor.
+    if (pfd != null) {
+      pfd.detachFd();
+    }
+
+    // Restore prior thread tags.
+    if (tag != oldTag) {
+      TrafficStats.setThreadStatsTag(oldTag);
+    }
+    //    if (uid != TrafficStatsUid.UNSET) {
```
alyssawilk, line 250: ?

alyssawilk (commented):
Sweet, very cool.  Next pass comments :-)

library/common/network/socket_tag_socket_option_impl.cc
RyanTheOptimist: Added a comment about multiplexed protocols. Thanks!

Can you say more about "internal sockets" or sockets without FDs?

library/common/network/socket_tag_socket_option_impl.cc
RyanTheOptimist: Whoops. Fixed.

library/common/network/socket_tag_socket_option_impl.h
RyanTheOptimist: Done

library/java/io/envoyproxy/envoymobile/engine/JniLibrary.java
RyanTheOptimist: OH! Shoot! I misunderstood. Hm. I'm happy to un-do part of this if that would be better?

library/java/org/chromium/net/AndroidNetworkLibrary.java
RyanTheOptimist: Ah, removed. Thanks. (In cronet that enum comes from some C code that we're not building in Envoy. )

docs/root/intro/version_history.rst
RyanTheOptimist: Sorry! As you uncovered below, I misunderstood the PR I was copying from. I'm happy to make this *not* part of the API if that would be better. In any case, I've added some comments to the proto to attempt to document this API.

But very happy to do something simpler!

RyanTheOptimist (commented):
Thanks for the great review! Sorry for misunderstanding the API.

docs/root/intro/version_history.rst
alyssawilk: cool.  and either repo link to the proto or reflink the docs here?

library/common/extensions/filters/http/socket_tag/filter.cc
alyssawilk: Oh but this is a failure case based on getting bad headers from the API - you should totally be able to test sending an invalid request with the filter configured, and getting the local response you expect, so let's do that!

you could also optionally add a test for the happy path, which would create the custom transport socket and make sure hashing works, if we don't have that tested upstream.  even if the tagging is a no-op it would at least cover that the memory model works.  

also clean up the commented out return value above this line?

library/common/network/socket_tag_socket_option_impl.cc
alyssawilk: so it's not uncommon Envoy users want to
1) decap connect
2) L7 process decapsulated HTTP (imagine a full HTTP/2 connection in here)
3) re-encap with connect
If either the inner or outer stream can be HTTP/2, you accomplish this via 3 passes through Envoy: one L7 pass which decapsulated, and forwards the inner payload to a loopback listener which does L7 processing, and forwards raw (maybe HTTP/2) bytes to another loopback listener which encapsulates.  The kernel context switches get cumbersome and keeping track of things like original IP get fraught, so three's an "internal listener" which lets you bypass the kernel, and basically do it all through userspace instead.  in this case there's no fd, and I'm not sure if the fd() call returns  -1 or ENVOY_BUG but tl;dr it's worth commenting that it only works if we can access a raw fd.

library/java/io/envoyproxy/envoymobile/engine/JniLibrary.java
alyssawilk: nah, fine to leave in unless Lyft folks object.

alyssawilk (commented):
Looking good!  FWIW I really only glanced over the Java as I'm not a useful reviewer for that.  tests (when we have the infra) will verify if it's all good.

docs/root/intro/version_history.rst
RyanTheOptimist: Hm. This sounds like a good idea, but I don't quite see how to do that based on existing entries in this file. Do you have an example I can start with?

library/common/extensions/filters/http/socket_tag/filter.cc
RyanTheOptimist: Ah, ok. Java tests added and removed the commented out code.

I looked at  test/common/integration/client_integration_test.cc, which we discussed offline. I wasn't quite sure how to configure it to add the socket tagging support. Maybe we can discuss that offine?

library/common/network/socket_tag_socket_option_impl.cc
RyanTheOptimist: Ah, I see. Done.

docs/root/intro/version_history.rst
alyssawilk: I don't know if there are exampes.  @jpsim thoughts here?

test/java/integration/AndroidEngineSocketTagTest.java
```diff
@@ -0,0 +1,392 @@
...
+  private final MockWebServer mockWebServer = new MockWebServer();
+  private Engine engine;
+
+  @BeforeClass
+  public static void loadJniLibrary() {
+    AndroidJniLibrary.loadTestLibrary();
+  }
+
+  @Before
+  public void setUpEngine() throws Exception {
```
alyssawilk, line 60: wow this seems like a lot of boilerplate to send a request and check for an error response. Maybe @jpsim or @goaway can comment if there's any low hanging fruit for clean up, or add "refactor for reuse" to the ever growing list of things we should do?

test/java/integration/AndroidEngineSocketTagTest.java
```diff
@@ -0,0 +1,392 @@
...
+              latch.countDown();
+              return null;
+            })
+            .setOnComplete((finalStreamIntel) -> {
+              response.get().addStreamIntel(finalStreamIntel);
+              response.get().setFinalStreamIntel(finalStreamIntel);
+              latch.countDown();
+              return null;
+            })
+            .setExplicitFlowControl(true)
```
alyssawilk, line 201: just thinking (and skip if it takes too long) I wonder if you kill off explicit flow control you can remove the window update functions.  and if you're not using intel you can probably remove most of these callback functions.  and if not just ignore me.

alyssawilk (commented):
Ok I think I'm hitting the end of my useful review comments, so let's see if we can get a Lyft volunteer to take it from  here :-)
LGTM modulo the few remaining comments

docs/root/intro/version_history.rst
Augustyniak: I think that `[starting_envoy.rst](https://github.com/envoyproxy/envoy-mobile/blob/main/docs/root/api/starting_envoy.rst)` file may be a good home for what you are planning to add.

test/java/integration/AndroidEngineSocketTagTest.java
RyanTheOptimist: I like the way you think! It looks like the `Response` and `RequestScenario` classes are used in a few different tests. How 'bout this:

In parallel with this PR (so as to avoid blocking it) I refactor those tests to move `Response` and `RequestScenario` into stand-alone helper classes, and depending on which PR lands first, make sure that after both land, the tests in this file use the new helper?

RyanTheOptimist:
@Augustyniak If you have time, I'd appreciate your full review now that I think this is sorted out.

alyssawilk (dismissed):

Augustyniak:
>@Augustyniak If you have time, I'd appreciate your full review now that I think this is sorted out.

Sure. Sorry for a delays response - was OOO on Friday due to an org-wide recharge day.

library/common/extensions/filters/http/socket_tag/filter.cc
```diff
@@ -0,0 +1,42 @@
...
+#include "library/common/network/socket_tag_socket_option_impl.h"
+
+namespace Envoy {
+namespace Extensions {
+namespace HttpFilters {
+namespace SocketTag {
+
+Http::FilterHeadersStatus SocketTagFilter::decodeHeaders(Http::RequestHeaderMap& request_headers,
+                                                         bool) {
+  auto socket_tag_header = Http::LowerCaseString("socket-tag");
```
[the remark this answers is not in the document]
Augustyniak: not sure about this anymore as the header is set by the user of the library so maybe _not_ having a prefix here is just OK.

test/java/integration/AndroidEngineSocketTagTest.java
Augustyniak: https://github.com/envoyproxy/envoy-mobile/pull/2444 landed so you should be able to benefit from it in your PR.

Augustyniak (commented):
Looks good in general. Left a few comments/questions.

Took only a brief look at tests as one of the comments suggest that they will change significantly (will become simpler) now that https://github.com/envoyproxy/envoy-mobile/pull/2444 has been merged. Will re-review after they are updated.

library/cc/engine_builder.h
```diff
@@ -37,6 +37,7 @@ class EngineBuilder {
   EngineBuilder& setStreamIdleTimeoutSeconds(int stream_idle_timeout_seconds);
   EngineBuilder& enableGzip(bool gzip_on);
   EngineBuilder& enableBrotli(bool brotli_on);
+  EngineBuilder& enableSocketTagging(bool socket_tagging_on);
```
Augustyniak, line 40: Should this be `enableSocketTag` to be consistent with naming in Kotlin/Java APIs? Or we can call it `enableSocketTagging` everywhere - my point is that it should probably be the same name everywhere :)

RyanTheOptimist (commented):
Thanks for the review! Github didn't seem to properly sign-off the commits it landed so I ended up having to force merge :/

library/common/extensions/filters/http/socket_tag/filter.cc
Augustyniak: Thank you. I am happy to help with the setup of the demo app if you run into any issues with it. It's also possible that the documentation is lacking or making some assumptions - let me know and I am happy to help.

Re experimental app - I think that you can go even one step further (sorry for not pointing this out in my previous comment) and adding a `socket-tag` header to requests performed by the experimental app. This way you would get an end-to-end confirmation that requests with socket-tag go through (we verify whether our experimental example is getting 200 from the server). You could add the header somewhere in [here](https://github.com/envoyproxy/envoy-mobile/pull/2423/files#diff-b15ec59c29644d8c6089ad0ea7370859a690563f75d02c819679270b61ab9520L116)

docs/root/intro/version_history.rst
Augustyniak: @RyanTheOptimist did you have a chance to look at the file I referenced above?

library/common/extensions/filters/http/socket_tag/filter.cc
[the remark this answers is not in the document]
Augustyniak: Thank you for a detailed response - the link to RFC is helpful. Looking at our codebase I can see that we use `x-envoy-mobile` prefix specifically and headers that follow this convention have special guards in place - i.e., the application cannot just set them using normal "addHeader" method as they are considered to be "restricted":

https://github.com/envoyproxy/envoy-mobile/blob/308f15cee1a132ab1341476ff5f553065bb416d8/library/kotlin/io/envoyproxy/envoymobile/HeadersBuilder.kt#L79-L81

Instead, the app needs to use strongly typed methods on `RequestHeaderBuilder` to add such headers:

https://github.com/envoyproxy/envoy-mobile/blob/308f15cee1a132ab1341476ff5f553065bb416d8/library/kotlin/io/envoyproxy/envoymobile/RequestHeadersBuilder.kt#L88-L95

I think that I should've commented with this earlier but if we want to follow an established pattern here we should add `addSocketTag` method to `RequestHeadersBuilder` and make this method add `x-envoy-mobile-socket-tag` header to the outgoing request (following the logic from the `addUpstreamHttpProtocol` method referenced above. 

I do not have a strong opinion about the existing pattern related to `x-envoy-mobile` prefix in headers but I think that there is some value in following it. On top of this, the existence of `addSocketTag` method in `RequestHeadersBuilder` class would help with the discoverability of the feature and act as a natural place for its documentation. I think that some concerns related to discoverability and lack of a place to document the feature were raised in https://github.com/envoyproxy/envoy-mobile/pull/2423#discussion_r929175850 and the proposed change could address both of them (the addition of `addSocketTag` method specifically).

Augustyniak (commented):
Thank you for addressing/responding to comments. Tests looks so much cleaner/simpler now.  💯

Left a few more comments

library/common/extensions/filters/http/socket_tag/filter.cc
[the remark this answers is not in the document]
Augustyniak: also - can we avoid re-creating this header instance (`Http::LowerCaseString("socket-tag")`) on each request?

library/common/extensions/filters/http/socket_tag/filter.cc
```diff
@@ -0,0 +1,43 @@
...
+  if (request_headers.get(socket_tag_header).empty()) {
+    return Http::FilterHeadersStatus::Continue;
+  }
+
+  // The socket-tag header must contain a pair of number separated by a comma, e.g.:
+  // socket-tag: 123,456
+  // The first number contains the UID and the second contains the traffic stats tag.
+  std::string tag_string(request_headers.get(socket_tag_header)[0]->value().getStringView());
```
Augustyniak, line 22: can we reorganize the code to avoid having to call `equest_headers.get(socket_tag_header)` twice?

RyanTheOptimist:
I'm *really* confused by GitHub. I see an email from 2:04 PM Pacific today (2022-08-11) with a few comments that were sent after I uploaded a few new commits. But github says the commits were yesterday, and some of my comments are missing from the conversation view, or at least hidden. And some seem completely lost. I'm not sure what happened.

In any case, @Augustyniak I tried to explain that I updated the sample app to set socket tags, but end up with a java class not found error for the class  org.chromium.net.AndroidNetworkLibrary which implements the tagSocket() method. I'm not quite sure how to resolve this. Any advice?

docs/root/api/starting_envoy.rst
```diff
@@ -426,6 +426,17 @@ Specify whether to enable transparent response Brotli decompression. Defaults to
   // Swift
   builder.enableBrotli(true)
 
+~~~~~~~~~~~~~~~~~~~~~~~
+``enableSocketTagging``
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Specify whether to enable support for Android socket tagging. Defaults to false.
```
Augustyniak, line 433:
```suggestion
Specify whether to enable support for Android socket tagging. Unavailable on iOS. Defaults to false.
```

Augustyniak:
@RyanTheOptimist I modified the example app locally to enable socket tagging and run it to see whether I can repro the crashes that you mentioned locally. This is what I got:

Screen Shot 2022-08-17 at 10 56 24 PM


That's despite your BUILD file changes in here https://github.com/envoyproxy/envoy-mobile/pull/2423/files#diff-bee4aeb6cd03b6684900f7c2d8bd51dec467011c33929eb3b34008a2a44deb23R19. This issue seems to be exactly the same issue as the one that's related `is_cleartext_permitted` and which was detailed in https://github.com/envoyproxy/envoy-mobile/issues/2432.

library/common/jni/android_jni_utility.cc
```diff
@@ -33,3 +33,17 @@ bool is_cleartext_permitted(absl::string_view hostname) {
   return true;
 #endif
 }
+
+void tag_socket(int ifd, int uid, int tag) {
+#if defined(__ANDROID_API__)
+  JNIEnv* env = get_env();
+  jclass jcls_AndroidNetworkLibrary = env->FindClass("org/chromium/net/AndroidNetworkLibrary");
```
Augustyniak, line 40: The one thing that you need to do here is to start using the `find_class` that was introduced in https://github.com/envoyproxy/envoy-mobile/pull/2483

```suggestion
  jclass jcls_AndroidNetworkLibrary = find_class("org.chromium.net.AndroidNetworkLibrary");
```

docs/root/api/starting_envoy.rst
Augustyniak: Can we add this note to make it clear that it's no intended for iOS?

test/java/integration/AndroidEngineSocketTagTest.java
```diff
@@ -0,0 +1,188 @@
...
+        assertThat(recordedRequest.getMethod()).isEqualTo(RequestMethod.GET.name());
+        assertThat(recordedRequest.getHeader("x-envoy-mobile-socket-tag")).isEqualTo(null);
+        return new MockResponse().setBody("This is my response Body");
+      }
+    });
+    mockWebServer.start();
+    RequestScenario requestScenario = new RequestScenario()
+                                          .setHttpMethod(RequestMethod.GET)
+                                          .setUrl(mockWebServer.url("post/flowers").toString())
+                                          .addHeader("x-envoy-mobile-socket-tag", "0,0");
```
Augustyniak, line 96: `RequestScenario` becomes a little bit of a leaky abstraction as in real application we are not supposed to call `addHeader("x-envoy-mobile-socket-tag", "0,0");` directly (since we added addSocketTag). 

In fact, I am not sure whether this test works the way it's intended too as it tries to add `x-envoy-mobile-socket-tag` header in https://github.com/envoyproxy/envoy-mobile/blob/0d7b43e30024980cb9cd438aecf93c56c3707e46/test/java/io/envoyproxy/envoymobile/engine/testing/RequestScenario.java#L53 and `add(...)` call does not accept headers that start with `x-envoy-mobile` prefix https://github.com/envoyproxy/envoy-mobile/blob/0d7b43e30024980cb9cd438aecf93c56c3707e46/library/kotlin/io/envoyproxy/envoymobile/HeadersBuilder.kt#L80

Maybe we could

test/java/integration/AndroidEngineSocketTagTest.java
```diff
@@ -0,0 +1,188 @@
...
+    mockWebServer.shutdown();
+  }
+
+  @Test
+  public void socket_tag() throws Exception {
+    mockWebServer.setDispatcher(new Dispatcher() {
+      @Override
+      public MockResponse dispatch(RecordedRequest recordedRequest) {
+        assertThat(recordedRequest.getMethod()).isEqualTo(RequestMethod.GET.name());
+        assertThat(recordedRequest.getHeader("x-envoy-mobile-socket-tag")).isEqualTo(null);
```
Augustyniak, line 88: why is it expected to be equal to `null` if we set a valid socket tag on the request?

Augustyniak (commented):
NICE! Great to see it working - this is awesome. 

Left a few comments related to tests - apologies for not catching whatever I commented on earlier but I was focused on the functionality more.

General comment: Can we update the description of the PR, please? In particular, for `testing` can we mention that the functionality is verified as part of the run of the experimentation example app?

test/java/integration/AndroidEngineSocketTagTest.java
Augustyniak: ah, because we remove it when we detect that it's present

RyanTheOptimist: Yup, exactly!

test/java/integration/AndroidEngineSocketTagTest.java
RyanTheOptimist: Blergh! Yes, good point. Ok, I've added an addSocketTag() method to RequestScenario which delegates to the underlying builder. I verified (via printf) that the socket tagging code is in fact called.

Augustyniak: great, thank you - I think that adding `addSocketTag` method to the test scenario is a way to go 👍 . That being said, I think that the right final solution in here is to change the underlying implementation of `TestScenario` so that it stores test state inside of `RequestHeadersBuilder` as opposed to on-the-side headers map. Only with `RequestHeadersBuilder` we can ensure that our tests simulate real-life condition.

I do not think that this has to be done as part of your PR so I am fine with it being merged the way it is now.

test/java/integration/AndroidEngineSocketTagTest.java
Augustyniak: Any way we could ensure that we are really adding `x-envoy-mobile-socket-tag` at some point? As it is now this test is always green - even if socket tagging is not enabled / does not end up adding a tag.

docs/root/api/starting_envoy.rst
Augustyniak: I do not want to be annoying but i still do not see it being updated.

Augustyniak (commented):
Left 3 more comments related to your responses/updates from the previous iterations of my review. Happy to approve once you respond to my comments.

test/java/integration/AndroidEngineSocketTagTest.java
RyanTheOptimist: Filed https://github.com/envoyproxy/envoy-mobile/issues/2494 as a result of offline discussions. Thanks for working through this!

test/java/integration/AndroidEngineSocketTagTest.java
Augustyniak: >Or is it not the socket tagging stuff that is problematic, but rather the general header stuff?

Yes, that's what I meant

>Are you thinking we should get rid of the headers member in RequestScenario and use the RequestHeadersBuilder directly?

That was my idea. This way our tests would be guaranteed to "get an update" every time we update `RequestHeadersBuilder`.

Augustyniak (approved):
Thanks for working on this. Really excited to see this landing 🎉  Congratz - great job.

PR #2424, opened 2022-07-15
tools: add arm64 debug run configuration

Augustyniak:
Description: Add an arm64 run configuration for Envoy example app so that it's possible to run and debug Envoy Mobile example app on ARM machines (i.e., M1 Macbooks). Following steps from https://envoy-mobile.github.io/docs/envoy-mobile/latest/development/debugging/android_local.html#adding-envoy-mobile-project-into-android-studio it's possible to set up an Android project that makes it possible to run and debug Envoy Mobile example app (put symbolic breakpoints in Envoy code). As it is now, the configured/added run configuration supports running of x86 builds only which do not work on ARM machines. 
Risk Level: None
Testing: Run new configuration locally on M1 Macbook.
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/examples/kotlin/hello_world/tools/android-studio-run-configurations/run_configuration_example_debug_arm64.xml
+++ b/examples/kotlin/hello_world/tools/android-studio-run-configurations/run_configuration_example_debug_arm64.xml
@@ -1,4 +1,4 @@
-<configuration name="Example App (Debug) [x86]"
+<configuration name="Example App (Debug) [arm64]"
     type="BlazeCommandRunConfigurationType"
     factoryName="Bazel Command"
     nameIsGenerated="false">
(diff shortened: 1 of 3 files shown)


examples/kotlin/hello_world/tools/android-studio-run-configurations/run_configuration_example_debug_arm64.xml
```diff
@@ -0,0 +1,38 @@
+<configuration name="Example App (Debug) [arm64]"
```
Augustyniak, line 1: this configuration is basically a copy-paste of the existing x86 run configuration

Augustyniak:
> Can you also update the docs to mention picking the architecture to match your host machine and reflect the x86 rename?
> 
> > From Android Studio select the Example App x86 (Debug) configuration

This part of the documentation is not up to date in general (as we removed `dist/` directory in https://github.com/envoyproxy/envoy-mobile/pull/2184) and I was planning to revamp it in one of the follow up PRs. Can start doing it in this PR too tho.

There are other things I will need to follow up on such as `- While breaking on a C++ function, Android Studio should present the source file and highlight the line where the breakpoint hit with all scope information` which I cannot make to work for me.

Augustyniak:
Pushed a documentation update in [684da7b](https://github.com/envoyproxy/envoy-mobile/pull/2424/commits/684da7b798376a93b9c87ccd8e4b4b22ed808105)

jpsim (approved):

PR #2429, opened 2022-07-18
tools: fix source mapping

Augustyniak:
Description: Per our documentation, the following should be true "While breaking on a C++ function, Android Studio should present the source file and highlight the line where the breakpoint hit with all scope information". That feature did not work - I didn't confirm this but it's possible that the functionality was broken when https://github.com/envoyproxy/envoy-mobile/pull/2184 was merged. Anyway, the issue was that https://github.com/envoyproxy/envoy-mobile/blob/581ba40f2f8597243efccb2bdccade8193ff4559/.bazelrc#L44 config setting was not applied when Envoy Mobile was built for the runs of example apps. This has been fixed by making run configurations that are used by Envoy Mobile example apps specify `--config=dbg` setting explicitly.
Risk Level: None
Testing: Confirmed that Android Studio opens the right file (and highlights the right line) when EnvoyMobile breakpoint is hit.
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -40,6 +40,7 @@ build --java_language_version=8
 # Override PGV validation with NOP functions
 build --@com_envoyproxy_protoc_gen_validate//bazel:template-flavor=nop
 
+build:dbg --compilation_mode=debug
 # Enable source map for debugging in IDEs
 build:dbg --copt="-fdebug-compilation-dir" --copt="/proc/self/cwd"
 
--- a/.bazelrc
+++ b/.bazelrc
@@ -40,7 +40,7 @@ build --java_language_version=8
 # Override PGV validation with NOP functions
 build --@com_envoyproxy_protoc_gen_validate//bazel:template-flavor=nop
 
-build:dbg --compilation_mode=debug
+build:dbg --compilation_mode=dbg
 # Enable source map for debugging in IDEs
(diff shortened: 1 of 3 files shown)


.bazelrc
```diff
@@ -40,6 +40,7 @@ build --java_language_version=8
 # Override PGV validation with NOP functions
 build --@com_envoyproxy_protoc_gen_validate//bazel:template-flavor=nop
 
+build:dbg --compilation_mode=debug
```
Reflejo, line 43: isn't it `--compilation_mode=dbg`?

Augustyniak: (I think that for whatever reason both work since the change from debug -> dbg did not result in a rebuild of the project)

jpsim (approved):

.bazelrc
Augustyniak: Mhm... looking into this.

Reflejo: Yea that's because android studio's bazel plugin is adding `-c dbg` when you hit the lil debug icon and is override your compilation mode, so I presume you can set any garbage and it won't fail because it'll get overriden

Augustyniak: ah, got it - makes sense. Thank you for the context.

In that case, I am going to keep `--compilation_mode=debug` for cases when we want to use `--config=dbg` from outside of the debugging flow in Android Studio UI.

PR #2430, opened 2022-07-19
cc: add support for registering a platform KV store

goaway:


library/cc/key_value_store.cc
```diff
@@ -0,0 +1,46 @@
+#include "library/cc/key_value_store.h"
+
+#include "library/common/data/utility.h"
+
+namespace Envoy {
+namespace Platform {
+
+namespace {
+
+envoy_data c_kv_store_read(envoy_data key, const void* context) {
```
Augustyniak, line 10: did you add `c_` prefix to avoid name conflicts? Can we document this?

library/cc/key_value_store.cc
```diff
@@ -0,0 +1,46 @@
...
+}
+
+} // namespace
+
+envoy_kv_store KeyValueStore::asEnvoyKeyValueStore() {
+  return envoy_kv_store{
+      &c_kv_store_read,
+      &c_kv_store_save,
+      &c_kv_store_remove,
+      new KeyValueStoreSharedPtr(shared_from_this()),
```
Augustyniak, line 41: So I needed to read the documentation of `enable_shared_from_this` to understand better what's happening here.  The documentation that I found https://en.cppreference.com/w/cpp/memory/enable_shared_from_this.

I am probably missing something but here is my question:
The documentation says: `It is permitted to call shared_from_this only on a previously shared object, i.e. on an object managed by std::shared_ptr<T>.`

When I look at the code from this PR I can see that we register a new KeyValueStore in [here](https://github.com/envoyproxy/envoy-mobile/pull/2430/files#diff-204f30e7778ed8d1f3e6008076942d0cd78f18df61826d44ef7f2d54b2b9d2fbR216) and that this code calls into [here](https://github.com/envoyproxy/envoy-mobile/pull/2430/files#diff-cbbc688841d6fdf3e2e1de0f4b000970b68527576c8928c653205fc2149cfe72R41) but I do not think that at this point our object is managed by any `shared_ptr` which documentation says is required for `shared_from_this` call to be safe. What am I missing here?

Augustyniak (commented):
Did the first pass. Have a question related to the use of `enable_shared_from_this`

library/cc/key_value_store.cc
Augustyniak: Discussed offline - the `KeyValueStorer` managed with `shared_ptr` is coming from a user of the library and it's passed in [here](https://github.com/envoyproxy/envoy-mobile/pull/2430/files#diff-204f30e7778ed8d1f3e6008076942d0cd78f18df61826d44ef7f2d54b2b9d2fbR87)

Augustyniak:
at the high level I do not see any issues with this PR other than a lack of an integration test. We talked offline and you mentioned that you are planning to address this issue in a follow up PR - please either add a test in this PR _or_ create a GH issue for adding an integration test.

library/cc/key_value_store.cc
goaway: There will be no naming conflicts, due to the enclosing anonymous namespace. That said, this is our current convention for all platform implementations. For iOS this is called `ios_kv_store_read`, for Android, this is called `jni_kv_store_read`, and the `c_` prefix is already used elsewhere for the `cc` platform (arguably the prefix could/should be `cc_`, but it doesn't seem worth breaking the convention in this PR).  As for where to document this - I'm not sure? It's perhaps worth having a general documentation page on how platform callbacks work, and that's something I can start on.

goaway:
https://github.com/envoyproxy/envoy-mobile/issues/2451

library/cc/key_value_store.h
```diff
@@ -0,0 +1,51 @@
...
+#include "envoy/common/pure.h"
+
+#include "absl/types/optional.h"
+#include "library/common/extensions/key_value/platform/c_types.h"
+
+namespace Envoy {
+namespace Platform {
+
+/**
+ *`KeyValueStore` is an interface that may be implemented to provide access to an arbitrary
```
jpsim, line 14: Nit, missing a space here:

```suggestion
 * `KeyValueStore` is an interface that may be implemented to provide access to an arbitrary
```

docs/root/intro/version_history.rst
```diff
@@ -32,7 +32,7 @@ Bugfixes:
 
 Features:
 
-- Android & iOS: add support for registering a platform KV store (:issue: `#2134 <2134>`) (:issue: `#2335 <2335>`)
+- Android, iOS, & C++: add support for registering a platform KV store (:issue: `#2134 <2134>`, :issue: `#2335 <2335>`, :issue: `#2430 <2430>`)
```
jpsim, line 35: Should we get into a habit of referring to the various APIs by language instead of a mix of platform and language?

`Android, iOS, & C++` -> `Kotlin, Swift & C++ `

jpsim (dismissed):

docs/root/intro/version_history.rst
goaway: No strong feelings. This labeling is purely historical, and we've just stuck with it to be consistent. To your point, the library itself is organized by language. The only point I can think of in favor of the platform names is that it perhaps highlights that the library is intended for use in these environments to the unfamiliar user. But the README can do that, too.

I'm good either way.

library/cc/key_value_store.h
goaway: Thanks, good catch.

jpsim (approved):

PR #2436, opened 2022-07-20
tests: make envoy_mobile_android_test macro work with kt files

Augustyniak:
Description: Add support for Kotlin robolectric tests by making `envoy_mobile_android_test` macro use `kt_android_local_test` maco as opposed to `android_local_test` macro as the former supports both Java and Kotlin files while the latter supports Java files only. The change should be additive when it comes to functionality. Bump `rules_kotlin` version to v1.7.0-RC-3 to pick up changes from https://github.com/bazelbuild/rules_kotlin/pull/810.
Risk Level: None
Testing: Existing tests using envoy_mobile_android_test macro.
Docs Changes: N/A
Release Notes: N/A


Bencodes (approved):

Augustyniak:
@Bencodes any idea as to why our java mac tests are failing after a move from `android_local_test` to `kt_android_local_test`? I thought that the transition from `android_local_test` to `kt_android_local_test` was supposed to be effortless but maybe we are running into some kind of a (known?) corner case. The logs for failing tests are available at https://github.com/envoyproxy/envoy-mobile/runs/7429257263?check_suite_focus=true

Augustyniak:
So I think that with https://github.com/envoyproxy/envoy-mobile/pull/2436/commits/54ecace49b7b67babe948f4322a71ba6dfb99e29 I was finally able to figure out what the issue with failing `android_tests / java_tests_mac (pull_request)` CI Job is. We currently pass `exec_properties` parameter to `android_local_test` macro and that parameter is not supported by `kt_android_local_test` which simply ignores it.

Working on adding support for `exec_properties` to `kt_android_local_test` in https://github.com/bazelbuild/rules_kotlin/pull/810.

jpsim (approved):

PR #2441, opened 2022-07-27
ios: create persistent UserDefaults-based KV store

goaway:
Description: Parallels Android implementation based on SharedPreferences.
Risk Level: Low
Testing: Application


library/swift/BUILD
```diff
@@ -46,6 +46,7 @@ swift_library(
         "UpstreamHttpProtocol.swift",
         "filters/*.swift",
         "grpc/*.swift",
+        "ios/*.swift",
```
jpsim, line 49: Can you move this to some other directory since UserDefaults aren't iOS-specific? They're available on all apple platforms.

Moving to the bare `library/swift/` directory would be fine.

library/swift/ios/UserDefaultsStore.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+
+/// Simple implementation of a `KeyValueStore` leveraging `UserDefaults` for persistence.
+@objcMembers
+public class UserDefaultsStore: KeyValueStore {
```
jpsim, line 5: There's no need for `@objcMembers` and since this class isn't useful to subclass, we should mark it as `final`:

```suggestion
public final class UserDefaultsStore: KeyValueStore {
```

library/swift/ios/UserDefaultsStore.swift
```diff
@@ -0,0 +1,23 @@
...
+  private let defaults: UserDefaults
+
+  public init(userDefaults: UserDefaults) {
+    self.defaults = userDefaults
+  }
+
+  public func readValue(forKey key: String) -> String? {
+    return defaults.string(forKey: key)
```
Augustyniak, line 13: I think that in general - even though it's not codified anywhere - we used to write `self.` when referring to instance variables. Not a blocker but consider adding `self.` for consistency.

test/swift/apps/experimental/ViewController.swift
```diff
@@ -17,6 +18,9 @@ final class ViewController: UITableViewController {
   override func viewDidLoad() {
     super.viewDidLoad()
 
+    // swiftlint:disable:next force_unwrapping
+    let userDefaults = UserDefaults(suiteName: kPersistenceKey)!
```
Augustyniak, line 22: can we just return if created `UserDefaults` is null as opposed to force unwrapping?

Augustyniak (commented):
Pls consider comments

test/swift/apps/experimental/ViewController.swift
jpsim: For this demo we should use the standard user defaults, which is what most users would want to reach for first anyway:

```suggestion
    let userDefaults = UserDefaults.standard
```

Augustyniak: Yes, good point

library/swift/BUILD
goaway: It's not part of OSS swift though right? Is that a distinction we should consider maintaining?

library/swift/ios/UserDefaultsStore.swift
goaway: Isn't @objcMembers necessary to call this from Objective-C code?

library/swift/BUILD
Augustyniak: I personally think that we should not take into account whether a given Swift API is OSS or not when deciding as to how we structure our code. @jpsim 's suggestion makes sense to me.

Augustyniak:
let's update version_history file with the user facing changes introduced in this PR.

library/swift/ios/UserDefaultsStore.swift
Augustyniak: yes, you need `@objcMembers` at the class level or `@objc` at method level if you want this class and its methods to be usable from objc.

Augustyniak (approved):

Augustyniak:
@goaway Android CI issues should've been fixed in https://github.com/envoyproxy/envoy-mobile/pull/2448. Merging master should help make this PR green.

library/swift/ios/UserDefaultsStore.swift
jpsim: > Isn't @objcMembers necessary to call this from Objective-C code?

No, this just stores properties (i.e. members) in the ObjC runtime, but we don't need to access the stored properties of this class at all.

library/swift/UserDefaultsStore.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+
+/// Simple implementation of a `KeyValueStore` leveraging `UserDefaults` for persistence.
+@objcMembers
+public final class UserDefaultsStore: KeyValueStore {
```
jpsim, line 5: Why not remove this type altogether and make `UserDefaults` conform directly to `KeyValueStore`?

```swift
extension UserDefaults: KeyValueStore {
  public func readValue(forKey key: String) -> String? {
    self.string(forKey: key)
  }

  public func saveValue(_ value: String, toKey key: String) {
    self.set(value, forKey: key)
  }

  public func removeKey(_ key: String) {
    self.removeObject(forKey: key)
  }  
}
```

goaway: That's pretty clever, and I'm not opposed to that approach. :) If you'd like to push a PR that does it, I'd be open to closing this in favor it, though otherwise I'd suggest we proceed with this for now.

jpsim: Sure, I'll put something up shortly.

jpsim: https://github.com/envoyproxy/envoy-mobile/pull/2452

goaway:
Closing in favor of https://github.com/envoyproxy/envoy-mobile/pull/2452

PR #2452, opened 2022-08-02
iOS: add `KeyValueStore` protocol conformance to `UserDefaults`

jpsim:
Continued from #2441.

Description: Parallels Android implementation based on SharedPreferences.
Risk Level: Low
Testing: Application


library/swift/BUILD
```diff
@@ -9,45 +9,7 @@ licenses(["notice"])  # Apache 2
...
-        "StreamClientImpl.swift",
-        "StreamIntel.swift",
-        "StreamPrototype.swift",
-        "TestEngineBuilder.swift",
-        "Trailers.swift",
-        "UpstreamHttpProtocol.swift",
-        "filters/*.swift",
-        "grpc/*.swift",
-        "mocks/*.swift",
-        "stats/*.swift",
```
goaway, line 50: Personally, I'd prefer we explicitly enumerate source files in build targets and avoid blobs. I feel it obscures information about what's being pulled in, and it has the potential to lead to surprises when files are moved around.

Upstream Envoy tends to follow the pattern of explicit enumeration of source files as well, even when that list is large.

goaway (changes requested):
Thanks @jpsim, this is a nifty approach! Left one comment, but otherwise looks good.

library/swift/BUILD
jpsim: Done. I don't feel strongly but the previous approach was inconsistent with a mix of explicit files and globs.

As of 4d77493 everything is being explicitly set.

Augustyniak (approved):

PR #2468, opened 2022-08-15
Add a binary build target for Cronvoy

RenjieTang:
Signed-off-by: Renjie Tang <renjietang@chromium.org>

Description: Add a Cronvoy binary target for the integration of Cronet-dependent projects
Risk Level: Low
Testing: build file only.
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/library/java/org/chromium/net/ChromiumNetManifest.xml
+++ b/library/java/org/chromium/net/ChromiumNetManifest.xml
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="org.chromium.net">
     <uses-sdk
-            android:minSdkVersion="9"
+            android:minSdkVersion="21"
             android:targetSdkVersion="27" />
 
 </manifest>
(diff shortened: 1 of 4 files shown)


library/java/org/chromium/net/ChromiumNetManifest.xml
```diff
@@ -2,7 +2,7 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
           package="org.chromium.net">
     <uses-sdk
-            android:minSdkVersion="9"
+            android:minSdkVersion="21"
```
RyanTheOptimist, line 5: What's the impact of moving this from 9 to 21? Does this prevent it from running on older Androids?

RyanTheOptimist (commented):
Nice!

library/java/org/chromium/net/ChromiumNetManifest.xml
RenjieTang: There's no impact since the Envoy-mobile [engine](https://github.com/envoyproxy/envoy-mobile/blob/main/library/java/io/envoyproxy/envoymobile/engine/AndroidEngineManifest.xml) requires 21+.
Cronvoy won't be able to support 9-21 anyway.

RyanTheOptimist: Ah, good point!

RyanTheOptimist (approved):

PR #2469, opened 2022-08-15
Correct version string for CronetUrlRequestContext

RenjieTang:
Signed-off-by: Renjie Tang <renjietang@chromium.org>

Description: Modify version string of the Cronvoy native engine so that it matches https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/android/java/src/org/chromium/net/impl/CronetUrlRequestContext.java;l=319?q=getversionstring&ss=chromium%2Fchromium%2Fsrc:components%2Fcronet%2F
Risk Level: Low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/library/java/org/chromium/net/impl/CronetUrlRequestContext.java
+++ b/library/java/org/chromium/net/impl/CronetUrlRequestContext.java
@@ -159,7 +159,7 @@ public final class CronetUrlRequestContext extends CronetEngineBase {
 
   @Override
   public String getVersionString() {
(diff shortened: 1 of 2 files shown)


RyanTheOptimist (dismissed):
LGTM! Any chance it would be *trivial* to add some sort of EXPECT_SOMETHING() of `getVersionString()` in test/java/org/chromium/net/impl/CronvoyEngineTest.java. If it's trivial let's add it. If not, we can land as is.

RyanTheOptimist (approved):

PR #2472, opened 2022-08-16
[Cronvoy] Plumb through enableQuic API call

RenjieTang:
Description: Connect Cronet's enableQuic() call with Envoy-mobile's QUIC knob.
Additional Description: Note that with the change, Cronvoy will no longer force any request into the h2 only cluster.
Risk Level: Low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a


library/java/org/chromium/net/impl/CronetUrlRequest.java
```diff
@@ -574,11 +574,12 @@ private void fireOpenConnection() {
...
       throw new IllegalArgumentException("Requests with upload data must have a Content-Type.");
     }
-    String protocol =
-        isHttp2Enabled && url.getProtocol().equalsIgnoreCase("https") ? "http2" : "http1";
 
-    headers.computeIfAbsent("x-envoy-mobile-upstream-protocol", unused -> new ArrayList<>())
-        .add(protocol);
+    if (isQuicEnabled) {
+      headers.computeIfAbsent("x-envoy-mobile-upstream-protocol", unused -> new ArrayList<>())
+          .add("http3");
```
RyanTheOptimist, line 580: If this header is missing, to which cluster does the request get routed? I guess base_h2 which then has ALPN enabled to it *can* speak HTTP/1, if needed?

RyanTheOptimist (commented):
This looks good, but makes me wonder if we could/should have any tests which verify this behavior. Do we have an h3 integration test we could leverage? (Maybe not?)

RenjieTang:
> This looks good, but makes me wonder if we could/should have any tests which verify this behavior. Do we have an h3 integration test we could leverage? (Maybe not?)

The closest thing we have might be https://github.com/envoyproxy/envoy-mobile/blob/c3fa0e24d9b462a269cb58dbcf7148fb6d9ea8ce/test/java/org/chromium/net/CronetUrlRequestContextTest.java#L1227. We'll need to implement something like https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/android/test/cronet_url_request_context_config_test.cc;drc=4068c47b995c9280e7df0272cd52d281b22e44fc;l=24 to verify it.

library/java/org/chromium/net/impl/CronetUrlRequest.java
RenjieTang: It will go to _base_ cluster.
With this change, no requests will go to _base_h2_ cluster. In fact, I think for our use case, we could eventually delete the base_h2 cluster since we won't do forced h2 connections.

RenjieTang:
> > This looks good, but makes me wonder if we could/should have any tests which verify this behavior. Do we have an h3 integration test we could leverage? (Maybe not?)
> 
> The closest thing we have might be
> 
> https://github.com/envoyproxy/envoy-mobile/blob/c3fa0e24d9b462a269cb58dbcf7148fb6d9ea8ce/test/java/org/chromium/net/CronetUrlRequestContextTest.java#L1227
> 
> . We'll need to implement something like https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/android/test/cronet_url_request_context_config_test.cc;drc=4068c47b995c9280e7df0272cd52d281b22e44fc;l=24 to verify it.

I found that there are multiple Cronvoy tests that are disabled. I think once Cronvoy is more seriously used, we should do some audits on them.

RyanTheOptimist (approved):

PR #2476, opened 2022-08-18
bazel: define custom Xcode schemes

jpsim:
This makes it much easier to find relevant schemes (there were over 2,000 schemes before this) and labels them with more descriptive and human-readable names than `__examples_swift_hello_world_app`.

```console
# Before
$ ls Envoy.xcodeproj/xcshareddata/xcschemes | wc -l
  2231
# After
$ ls Envoy.xcodeproj/xcshareddata/xcschemes | wc -l
  7
```

New scheme picker looks like this:

<image  src=https://user-images.githubusercontent.com/474794/185428361-cb8b42a9-e7e1-4ddd-85db-a834b9796780.png width=250 />

I've separated the "Swift Library" scheme from the "iOS Tests" one because the Swift Library can be built for macOS too, but the tests only run on iOS.

If you want to build a scheme that's not in this list, you can quickly revert back to the previous behavior to generate schemes for all transitive dependencies by setting `scheme_autogeneration_mode = "all"`.

Depends on https://github.com/envoyproxy/envoy-mobile/pull/2475.

Risk Level: Low, local dev only
Testing: Local
Docs Changes: Updated
Release Notes: N/A

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -120,7 +120,7 @@ xcodeproj(
     build_mode = "bazel",
     project_name = "Envoy",
     tags = ["manual"],
-    targets = [
+    top_level_targets = [
         # Libraries
         "//library/swift:ios_lib",
         "//library/objective-c:envoy_engine_objc_lib",
--- a/BUILD
+++ b/BUILD
@@ -123,7 +123,7 @@ xcodeproj(
     bazel_path = "./bazelw",
     build_mode = "bazel",
     project_name = "Envoy",
-    scheme_autogeneration_mode = "auto", # Switch to "all" to generate schemes for all deps
+    scheme_autogeneration_mode = "auto",  # Switch to "all" to generate schemes for all deps
     schemes = [
         xcode_schemes.scheme(
             name = "Async Await App",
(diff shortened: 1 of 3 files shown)


Augustyniak (approved):

BUILD
```diff
@@ -119,12 +123,52 @@ xcodeproj(
...
+            launch_action = xcode_schemes.launch_action("//test/swift/apps/experimental:app"),
+        ),
+        xcode_schemes.scheme(
+            name = "Swift Library",
+            build_action = xcode_schemes.build_action(["//library/swift:ios_lib"]),
+        ),
+        xcode_schemes.scheme(
+            name = "iOS Tests",
+            test_action = xcode_schemes.test_action([
+                "//experimental/swift:quic_stream_test.__internal__.__test_bundle",
```
keith, line 156: would be nice to not have to rely on this implementation detail

keith (approved):
very cool!

BUILD
jpsim: Agreed. https://github.com/buildbuddy-io/rules_xcodeproj/issues/215

Issue #2432, opened 2022-07-19
is_cleartext_permitted method always crashes


Augustyniak:
The implementation of `is_cleartext_permitted` method always crashes.

That's because the `jcls_AndroidNetworkLibrary` variable in https://github.com/envoyproxy/envoy-mobile/blob/6eb67e6cd004665a01a35981b5728450ba6a179c/library/common/jni/android_jni_utility.cc#L23 is NULL.

See failing CI jobs in https://github.com/envoyproxy/envoy-mobile/pull/2431 PR draft.

<img width="1449" alt="Screen Shot 2022-07-19 at 3 40 38 PM" src="https://user-images.githubusercontent.com/4410346/179764752-c663ebc0-7155-411d-83ec-db99092ac31c.png">

```
* thread #25, name = 'EnvoyMain', stop reason = signal SIGABRT
  * frame #0: 0x0000007cd11fdacc libc.so`abort + 164
    frame #1: 0x0000007a2c2d7bb4 libart.so`art::Runtime::Abort(char const*) + 672
    frame #2: 0x0000007a00e16960 libbase.so`android::base::SetAborter(std::__1::function<void (char const*)>&&)::$_3::__invoke(char const*) + 80
    frame #3: 0x0000007a00e15f90 libbase.so`android::base::LogMessage::~LogMessage() + 368
    frame #4: 0x0000007a2bfd6f50 libart.so`art::JavaVMExt::JniAbort(char const*, char const*) + 2444
    frame #5: 0x0000007a2c18d6e4 libart.so`art::JavaVMExt::JniAbortV(char const*, char const*, std::__va_list) + 112
    frame #6: 0x0000007a2c18877c libart.so`art::(anonymous namespace)::ScopedCheck::AbortF(char const*, ...) + 144
    frame #7: 0x0000007a2c18c990 libart.so`art::(anonymous namespace)::ScopedCheck::CheckThread(_JNIEnv*) + 444
    frame #8: 0x0000007a2c169b4c libart.so`art::(anonymous namespace)::ScopedCheck::Check(art::ScopedObjectAccess&, bool, char const*, art::(anonymous namespace)::JniValueType*) + 572
    frame #9: 0x0000007a2c175850 libart.so`art::(anonymous namespace)::CheckJNI::GetMethodIDInternal(char const*, _JNIEnv*, _jclass*, char const*, char const*, bool) + 380
[...]

PR #2483, opened 2022-08-19
jni: introduce custom find_class method

Augustyniak:
Description: The change is needed in order for JNI to be able to find classes which are defined in our repository (as opposed to built in Java types). Google documents in [here](https://developer.android.com/training/articles/perf-jni#native-libraries) touches on this issue. The implementation could be improved by adding some caching to it but I am inclined to keep it simple unless we find out that its performance is an issue. Needed  for https://github.com/envoyproxy/envoy-mobile/issues/2432.
Risk Level: None, not used anywhere as of now.
Testing: Manual
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/library/common/jni/jni_utility.h
+++ b/library/common/jni/jni_utility.h
@@ -22,7 +22,7 @@ void set_class_loader(jobject class_loader);
  * application's context and should be associated with project's code - Java classes
  * defined by the project. For finding classes of Java built in-types use
  * `env->FindClass(...)` method instead as it is lighter to use.
- * 
+ *
  * The method works on Android targets only as the `set_class_loader` method is not
  * called by JVM-only targets.
  *
--- a/library/common/jni/jni_utility.h
+++ b/library/common/jni/jni_utility.h
@@ -14,6 +14,22 @@ JavaVM* get_vm();
 
 JNIEnv* get_env();
 
+void set_class_loader(jobject class_loader);
+
+/**
+ * Finds a class with a given name using a class loader provided with the use
+ * of `set_class_loader` function. The class loader is supposed to come from 
+ * application's context and should be associated with project's code - Java classes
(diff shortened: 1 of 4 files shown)


library/common/jni/jni_utility.h
```diff
@@ -14,6 +14,22 @@ JavaVM* get_vm();
...
+void set_class_loader(jobject class_loader);
+
+/**
+ * Finds a class with a given name using a class loader provided with the use
+ * of `set_class_loader` function. The class loader is supposed to come from
+ * application's context and should be associated with project's code - Java classes
+ * defined by the project. For finding classes of Java built in-types use
+ * `env->FindClass(...)` method instead as it is lighter to use.
+ *
+ * @param class_name, the name of the class to find (i.e. "org.chromium.net.AndroidNetworkLibrary").
```
snowp, line 26: its / delimited, not . right?

snowp (dismissed):
Maybe document somewhere that this only gets set up for Android? if we tried to use this for the jvm-only targets the static CL would be null

library/common/jni/jni_utility.h
Augustyniak: It requires `.` - I added an example to avoid confusion since I realize that `/'` would be a more obvious choice here.

Augustyniak:
> Maybe document somewhere that this only gets set up for Android? if we tried to use this for the jvm-only targets the static CL would be null

Good idea, updated.

library/common/jni/jni_utility.h
snowp: oh i see, you're calling loadClass and not findClass

nice consistency java

jpsim (approved):
Tests can come when we actually start using this.

RyanTheOptimist (approved):
Looks great! Thanks for doing this.

PR #2485, opened 2022-08-19
jni: release strings after they are no longer needed

Augustyniak:
Description: Avoid leaking config/log_level related strings. The leak happens during engine initialization.
Risk Level: Low
Testing: Manual
Docs Changes:  N/A
Release Notes: N/A

### DIFF
--- a/library/common/jni/jni_interface.cc
+++ b/library/common/jni/jni_interface.cc
@@ -120,8 +120,12 @@ extern "C" JNIEXPORT jlong JNICALL Java_io_envoyproxy_envoymobile_engine_JniLibr
 
 extern "C" JNIEXPORT jint JNICALL Java_io_envoyproxy_envoymobile_engine_JniLibrary_runEngine(
     JNIEnv* env, jclass, jlong engine, jstring config, jstring log_level) {
-  return run_engine(engine, env->GetStringUTFChars(config, nullptr),
-                    env->GetStringUTFChars(log_level, nullptr), "");
+  const char* string_config = env->GetStringUTFChars(config, nullptr);
+  const char* string_log_level = env->GetStringUTFChars(log_level, nullptr);
+  jint result = run_engine(engine, string_config, string_log_level, "");
+  env->ReleaseStringUTFChars(config, string_config);
+  env->ReleaseStringUTFChars(log_level, string_log_level);
+  return result;
 }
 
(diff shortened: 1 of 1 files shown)


library/common/jni/jni_interface.cc
```diff
@@ -120,8 +120,12 @@ extern "C" JNIEXPORT jlong JNICALL Java_io_envoyproxy_envoymobile_engine_JniLibr
 
 extern "C" JNIEXPORT jint JNICALL Java_io_envoyproxy_envoymobile_engine_JniLibrary_runEngine(
     JNIEnv* env, jclass, jlong engine, jstring config, jstring log_level) {
-  return run_engine(engine, env->GetStringUTFChars(config, nullptr),
-                    env->GetStringUTFChars(log_level, nullptr), "");
+  const char* string_config = env->GetStringUTFChars(config, nullptr);
+  const char* string_log_level = env->GetStringUTFChars(log_level, nullptr);
+  jint result = run_engine(engine, string_config, string_log_level, "");
+  env->ReleaseStringUTFChars(config, string_config);
```
snowp, line 126: If you want bonus points and more reusable code, the C++ way of doing this stuff would be to define a RAII wrapper, something like 

```
class JniStringHandle {
public:
  JniStringHandle(JNIEnv& env, jstring string) : 
    env_(env),
    jni_string_(string),
    string_(env.GetStringUTFChars(config, nullptr)) {}
  ~JniStringHandle() { env.ReleaseStringUTFChars(jni_string, string); }

  absl::string_view string() { return string_; }

private:
  JNIEnv& env_;
  jstring jni_string;
  const char* string_;
}
```

which would provide automatic cleanup and wraps up all the initialization of the string into the constructor

snowp (approved):

library/common/jni/jni_interface.cc
Augustyniak: Yes, I think that this is the way we will need to follow. I'm not going to do this as part of this PR tho and there is one thing that I would like to hear your thoughts on - so we have a few places where we allocate string UTF chars and allow them to leak (i.e. [here](https://github.com/envoyproxy/envoy-mobile/blob/main/library/common/jni/jni_interface.cc#L951)). Do you think that for these cases - at least for now - we should just not use the RAII wrapper or should we just always ensure that we copy the const char* on the receiving end (if we need it to live for longer than the duration of the receiving call)?

PR #2486, opened 2022-08-19
core: pass strings as references

Augustyniak:
Description: Passing these strings as references may help us to avoid doing copying during initialization.
Risk Level: Low
Testing: Manual
Docs Changes: N/A
Release Notes: N/A


goaway:
You need to be absolutely sure of the lifetime of these strings if you're going to pass them as references. `main` is used to spawn a new thread, and the call stack from `run` will return long before it's complete. Unless these are managed safely somewhere, I don't see how this can be safe to do. On the other hand, when passing as raw strings, there's a chance the compiler/optimizer can still manage the string by default relatively efficiently and avoid some unnecessary copies (the standard mentions a few cases copy elision is allowed - that may be a stretch though, but you could of course examine the compiled code to see).

Issue #2450, opened 2022-08-02
An ability to run tests on Android emulator


Augustyniak:
Add a support for running Android tests on Android emulator to properly test all code that runs when Envoy Mobile runs on real Android devices.

Some of the code that's executed when Envoy Mobile runs on Android is hidden behind a compiler flag check that looks as follow: `#if defined(__ANDROID_API__)`. For example: https://github.com/envoyproxy/envoy-mobile/blob/4cf9236a0e1ce23d48d6202bbfc25820d88d7f03/library/common/jni/android_jni_utility.cc#L19

To properly test this code Envoy Mobile has to be build under android cc tool chain and the app (or test host) has to run on an emulator or a real device. As of now, there are no option to run tests like this in Envoy Mobile repo. Since testing of such code is impossible it's currently untested which leads to issues like https://github.com/envoyproxy/envoy-mobile/issues/2432.

On the surface it seems that we could add support for running Android tests on emulator by using `android_instrumentation_test` macro that's documented in [here](https://docs.bazel.build/versions/main/android-instrumentation-test.html). There is a GH issue https://github.com/bazelbuild/bazel/issues/7069 which seems to suggest that this specific bazel macro does not currently work on macOS. That should not be a problem if we are OK with running these tests on Linux machine.

Augustyniak:
We talked with @jpsim and Lyft client tooling folks to see what the next steps here could be. A few updates/notes from what we talked about and eventually decided on:

- we decided not to go with `android_instrumentation_test` macro. It not working on mac os being one of the reasons.
- we are planning to have a setup that's similar to our set up for running Envoy Mobile example apps. We will write custom scripts for compiling test files as part of a test example app that will be then run on an actual Android emulator. There are some concerns here related to state sharing between different tests (we would like tests to be isolated from each other) and the ability to use Android testing framework when our tests are compiled as part of a 'normal' Android app/target (and not a special testing one).

PR #2488, opened 2022-08-22
eventually fixing cleartext code

alyssawilk:
Signed-off-by: Alyssa Wilk <alyssar@chromium.org>

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]


Augustyniak:
@alyssawilk just FYI: as they are now, Robolectric tests are _not_ enough to test the implementation of `is_cleartext_permitted` since the `#if defined(__ANDROID_API__)` that's a part of the implementation of that method evaluates to `false` in Roboelectric tests. For `#if defined(__ANDROID_API__)` to evaluate to `true` we need to build EM with Android tool chain and run the app on the actual emulator - that's something that we are working on as part of https://github.com/envoyproxy/envoy-mobile/issues/2450.

PR #2491, opened 2022-08-22
ci: use 'jpsim/retest' GitHub Action for `/retest` command

jpsim:
Defined here: https://github.com/jpsim/retest

It has a number of advantages over the previous in-repo implementation:

1. Runs faster because it runs in a node.js environment and doesn't require pulling and running a Docker container
2. Written in Typescript instead of bash/jq
3. Has unit tests
4. _Potentially_ will be better maintained if it ends up being used & improved by other people unrelated to their relationship with Envoy Mobile

Otherwise should be at feature parity with the previous implementation.


Augustyniak:
Should we move this to be a part of one of the existing orgs such as `envoy-mobile`? While `envoy-mobile` is not as good as `envoyproxy` would be it's better than associating the tools we use with privately owned repos I think. That of course assumes that you did not work on this in your private time.

jpsim:
> Should we move this to be a part of one of the existing orgs such as `envoy-mobile`? While `envoy-mobile` is not as good as `envoyproxy` would be it's better than associating the tools we use with privately owned repos I think. That of course assumes that you did not work on this in your private time.

I did work on this over the weekend and consider it mostly unrelated to Envoy, Envoy Mobile & CNCF projects, despite a primary motivation being to use it here.

I don't think it makes sense to have it be hosted and maintained as part of Envoy Mobile, whereas I would intend to keep maintaining it even if I were to stop working on Envoy Mobile.

The code is permissively licensed under MIT and so I could inline it in this project, but I think there are more benefits to treating this as an external dependency.

Augustyniak:
> I don't think it makes sense to have it be hosted and maintained as part of Envoy Mobile, whereas I would intend to keep maintaining it even if I were to stop working on Envoy Mobile.

I guess `Lyft` org would be a better proposal than `Envoy Mobile` org as Lyft will probably continue to be interested in maintaining whatever needs to continue to work for EM project to continue to operate. Personally, I do not see benefits of having this tool being attached to your personal GH account but if it's MIT based we can always copy-paste and modify it as needed.

Augustyniak (approved):

jpsim:
> I guess `Lyft` org would be a better proposal than `Envoy Mobile` org as Lyft will probably continue to be interested in maintaining whatever needs to continue to work for EM project to continue to operate.

Lyft also has an involved process for approving open source projects that fall under its control that seems like overkill for such a simple and focused utility. I created the repo under my personal GitHub org to avoid that overhead.

PR #2492, opened 2022-08-22
ci: update GitHub Actions Linux runners to run on `ubuntu-latest`

jpsim:
The `ubuntu-18.04` image has been deprecated: https://github.com/actions/runner-images/issues/6002

May explain why our Linux CI jobs started being cancelled before even starting. https://envoyproxy.slack.com/archives/C02F93EEJCE/p1661181247759129


jpsim:
Since (almost) all these jobs all specify a Docker container, I think it doesn't matter much what version of Linux actually hosts the Docker instance, which is why I'm specifying `ubuntu-latest` instead of a pinned but more recent version of Ubuntu, so this keeps working for the foreseeable future and we don't hit another surprise deprecation when the next version is dropped.

keith (approved):

PR #2495, opened 2022-08-23
docs: update find_class documentation

Augustyniak:
Description: Update the documentation of `find_class` method as it seems to accept `/` and `.` and using `/` should be less confusing as this is what `env->findClass` expects.
Risk Level: None
Testing: Manual confirmed that the current implementing of tagging works and it does uses `/` as opposed to `.`.
Docs Changes: N/A
Release Notes: N/A

Signed-off-by: Rafał Augustyniak <Augustyniak@users.noreply.github.com>


RyanTheOptimist (approved):
LGTM. Thanks for doing this!

PR #2497, opened 2022-08-24
bazel: add iOS dynamic framework options

keith:
If folks prefer to use dynamic frameworks to more easily share between
apps and extensions, this provides the targets to do so, while also
limiting their symbols to only those required to enable as much
stripping as possible.


keith:
we need to fill out the actual symbols here, but i think this will work as expected once we do. we might need to make that file configurable as well

keith:
we can redo this on the envoy repo when we want it

PR #2503, opened 2022-08-30
testing: make find_class method work in tests

Augustyniak:
Description: Update the implementation of `Java_io_envoyproxy_envoymobile_engine_AndroidJniLibrary_initialize` so that it takes a `class_loader` argument which was introduced in https://github.com/envoyproxy/envoy-mobile/pull/2483. It turns out that this method is implemented in two separate places - `android_jni_interface.cc` and  and `android_test_jni_interface.cc` files - and my previous PR updated only the former. This is needed to make `find_class` method work in tests as tests depend on  the implementation from `android_test_jni_interface.cc` file.
Risk Level: None, additive change for test targets only.
Testing: None, a follow up tests that depend on `find_class` method are being worked on.
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/library/common/jni/android_jni_utility.cc
+++ b/library/common/jni/android_jni_utility.cc
@@ -37,7 +37,7 @@ bool is_cleartext_permitted(absl::string_view hostname) {
 void tag_socket(int ifd, int uid, int tag) {
 #if defined(__ANDROID_API__)
   JNIEnv* env = get_env();
-  jclass jcls_AndroidNetworkLibrary = find_class("org/chromium/net/AndroidNetworkLibrary");
+  jclass jcls_AndroidNetworkLibrary = find_class("org.chromium.net.AndroidNetworkLibrary");
   jmethodID jmid_tagSocket =
       env->GetStaticMethodID(jcls_AndroidNetworkLibrary, "tagSocket", "(III)V");
   env->CallStaticVoidMethod(jcls_AndroidNetworkLibrary, jmid_tagSocket, ifd, uid, tag);
--- a/library/common/jni/android_test_jni_interface.cc
+++ b/library/common/jni/android_test_jni_interface.cc
@@ -10,7 +10,9 @@
 extern "C" JNIEXPORT jint JNICALL
 Java_io_envoyproxy_envoymobile_engine_AndroidJniLibrary_initialize(JNIEnv* env,
                                                                    jclass, // class
+                                                                   jobject class_loader,
                                                                    jobject connectivity_manager) {
+  set_class_loader(env->NewGlobalRef(class_loader));
 
   return 0;
 }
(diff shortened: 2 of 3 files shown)


library/common/jni/android_jni_utility.cc
```diff
@@ -37,7 +37,7 @@ bool is_cleartext_permitted(absl::string_view hostname) {
 void tag_socket(int ifd, int uid, int tag) {
 #if defined(__ANDROID_API__)
   JNIEnv* env = get_env();
-  jclass jcls_AndroidNetworkLibrary = find_class("org/chromium/net/AndroidNetworkLibrary");
+  jclass jcls_AndroidNetworkLibrary = find_class("org.chromium.net.AndroidNetworkLibrary");
```
Augustyniak, line 40: @RyanTheOptimist It turns out after all that we need `.` in here as @danzh2010 ran into a case where the use `/` was causing a crash (and `.` worked just fine)

RyanTheOptimist: *boggle* :)

library/common/jni/android_test_jni_interface.cc
```diff
@@ -10,7 +10,9 @@
 extern "C" JNIEXPORT jint JNICALL
 Java_io_envoyproxy_envoymobile_engine_AndroidJniLibrary_initialize(JNIEnv* env,
                                                                    jclass, // class
+                                                                   jobject class_loader,
```
RyanTheOptimist, line 13: Since this method has a new signature, I would have expected the callers to need to change. Is there a reason they don't have to?

RyanTheOptimist (commented):
Should the PR title be modified so that "find_test" => "find_class"?

library/common/jni/android_test_jni_interface.cc
Augustyniak: Done - thanks!

Augustyniak: Yea, the issue is that the current signature - from before this PR - is the one that's wrong. AFAIk the only reason for why this compiles is that JNI does not enforce the existence of the implementation that matches whatever we expose in JNI header files. It's probably currently not crashing in main only because we do not use any of the passed arguments in here (or by luck).

library/common/jni/android_jni_utility.cc
Augustyniak: it's like tabs versus spaces discussion all over again.  🙃

library/common/jni/android_test_jni_interface.cc
RyanTheOptimist: Bananas! :)

RyanTheOptimist (approved):

PR #2505, opened 2022-08-31
android: add an empty implementation of isCleartextTrafficPermitted

Augustyniak:
Description: Fixes a crash https://github.com/envoyproxy/envoy-mobile/issues/2432
Risk Level: Low, it effectively does nothing other than returning `true` and should prevent the app from crashing
Testing: Manual
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/library/java/org/chromium/net/AndroidNetworkLibrary.java
+++ b/library/java/org/chromium/net/AndroidNetworkLibrary.java
@@ -104,6 +104,14 @@ public final class AndroidNetworkLibrary {
     }
   }
 
+  /*
+   * Returns true if cleartext traffic to a given host is allowed by the current app.
+   */ 
+  public static boolean isCleartextTrafficPermitted(String host) {
+    // TODO(Augustyniak): Implement this method.
+    return true;
+  }
+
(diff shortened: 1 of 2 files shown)


library/java/org/chromium/net/AndroidNetworkLibrary.java
```diff
@@ -104,6 +104,14 @@ public static void clearTestRootCertificates()
     }
   }
 
+  /*
+   * Returns true if cleartext traffic to a given host is allowed by the current app.
+   */
+  public static boolean isCleartextTrafficPermitted(String host) {
+    // TODO(Augustyniak): Implement this method.
```
alyssawilk, line 111: feel free to move this TODO to me

alyssawilk (approved):

library/java/org/chromium/net/AndroidNetworkLibrary.java
Augustyniak: done. thank you!

alyssawilk (approved):

PR #2510, opened 2022-09-01
android: enable forceIPv6 by default

Augustyniak:
Description: Forcing the use of IPv6 socket addresses is required to make Envoy Mobile work with some carriers when running on an Android device. The option was introduced upstream in https://github.com/envoyproxy/envoy/pull/21803.
Risk Level: Low, enabled a well tested option.
Testing: Manual, launched the example app
Docs Changes: Updated
Release Notes: Updated

### DIFF
--- a/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
+++ b/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
@@ -44,7 +44,6 @@ public class EnvoyConfiguration {
   public final Boolean enableSocketTagging;
   public final Boolean enableHappyEyeballs;
   public final Boolean enableInterfaceBinding;
-  public final Boolean forceIPv6;
   public final Integer h2ConnectionKeepaliveIdleIntervalMilliseconds;
   public final Integer h2ConnectionKeepaliveTimeoutSeconds;
   public final Boolean h2ExtendKeepaliveTimeout;
@@ -88,7 +87,6 @@ public class EnvoyConfiguration {
    * @param enableSocketTagging          whether to enable socket tagging.
    * @param enableHappyEyeballs          whether to enable RFC 6555 handling for IPv4/IPv6.
    * @param enableInterfaceBinding       whether to allow interface binding.
-   * @param forceIPv6                    whether to force connections to use IPv6.
    * @param h2ConnectionKeepaliveIdleIntervalMilliseconds rate in milliseconds seconds to send h2
    *     pings on stream creation.
    * @param h2ConnectionKeepaliveTimeoutSeconds rate in seconds to timeout h2 pings.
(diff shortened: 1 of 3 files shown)


docs/root/intro/version_history.rst
```diff
@@ -47,6 +47,7 @@ Features:
 - api: improved C++ APIs compatibility with Java / Kotlin / Swift (:issue `#2362 <2362>`)
 - api: add option to use the a ``getaddrinfo``-based system DNS resolver instead of c-ares (:issue: `#2419 <2419>`)
 - iOS: add ``KeyValueStore`` protocol conformance to ``UserDefaults`` (:issue: `#2452 <2452>`)
+- android: enable forcing of IPv6 socket addresses by default and remove ``forceIPv6`` method from the engine builder. (:issue: `#2510 <2510>`)
```
jpsim, line 50: I wonder if we should adjust this line to reflect this:

```
- api: add experimental option to force all connections to use IPv6 (:issue: `#2379 <2379>`, :issue: `#2396 <2396>`)
```

Augustyniak: yep, great point - done. I still use 'ios' / 'android' to stay consistent for this release but planning to move to swift etc in the next release.

library/cc/engine_builder.cc
```diff
@@ -146,29 +146,31 @@ std::string EngineBuilder::generateConfigStr() {
...
-      {"per_try_idle_timeout", fmt::format("{}s", this->per_try_idle_timeout_seconds_)},
-      {"virtual_clusters", this->virtual_clusters_},
+  std::vector<std::pair<std::string, std::string>> replacements {
+    {"connect_timeout", fmt::format("{}s", this->connect_timeout_seconds_)},
+        {"dns_fail_base_interval", fmt::format("{}s", this->dns_failure_refresh_seconds_base_)},
+        {"dns_fail_max_interval", fmt::format("{}s", this->dns_failure_refresh_seconds_max_)},
+        {"dns_preresolve_hostnames", this->dns_preresolve_hostnames_},
+        {"dns_refresh_rate", fmt::format("{}s", this->dns_refresh_seconds_)},
+        {"dns_query_timeout", fmt::format("{}s", this->dns_query_timeout_seconds_)},
+        {"dns_resolver_name", dns_resolver_name}, {"dns_resolver_config", dns_resolver_config},
```
jpsim, line 156: Can we keep these on two lines?
```suggestion
        {"dns_resolver_name", dns_resolver_name},
        {"dns_resolver_config", dns_resolver_config},
```

jpsim (approved):

library/cc/engine_builder.cc
Augustyniak: it's a linter tool that made a change in here. It does not allow me to make a change from your suggestion.

For reference, I ran `./tools/code_format/check_format.py fix ../library/` from within `envoy` directory locally.

jpsim: huh, interesting that it was previously on two lines.

PR #2511, opened 2022-09-01
android: default to use the system DNS resolver

jpsim:
Description: Rather than c-ares. In our tests, the system DNS resolver consumes less CPU and memory while otherwise performing similarly overall.
Risk Level: Medium, although it's performed well in our tests, it's possible that some situations we haven't tested perform worse than c-ares. Users can still disable this in the engine builder though.
Testing: Lyft apps & unit tests
Docs Changes: Docstrings updated to reflect new default
Release Notes: Added

### DIFF
--- a/docs/root/intro/version_history.rst
+++ b/docs/root/intro/version_history.rst
@@ -45,7 +45,7 @@ Features:
 - android: create simple persistent SharedPreferencesStore (:issue: `#2319 <2319>`)
 - iOS: A documentation archive is now included in the GitHub release artifact (:issue: `#2335 <2335>`)
 - api: improved C++ APIs compatibility with Java / Kotlin / Swift (:issue `#2362 <2362>`)
-- api: add option to use the a ``getaddrinfo``-based system DNS resolver instead of c-ares (:issue: `#2419 <2419>`)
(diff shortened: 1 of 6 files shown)


docs/root/intro/version_history.rst
```diff
@@ -44,7 +44,7 @@ Features:
 - android: create simple persistent SharedPreferencesStore (:issue: `#2319 <2319>`)
 - iOS: A documentation archive is now included in the GitHub release artifact (:issue: `#2335 <2335>`)
 - api: improved C++ APIs compatibility with Java / Kotlin / Swift (:issue `#2362 <2362>`)
-- api: add option to use the a ``getaddrinfo``-based system DNS resolver instead of c-ares (:issue: `#2419 <2419>`)
+- Android: default to use a ``getaddrinfo``-based system DNS resolver instead of c-ares (:issue: `#2419 <2419>`)
```
Augustyniak, line 47: for whatever reason Android is not capitalized in other occurences in this file. Up to you whether you want to stay consistent or not.

Augustyniak (approved):

docs/root/intro/version_history.rst
jpsim: Seems like we should honour brand name capitalization. As someone whose name is frequently mis-capitalized this might be a touchy issue for me 😅.

PR #2516, opened 2022-09-02
Fix Android certificate validation JNI calling convention

StefanoDuo:
Use JNI parameter ordering defined for instance methods (instead of the
static methods one).
Add tests to avoid regressions.

#1575

### DIFF
--- a/library/common/jni/jni_interface.cc
+++ b/library/common/jni/jni_interface.cc
@@ -1209,3 +1209,15 @@ Java_io_envoyproxy_envoymobile_engine_JniLibrary_callClearTestRootCertificateFro
                                                                                         jclass) {
   jvm_clear_test_root_certificate();
 }
+
+extern "C" JNIEXPORT envoy_cert_verify_status_t JNICALL
+Java_io_envoyproxy_envoymobile_engine_JniLibrary_callAndroidCertVerifyResultGetStatusFromNative(
+    JNIEnv* env, jclass, jobject result) {
+  return jvm_cert_get_status(env, result);
+}
+
+extern "C" JNIEXPORT bool JNICALL
+Java_io_envoyproxy_envoymobile_engine_JniLibrary_callAndroidCertVerifyResultIsIssuedByKnownRootFromNative(
+    JNIEnv* env, jclass, jobject result) {
+  return jvm_cert_is_issued_by_known_root(env, result);
+}
--- a/library/java/io/envoyproxy/envoymobile/engine/JniLibrary.java
+++ b/library/java/io/envoyproxy/envoymobile/engine/JniLibrary.java
@@ -381,7 +381,7 @@ public class JniLibrary {
    * @param certChain The ASN.1 DER encoded bytes for certificates.
    * @param authType The key exchange algorithm name (e.g. RSA).
    * @param host The hostname of the server.
-   * @return Android certificate verification result code.
+   * @return AndroidCertVerifyResult representing the certificate verification result.
    */
   public static native Object callCertificateVerificationFromNative(byte[][] certChain,
                                                                     byte[] authType, byte[] host);
@@ -396,7 +396,23 @@ public class JniLibrary {
   /**
    * Mimic a call to AndroidNetworkLibrary#clearTestRootCertificate from native code.
    * To be used for testing only.
-   *
    */
   public static native void callClearTestRootCertificateFromNative();
+
+  /**
+   * Mimic a call to AndroidCertVerifyResult#getStatus from native code.
+   * To be used for testing only.
+   *
+   * @param result AndroidCertVerifyResult to call getStatus onto.
+   */
+  public static native int callAndroidCertVerifyResultGetStatusFromNative(Object result);
+
+  /**
+   * Mimic a call to AndroidCertVerifyResult#isIssuedByKnownRoot from native code.
+   * To be used for testing only.
+   *
+   * @param result AndroidCertVerifyResult to call isIssuedByKnownRoot onto.
+   */
+  public static native boolean
+  callAndroidCertVerifyResultIsIssuedByKnownRootFromNative(Object result);
 }
--- a/library/java/org/chromium/net/AndroidCertVerifyResult.java
+++ b/library/java/org/chromium/net/AndroidCertVerifyResult.java
@@ -14,7 +14,7 @@ public final class AndroidCertVerifyResult {
   /**
    * The verification status. One of the values in CertVerifyStatusAndroid.
    */
-  private final int mStatus;
+  @CertVerifyStatusAndroid private final int mStatus;
 
   /**
    * True if the root CA in the chain is in the system store.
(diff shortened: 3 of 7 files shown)


StefanoDuo:
@danzh2010 Would you like to review this yourself or should I send this to someone else?

library/common/jni/jni_interface.cc
```diff
@@ -1146,7 +1145,7 @@ envoy_cert_verify_status_t jvm_cert_get_status(JNIEnv* env, jobject j_result) {
   jclass jcls_AndroidCertVerifyResult = env->FindClass("org/chromium/net/AndroidCertVerifyResult");
   jmethodID jmid_getStatus = env->GetMethodID(jcls_AndroidCertVerifyResult, "getStatus", "()I");
   envoy_cert_verify_status_t result = static_cast<envoy_cert_verify_status_t>(
-      env->CallIntMethod(jcls_AndroidCertVerifyResult, jmid_getStatus, j_result));
+      env->CallIntMethod(j_result, jmid_getStatus, j_result));
```
danzh2010, line 1148: Is the 3rd argument needed?

library/java/org/chromium/net/AndroidCertVerifyResult.java
```diff
@@ -14,7 +14,7 @@ public final class AndroidCertVerifyResult {
   /**
    * The verification status. One of the values in CertVerifyStatusAndroid.
    */
-  private final int mStatus;
+  @CertVerifyStatusAndroid private final int mStatus;
```
danzh2010, line 17: To educate myself, what is `@CertVerifyStatusAndroid` annotation for?

test/java/org/chromium/net/CertificateVerificationTest.java
```diff
@@ -143,4 +145,24 @@ public void testChainWithMultipleRootCertificates() throws Exception {
                                                                                   authType, host);
     assertEquals(result.getStatus(), CertVerifyStatusAndroid.OK);
   }
+
+  @Test
```
danzh2010, line 149: Nice tests! 

It's still unclear to me how the previous tests didn't run into any issue with the incorrect jobject being passed in. Can you explain a bit?

library/common/jni/jni_interface.cc
StefanoDuo: It's not, thanks for catching this!

library/java/org/chromium/net/AndroidCertVerifyResult.java
StefanoDuo: The `@CertVerifyStatusAndroid` annotation is declared at [java/org/chromium/net/CertVerifyStatusAndroid.java](https://sourcegraph.com/github.com/envoyproxy/envoy-mobile/-/blob/library/java/org/chromium/net/CertVerifyStatusAndroid.java).

It basically makes use of Android's [IntDef](https://developer.android.com/reference/androidx/annotation/IntDef) annotation to say: this is a "special int" and it's values should be constrained to those declared by `@CertVerifyStatusAndroid`.

I've taken this approach from Chromium's code, I think it makes sense in this case (compared for example to using an enum) because we will need an actual int to pass it through the JNI layer.

test/java/org/chromium/net/CertificateVerificationTest.java
StefanoDuo: I was also confused, but if you look closely at the tests above they don't exercise the path where C++ unpacks AndroidCertVerifyResult through JNI. They just construct the AndroidCertVerifyResult through JNI (`JniLibrary.callCertificateVerificationFromNative`) and then unpack that directly in Java, skipping the wrong JNI code...

The new tests generate the AndroidCertVerifyResult directly in Java, then unpack it through JNI calls, so now we should be covering the entirety of the JNI call surface.

danzh2010: Wow, that's tricky. But nice that we found it out!

danzh2010 (dismissed):
LGTM

RyanTheOptimist (approved):
Thanks for doing this!

library/java/org/chromium/net/AndroidCertVerifyResult.java
```diff
@@ -26,29 +26,35 @@ public final class AndroidCertVerifyResult {
...
     mCertificateChain = Collections.<X509Certificate>emptyList();
   }
 
-  // TODO(stefanoduo): Hook envoy-mobile JNI.
-  //@CalledByNative
-  public int getStatus() { return mStatus; }
+  /**
+   * Called from native.
+   */
+  @CertVerifyStatusAndroid
```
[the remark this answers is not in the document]
StefanoDuo: Syntactically both work. I've personally always seen annotation been put before public (e.g. https://developer.android.com/studio/write/annotations#adding-nullness).

Having said that, if you prefer the other way around (or envoy-mobile styleguide requires that) I can move it. Feel free to reopen if that's the case.

library/java/org/chromium/net/AndroidCertVerifyResult.java
```diff
@@ -26,29 +26,35 @@ public final class AndroidCertVerifyResult {
...
+  public int getStatus() {
+    return mStatus;
+  }
```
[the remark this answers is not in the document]
StefanoDuo: That's what I've discovered:
* If CertVerifyStatusAndroid is before public, that's what we get (not a one liner)
* if CertVerifyStatusAndroid is moved before int, this stays a one liner

library/java/io/envoyproxy/envoymobile/engine/JniLibrary.java
```diff
@@ -378,7 +378,23 @@ public static native Object callCertificateVerificationFromNative(byte[][] certC
...
    */
   public static native void callClearTestRootCertificateFromNative();
+
+  /**
+   * Mimic a call to AndroidCertVerifyResult#getStatus from native code.
+   * To be used for testing only.
+   *
+   * @param result AndroidCertVerifyResult to call getStatus onto.
+   */
+  public static native int callAndroidCertVerifyResultGetStatusFromNative(Object result);
```
Augustyniak, line 390: If this is supposed to be called from within tests only can we make it a part of a separate JNI interface - something such as `JNITestLibrary` or similar and add these 2 methods in there (maybe we could make JNITestLibrary to inherit from `JNILibrary`)? We could then include `JNILibrary` and these 2 methods in a test target only. That would allow us to avoid leaking testing utility methods in production code (and prevent accidental prod use of these methods).

Augustyniak (commented):
Looks good, thank you for fixing these! let one comment

I've seen a comment related to this already but just to reiterate - we will need to change `env->FindClass` to `find_class` to make this work on an actual device I think.

RyanTheOptimist:
@StefanoDuo is this ready to merge?

library/java/io/envoyproxy/envoymobile/engine/JniLibrary.java
StefanoDuo: I've been trying to split the JNI library, but I've been hitting a bunch of Java runtime errors (I'm not entirely sure if I'm supposed to register the additional testing library somewhere?).

How strongly do you feel about this? As in, I completely agree that it's a good thing to do, but it's also relatively obvious that these are methods to be used only for testing. So, I'm not sure how much time is worth spending into this to get it "right/bullet proof".

StefanoDuo:
P.S. Sorry for the late follow-up. I've been OOO for the past month.

AFAIU, most of these CL has been merged by @danzh2010 through https://github.com/envoyproxy/envoy-mobile/pull/2554. What's missing is just these additional tests, that prevent the calling convention bug to resurface in the future.

I've rebased and updated the code to reflect that.

jpsim:
Closing this PR as this repo is now archived. You can open a new PR in the Envoy repo if you'd like: https://github.com/envoyproxy/envoy

PR #2518, opened 2022-09-02
Update swift_flatbuffers to 2.0.8

jpsim:
Description: It now has a bazel build file for the Swift library so we can remove our own.
Risk Level: Low, we don't actually use this (for now), but it's used by the Lyft downstream integration.
Testing: CI
Docs Changes: N/A
Release Notes: N/A


jpsim:
We need to keep this in sync with Envoy's version: https://github.com/envoyproxy/envoy/blob/b08ca2e5831925f55611af2dd156c58ea27871b7/bazel/repository_locations.bzl#L1022

jpsim:
Trying again with https://github.com/envoyproxy/envoy/pull/22998

RyanTheOptimist (commented):
This seems to include Envoy change (or maybe an envoy import). Is that expected?

jpsim:
> This seems to include Envoy change (or maybe an envoy import). Is that expected?

Yes, pulls in https://github.com/envoyproxy/envoy/pull/22998

jpsim:
@goaway you were doing work with flat buffers recently, can you take a look?

Augustyniak (approved):

goaway:
I'll confirm that this works on my flatbuffers branch, but at a glance I don't see anything concerning.

PR #2520, opened 2022-09-03
android: fix admin interface

Augustyniak:
Description: Fix admin interface on Android. Enable admin interface in experimental example app.
Risk Level: low, admin interface is disabled by default and it's crashing on app launch on Android without the change from this PR anyway.
Testing: manual on iOS and Android
Docs Changes: none
Release Notes: updated

### DIFF
--- a/docs/root/intro/version_history.rst
+++ b/docs/root/intro/version_history.rst
@@ -10,7 +10,7 @@ Breaking changes:
 
 Bugfixes:
 
-- android: fix app launch crash for when admin interface is enabled. (:issue:`#2520 <2520>`)
+- android: fix engine startup crash for when admin interface is enabled. (:issue:`#2520 <2520>`)
 
 Features:
 
--- a/docs/root/intro/version_history.rst
+++ b/docs/root/intro/version_history.rst
@@ -10,7 +10,7 @@ Breaking changes:
 
 Bugfixes:
 
--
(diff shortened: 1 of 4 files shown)


library/common/config/config.cc
```diff
@@ -143,7 +143,7 @@ R"(- &enable_drain_post_dns_refresh false
 !ignore admin_interface_defs: &admin_interface
     address:
       socket_address:
-        address: 127.0.0.1
```
Augustyniak, line 146: @snowp @mattklein123 the IPv4 address in here stopped working after we started enforcing the use of IPv6 addresses on Android. Is moving to IPv6 address an acceptable solution here (it does seem to work)?

Augustyniak: This is used for when we enable Envoy admin interface so that we are able to do things such as `curl localhost:9901/stats` in the terminal to get the state of EM stats.

docs/root/intro/version_history.rst
```diff
@@ -10,7 +10,7 @@ Breaking changes:
 
 Bugfixes:
 
--
+- android: fix app launch crash for when admin interface is enabled. (:issue:`#2520 <2520>`)
```
jpsim, line 13: Nit: not crashing on launch, but on engine build?

jpsim (approved):

jpsim (approved):

Issue #2506, opened 2022-08-31
android: implement AndroidNetworkLibrary's isCleartextTrafficPermitted method


Augustyniak:
The method was introduced in https://github.com/envoyproxy/envoy-mobile/pull/2505 and just returns `true` now. The dummy implementation was introduced as the lack of this method was causing the application to crash when Envoy attempted to perform an HTTP (not s) request. Read more about the crash in https://github.com/envoyproxy/envoy-mobile/issues/2432.

We need to provide the proper implementation of the mthod - this is how [cronet's implementation](https://github.com/lizhangqu/cronet/blob/master/cronet-native/src/main/java/org/chromium/net/AndroidNetworkLibrary.java#L328-L335) looks like.

Fixed in #2528

PR #2528, opened 2022-09-07
android: fix cleartext implementation checks

alyssawilk:
The java changes from the initial PR (https://github.com/envoyproxy/envoy-mobile/pull/2318/commits/9759f4edb781f45cf0ef57bbacc5fb07e6a9c6c3) got lost in one of the subsequent pushes :-(

Signed-off-by: Alyssa Wilk <alyssar@chromium.org>

### DIFF
--- a/library/java/org/chromium/net/AndroidNetworkLibrary.java
+++ b/library/java/org/chromium/net/AndroidNetworkLibrary.java
@@ -109,15 +109,16 @@ public final class AndroidNetworkLibrary {
    * Returns true if cleartext traffic to a given host is allowed by the current app.
    */
   public static boolean isCleartextTrafficPermitted(String host) {
-    if (Build.VERSION.SDK_INT == Build.VERSION_CODES.M) {
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+      // This API was not implemented before Android M.
+      return true;
+    } else if (Build.VERSION.SDK_INT == Build.VERSION_CODES.M) {
       // M only supported global checks.
       return NetworkSecurityPolicy.getInstance().isCleartextTrafficPermitted();
+    } else {
+      // The host-specific API was implemented in Android N (which came after Android M).
+      return NetworkSecurityPolicy.getInstance().isCleartextTrafficPermitted(host);
     }
-    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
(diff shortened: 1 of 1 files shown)


library/java/org/chromium/net/AndroidNetworkLibrary.java
```diff
@@ -108,8 +109,15 @@ public static void clearTestRootCertificates()
...
    */
   public static boolean isCleartextTrafficPermitted(String host) {
-    // TODO(alyssawilk) Implement this method.
-    return true;
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
+      // This API was not implemented before Android N.
+      return true;
+    }
+    // M only supported global checks.
+    if (Build.VERSION.SDK_INT == Build.VERSION_CODES.M) {
```
Augustyniak, line 117: Isn't `M` before `N` - in that case I would expect this API to be not implemented for `M` per your comment from above.

Augustyniak: in other words - this condition will be never true as we will return early in line 114

alyssawilk: yes sorry, I'd collapsed https://source.chromium.org/chromium/chromium/src/+/main:net/android/java/src/org/chromium/net/AndroidNetworkLibrary.java;l=357?q=isCleartextTrafficPermitted&ss=chromium into one function and done it incorrectly.  thanks!

Augustyniak: Thank you for fixing.

If I may suggest something to make it easier to read this code since the flow is still a little bit confusing I think:

```Java
if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
    // This API was not implemented before Android M.
    return true;
} else if (Build.VERSION.SDK_INT == Build.VERSION_CODES.M) {
    // M only supported global checks.
    return NetworkSecurityPolicy.getInstance().isCleartextTrafficPermitted();
} else {
    // The host-specific API was implemented in Android N (which came after Android M).
    return NetworkSecurityPolicy.getInstance().isCleartextTrafficPermitted(host);
}
```

What do you think?

Augustyniak (approved):
🚢

Augustyniak:
fixes https://github.com/envoyproxy/envoy-mobile/issues/2506

PR #2529, opened 2022-09-07
changing   kotlin baseline to test HTTP

alyssawilk:
This changes the kotlin baseline app to test HTTP requests rather than HTTPS
To do this we disable the async filter and buffer filter, as they do not handle responses without bodies properly.

Risk Level: low
Testing: yes
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/test/kotlin/apps/baseline/MainActivity.kt
+++ b/test/kotlin/apps/baseline/MainActivity.kt
@@ -127,9 +127,7 @@ class MainActivity : Activity() {
         }
 
         if (status != 100) {
-          val fake_message = "received headers with status 200"
-          Log.d("MainActivity", message)
-          recyclerView.post { viewAdapter.add(Success(fake_message, headerText)) }
+          recyclerView.post { viewAdapter.add(Success(message, headerText)) }
         } else {
           recyclerView.post { viewAdapter.add(Failure(message)) }
         }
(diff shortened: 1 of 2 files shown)


Augustyniak (commented):
Let's update the description of the PR before we merge it.

test/kotlin/apps/baseline/MainActivity.kt
```diff
@@ -128,7 +126,7 @@ class MainActivity : Activity() {
           Log.d("MainActivity", "filter-demo: $filterDemoValue")
         }
 
-        if (status == 200) {
+        if (status != 100) {
```
Augustyniak, line 129: ah, I see - we want to show 301 requests as successes in the UI even though it does not matter from the perspective of the CI job itself - makes sense, that will make it easier for us by human beings who want to run these apps manually. 👍

In that case, can we have a check that looks as follows `status >= 200 && status <= 301` + add some comment on top of this code explaining that it's expected that `http://api.lyft.com/ping` returns us a 301? Just to make this code easier to understand for the potential reader (and future us 🙃)

alyssawilk: given only 301 will work I'm going to update to that, and comment why.

Augustyniak (approved):

Issue #2530, opened 2022-09-07
proposal: remove the h2 raw feature


Augustyniak:
Proposal: Remove the `addH2RawDomains` feature from Envoy engine builders.

Envoy engine builders allow its users to specify the list of h2 raw domains. For requests that go to domains defined as 'h2 raw domains' Envoy Mobile skips the ALPN negotiation and goes straight to enforcing h2 connections.

https://github.com/envoyproxy/envoy-mobile/blob/3b761c6ac51e04d6b645cb8de85eb5abd51a86bc/library/swift/EngineBuilder.swift#L290

https://github.com/envoyproxy/envoy-mobile/blob/3b761c6ac51e04d6b645cb8de85eb5abd51a86bc/library/swift/EngineBuilder.swift#L290

The feature was introduced originally back when Lyft was trying to address performance issues related to a potentially unlimited number of connections being created in response to application layer starting an unlimited number of requests to a new domain in a short period of time. Enforcing HTTP/2 for domains that we had a full control over allowed us to ensure that the number of connections was capped and kept at a reasonably low number. Soon after we added support for h2 raw domains @alyssawilk introduced an option for us to specify the maximum number of connections on per host basis. Envoy Mobile quickly adopted this feature as most apps either do not use HTTP/2 domains exclusively or do not have control over _all_ domains their applications talks to.

TL;DR
If an experiment - mentioned below in *cons* section - confirms that h2 raw can be disabled at Lyft without degrading any of our business metrics we will disable it. Since there are no other companies that currently depend on the feature or are planning to use it in the near future we are going to remove the feature from the engine to not have to support it anymore (it's not free) and to make it easier for ourselves to start using stats coming from Envoy's virtual clusters (more about this in *pros* section).

Detailed discussion of the pros and cons of the removal of h2 raw domains feature
Pros:
- Simpler code, less code is simpler code. The maintenance of EM's public API becomes cheaper.
- Simpler setup of Envoy's virtual clusters. In fact, the https://github.com/envoyproxy/envoy-mobile/pull/2088 that introduced the h2 raw domains feature broken (disabled) the emission of stats coming from Envoy Mobile virtual clusters. That's because it renamed the virtual host from `api` to another name and increase the number of virtual clusters from 1 to 2 without updating the list of enabled EM stats in https://github.com/envoyproxy/envoy-mobile/blob/main/library/common/config/config.cc#L486-L487.
[...]

PR #2548, opened 2022-09-13
config: update stats regex

Augustyniak:
Description: The change is needed for EM to emit virtual cluster stats. It used to work but was broken as part of https://github.com/envoyproxy/envoy-mobile/issues/2530 where we increase the number of virtual hosts from 1 to 2 and changed the name of the existing virtual host (which used to be called `api`).
Risk Level: Low, enables additional instrumentation
Testing: Manual using admin interface and curling `stats` endpoint.
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/library/common/config/config.cc
+++ b/library/common/config/config.cc
@@ -484,7 +484,7 @@ stats_config:
         - safe_regex:
             regex: '^pulse.*'
         - safe_regex:
-            regex: '^vhost.*.vcluster\.[\w]+?\.upstream_rq_(?:[12345]xx|retry.*|time|timeout|total)'
+            regex: '^vhost\.[\w]+\.vcluster\.[\w]+?\.upstream_rq_(?:[12345]xx|retry.*|time|timeout|total)'
   use_all_default_tags:
     false
 watchdogs:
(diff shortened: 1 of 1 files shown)


library/common/config/config.cc
```diff
@@ -484,7 +484,7 @@ stats_sinks: *stats_sinks
         - safe_regex:
             regex: '^pulse.*'
         - safe_regex:
-            regex: '^vhost.api.vcluster\.[\w]+?\.upstream_rq_(?:[12345]xx|retry.*|time|timeout|total)'
+            regex: '^vhost.*.vcluster\.[\w]+?\.upstream_rq_(?:[12345]xx|retry.*|time|timeout|total)'
```
jpsim, line 487: This will match `vhostXYZfoo.vcluster..." is that intended?

Should we match a period before the vhost name?

Maybe something like this? `^vhost\.\w+\.vcluster`

Augustyniak: Good points - fixed! Thank you!

jpsim:
It'd be nice to find a way to unit test this.

jpsim (approved):

Issue #1622, opened 2021-07-22
Cronvoy: support HTTP Proxy


carloseltuerto:
Cronet supports HTTP Proxy, and the [public API reports](https://source.chromium.org/chromium/chromium/src/+/main:components/cronet/android/api/src/org/chromium/net/UrlResponseInfo.java;l=102) it.

Note: inside Google, there is only one production call to ```UrlResponseInfo.getProxyServer()```. This is kind of expected since this is informational only.

alyssawilk:
definitely only when needed.  don't want to lose HTTP/2 or HTTP/3 for the common case!

mattklein123:
OK sounds good. I guess my question is how are we doing to know when to route to the proxy? I know that @Augustyniak has been looking into this a lot lately to make sure we exclude people that require proxies as that is totally broken today. Probably worth syncing up with him if you haven't already.

PR #2550, opened 2022-09-14
proxy monitor async version

### LINKED ISSUE #2532: android proxy: add support for proxies with host defined with the use of a domain
Proxy settings on Android consist of a `host` and a `port`. `Host` can be either an IP address or a URL host such as `api.lyft.com `. This issue is about adding support for the latter types of hosts. Supporting them will require Envoy Mobile to perform a DNS resolution to learn the address the library needs to connect to.

alyssawilk:
modifying the NetworkConfigurationFilter to do asynchronous DNS lookups for non-IP-address proxies.
If address resolution fails, the request will be short circuited by sendLocalReply rather than being sent upstream without being proxied.

This PR also changes ConnectivityManager to have a pure API and an Impl for testing.

Risk Level: low
Testing: unit tests, integration test
Docs Changes: n/a
Release Notes: n/a
part of https://github.com/envoyproxy/envoy-mobile/issues/1622 
 fixes #2532

### DIFF
--- a/library/common/extensions/filters/http/network_configuration/filter.cc
+++ b/library/common/extensions/filters/http/network_configuration/filter.cc
@@ -58,6 +58,7 @@ NetworkConfigurationFilter::decodeHeaders(Http::RequestHeaderMap& request_header
     return Http::FilterHeadersStatus::Continue;
   }
 
+  ENVOY_LOG(trace, "netconf_filter_processing_proxy_for_request", proxy_settings->asString());
   // If there is a proxy with a raw address, set the information, and continue.
   const auto proxy_address = proxy_settings->address();
   if (proxy_address != nullptr) {
--- a/library/common/extensions/filters/http/network_configuration/filter.cc
+++ b/library/common/extensions/filters/http/network_configuration/filter.cc
@@ -32,15 +32,25 @@ void NetworkConfigurationFilter::setDecoderFilterCallbacks(
 
 void NetworkConfigurationFilter::onLoadDnsCacheComplete(
     const Common::DynamicForwardProxy::DnsHostInfoSharedPtr& host_info) {
-  dns_cache_handle_.reset();
+  if (onAddressResolved(host_info)) {
+    continue_decoding_callback_ = decoder_callbacks_->dispatcher().createSchedulableCallback(
+        [this]() { decoder_callbacks_->continueDecoding(); });
+    continue_decoding_callback_->scheduleCallbackNextIteration();
+    return;
+  }
+}
+
+bool NetworkConfigurationFilter::onAddressResolved(
+    const Common::DynamicForwardProxy::DnsHostInfoSharedPtr& host_info) {
   if (host_info->address()) {
     setInfo(decoder_callbacks_->streamInfo().getRequestHeaders()->getHostValue(),
             host_info->address());
-    decoder_callbacks_->continueDecoding();
+    return true;
   }
   decoder_callbacks_->sendLocalReply(Http::Code::BadRequest,
                                      "Proxy configured but DNS resolution failed", nullptr,
                                      absl::nullopt, "no_dns_address_for_proxy");
+  return false;
 }
 
 Http::FilterHeadersStatus
@@ -94,8 +104,11 @@ NetworkConfigurationFilter::decodeHeaders(Http::RequestHeaderMap& request_header
 
   // If the hostname is in cache, set the info and continue.
   if (result.host_info_.has_value()) {
-    onLoadDnsCacheComplete(*result.host_info_);
-    return Http::FilterHeadersStatus::Continue;
+    if (onAddressResolved(*result.host_info_)) {
+      return Http::FilterHeadersStatus::Continue;
+    } else {
+      return Http::FilterHeadersStatus::StopIteration;
+    }
   }
 
   // If DNS lookup straight up fails, fail the request.
--- a/library/common/extensions/filters/http/network_configuration/filter.cc
+++ b/library/common/extensions/filters/http/network_configuration/filter.cc
@@ -30,34 +30,87 @@ void NetworkConfigurationFilter::setDecoderFilterCallbacks(
   decoder_callbacks_->addUpstreamSocketOptions(options);
 }
 
+void NetworkConfigurationFilter::onLoadDnsCacheComplete(
+    const Common::DynamicForwardProxy::DnsHostInfoSharedPtr& host_info) {
+  dns_cache_handle_.reset();
+  if (host_info->address()) {
+    setInfo(decoder_callbacks_->streamInfo().getRequestHeaders()->getHostValue(),
+            host_info->address());
+    decoder_callbacks_->continueDecoding();
+  }
+  decoder_callbacks_->sendLocalReply(Http::Code::BadRequest,
+                                     "Proxy configured but DNS resolution failed", nullptr,
+                                     absl::nullopt, "no_dns_address_for_proxy");
+}
+
 Http::FilterHeadersStatus
 NetworkConfigurationFilter::decodeHeaders(Http::RequestHeaderMap& request_headers, bool) {
-  const auto proxy_settings = connectivity_manager_->getProxySettings();
+  ENVOY_LOG(debug, "NetworkConfigurationFilter::decodeHeaders", request_headers);
 
-  ENVOY_LOG(trace, "NetworkConfigurationFilter::decodeHeaders", request_headers);
+  const auto authority = request_headers.getHostValue();
+  if (authority.empty()) {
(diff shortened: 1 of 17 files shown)


alyssawilk:
@Augustyniak I'm going to let this sit until your PR lands, but putting it out as a draft now just for reference

Original commit is your PR
Next is adding C++ unit tests to your PR
Finally making everything async, and fixing up tests for the new code paths.

alyssawilk:
I think the new kotlin tests passed, and most of the failures look like fetch failures, so the code is ready for at least a first pass when you have a few!

Augustyniak:
> I think the new kotlin tests passed, and most of the failures look like fetch failures, so the code is ready for at least a first pass when you have a few!

Great! Will take a look first thing in the morning tomorrow.

library/common/network/proxy_settings.h
```diff
@@ -62,6 +70,9 @@ struct ProxySettings {
     if (address_ != nullptr) {
       return address_->asString();
     }
+    if (!hostname_.empty()) {
```
Augustyniak, line 73: we need to update the implementation of `==` operator in this file too. Some logic in connectivity_manager does depend on it being accurate.

library/common/extensions/filters/http/network_configuration/filter.cc
```diff
@@ -30,34 +30,101 @@ void NetworkConfigurationFilter::setDecoderFilterCallbacks(
...
+  if (!connectivity_manager_->dnsCache()) {
+    decoder_callbacks_->sendLocalReply(Http::Code::BadRequest,
+                                       "Proxy configured but no DNS cache available", nullptr,
+                                       absl::nullopt, "no_dns_cache_for_proxy");
+    return Http::FilterHeadersStatus::StopIteration;
+  }
+
+  // Attempt to load the proxy's hostname from the DNS cache.
+  auto result =
+      connectivity_manager_->dnsCache()->loadDnsCacheEntry(proxy_settings->hostname(), 80, *this);
```
Augustyniak, line 96: So we have a hostname that's a composition of a `hostname + port` and we need to specify port 80 too?

library/common/extensions/filters/http/network_configuration/filter.cc
```diff
@@ -30,34 +30,101 @@ void NetworkConfigurationFilter::setDecoderFilterCallbacks(
...
+
+  // If the hostname is not in the cache, pause filter iteration and wait for onLoadDnsCacheComplete
+  // to be called.
+  if (result.status_ == Common::DynamicForwardProxy::DnsCache::LoadDnsCacheEntryStatus::Loading) {
+    dns_cache_handle_ = std::move(result.handle_);
+    return Http::FilterHeadersStatus::StopAllIterationAndWatermark;
```
Augustyniak, line 102: how will be the filter resumed once the DNS results become available?

library/common/network/proxy_settings.h
```diff
@@ -52,6 +53,13 @@ struct ProxySettings {
    */
   const Envoy::Network::Address::InstanceConstSharedPtr& address() const { return address_; }
 
+  /**
+   * @brief Returns the hostname of a proxy.
+   *
+   * @return Hostname of a proxy or the empty string if there is no hostname.
+   */
+  const std::string& hostname() const { return hostname_; }
```
Augustyniak, line 61: Per my above comment - should we mention that this contains a port too?

test/kotlin/integration/proxying/PerformHTTPSRequestUsingAsyncProxyTest.kt
```diff
@@ -0,0 +1,107 @@
...
+//                                                │                  │
+//                                                └──────────────────┘
+@RunWith(RobolectricTestRunner::class)
+class PerformHTTPSRequestUsingProxy {
+  init {
+    JniLibrary.loadTestLibrary()
+  }
+
+  @Test
+  fun `performs an HTTPs request through a proxy`() {
```
Augustyniak, line 49: let's rename `performs an HTTPs request through a proxy using an async DNS resolution`

test/kotlin/integration/proxying/PerformHTTPSRequestBadHostname.kt
```diff
@@ -0,0 +1,105 @@
...
+// ┌─────────────────────────┐                  ┌─┼─►listener_proxy│ │
+// │https://api.lyft.com/ping│  ┌──────────────┬┘ │ └──────┬───────┘ │ ┌────────────┐
+// │         Request         ├──►Android Engine│  │        │         │ │api.lyft.com│
+// └─────────────────────────┘  └──────────────┘  │ ┌──────▼──────┐  │ └──────▲─────┘
+//                                                │ │cluster_proxy│  │        │
+//                                                │ └─────────────┴──┼────────┘
+//                                                │                  │
+//                                                └──────────────────┘
+@RunWith(RobolectricTestRunner::class)
+class PerformHTTPSRequestUsingProxy {
```
Augustyniak, line 43: let's rename it so that it matches the name of the file

test/kotlin/integration/proxying/PerformHTTPSRequestBadHostname.kt
```diff
@@ -0,0 +1,105 @@
...
+  init {
+    JniLibrary.loadTestLibrary()
+  }
+
+  @Test
+  fun `performs an HTTPs request through a proxy`() {
```
Augustyniak, line 49: let's rename `performs an HTTPs request through a proxy using an async DNS resolution that fails`

Augustyniak (commented):
Nice!

Left a few comments/suggestions but overall it looks good

library/common/network/proxy_settings.h
```diff
@@ -71,14 +78,14 @@ struct ProxySettings {
...
-      return hostname_;
+      return absl::StrCat(hostname_, ":", port_);
     }
     return "no_proxy_configured";
   }
 
   bool operator==(ProxySettings const& rhs) const {
     if (this->address() == nullptr || rhs.address() == nullptr) {
-      return this->address() == nullptr && rhs.address() == nullptr;
+      return this->hostname() == rhs.hostname() && this->port() == rhs.port();
```
Augustyniak, line 88: that's not always correct - 

lhs: address = 192.168.0.1, host = "", port = 0
rhs: address = nullptr, host = "", port = ""

As is, the function returns `true` even though it should return `false`. I think that your conditions need to be && with `this->address() == rhs.address()`.

let's add a unit case for this particular case.

library/common/extensions/filters/http/network_configuration/filter.cc
Augustyniak: thank you - I understand the flow now

library/common/extensions/filters/http/network_configuration/filter.cc
```diff
@@ -30,34 +30,101 @@ void NetworkConfigurationFilter::setDecoderFilterCallbacks(
...
+  // If DNS lookup straight up fails, fail the request.
+  decoder_callbacks_->sendLocalReply(Http::Code::BadRequest,
+                                     "Proxy configured but DNS resolution failed", nullptr,
+                                     absl::nullopt, "no_dns_address_for_proxy");
+  return Http::FilterHeadersStatus::StopIteration;
+}
+
+void NetworkConfigurationFilter::setInfo(absl::string_view authority,
+                                         Network::Address::InstanceConstSharedPtr address) {
+  ENVOY_LOG(trace, "set_request_proxy_info", authority, address->asString());
```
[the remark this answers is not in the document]
Augustyniak: ping

library/common/extensions/filters/http/network_configuration/filter.cc
Augustyniak: Just checking - is it fine to keep the port hardcoded to 80 in here?

library/common/network/proxy_settings.h
alyssawilk: Added the test case.  It passes because one of the addresses is null but the hostnames are then 
192.168.0.1 and "" respectively.
Honestly we could just always compare pure hostname and skip the address bits for clarity if you'd like?

Augustyniak: >Added the test case. It passes because one of the addresses is null but the hostnames are then
192.168.0.1 and "" respectively.
Ah, I see - my bad.

>Honestly we could just always compare pure hostname and skip the address bits for clarity if you'd like?
Good idea. Let's do that

alyssawilk:
disabling that test with a TODO - I'll fix it in a follow-up once the remote address change is merged.

Augustyniak (approved):
Great progress!

PR #2551, opened 2022-09-14
key_value: Use known registered name to create global PlatformInterfaceImpl

abeyad:
The PlatformKeyValueStoreFactory used to take a PlatformInterfaceImpl where the registered KV store was found in the API registry and used to bridge the PlatformKeyValueStore with the native underlying store (e.g. SharedPrefs on Android).

However, registered factory classes only have their default constructor invoked, so the current PlatformKeyValueStoreFactory constructor taking in the PlatformInterfaceImpl would not have worked.

Instead, we use a globally known registered API name to fetch the registered KV store, and use it to create a static global PlatformInstanceImpl.

### DIFF
--- a/library/common/extensions/key_value/platform/config.cc
+++ b/library/common/extensions/key_value/platform/config.cc
@@ -12,7 +12,7 @@ namespace Envoy {
 namespace Extensions {
 namespace KeyValue {
 
-class PlatformInterfaceImpl : PlatformInterface, public Logger::Loggable<Logger::Id::filter> {
+class PlatformInterfaceImpl : public PlatformInterface, public Logger::Loggable<Logger::Id::filter> {
 public:
   PlatformInterfaceImpl(const std::string& name)
       : bridged_store_(*static_cast<envoy_kv_store*>(Api::External::retrieveApi(name))) {}
@@ -35,13 +35,13 @@ private:
   envoy_kv_store bridged_store_;
 };
 
-PlatformKeyValueStore::PlatformKeyValueStore(Event::Dispatcher& dispatcher,
-                                             std::chrono::milliseconds save_interval,
-                                             PlatformInterface& platform_interface,
-                                             uint64_t max_entries, const std::string& key)
+PlatformKeyValueStore::PlatformKeyValueStore(
+    Event::Dispatcher& dispatcher, std::chrono::milliseconds save_interval,
+    std::unique_ptr<PlatformInterface>&& platform_interface, uint64_t max_entries,
+    const std::string& key)
     : KeyValueStoreBase(dispatcher, save_interval, max_entries),
-      platform_interface_(platform_interface), key_(key) {
-  const std::string contents = platform_interface.read(key);
+      platform_interface_(std::move(platform_interface)), key_(key) {
+  const std::string contents = platform_interface_->read(key);
   if (!parseContents(contents)) {
     ENVOY_LOG(warn, "Failed to parse key value store contents {}", key);
   }
@@ -53,7 +53,7 @@ void PlatformKeyValueStore::flush() {
     absl::StrAppend(&output, it.first.length(), "\n", it.first, it.second.length(), "\n",
                     it.second);
   }
-  platform_interface_.save(key_, output);
+  platform_interface_->save(key_, output);
 }
 
 KeyValueStorePtr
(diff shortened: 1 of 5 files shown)


abeyad:
cc @alyssawilk @goaway

library/common/extensions/key_value/platform/config.cc
```diff
@@ -63,14 +64,16 @@ PlatformKeyValueStoreFactory::createStore(const Protobuf::Message& config,
...
       typed_config.config().typed_config(), validation_visitor);
-  auto milliseconds =
-      std::chrono::milliseconds(DurationUtil::durationToMilliseconds(file_config.save_interval()));
-  return std::make_unique<PlatformKeyValueStore>(dispatcher, milliseconds,
-                                                 platform_interface_.value().get(),
-                                                 file_config.max_entries(), file_config.key());
+  auto milliseconds = std::chrono::milliseconds(
+      DurationUtil::durationToMilliseconds(platform_kv_store_config.save_interval()));
+  std::unique_ptr<PlatformInterface> platform_interface =
+      std::make_unique<PlatformInterfaceImpl>(platform_kv_store_config.kv_store_name());
```
alyssawilk, line 73: I think I need Mike to look at this, but it looks like this gets passed to 
*static_cast<envoy_kv_store*>(Api::External::retrieveApi(name))
and there's a second key used for read and write
so my mental model of this (which could be 100% wrong) would be that we should be calling registerAPI early on in the engine as we do for the event tracker, and we'd register it with a global well known name, and then the individual proto config keys would be passed into the read/write calls.

abeyad: If we use a single global well-known name, then we can only use the PlatformKeyValueStore for one purpose, not multiple purposes.  Enabling the registered KV store to be specified in the PlatformKeyValueStoreConfig allows the PlatformKeyValueStore to be used for multiple purposes.

builder.addKeyValueStore("dns_cache", new SharedPreferencesStore())
builder.addKeyValueStore("xds_cache", new SharedPreferencesStore())

Then for DNS cache, the PlatformKeyValueStoreConfig would have `kv_store_name` set to `dns_cache` and for the xDS cache, the PlatformKeyValueStoreConfig would have `kv_store_name` set to `xds_cache`.

alyssawilk: "If we use a single global well-known name, then we can only use the PlatformKeyValueStore for one purpose, not multiple purposes. "
Why?  I thought the name here was a handle on how one gets the KV store platform, and then later on we'd 
platform.write("dns_cache", dns_cache_contents)
platform.read("dns_cache")
platform.read("xds_cache")

abeyad: Alyssa and I chatted over VC, I changed the code to use a single globally-known name for the KV store.

library/common/extensions/key_value/platform/config.h
```diff
@@ -26,21 +26,21 @@ class PlatformInterface {
...
+                        std::unique_ptr<PlatformInterface>&& platform_interface,
+                        uint64_t max_entries, const std::string& key);
   // KeyValueStore
   void flush() override;
 
 private:
-  PlatformInterface& platform_interface_;
+  // TODO(alyssawilk, goaway) the default PlatformInterface should do up calls through Java and this
+  // can be moved to a non-optional reference.
+  std::unique_ptr<PlatformInterface> platform_interface_;
```
alyssawilk, line 37: If we're making this optional, I think we want to error handle it being null (ENVOY_BUG a warning), WDYT?

abeyad: Changed to a non-optional static singleton.

library/common/extensions/key_value/platform/config.h
```diff
@@ -32,15 +32,15 @@ class PlatformKeyValueStore : public KeyValueStoreBase {
   void flush() override;
 
 private:
+  // TODO(alyssawilk, goaway) the default PlatformInterface should do up calls through Java and this
+  // can be moved to a non-optional reference.
```
alyssawilk, line 36: this is a non-optional reference :-)
Maybe remove this TODO and add one by Api::External::retrieveApi to gracefully handle the API not being registered?

alyssawilk (commented):
One nit, otherwise LGTM.  
Are the test failures unrelated or are there other things which need to be fixed up?

library/common/extensions/key_value/platform/config.h
abeyad: Oops yes, done!

alyssawilk (approved):

abeyad:
> One nit, otherwise LGTM. Are the test failures unrelated or are there other things which need to be fixed up?

The test failures seemed unrelated (couldn't set up the iOS or Android environment) so I triggered a retest.  Previous runs had some of these tests "cancelled" instead of failed, so if it happens again, I'll need to dig into what's happening, though I don't believe it's related to the code changes in this PR

abeyad:
> > One nit, otherwise LGTM. Are the test failures unrelated or are there other things which need to be fixed up?
> 
> The test failures seemed unrelated (couldn't set up the iOS or Android environment) so I triggered a retest. Previous runs had some of these tests "cancelled" instead of failed, so if it happens again, I'll need to dig into what's happening, though I don't believe it's related to the code changes in this PR

The failures don't seem to be related to this PR but persist nevertheless.  Issues with `IDESimulatorAvailability` and tests timing out: https://envoy.cluster.engflow.com/invocation/e1152648-4588-4767-a7eb-e6447396cedd.

goaway (approved):
LGTM - I like the 'reserved' namespace for the hardcoded string.

PR #2554, opened 2022-09-16
android: fix JNI call to AndroidNetworkLibrary

danzh2010:
Signed-off-by: Dan Zhang <danzh@google.com>

Description: address https://github.com/envoyproxy/envoy-mobile/pull/2516#issuecomment-1237169849 to use find_class() instead of FindClass() as AndroidNetworkLibrary is defined in Envoy Mobile repo.
Risk Level: low
Testing: also fixed existing unit tests
Docs Changes: N/A
Release Notes: N/A


danzh2010:
@Augustyniak for review

test/java/org/chromium/net/CertificateVerificationTest.java
```diff
@@ -23,7 +24,8 @@
 public final class CertificateVerificationTest {
   static {
     AndroidJniLibrary.loadTestLibrary();
```
Augustyniak, line 26: I think that we should make it harder for folks to forget to call both `AndroidJniLibrary.load` and `AndroidJniLibrary.loadTestLibrary` - it's a lot of things to remember. Can we change the signature and implementation of `AndroidJniLibrary.loadTestLibrary` method so that:

1. It takes `context` as its argument
1. It calls `AndroidJniLibrary.load` internally.

That would allow us to simplify our test code. What do you think?

library/common/jni/jni_utility.cc
```diff
@@ -299,6 +302,8 @@ jbyteArray ToJavaByteArray(JNIEnv* env, const std::string& str) {
 
 void JavaArrayOfByteArrayToStringVector(JNIEnv* env, jobjectArray array,
                                         std::vector<std::string>* out) {
+  ASSERT(out);
+  ASSERT(array);
```
Augustyniak, line 306: not sure whether this was added as part of this PR intentionally

library/common/jni/jni_utility.cc
```diff
@@ -20,7 +20,10 @@ JavaVM* get_vm() { return static_jvm; }
 
 void set_class_loader(jobject class_loader) { static_class_loader = class_loader; }
 
-jobject get_class_loader() { return static_class_loader; }
+jobject get_class_loader() {
+  ASSERT(static_class_loader);
```
Augustyniak, line 24: I wonder whether this should be a `RELEASE_ASSERT` - it should basically always crash, even in prod. That way we could find out about potential issues as soon as possible. What do you think?

Augustyniak (commented):
Thanks for fixing these! Left a few comments/ suggestions

test/java/org/chromium/net/CertificateVerificationTest.java
danzh2010: AndroidJniLibrary.load() is already called in AndroidEngineImpl constructor, and any tests which create an engine which is most of the tests should have already loaded the app context and won't need to do it twice. We missed it in this test because it is unit-testing the JNI calls. So RELEASE_ASSERT in `get_class_loader()` should be sufficient to avoid similar mistakes in the future.

library/common/jni/jni_utility.cc
danzh2010: done

library/common/jni/jni_utility.cc
danzh2010: They are intentional. I just found neither the caller nor callee check these.

test/java/org/chromium/net/CertificateVerificationTest.java
Augustyniak: OK, that makes sense (the fact that we call this method from within the Android Engine builder anyway). Let's leave it the way it is and revisit if it turns out that calling 2 methods as opposed to 1 will be confusing to people.

Augustyniak (approved):

danzh2010:
@Augustyniak mind merging it?

PR #2563, opened 2022-09-20
support configuring multiple stat sinks

snowp:
This introduces a `addExtraStatSinks(List<String>)` function to the builders, allowing for the injection of arbitrary stat sink configuration to use. The function accepts a list of YAML strings, which allows callers to inject any stat sink compiled into the Envoy configuration.

The motivation for this YAML-based API is to support injecting stat sinks that aren't available in the OSS.

One plus of this is that we are able to deprecate the statsd API on the JVM side (only used for test) and instead use this more flexible API.

Description:
Risk Level:
Testing:
Docs Changes:
Release Notes:
[Optional Fixes #Issue]
[Optional Deprecated:]

### DIFF
--- a/library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
+++ b/library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
@@ -91,7 +91,7 @@ open class EngineBuilder(
   }
 
   /**
-   * Specifies the the domain (e.g. `example.com`) to use in the default gRPC stat sink to flush
+   * Specifies the domain (e.g. `example.com`) to use in the default gRPC stat sink to flush
    * stats.
    *
    * Setting this value enables the gRPC stat sink, which periodically flushes stats via the gRPC
--- a/library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
+++ b/library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
@@ -76,7 +76,7 @@ open class EngineBuilder(
   private var nativeFilterChain = mutableListOf<EnvoyNativeFilterConfig>()
   private var stringAccessors = mutableMapOf<String, EnvoyStringAccessor>()
   private var keyValueStores = mutableMapOf<String, EnvoyKeyValueStore>()
-  private var statSinks = listOf<String>()
+  private var statsSinks = listOf<String>()
 
   /**
    * Add a log level to use with Envoy.
@@ -91,7 +91,8 @@ open class EngineBuilder(
   }
 
   /**
-   * Specifies the the domain (e.g. `example.com`) to use in the default gRPC stat sink to flush stats.
+   * Specifies the the domain (e.g. `example.com`) to use in the default gRPC stat sink to flush
+   * stats.
    *
    * Setting this value enables the gRPC stat sink, which periodically flushes stats via the gRPC
    * MetricsService API. The flush interval is specified via addStatsFlushSeconds.
@@ -114,8 +115,8 @@ open class EngineBuilder(
    *
    * @return this builder.
    */
-  fun addStatsSinks(statSinks: List<String>): EngineBuilder {
-    this.statSinks = statSinks
+  fun addStatsSinks(statsSinks: List<String>): EngineBuilder {
+    this.statsSinks = statsSinks
     return this
   }
 
@@ -648,7 +649,7 @@ open class EngineBuilder(
       platformFilterChain,
       stringAccessors,
       keyValueStores,
-      statSinks
+      statsSinks
     )
 
     return when (configuration) {
--- a/library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
+++ b/library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
@@ -43,7 +43,6 @@ open class EngineBuilder(
   private var logLevel = LogLevel.INFO
   private var adminInterfaceEnabled = false
   private var grpcStatsDomain: String? = null
-  private var statsDPort: Int? = null
   private var connectTimeoutSeconds = 30
   private var dnsRefreshSeconds = 60
   private var dnsFailureRefreshSecondsBase = 2
@@ -77,6 +76,7 @@ open class EngineBuilder(
   private var nativeFilterChain = mutableListOf<EnvoyNativeFilterConfig>()
   private var stringAccessors = mutableMapOf<String, EnvoyStringAccessor>()
   private var keyValueStores = mutableMapOf<String, EnvoyKeyValueStore>()
+  private var statSinks = listOf<String>()
 
   /**
    * Add a log level to use with Envoy.
@@ -91,12 +91,12 @@ open class EngineBuilder(
   }
 
   /**
-   * Add a domain to flush stats to.
-   * Passing nil disables stats emission via the gRPC stat sink.
+   * Specifies the the domain (e.g. `example.com`) to use in the default gRPC stat sink to flush stats.
    *
-   * Only one of the statsd and gRPC stat sink can be enabled.
+   * Setting this value enables the gRPC stat sink, which periodically flushes stats via the gRPC
+   * MetricsService API. The flush interval is specified via addStatsFlushSeconds.
    *
-   * @param grpcStatsDomain The domain to use for stats.
+   * @param grpcStatsDomain The domain to use for the gRPC stats sink.
    *
    * @return this builder.
    */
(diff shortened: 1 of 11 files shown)


library/objective-c/EnvoyConfiguration.m
```diff
@@ -198,11 +202,29 @@ - (nullable NSString *)resolveTemplate:(NSString *)templateYAML {
...
     [definitions appendString:@"- &stats_sinks [ *base_metrics_service ]\n"];
+    [stat_sinks addObject:@"*base_metrics_serivce"];
+  }
+
+  bool first = true;
+  if ([stat_sinks count] > 0) {
+    [definitions appendString:@"- &stats_sinks ["];
+
+    for (id sink in stat_sinks) {
+      if (!first) {
```
Augustyniak, line 222: this is never true?

Augustyniak: Can we check whether `definitions` is empty instead of using a  `first` variable?

test/kotlin/integration/StatFlushIntegrationTest.kt
```diff
@@ -49,9 +49,12 @@ class StatFlushIntegrationTest {
     val countDownLatch = CountDownLatch(1)
     engine = EngineBuilder()
       .addLogLevel(LogLevel.DEBUG)
-      .addStatsDPort(8125)
       // Really high flush interval so it won't trigger during test execution.
       .addStatsFlushSeconds(100)
+      .addExtraStatSinks(listOf("""{ name: envoy.stat_sinks.statsd,
```
Augustyniak, line 54: any chance that we could have a similar test for Swift?

library/objective-c/EnvoyConfiguration.m
Augustyniak: actually, I know that you asked me for this but I just realized that objective-c has an API for joining all elements of an array with a given separator. Try:
```
[array componentsJoinedByString:@","]
```

Augustyniak (commented):
I do not have a strong opinion but I think that we should change the `addExtraStatsSinks` naming to `addStatsSinks`. Word `add` implies that we are adding stuff and the possibility of an existence of some kind of a default/base stats sink.

library/objective-c/EnvoyConfiguration.m
```diff
@@ -1,4 +1,6 @@
 #import "library/objective-c/EnvoyEngine.h"
+#include <Foundation/NSArray.h>
+#include <objc/objc.h>
```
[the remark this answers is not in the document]
snowp: auto-import thanks to VS code, let me remove

test/kotlin/integration/StatFlushIntegrationTest.kt
snowp: I can do one that just initializes one with the statsd sink without any extra tests just to see if the configuration bits work right

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -91,12 +91,12 @@ open class EngineBuilder(
   }
 
   /**
-   * Add a domain to flush stats to.
-   * Passing nil disables stats emission via the gRPC stat sink.
+   * Specifies the the domain (e.g. `example.com`) to use in the gRPC stat sink to flush stats.
```
Augustyniak, line 94:
```suggestion
   * Specifies the the domain (e.g. `example.com`) to use in the default gRPC stat sink to flush stats.
```

library/objective-c/EnvoyConfiguration.m
```diff
@@ -198,11 +200,20 @@ - (nullable NSString *)resolveTemplate:(NSString *)templateYAML {
...
+
   if (self.grpcStatsDomain != nil) {
     [definitions appendFormat:@"- &stats_domain %@\n", self.grpcStatsDomain];
-    [definitions
-        appendFormat:@"- &stats_flush_interval %lus\n", (unsigned long)self.statsFlushSeconds];
-    [definitions appendString:@"- &stats_sinks [ *base_metrics_service ]\n"];
+    [stat_sinks_config addObject:@"*base_metrics_serivce"];
+  }
+
+  if ([stat_sinks_config count] > 0) {
```
Augustyniak, line 213:
```suggestion
  if (stat_sinks_config.count > 0) {
```
Missed previously - dot accessor is a preferred one for things that are implemented as properties (even though your implementation works just fine).

library/objective-c/EnvoyConfiguration.m
```diff
@@ -198,11 +200,20 @@ - (nullable NSString *)resolveTemplate:(NSString *)templateYAML {
...
+    [definitions appendString:@"- &stats_sinks ["];
+    [definitions appendString:[stat_sinks_config componentsJoinedByString:@","]];
+    [definitions appendString:@"] \n"];
```
Augustyniak, line 216:
```suggestion
    [definitions appendString:@"]\n"];
```

library/swift/EngineBuilder.swift
```diff
@@ -77,6 +78,19 @@ open class EngineBuilder: NSObject {
     return self
   }
 
+  /// Adds additional stats sink, in the form of the raw YAML/JSON configuration.
+  /// Sinks added in this fashion will be included in addition to the gRPC stats sink
+  /// that may be configured via addGrpcStatsDomain.
```
Augustyniak, line 83:
```suggestion
  /// that may be enabled via addGrpcStatsDomain.
```

Augustyniak: optional nit

library/swift/EngineBuilder.swift
```diff
@@ -77,6 +78,19 @@ open class EngineBuilder: NSObject {
...
+  ///
+  /// - parameter statsSinks: Array of stat sink configurations to include.
```
Augustyniak, line 85:
```suggestion
  /// - parameter statsSinks: Configurations of stat sinks to add.
```
optional

test/kotlin/integration/StatFlushIntegrationTest.kt
```diff
@@ -44,14 +44,29 @@ class StatFlushIntegrationTest {
     }
   }
 
+  @Test
+  fun `multipe stat sinks configured`() {
```
Augustyniak, line 48:
```suggestion
  fun `multiple stat sinks configured`() {
```

test/swift/integration/StatFlushIntegrationTest.swift
```diff
@@ -29,4 +29,33 @@ final class StatFlushIntegrationTest: XCTestCase {
...
+      .addGrpcStatsDomain("example.com")
+      .addStatsSinks(
+        [statsdSinkConfig(port: 1234), statsdSinkConfig(port: 5555)]
+      )
+      .setOnEngineRunning {
+        engineExpectation.fulfill()
+      }
+      .build()
+
+    XCTAssertEqual(XCTWaiter.wait(for: [engineExpectation], timeout: 10), .completed)
```
Augustyniak, line 48: let's call `engine.terminate()` at the end  to be consistent with other tests.

test/swift/integration/StatFlushIntegrationTest.swift
```diff
@@ -29,4 +29,33 @@ final class StatFlushIntegrationTest: XCTestCase {
...
+
+  func testMultipleStatSinks() throws {
+    let engineExpectation = self.expectation(description: "Engine Running")
+
+    let engine = EngineBuilder()
+      .addLogLevel(.debug)
+      .addStatsFlushSeconds(1)
+      .addGrpcStatsDomain("example.com")
+      .addStatsSinks(
+        [statsdSinkConfig(port: 1234), statsdSinkConfig(port: 5555)]
```
Augustyniak, line 41: looks so nice :)

test/kotlin/integration/StatFlushIntegrationTest.kt
```diff
@@ -44,14 +44,29 @@ class StatFlushIntegrationTest {
...
+
   @Test
   fun `flush flushes to stats sink`() {
     val countDownLatch = CountDownLatch(1)
     engine = EngineBuilder()
       .addLogLevel(LogLevel.DEBUG)
-      .addStatsDPort(8125)
       // Really high flush interval so it won't trigger during test execution.
       .addStatsFlushSeconds(100)
+      .addStatSinks(listOf(statsdSinkConfig(8125), statsdSinkConfig(5000)))
```
Augustyniak, line 69: Could we update the assertion/expectation in this test so that it awaits multiple stats uploads?

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
Augustyniak, line 94:
```suggestion
   * Specifies the the domain (e.g. `example.com`) to use in the default gRPC stat sink to flush stats.
```

library/kotlin/io/envoyproxy/envoymobile/EngineBuilder.kt
```diff
@@ -606,6 +591,15 @@ open class EngineBuilder(
     return this
   }
 
+  /**
+   * Injects additional stat sink configurations. This should be a YAML/JSON formatted extension
```
Augustyniak, line 595: Make this doc string consistent with the Swift one (which currently has more information)

test/kotlin/integration/StatFlushIntegrationTest.kt
Augustyniak: that's better than nothing but is it possible to launch a mock server in the swift test (we have something like that in the Java/kotlin one) and see whether a metric is received by it? If that's easy to do then it should be considered to be outside of the scope of this PR

test/kotlin/integration/StatFlushIntegrationTest.kt
snowp: Yeah let me give that a go

snowp:
@Augustyniak this is green, ptal

test/kotlin/integration/StatFlushIntegrationTest.kt
snowp: Talked offline, not that easy to do since we need a statsd server written in Swift

Augustyniak (commented):
add a few nit comments. It looks great

Augustyniak (approved):

PR #2572, opened 2022-09-27
bazel: update rules_xcodeproj to 0.8.0

jpsim:
* Add Hello World Objective-C example apps to the generated project
* The internal unit test labels are no longer required
* There's now an easier way to add device and simulator targets
* We can remove `archived_bundles_allowed` which was deprecated

Finally instead of thousands of schemes to pick from, we only expose the top-level ones to make it easier to use:

image

Release info: https://github.com/buildbuddy-io/rules_xcodeproj/releases/tag/0.8.0


jpsim:
Depends on #2571 to fully work.

Augustyniak (approved):

bazel/framework_imports_extractor.bzl
```diff
@@ -36,7 +36,8 @@ def _framework_imports_extractor(ctx):
     ctx.actions.run_shell(
         inputs = [framework_zip],
         outputs = outputs,
-        command = "unzip -qq {} -d {}".format(framework_zip.path, ctx.bin_dir.path),
+        # Workaround for https://github.com/bazelbuild/rules_apple/issues/1489
+        command = "unzip -qq {} -d {} || true".format(framework_zip.path, ctx.bin_dir.path),
```
keith, line 40: If the unzip fails won't the required output files not be created

jpsim: In my experience the CRC failures cause the process to exit with a non-zero code, but the resulting files still match.

PR #2580, opened 2022-09-29
build: Upgrade Abseil library to July 2022 release

abeyad:
When enabling google_grpc in Envoy Mobile (which is off by default), the build fails because Google gRPC depends on
absl/functional:any_invocable, which is not part of the Abseil library that Envoy Mobile is on.

It turns out that Envoy Mobile's Abseil library is from a June 2021 release, so this commit upgrades the Abseil library used by Envoy Mobile to the same one used by Envoy, which is from the July 5, 2022 release.


bazel/envoy_mobile_repositories.bzl
```diff
@@ -52,10 +52,10 @@ def upstream_envoy_overrides():
     http_archive(
         name = "com_google_absl",
         patches = ["@envoy_mobile//bazel:abseil.patch"],
```
jpsim, line 54: Can we remove this patch now? Removes linking Foundation on macOS, but that should be safe? https://github.com/envoyproxy/envoy-mobile/blob/main/bazel/abseil.patch

abeyad: just pushed up a commit to try that out

jpsim: Ah it looks like https://github.com/abseil/abseil-cpp/issues/326 is still open and this could still be a problem.

jpsim (dismissed):

bazel/envoy_mobile_repositories.bzl
jpsim: If we can indeed remove the patch, that means we can remove this dep override altogether and implicitly use the one defined in Envoy.

abeyad: Given that the issue is still open, it would seem we should not remove the patch.  But I would assume CI would catch it?  (it's still running).  If we inherit from the Envoy dependencies by default and this is just overrides, then yes, would be great to remove this all together!  I'll wait to see if CI passes.  If it does, I'll remove both the patch and the dependency override.

jpsim: I'm a bit concerned that because we build the aar on a macOS host and run the app verifications on a macOS host too, that it could mask the issue that could show up if we ran the aar on an emulator on a non-Darwin host (i.e. Linux) or on a physical Android device.

abeyad: I think what I'll do is put the patch back in, and add a comment referencing the open GH issue, so people know to keep the patch around until the GH issue is resolved.

jpsim: There's already a link in a comment above, that's how I knew to check. But yes, let's be safe and re-add the patch.

abeyad: Done

jpsim (approved):

Issue #2581, opened 2022-09-29
Disable GuardDogImpl


jpsim:
https://github.com/envoyproxy/envoy/blob/main/source/server/guarddog_impl.cc

The purpose of this type is to terminate the current process (generally an Envoy server) if the system isn't healthy.

We should never do this in a mobile app setting, so we should just turn this off when building Envoy Mobile.

jpsim:
Envoy PR: https://github.com/envoyproxy/envoy/pull/23316

jpsim:
> Hmmm, I think the watch dog is currently responsible for actually running the main dispatch loop, so to disable it I'll actually need to run the main dispatch loop some other way.
> I've marked the Envoy PR as being a draft until I can do that.

PR #2582, opened 2022-09-29
build: disable guard dog

jpsim:
Commit Message: build: disable guard dog
Additional Description: Which Envoy Mobile should never enable, because terminating the Envoy process would mean terminating the host iOS or Android app. See https://github.com/envoyproxy/envoy-mobile/issues/2581 for more details.
Risk Level: Low, compile-time flag, no-op by default.
Testing: Building Envoy Mobile locally.
Docs Changes: None.
Release Notes: None.
Platform Specific Features: Will be disabled when building Envoy Mobile.
Fixes: https://github.com/envoyproxy/envoy-mobile/issues/2581

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -2,6 +2,7 @@
 try-import ./envoy/.bazelrc
 
 # Common flags for all builds
+build --@envoy//bazel:guarddog=False
 build --platform_mappings=envoy/bazel/platform_mappings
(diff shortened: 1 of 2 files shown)


jpsim:
Closing as stale now that we've migrated to the Envoy repo: https://github.com/envoyproxy/envoy/blob/main/mobile/README.md

PR #2590, opened 2022-10-03
api: remove h2 raw feature

Augustyniak:
Description: Remove the support for h2 raw domains. Fixes https://github.com/envoyproxy/envoy-mobile/issues/2530.
Risk Level: Low, remove features. Setting an HTTP header using `addUpstreamHttpProtocol` should allow users to continue to enforce the use of HTTP/2 as needed.
Testing: Integration tests in example apps.
Docs Changes: N/A 
Release Notes: Yes


library/common/config/config.cc
```diff
@@ -330,25 +329,7 @@ const char* config_template = R"(
...
-                route:
-                  cluster: base_h2
-                  timeout: 0s
-                  retry_policy:
-                    per_try_idle_timeout: *per_try_idle_timeout
-                    retry_back_off:
-                      base_interval: 0.25s
-                      max_interval: 60s
-            - name: primary
+            - name: api
```
jpsim, line 332: Should there be a comment here explaining why there needs to be a single virtual host, and that it must be named `api`?

Augustyniak: It does not need to named `api` anymore after we merged https://github.com/envoyproxy/envoy-mobile/pull/2548. Let me leave some comment with regard to the number of clusters.

Augustyniak: added a comment

jpsim: A single `R"()"` is now split in two with a C++ comment in between. Do these get concatenated?

I was thinking adding a YAML comment (`# comment...`) directly in the YAML string which doesn't have this risk, but I just don't know how C++ `R`-strings work here.

jpsim (approved):
If you're indeed sure that in C++ `R"()" // comment R"()"` concatenates the two string literals into one.

Augustyniak:
> If you're indeed sure that in C++ `R"()" // comment R"()"` concatenates the two string literals into one.

I am not sure to be honest but green tests make me confident enough that this is the case

Augustyniak:
I did some extra testing on Android to confirm that we were not creating too many connections for when we disable h2 raw. I tested the app in two scenarios - for when we have a `.addUpstreamHttpProtocol(UpstreamHttpProtocol.HTTP2)` and without that call.

Next, I made the app start 100 requests every 1 second. With happy eyeballs disabled I saw only 1 connection being created in both of the aforementioned testing scenarios and hundreds of requests finishing with 200 status code.

PR #2594, opened 2022-10-06
build: remove FlatBuffers integration

jpsim:
This was unused, and was causing issues migrating to Xcode 14, and recent releases aren't available on Maven.

When we have a compelling reason to re-introduce this dependency in the future, we can address these issues at that time.


Augustyniak (approved):
I think that this is a reasonable path forward that unlocks our migration to Xcode 14.0

Augustyniak:
At this point the merge of this PR is blocking the merge of the upstream Envoy bump https://github.com/envoyproxy/envoy-mobile/pull/2598 which includes a change that's supposed to help with the performance of proxied requests and blocks https://github.com/envoyproxy/envoy-mobile/pull/2602. The lack of https://github.com/envoyproxy/envoy-mobile/pull/2602 means a delay in the rollout of proxy support on Android which is something that I would like to avoid.

Let's sync on this tmrw and decide whether it's safe to merge all of the aforementioned PRs between the EOD on Tuesday.

jpsim:
After discussing with @Augustyniak today and other maintainers at last week's community call, I'm choosing to merge this now in order to unblock the Envoy and Xcode updates.

If anyone wants to re-integrate FlatBuffers again in the future, I'm happy to help address these integration issues at that time.

PR #2597, opened 2022-10-06
Update link to Slack invite

filip-doordash:
The old link (https://envoyslack.cncf.io) seems to be broken. 

Signed-off-by: Filip Busic <61715767+filip-doordash@users.noreply.github.com>


jpsim (approved):
Thanks!

Issue #2498, opened 2022-08-25
Simplify Envoy Config Generation


RyanTheOptimist:
The system that Envoy Mobile uses to generate Envoy's config is fairly complicated. It involves snippets of YAML defined as C strings in [library/common/config/config.cc](https://sourcegraph.com/github.com/envoyproxy/envoy-mobile/-/blob/library/common/config/config.cc). In order to avoid duplication, these snippets make use of YAML anchors and aliases. These YAML snippets are conditionally stitched together along with in Swift via [EnvoyConfiguration.m](https://sourcegraph.com/github.com/envoyproxy/envoy-mobile/-/blob/library/objective-c/EnvoyConfiguration.m) and in Java via [EnvoyConfiguration.java](https://sourcegraph.com/github.com/envoyproxy/envoy-mobile/-/blob/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java) and in C++ via [library/cc/engine_builder.h](https://sourcegraph.com/github.com/envoyproxy/envoy-mobile/-/blob/library/cc/engine_builder.h). 

It seems undesirable to have 3 different builders with different functionality. The Alternate Protocols cache filter is only supported in Java. Custom filters are not supported in the C++ builder.

In addition, the use of YAML anchor and aliases interacts a bit poorly with the builders. Some of the anchors are used by aliases in `config_header` which prevents them from being overridden by builder-supplied values. Of course, other anchor are overrideable and some are currently overrideable in some places but not in others.

I think we should simplify this. My strawman would be that config generation is implemented in a single place in (C or C++). The Java, Swift and C++ Builder would be responsible for calling into this single config generation method with whatever state is required to generate a valid config. (This includes numerous booleans and integers as well as a number of strings, lists and maps). However it is important that this not regress app startup so perhaps another approach would be warranted.

snowp:
I think it would be interesting to see how constructing the config from the C++ config compares to the current YAML code, I don't have a great sense of which one would scale better. 

[...]

PR #2601, opened 2022-10-10
Add a setPerTryIdleTimeoutSeconds method to EngineBuilder.

RyanTheOptimist:
Add a setPerTryIdleTimeoutSeconds method to EngineBuilder.
Add tests for setPerTryIdleTimeoutSeconds and setStreamIdleTimeoutSeconds.
Minor cleanup of envoy_config_test.cc.

Part of #2498 

Risk Level: Low
Testing: Added unit tests
Docs Changes: N/A
Release Notes: Updated version_history.txt

PR #2603, opened 2022-10-11
Add various methods to C++ EngineBuilder to bring it to parity with the Java and Obj-C builders

RyanTheOptimist:
Add various methods to C++ EngineBuilder to bring it to parity with the Java and Obj-C builders
-  addStatsSinks()
-  addDnsMinRefreshSeconds()
-  addMaxConnectionsPerHost()
-  enableAdminInterface()
-  enableHappyEyeballs()
-  enableHttp3()
-  enableInterfaceBinding()
-  enableDrainPostDnsRefresh()
-  enableH2ExtendKeepaliveTimeout()
-  enforceTrustChainVerification()

Part of #2498

Risk Level: Low
Testing: New unit tests
Docs Changes: N/A
Release Notes: Updated version_history.txt


RyanTheOptimist:
@Augustyniak there are still 5 more knobs to add but those are going to require a bit more work. In the meantime, I think this is progress.

library/cc/engine_builder.h
```diff
@@ -20,12 +22,15 @@ class EngineBuilder {
   EngineBuilder& addLogLevel(LogLevel log_level);
   EngineBuilder& setOnEngineRunning(std::function<void()> closure);
 
+  EngineBuilder& addStatsSinks(const std::vector<std::string>& stat_sinks);
   EngineBuilder& addGrpcStatsDomain(const std::string& stats_domain);
   EngineBuilder& addConnectTimeoutSeconds(int connect_timeout_seconds);
   EngineBuilder& addDnsRefreshSeconds(int dns_refresh_seconds);
   EngineBuilder& addDnsFailureRefreshSeconds(int base, int max);
   EngineBuilder& addDnsQueryTimeoutSeconds(int dns_query_timeout_seconds);
+  EngineBuilder& addDnsMinRefreshSeconds(int dns_min_refresh_seconds);
```
Augustyniak, line 31: related to https://github.com/envoyproxy/envoy-mobile/pull/2601#discussion_r992520428 - mhm, we are adding `add` methods now 🤔 I think that we should just change the following methods to use `add...` prefix too then:

```
  EngineBuilder& setAppVersion(const std::string& app_version);
  EngineBuilder& setAppId(const std::string& app_id);
  EngineBuilder& setDeviceOs(const std::string& app_id);
  EngineBuilder& setStreamIdleTimeoutSeconds(int stream_idle_timeout_seconds);
  EngineBuilder& setPerTryIdleTimeoutSeconds(int per_try_idle_timeout_seconds);
```

does not have to be done as part of this PR.

library/cc/engine_builder.cc
```diff
@@ -185,6 +245,15 @@ std::string EngineBuilder::generateConfigStr() {
   for (const auto& [key, value] : replacements) {
     config_builder << "- &" << key << " " << value << std::endl;
   }
+  std::vector<std::string> stat_sinks = stat_sinks_;
+  if (!stats_domain_.empty()) {
+    stat_sinks.push_back("*base_metrics_service");
```
Augustyniak, line 250: are we missing something along these lines in here
https://github.com/envoyproxy/envoy-mobile/blob/b5dc941df120989327e34f69509dd32b8f0b4926/library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java#L305

library/cc/engine_builder.h
Augustyniak: Thanks for pointing that out. 👍 And thank you for working on improving the state of things in here - much appreciated.

I do not have a strong opinion as to whether to use `add` or `set` but going to share a few thoughts with regard to what I think the most optimal state could be. Side note: while I do not have a strong opinion as to whether it should be `set` or `add`  I think that we should clean up the APIs so that the name of a given method stays consistent between engine builder across Kotlin/Java/c++.

The name should depend on the nature of a given method:
* whether it's adding something (add) or just overriding a default value (set)
* what are the effects of calling a given method multiple times (whether `x+1`th call overrides the effects of `x`th call). If `x+1`th  call always overrides the result of `x`th call then `set`, otherwise consider `add`.

Augustyniak (dismissed):

library/cc/engine_builder.h
RyanTheOptimist: Good advice. Will do!

Augustyniak (approved):

PR #2611, opened 2022-10-17
Update rules_swift

jpsim:
For https://github.com/bazelbuild/rules_swift/pull/922


jpsim:
I've seen `output '*-Swift.h' was not created` a number of times in local development, this should fix that.

Augustyniak (approved):

PR #2619, opened 2022-10-19
Add support for String Accessors to the C++ engine builder

RyanTheOptimist:
Add support for String Accessors to the C++ engine builder

Introduces a C++ StringAccessor interface and method to convert to an envoy_string_accessor.
Minor cleanup of key_value_store handling in engine_builder.cc

Part of: #2498

Risk Level: Low
Testing: New unit tests
Docs Changes: N/A
Release Notes: Updated version_history.rst


library/cc/engine_builder.cc
```diff
@@ -322,14 +328,22 @@ EngineSharedPtr EngineBuilder::build() {
...
+  }
+
+  for (const auto& [name, accessor] : string_accessors_) {
     // TODO(goaway): This leaks, but it's tied to the life of the engine.
-    envoy_kv_store* api = static_cast<envoy_kv_store*>(safe_malloc(sizeof(envoy_kv_store)));
-    *api = it->second->asEnvoyKeyValueStore();
-    register_platform_api(it->first.c_str(), api);
+    auto* api = new envoy_string_accessor();
+    *api = StringAccessor::asEnvoyStringAccessor(accessor);
+    std::cout << "registering API: " << name << std::endl;
```
alyssawilk, line 345: ENVOY LOG or ermove?

docs/root/intro/version_history.rst
```diff
@@ -21,7 +21,7 @@ Features:
 - kotlin: add a way to tell Envoy Mobile to respect system proxy settings by calling an ``enableProxying(true)`` method on the engine builder. (:issue:`#2416 <2416>`)
 - kotlin: add a ``enableSkipDNSLookupForProxiedRequests(true)`` knob for controlling whether Envoy waits on DNS response in the dynamic forward proxy filter for proxied requests. (:issue:`#2602 <2602>`)
 - api: Add various methods to C++ EngineBuilder to bring it to parity with the Java and Obj-C builders. (:issue:`#2498 <2498>`)
-
+- api: Add support for String Accessors to the C++ EngineBuilder. (:issue:`#2498 <2498>`)
```
alyssawilk, line 24: I'm not sure I understand what the intended use of this is. Will it be more clear in a future PR or should I push for documentation clarity here?

library/cc/engine_builder.cc
RyanTheOptimist: Bah! Removed. Thanks.

docs/root/intro/version_history.rst
RyanTheOptimist: Good question. So String Accessors are already "a thing" in the Java and Obj-C engine builders. Rafal explained it to me as "so with string accessor you get an access to the API that you can call from within an EM core and which will you return a string that comes from the platform layer. ... I want to implement a filter in c++ that requires me having access to an  api_token. EM does not now anything about api_token but I can register a string accessor that returns it if needed"

So this PR merely adds a C++ implementation of an Accessor and adds a method to the C++ Builder to register them when creating an engine. I *think* improved docs for this feature would be great, but is probably out of scope for this PR. WDYT?

alyssawilk: fair enough :-P

alyssawilk (approved):

library/cc/engine_builder.cc
```diff
@@ -322,14 +328,21 @@ EngineSharedPtr EngineBuilder::build() {
   } else {
     config_str = config_override_for_tests_;
   }
-  auto envoy_engine =
+  envoy_engine_t envoy_engine =
       init_engine(this->callbacks_->asEnvoyEngineCallbacks(), null_logger, null_tracker);
 
-  for (auto it = key_value_stores_.begin(); it != key_value_stores_.end(); ++it) {
+  for (const auto& [name, store] : key_value_stores_) {
+    // TODO(goaway): This leaks, but it's tied to the life of the engine.
```
Augustyniak, line 335: not sure whether we should keep assigning this to goaway - no strong opinion

library/cc/string_accessor.h
```diff
@@ -0,0 +1,36 @@
...
+ * string from the application.
+ */
+class StringAccessor : public std::enable_shared_from_this<StringAccessor> {
+public:
+  virtual ~StringAccessor() = default;
+
+  /**
+   * Returns the string associated with this accessor.
+   */
+  virtual const std::string& getString() const PURE;
```
Augustyniak, line 24: maybe let's just call it `get()`? `String` part is clear based on the signature of the method. 

No strong opinion

Augustyniak (dismissed):
2 optional comments

library/cc/engine_builder.cc
RyanTheOptimist: Done!

library/cc/string_accessor.h
RyanTheOptimist: Done. (I'd considered this initially but the Java impl had string in the name. But if you and I both prefer simple "get()", then I'm sold :>)

RyanTheOptimist (commented):
Thanks for the review! PTAL

PR #2624, opened 2022-10-20
android: fix engine builder crash on Android 6/7

### LINKED ISSUE #2622: Use of unavailable String.join function in older Android OS versions
Introduced here: https://github.com/envoyproxy/envoy-mobile/pull/2563

`String.join` is not available in Android OS < 8.

jpsim:
Description: `String.join` is only available in Android 8 and later.
Risk Level: Low.
Testing: Unit tests.
Docs Changes: N/A.
Release Notes: Added.
Fixes #2622

### DIFF
--- a/docs/root/intro/version_history.rst
+++ b/docs/root/intro/version_history.rst
@@ -12,7 +12,6 @@ Breaking changes:
 Bugfixes:
 
 - android: fix engine startup crash for when admin interface is enabled. (:issue:`#2520 <2520>`)
-- android: fix engine builder crash on Android 6/7 when using stats sinks. (:issue:`#2622 <2622>`)
 - android: respect system security policy when determining whether clear text requests are allowed. (:issue:`#2528 <2528>`)
 
 Features:
--- a/docs/root/intro/version_history.rst
+++ b/docs/root/intro/version_history.rst
@@ -12,6 +12,7 @@ Breaking changes:
 Bugfixes:
 
 - android: fix engine startup crash for when admin interface is enabled. (:issue:`#2520 <2520>`)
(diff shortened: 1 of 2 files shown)


docs/root/intro/version_history.rst
```diff
@@ -12,6 +12,7 @@ Breaking changes:
 Bugfixes:
 
 - android: fix engine startup crash for when admin interface is enabled. (:issue:`#2520 <2520>`)
+- android: fix engine builder crash on Android 6/7 when using stats sinks. (:issue:`#2622 <2622>`)
```
Augustyniak, line 15: that crash was not a part of any of the manually published releases so not sure whether we need to list it as something that changed.

library/java/io/envoyproxy/envoymobile/engine/EnvoyConfiguration.java
```diff
@@ -309,7 +309,10 @@ String resolveTemplate(final String configTemplate, final String platformFilterT
 
     if (!stat_sinks_config.isEmpty()) {
       configBuilder.append("- &stats_sinks [");
-      configBuilder.append(String.join(",", stat_sinks_config));
+      configBuilder.append(stat_sinks_config.get(0));
+      for (int i = 1; i < stat_sinks_config.size(); i++) {
```
Augustyniak, line 313: that does not work with one stat sink?

jpsim: With just one stat sink, this for loop won't run at all, since the condition starts at `i = 1` and runs until `i < 1` which is never.

Augustyniak (approved):

PR #2626, opened 2022-10-20
Add support for Platform and Native filters to C++ EngineBuilder

RyanTheOptimist:
As discussed in the weekly meeting, this does not provide a C++ implementation of Platform filters, merely the ability to configure Envoy to use them.
    
Part of: #2498
    
Risk Level: Low
Testing: New unit tests
Docs Changes: N/A
Release Notes: Updated version_history.rst
    
### DIFF
--- a/library/cc/engine_builder.cc
+++ b/library/cc/engine_builder.cc
@@ -212,6 +212,11 @@ EngineBuilder& EngineBuilder::addNativeFilter(const std::string& name, const std
   return *this;
 }
 
+EngineBuilder& EngineBuilder::addPlatformFilter(const std::string& name) {
+  platform_filters_.push_back(name);
+  return *this;
+}
+
 std::string EngineBuilder::generateConfigStr() const {
 #if defined(__APPLE__)
   std::string dns_resolver_name = "envoy.network.dns_resolver.apple";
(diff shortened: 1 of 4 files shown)


library/cc/engine_builder.cc
```diff
@@ -207,6 +207,17 @@ EngineBuilder& EngineBuilder::addStringAccessor(const std::string& name,
...
 }
 
+EngineBuilder& EngineBuilder::addNativeFilter(const std::string& name,
+                                              const std::string& typed_config) {
+  native_filter_chain_.emplace_back(name, typed_config);
+  return *this;
+}
+
+EngineBuilder& EngineBuilder::addPlatformFilter(const std::string& name) {
+  platform_filters_.push_back(name);
```
Augustyniak, line 217: 1. any reason for why we use both `emplace_back` and `push_back` in the implementations of these 2 methods or could we use just one of them instead?
1. Not entirely sure how is `addPlatformFilter` supported to work given that it does take only 1 argument as opposed to its Swift and Kotlin counterparts that take 2. I realize that we discussed this briefly in our last community meeting but I thought that we would not add `addPlatformFilter` method at all.

Augustyniak: >this implementation of addPlatformFilter simply adds the platform filter name to the envoy config. It does not, however, register the PlatformFilter into the Envoy API. The Swift and Kotlin builders do both of those things. Once the Kotllin and Swift builder call into the C++ builder, they will still do the language specific Filter registration, but they will need the C++ builder to configure Envoy to use it. I hope that helps!

That does definitely help and makes sense! Maybe it would be a good idea to drop some comment with regard to that in `addPlatformFilter` method's implementation? leaving the decision up to you.

Augustyniak (approved):

PR #2627, opened 2022-10-20
Register getaddrinfo in extension_registry

caschoener:
This change seems to be needed to get envoy-mobile client integration tests to build inside Google3.  It seems safe to upstream it.


alyssawilk:
Looks like you need a fix format locally but otherwise LGTM

jpsim (dismissed):
Good catch, please fix the format and I'll re-approve.

alyssawilk (approved):

PR #2634, opened 2022-10-25
Add a comment to C++ Engine Builder addPlatformFilter

RyanTheOptimist:
Follow up from #2626.

Risk Level: None
Testing: N/A
Docs Changes: N/A
Release Notes: N/A


RyanTheOptimist:
@Augustyniak please take a look. I added this to the declaration not the implementation since this seems like a comment that callers need to know about (as they need to do the platform API registration).

Augustyniak (approved):
Thank you!

PR #2638, opened 2022-10-27
Implement an iOS platform certificate verifier.

RyanTheOptimist:
Implement an iOS platform certificate verifier 
and Swift builder APIs for enabling it.

Risk Level: Low
Testing: Help?
Docs Changes: N/A
Release Notes: Updated version_history.txt


RyanTheOptimist:
@jpsim this is finally passing CI. Woo hoo! I'm 100% confident that it needs work, but I'm not clear enough on how iOS work in Envoy Mobile to really have a good sense for where this code should go/what it should look like. So this is really more of rough draft than a ready-to-review PR, but if you could take a look, I would appreciate it.

docs/root/intro/version_history.rst
```diff
@@ -17,6 +17,7 @@ Bugfixes:
 
 Features:
 
+- java/switf: add option to support platform provided certificates validation interfaces on iOS. (:issue `#2144 <2144>`)
 - kotlin/c++: add option to support platform provided certificates validation interfaces on Android. (:issue `#2144 <2144>`)
```
jpsim, line 21: Let's combine these?

```suggestion
- api: add option to support platform provided certificates validation interfaces on iOS and Android. (:issue `#2144 <2144>`)
```

library/objective-c/EnvoyConfiguration.m
```diff
@@ -188,7 +190,10 @@ - (nullable NSString *)resolveTemplate:(NSString *)templateYAML {
       appendFormat:@"- &stats_flush_interval %lus\n", (unsigned long)self.statsFlushSeconds];
 
   NSString *cert_validator_template =
-      [[NSString alloc] initWithUTF8String:default_cert_validation_context_template];
+      self.enablePlatformCertificateValidation
+          ? [[NSString alloc] initWithUTF8String:platform_cert_validation_context_template]
+          : [[NSString alloc] initWithUTF8String:default_cert_validation_context_template];
```
jpsim, line 195: Nit: Objective-C has a shorthand syntax for this:

```suggestion
      self.enablePlatformCertificateValidation
          ? @(platform_cert_validation_context_template)
          : @(default_cert_validation_context_template);
```

library/objective-c/EnvoyEngineImpl.m
```diff
@@ -430,6 +430,119 @@ static void ios_track_event(envoy_map map, const void *context) {
...
 
+#include <CoreFoundation/CoreFoundation.h>
+#include <CoreFoundation/CFArray.h>
+#include "library/common/extensions/cert_validator/platform_bridge/c_types.h"
+#include "openssl/ssl.h"
+
+// Returns a new CFMutableArrayRef containing a series of SecPolicyRefs to be
+// added to a SecTrustRef used to validate a certificate for an SSL server,
+// or NULL on failure.
+CFMutableArrayRef CreateTrustPolicies() {
```
jpsim, line 441: Since these are just C APIs, why not move them a layer down into the `library/common` layer so they can be used directly from interfaces that don't delegate to the ObjC implementation, like the C++ builder?

library/swift/EngineBuilder.swift
```diff
@@ -27,6 +27,7 @@ open class EngineBuilder: NSObject {
   private var enableBrotli: Bool = false
   private var enableInterfaceBinding: Bool = false
   private var enforceTrustChainVerification: Bool = true
+  private var enablePlatformCertificateValidation: Bool = true
```
jpsim, line 30: Should this default to `false` until it's been vetted in a production experiment?

```suggestion
  private var enablePlatformCertificateValidation: Bool = false
```

test/swift/EngineBuilderTests.swift
```diff
@@ -586,6 +594,13 @@ final class EngineBuilderTests: XCTestCase {
     XCTAssertTrue(resolvedYAML.contains("&dns_multiple_addresses false"))
     XCTAssertTrue(resolvedYAML.contains("&enable_interface_binding false"))
     XCTAssertTrue(resolvedYAML.contains("&trust_chain_verification VERIFY_TRUST_CHAIN"))
+    XCTAssertTrue(resolvedYAML.contains(
+"""
+&validation_context
+  custom_validator_config:
+    name: \"envoy_mobile.cert_validator.platform_bridge_cert_validator
```
jpsim, line 601: Is this the expected string? Has an over-escaped leading quote and no trailing quote? There's no need to escape quotes like this in a multiline string literal.

jpsim:
It's great to see this @RyanTheOptimist! My main thought is if it's possible to move the bulk of this into the `library/common` layer since the iOS platform APIs are in C.

I look forward to being able to stop bundling certs with Envoy Mobile in the long run!

test/swift/EngineBuilderTests.swift
RyanTheOptimist: Hm, yeah that's confusing. I had some trouble sorting out multi-line strings in Swift and I think it lost the trailing quote. Fixed! Thanks,

RyanTheOptimist (commented):
Thanks for the review!

.bazelrc
```diff
@@ -18,7 +18,7 @@ build --ios_simulator_device="iPhone 13"
 build --ios_simulator_version=16.0
 build --verbose_failures
 build --workspace_status_command=envoy/bazel/get_workspace_status
-build --xcode_version=14.0
+build --xcode_version=14.1
```
jpsim, line 21: We don't have Xcode 14.1 installed on CI, only Xcode 14.0

RyanTheOptimist: Bah! Sorry, didn't mean to git push that.

library/common/network/BUILD
```diff
@@ -76,3 +76,23 @@ envoy_cc_library(
         "@envoy//source/common/network:socket_interface_lib",
     ],
 )
+
+cc_library(
+    name = "ios_platform_verifier",
+    srcs = select({
```
jpsim, line 82: Any reason to put these selects in multiple places turning this into an empty `cc_library` vs using a single `select` when this is used in `deps`?

library/common/network/ios_platform_verifier.cc
```diff
@@ -0,0 +1,122 @@
...
+envoy_cert_validation_result make_result(envoy_status_t status, uint8_t tls_alert,
+                                         const char* error_details) {
+  envoy_cert_validation_result result;
+  result.result = status;
+  result.tls_alert = tls_alert;
+  result.error_details = error_details;
+  return result;
+}
+
+#include <iostream>
```
jpsim, line 73: Nit: should this be with the other includes at the top of the file?

jpsim (commented):
This is really shaping up! Did you want to hook this up in the C++ builder in this PR or as a followup?

jpsim:
I think I'm seeing a crash with this enabled:

```
* thread #14, stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x00000001052583b0 app`Envoy::Extensions::TransportSockets::Tls::PlatformBridgeCertValidator::PendingValidation::postVerifyResultAndCleanUp(this=0x00006000019f4ef0, success=true, error_details="", tls_alert='.', error_counter=OptRef<Envoy::Stats::Counter> @ 0x000000016d9567c8) at platform_bridge_cert_validator.cc:142:21
    frame #1: 0x000000010525821c app`Envoy::Extensions::TransportSockets::Tls::PlatformBridgeCertValidator::verifyCertChainByPlatform(this=0x00006000000c61c0, cert_chain=size=0, host_name="", subject_alt_names=size=1, pending_validation=0x00006000019f4ef0) at platform_bridge_cert_validator.cc:125:22
    frame #2: 0x0000000105257360 app`Envoy::Extensions::TransportSockets::Tls::PlatformBridgeCertValidator::PendingValidation::verifyCertsByPlatform(this=0x00006000019f4ef0) at platform_bridge_cert_validator.cc:130:11
    frame #3: 0x0000000105263864 app`decltype(__f=0x00006000032291a8, __a0=0x00006000032291b8)).*fp()) std::__1::__invoke<void (Envoy::Extensions::TransportSockets::Tls::PlatformBridgeCertValidator::PendingValidation::*)(), Envoy::Extensions::TransportSockets::Tls::PlatformBridgeCertValidator::PendingValidation*, void>(void (Envoy::Extensions::TransportSockets::Tls::PlatformBridgeCertValidator::PendingValidation::*&&)(), Envoy::Extensions::TransportSockets::Tls::PlatformBridgeCertValidator::PendingValidation*&&) at type_traits:3859:1
    frame #4: 0x00000001052637ac app`void std::__1::__thread_execute<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void (Envoy::Extensions::TransportSockets::Tls::PlatformBridgeCertValidator::PendingValidation::*)(), Envoy::Extensions::TransportSockets::Tls::PlatformBridgeCertValidator::PendingValidation*, 2ul>(__t=size=3, (null)=__tuple_indices<2> @ 0x000000016d956f7f)(), Envoy::Extensions::TransportSockets::Tls::PlatformBridgeCertValidator::PendingValidation*>&, std::__1::__tuple_indices<2ul>) at thread:287:5
    frame #5: 0x0000000105262f38 app`void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void (Envoy::Extensions::TransportSockets::Tls::PlatformBridgeCertValidator::PendingValidation::*)(), Envoy::Extensions::TransportSockets::Tls::PlatformBridgeCertValidator::PendingValidation*> >(__vp=0x00006000032291a0) at thread:298:5
    frame #6: 0x00000001cba8f4e4 libsystem_pthread.dylib`_pthread_start + 116
```

jpsim:
The crash happened when I was stepping through code in lldb breakpoints, so it may have influenced the crash. I haven't seen the crash since running without stopping at a breakpoint.

Can you add this to the experimental app so we get some integration test coverage?

library/common/network/ios_platform_verifier.cc
```diff
@@ -0,0 +1,122 @@
...
+
+#include <CoreFoundation/CFArray.h>
+#include <CoreFoundation/CoreFoundation.h>
+#include <Security/SecCertificate.h>
+#include <Security/SecPolicy.h>
+#include <Security/SecTrust.h>
+
+#include "library/common/extensions/cert_validator/platform_bridge/c_types.h"
+#include "library/common/main_interface.h"
+#include "openssl/ssl.h"
```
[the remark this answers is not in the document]
RyanTheOptimist: I've removed the direct SSL dependency, and used a constant as per your suggestion. That being said, BoringSSL is definitely required by Envoy so I'm not sure we're going to be able to get rid of it. But I've remove it here :)

library/common/network/ios_platform_verifier.cc
RyanTheOptimist: Whoops! Thanks. Should have just removed that. I added it for some debug logging that I removed. Forgot to remove this though. Removed.

RyanTheOptimist:
> This is really shaping up! Did you want to hook this up in the C++ builder in this PR or as a followup?

I'd prefer to hook up in a followup, just to keep this smaller, if that works for you?

RyanTheOptimist:
> The crash happened when I was stepping through code in lldb breakpoints, so it may have influenced the crash. I haven't seen the crash since running without stopping at a breakpoint.
> 
> Can you add this to the experimental app so we get some integration test coverage?

Is the experimental all the same thing as hello_world? I enabled it there, hoping that's what you're looking for, but maybe not?

RyanTheOptimist (commented):
Thanks for the review!

library/common/network/BUILD
jpsim: Ah I didn't think about keeping `//library/common/network/...` compiling on every platform, that's a good enough reason to do this I guess.

jpsim:
> Is the experimental all the same thing as hello_world? I enabled it there, hoping that's what you're looking for, but maybe not?

No, we try to keep the "hello world" type apps focused, the "baseline" apps as using the default engine builder configurations and the "experimental" apps using non-default engine builder configurations.

Can you set `.enablePlatformCertificateValidation(true)` here? https://github.com/envoyproxy/envoy-mobile/blob/b22710bfa3ff50ec10a65c80634ce372ed820859/test/swift/apps/experimental/ViewController.swift#L20

library/common/network/ios_platform_verifier.cc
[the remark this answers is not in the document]
jpsim: I was under the impression that OpenSSL was primarily a library for exposing cryptographic functionality, which we can probably inject from the platform instead, but reading more about it I now understand that there's higher level TLS functionality that we probably wouldn't as easily be able to get from the platform layer, so there goes probably any hope of ever allowing Envoy Mobile to omit OpenSSL in any configuration.

Sorry for the back and forth but given this realization, I think we'd be better off going back to the `SSL_AD_CERTIFICATE_UNKNOWN` constant instead. 🤦

library/common/network/BUILD
RyanTheOptimist: Oh, interesting. That's basically the only way I build stuff :) FWIW, Envoy has `envoy_cc_linux_library`, `envoy_cc_posix_library`, `envoy_cc_win32_library` which wrap these select statements. We could use this in Envoy Mobile, if we wished?

library/common/network/ios_platform_verifier.cc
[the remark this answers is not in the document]
RyanTheOptimist: Ah, I see where you were going with that. Such a thing would potentially be possible, but as you say, probably not practical, alas. Ok, back to using SSL_AD_CERTIFICATE_UNKNOWN.

RyanTheOptimist:
> > Is the experimental all the same thing as hello_world? I enabled it there, hoping that's what you're looking for, but maybe not?
> 
> No, we try to keep the "hello world" type apps focused, the "baseline" apps as using the default engine builder configurations and the "experimental" apps using non-default engine builder configurations.
> 
> Can you set `.enablePlatformCertificateValidation(true)` here?
> 
> https://github.com/envoyproxy/envoy-mobile/blob/b22710bfa3ff50ec10a65c80634ce372ed820859/test/swift/apps/experimental/ViewController.swift#L20

Oh, I see. Done!

library/common/network/BUILD
jpsim: An empty library seems fine for now.

library/common/network/BUILD
```diff
@@ -76,3 +76,23 @@ envoy_cc_library(
         "@envoy//source/common/network:socket_interface_lib",
     ],
 )
+
+cc_library(
+    name = "ios_platform_verifier",
```
jpsim, line 81: Nit: can we call this `apple_platform_cert_verifier` since "verifier" could imply it verifies a lot of things, and this applies to all Apple platforms, not just iOS, and mirrors the naming of the apple_dns_resolver.

```suggestion
    name = "apple_platform_cert_verifier",
```

library/common/network/BUILD
```diff
@@ -76,3 +76,23 @@ envoy_cc_library(
...
+    }),
+    deps = select({
+        "@envoy//bazel:apple": [
+            "//library/common/extensions/cert_validator/platform_bridge:c_types_lib",
+            "//library/common:envoy_main_interface_lib",
+            "@envoy//bazel:boringssl",
+        ],
+        "//conditions:default": [],
+    }),
+)
```
jpsim, line 98: Can you add a trailing newline to match unix conventions?

jpsim (approved):
Hooray, great job setting this up.

library/common/network/BUILD
RyanTheOptimist: Good idea. Done.

library/common/network/BUILD
RyanTheOptimist: Done. Not sure how that crept back :/ (Also not sure why clang-format didn't fix it. But c'est la vie)

library/common/network/apple_platform_cert_verifier.h
```diff
@@ -9,7 +9,7 @@ extern "C" {
 /**
  * Registers the iOS platform cert verifier API.
```
jpsim, line 10:
```suggestion
 * Registers the Apple platform cert verifier API.
```

jpsim (approved):

library/common/network/apple_platform_cert_verifier.h
RyanTheOptimist: Facepalm! Done. Thanks.

jpsim (approved):

Issue #1274, opened 2021-03-01
support binary size analysis tool with split DWARF support


junr03:
After this [change](https://github.com/envoyproxy/envoy/pull/12569) Envoy starting  using split DWARF capabilities to reduce link time and binary size when compiling debug symbols. However, Bloaty (the analysis tool documented in the Envoy Mobile docs) does not support using `.dwp` files as a source for debug symbols (tracked in this [issue](https://github.com/google/bloaty/issues/156)

junr03:
Applying this patch to the Envoy submodule reverts split DWARF support in upstream envoy and allows the analysis documented in [Envoy Mobile](https://envoy-mobile.github.io/docs/envoy-mobile/latest/development/performance/binary_size.html).

```diff
diff --git a/.bazelrc b/.bazelrc
index c7a6e779b..5c57d6675 100644
--- a/.bazelrc
+++ b/.bazelrc
@@ -24,8 +24,6 @@ build --enable_platform_specific_config
 build:linux --copt=-fPIC
 build:linux --cxxopt=-std=c++17
 build:linux --conlyopt=-fexceptions
-build:linux --fission=dbg,opt
-build:linux --features=per_object_debug_info
 
 # We already have absl in the build, define absl=1 to tell googletest to use absl for backtrace.
 build --define absl=1
diff --git a/bazel/envoy_internal.bzl b/bazel/envoy_internal.bzl
index fc41557fe..b19656466 100644
--- a/bazel/envoy_internal.bzl
+++ b/bazel/envoy_internal.bzl
@@ -50,9 +50,9 @@ def envoy_copts(repository, test = False):
                # debugging info detailing some 1600 test binaries would be wasteful.
                # targets listed in order from generic to increasing specificity.
                # Bazel adds an implicit -DNDEBUG for opt targets.
-               repository + "//bazel:opt_build": [] if test else ["-ggdb3", "-gsplit-dwarf"],
+               repository + "//bazel:opt_build": [] if test else ["-ggdb3"],
                repository + "//bazel:fastbuild_build": [],
-               repository + "//bazel:dbg_build": ["-ggdb3", "-gsplit-dwarf"],
+               repository + "//bazel:dbg_build": ["-ggdb3"],
                repository + "//bazel:windows_opt_build": [] if test else ["-Z7"],
                repository + "//bazel:windows_fastbuild_build": [],
                repository + "//bazel:windows_dbg_build": [],
```

PR #2645, opened 2022-10-31
Remove note about DWARF patch being required

RyanTheOptimist:
As per https://github.com/envoyproxy/envoy-mobile/issues/1274#issuecomment-954305795, this patch is no longer required.

### DIFF
--- a/docs/root/development/performance/binary_size.rst
+++ b/docs/root/development/performance/binary_size.rst
@@ -49,17 +49,6 @@ necessary tools::
   make -j6
   cp bloaty /usr/local/bin/bloaty
 
-.. attention::
-
-    After this `change <https://github.com/envoyproxy/envoy/pull/12569>`_ Envoy started
-    using split DWARF capabilities to reduce link time and binary size when compiling debug symbols.
-    However, Bloaty (the analysis tool described below)
-    does not support using ``.dwp`` files as a source for debug symbols
-    (tracked in this `issue <https://github.com/google/bloaty/issues/156>`_). Therefore, a patch
-    `like this <https://github.com/envoyproxy/envoy-mobile/issues/1274#issuecomment-788345216>`_
-    must be applied to the Envoy submodule before compiling the binary described
-    below.
-
 The binary being compiled is ``//test/performance:test_binary_size``.
 The binary is getting built with the following build command::
 


jpsim:
Envoy's `linux` build config still passes `--features=per_object_debug_info`: https://github.com/envoyproxy/envoy/blob/66e5e6f9ee201a4ebdb7c33cd83d30a72ca357d4/.bazelrc#L44

And that passes `-gsplit-dwarf`: https://github.com/bazelbuild/bazel/blob/19b8d24531fb728701a85ec050361d434600b5d2/tools/cpp/unix_cc_toolchain_config.bzl#L406

And bloaty still doesn't support that. So from what I can tell, this is still needed?

RyanTheOptimist:
> Envoy's `linux` build config still passes `--features=per_object_debug_info`: https://github.com/envoyproxy/envoy/blob/66e5e6f9ee201a4ebdb7c33cd83d30a72ca357d4/.bazelrc#L44
> 
> And that passes `-gsplit-dwarf`: https://github.com/bazelbuild/bazel/blob/19b8d24531fb728701a85ec050361d434600b5d2/tools/cpp/unix_cc_toolchain_config.bzl#L406
> 
> And bloaty still doesn't support that. So from what I can tell, this is still needed?

Hm! That's curious. Both @alyssawilk and I were able to follow the instructions on that page without applying the patch and got bloaty output successfully. As a result, it seemed like it wasn't required.

On the other hand, @abeyad did have some issue getting bloaty to work. Ali, if you apply the patch, are you able to get bloaty to work?

abeyad:
Even applying that patch, I still get:
```
bloaty: Don't know how to parse DWARF form: 39
```

Note that the patch doesn't seem to be fully relevant anymore: I could [comment out the lines](https://github.com/envoyproxy/envoy/blob/main/.bazelrc#L43-L44) in `.bazelrc` as JP mentioned, but in `bazel/envoy_internal.bzl`, the `-gsplit-dwarf` option no longer exists in that file, so there is nothing to patch there.

Regardless, applying what's left of the patch still gives me the error above.

RyanTheOptimist:
@jpsim WDYT?

abeyad:
I forgot to follow up here, but after running `bazel build` with `--config=release-android`, the DWARF problem went away, so I don't think we need the patch, and it's not fully accurate now since some of the patch doesn't apply anymore.

jpsim (approved):
I'm ok removing this if it's not helpful.

PR #2649, opened 2022-11-01
fix one issue blocking bumping Envoy

alyssawilk:
Fixing issue merging in https://github.com/envoyproxy/envoy/pull/23676


keith (approved):

alyssawilk:
ah yeah, this only fixes the issue I introduced :-/  Can someone more EM savvy look info the other failure?

keith (approved):

Issue #471, opened 2019-09-28
bazel: remove BoringSSL workaround


keith:
After https://boringssl-review.googlesource.com/c/boringssl/+/37804 is merged upstream and part of master-with-bazel, we can update BoringSSL in envoy, and then revert https://github.com/lyft/envoy-mobile/pull/470

keith:
Repro case for reference https://github.com/keith/BoringSSLWeakRepro

PR #470, opened 2019-09-28
Work around BoringSSL bitcode linker warning

keith:
When linking envoy-mobile to an iOS app that ships with bitcode, you get
these warnings from the linker:

```
ld: warning: Linker asked to preserve internal global: 'sk_CRYPTO_BUFFER_call_free_func'
ld: warning: Linker asked to preserve internal global: 'sk_X509_call_free_func'
ld: warning: Linker asked to preserve internal global: 'sk_X509_call_free_func'
ld: warning: Linker asked to preserve internal global: 'sk_X509_call_free_func'
ld: warning: Linker asked to preserve internal global: 'sk_X509_call_free_func'
```

I have submitted an upstream patch to fix this
https://boringssl-review.googlesource.com/c/boringssl/+/37804 in the
meantime I've applied that patch here.

Signed-off-by: Keith Smiley <keithbsmiley@gmail.com>

PR #2651, opened 2022-11-01
build: revert boringssl patch

alyssawilk:
Reverting https://github.com/envoyproxy/envoy-mobile/pull/470 which is a patch to boringssl
https://boringssl-review.googlesource.com/c/boringssl/+/37804 indicates that this patch was added to avoid a linker error, so given CI passes I assume it can be reverted.

Risk Level: low
Testing: n/a
Docs Changes: n/a
Release Notes: n/a

Part of https://github.com/envoyproxy/envoy/issues/23758
Fixes https://github.com/envoyproxy/envoy-mobile/issues/471


keith:
So the linker failures for this were downstream, so I don't think our tests here ever really validated it was an issue / fixed. I verified with my repro case https://github.com/keith/BoringSSLWeakRepro that it does still repro with HEAD BoringSSL. But the issue was only around when you used this with bitcode, which was deprecated starting with Xcode 14. So we might be able to merge this and just ignore this moving forward. I feel like it should be reproducible if you're using LTO, but my repro case doesn't work with just testing that quickly now.

keith (approved):

alyssawilk:
Ok in that case let's try landing and seeing if it sticks.  If not we can roll back and revisit - I had an email drafted to boringssl folks to ask they bump priority but I wanted to repro first and this didn't :-P

keith:
I just rebased and commented again on my upstream patch, we'll see if it goes anywhere.

PR #2663, opened 2022-11-07
set enablePlatformCertificateValidation to false on iOS by default

RyanTheOptimist:

Was set to true inadvertently in #2638.


jpsim (approved):
+1 although hopefully tests fail with this and need to be updated to reflect the new defaults

PR #2667, opened 2022-11-09
Use std::string instead of const std::string& in C++ APIs where the string is being copied.

RyanTheOptimist:
Use std::string instead of const std::string& in C++ APIs where the string is being copied.

This offers the potential for a perf win because the caller can std::move() their string and avoid a copy.

Risk Level: Low
Testing: Existing
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/library/cc/engine_builder.cc
+++ b/library/cc/engine_builder.cc
@@ -40,7 +40,7 @@ EngineBuilder& EngineBuilder::addStatsSinks(const std::vector<std::string>& stat
 }
 
 EngineBuilder& EngineBuilder::addGrpcStatsDomain(std::string stats_domain) {
-  this->stats_domain_ = stats_domain;
+  this->stats_domain_ = std::move(stats_domain);
   return *this;
 }
 
@@ -72,7 +72,7 @@ EngineBuilder& EngineBuilder::addDnsQueryTimeoutSeconds(int dns_query_timeout_se
 
 EngineBuilder&
 EngineBuilder::addDnsPreresolveHostnames(std::string dns_preresolve_hostnames) {
-  this->dns_preresolve_hostnames_ = dns_preresolve_hostnames;
+  this->dns_preresolve_hostnames_ = std::move(dns_preresolve_hostnames);
   return *this;
 }
 
@@ -105,28 +105,28 @@ EngineBuilder& EngineBuilder::addStatsFlushSeconds(int stats_flush_seconds) {
 }
 
 EngineBuilder& EngineBuilder::addVirtualClusters(std::string virtual_clusters) {
-  this->virtual_clusters_ = virtual_clusters;
+  this->virtual_clusters_ = std::move(virtual_clusters);
   return *this;
 }
 
 EngineBuilder& EngineBuilder::addKeyValueStore(std::string name,
                                                KeyValueStoreSharedPtr key_value_store) {
-  this->key_value_stores_[name] = key_value_store;
+  this->key_value_stores_[std::move(name)] = key_value_store;
   return *this;
 }
 
 EngineBuilder& EngineBuilder::setAppVersion(std::string app_version) {
-  this->app_version_ = app_version;
+  this->app_version_ = std::move(app_version);
   return *this;
 }
 
 EngineBuilder& EngineBuilder::setAppId(std::string app_id) {
-  this->app_id_ = app_id;
+  this->app_id_ = std::move(app_id);
   return *this;
 }
 
 EngineBuilder& EngineBuilder::setDeviceOs(std::string device_os) {
-  this->device_os_ = device_os;
+  this->device_os_ = std::move(device_os);
   return *this;
 }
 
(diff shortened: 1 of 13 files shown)


RyanTheOptimist:
@mjbshaw, PTAL

library/cc/engine_builder.cc
```diff
@@ -104,29 +104,29 @@ EngineBuilder& EngineBuilder::addStatsFlushSeconds(int stats_flush_seconds) {
...
+EngineBuilder& EngineBuilder::addVirtualClusters(std::string virtual_clusters) {
+  this->virtual_clusters_ = std::move(virtual_clusters);
   return *this;
 }
 
-EngineBuilder& EngineBuilder::addKeyValueStore(const std::string& name,
+EngineBuilder& EngineBuilder::addKeyValueStore(std::string name,
                                                KeyValueStoreSharedPtr key_value_store) {
-  this->key_value_stores_[name] = key_value_store;
+  this->key_value_stores_[std::move(name)] = key_value_store;
```
mjbshaw, line 114: The key_value_store (which is a std::shared_ptr) should be moved to avoid a temporary shared_ptr (which results in an unnecessary atomic increment and immediate decrement).

library/cc/headers_builder.cc
```diff
@@ -3,20 +3,20 @@
 namespace Envoy {
 namespace Platform {
 
-HeadersBuilder& HeadersBuilder::add(const std::string& name, const std::string& value) {
+HeadersBuilder& HeadersBuilder::add(std::string name, std::string value) {
   if (this->isRestrictedHeader(name)) {
     return *this;
   }
-  this->headers_[name].push_back(value);
+  this->headers_[std::move(name)].push_back(std::move(value));
```
mjbshaw, line 10: Moving the key (here and elsewhere) is kinda interesting. If they key doesn't exist in the map then it's optimal. If it does then we've created an unnecessary copy and string_view would have been better.

I don't have any particular opinions here, so I'm not opposed to taking the key by value and moving it like this. Intuitively it seems like users would most likely be calling these methods new keys that aren't already in the map, in which case this new change is ideal. But I don't have any evidence for my intuition so I'll leave it up to you/others.

library/cc/engine_builder.h
```diff
@@ -73,7 +73,7 @@ class EngineBuilder {
 
 private:
   struct NativeFilterConfig {
-    NativeFilterConfig(const std::string& name, const std::string& typed_config)
+    NativeFilterConfig(std::string name, std::string typed_config)
         : name_(name), typed_config_(typed_config) {}
```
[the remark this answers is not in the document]
RyanTheOptimist: Oh, yes. Good point. Done.

library/cc/headers_builder.h
```diff
@@ -9,17 +9,17 @@ class HeadersBuilder {
 public:
   virtual ~HeadersBuilder() {}
 
-  HeadersBuilder& add(const std::string& name, const std::string& value);
-  HeadersBuilder& set(const std::string& name, const std::vector<std::string>& values);
+  HeadersBuilder& add(std::string name, std::string value);
+  HeadersBuilder& set(std::string name, const std::vector<std::string>& values);
```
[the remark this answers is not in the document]
RyanTheOptimist: Ah, good point. Done.

library/cc/headers_builder.h
```diff
@@ -9,17 +9,17 @@ class HeadersBuilder {
...
   HeadersBuilder& remove(const std::string& name);
 
 protected:
   HeadersBuilder();
-  HeadersBuilder& internalSet(const std::string& name, const std::vector<std::string>& values);
+  HeadersBuilder& internalSet(std::string name, const std::vector<std::string>& values);
```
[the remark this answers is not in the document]
RyanTheOptimist: Ah, good point. Done.

library/cc/headers_builder.h
```diff
@@ -9,17 +9,17 @@ class HeadersBuilder {
...
   const RawHeaderMap& allHeaders() const;
 
 private:
-  bool isRestrictedHeader(const std::string& name) const;
+  bool isRestrictedHeader(std::string name) const;
```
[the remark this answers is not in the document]
RyanTheOptimist: Done.

library/cc/engine_builder.cc
RyanTheOptimist: Done.

library/cc/engine_builder.cc
```diff
@@ -201,20 +201,20 @@ EngineBuilder::enablePlatformCertificatesValidation(bool platform_certificates_v
   return *this;
 }
 
-EngineBuilder& EngineBuilder::addStringAccessor(const std::string& name,
+EngineBuilder& EngineBuilder::addStringAccessor(std::string name,
                                                 StringAccessorSharedPtr accessor) {
-  string_accessors_[name] = accessor;
+  string_accessors_[std::move(name)] = accessor;
```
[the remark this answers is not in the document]
RyanTheOptimist: Done!

library/cc/headers_builder.cc
RyanTheOptimist: Yeah, I debated this myself. My intuition matches yours that this is likely called when the key is not in the map, but I don't feel strongly. Let's see what the other Envoy Mobile reviewers have to say?

RyanTheOptimist (commented):
Thanks for the great comments!

RyanTheOptimist:
@jpsim, Alyssa is out today. Would you mind taking a look?

jpsim (approved):

library/cc/headers_builder.cc
```diff
@@ -3,20 +3,20 @@
...
 
-HeadersBuilder& HeadersBuilder::set(const std::string& name,
-                                    const std::vector<std::string>& values) {
+HeadersBuilder& HeadersBuilder::set(std::string name,
+                                    std::vector<std::string> values) {
   if (this->isRestrictedHeader(name)) {
     return *this;
   }
-  this->headers_[name] = values;
+  this->headers_[std::move(name)] = values;
```
mjbshaw, line 19: std::move the `values`

RyanTheOptimist: Oof! Good point. Sent out https://github.com/envoyproxy/envoy-mobile/pull/2694

Issue #2678, opened 2022-11-16
xDS integration tests hang for 15s before receiving DiscoveryResponse


abeyad:
This happens when running //test/common/integration:rtds_integration_test and //test/common/integration:sds_integration_test.

The cause seems to be the initial config fetch timer expiring after 15s: https://github.com/envoyproxy/envoy/blob/3059a76977311b43dc993720f7fda49c8e4dfbee/source/common/config/grpc_subscription_impl.cc#L35.  This seems to unblock the FakeUpstream from sending a xDS response, allowing the gRPC stream to proceed and the response is received in the Envoy engine.

abeyad:
The reason for waiting 15s and then getting a gRPC initial fetch timeout is because the default [initial_fetch_timeout](https://github.com/envoyproxy/envoy/blob/dede92e6e5888cdf27320bf3b6bc9987657d6e22/api/envoy/api/v2/core/config_source.proto#L180) for the config source is 15s (unless one explicitly sets it).  Until the initial fetch completes (either by success, failure, or timeout), the [PostInit function](https://github.com/envoyproxy/envoy-mobile/blob/7a0ca64eaf6e60b90df2f2f52a97e000f476dca4/library/common/engine.cc#L130) will not get called, and if the `PostInit` function doesn't get called, the Envoy Mobile engine will [wait](https://github.com/envoyproxy/envoy-mobile/blob/7a0ca64eaf6e60b90df2f2f52a97e000f476dca4/test/common/integration/base_client_integration_test.cc#L166) without receiving a notification on the mutex.  This is what causes the test to hang for 15s before proceeding.

The integration tests will not receive an initial fetch for xDS because it has to wait for it to get sent by the FakeUpstream via a [sendDiscoveryResponse](https://github.com/envoyproxy/envoy/blob/dede92e6e5888cdf27320bf3b6bc9987657d6e22/test/integration/base_integration_test.h#L190) call.  The solution for integration tests is to set the `initial_fetch_timeout`, like what we did in https://github.com/envoyproxy/envoy-mobile/pull/2679.

The solution for Envoy Mobile in general is to make sure the builder APIs for xDS add an `initial_fetch_timeout` to the configured xDS source and ensure it's a reasonable value (e.g. 5s instead of 15s).

PR #2679, opened 2022-11-16
test: Set initial_fetch_timeout on the xDS integration tests

abeyad:
While we figure out and fix
https://github.com/envoyproxy/envoy-mobile/issues/2678, this commit reduces the initial_fetch_timeout on the xDS config sources in the integration tests from 15s to 1s, preventing issues like test timeouts.


jpsim (approved):
+1 to unblock tests

PR #2685, opened 2022-11-18
Allow library/common/jni/ to be built on non-Android platforms

RyanTheOptimist:
Allow library/common/jni/ to be built on non-Android platforms
Use bazel select to make a number of android-specific targets conditional on Android.

Risk Level: Low
Testing: N/A
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/library/common/jni/BUILD
+++ b/library/common/jni/BUILD
@@ -231,10 +231,8 @@ cc_library(
 cc_library(
     name = "android_network_utility_lib",
     srcs = [
-    ] + select({
-        "@envoy//bazel:android": ["android_network_utility.cc"],
-        "//conditions:default": [],
-    }),
+        "android_network_utility.cc",
+    ],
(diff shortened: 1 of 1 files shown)


library/common/jni/BUILD
```diff
@@ -65,9 +65,9 @@ cc_binary(
     copts = ["-std=c++17"],
     linkopts = [
         "-lm",
-        "-llog",
     ] + select({
         "@envoy//bazel:dbg_build": ["-Wl,--build-id=sha1"],
+        "@envoy//bazel:android": ["-llog"],
```
jpsim, line 70: Is select mutually exclusive? I.e. should this be two selects?

jpsim:
Sorry for the delay in reviewing this, I missed it while I was out. Would you be ok to close this and re-open in Envoy?

PR #2686, opened 2022-11-20
test: Wait for RTDS updates to be applied in the integration test

abeyad:
This fixes a bug where sometimes the RtdsIntegrationTest would fail because we would check the updates after the DiscoveryResponse was sent but before ensuring the RTDS updates in the response have been applied by the Envoy engine.

### DIFF
--- a/test/common/integration/rtds_integration_test.cc
+++ b/test/common/integration/rtds_integration_test.cc
@@ -85,6 +85,7 @@ TEST_P(RtdsIntegrationTest, RtdsReload) {
   EXPECT_EQ("yar", getRuntimeKey("bar"));
   EXPECT_EQ("", getRuntimeKey("baz"));
 
+  const uint64_t initial_load_success = getCounterValue("runtime.load_success");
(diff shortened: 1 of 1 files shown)


test/common/integration/rtds_integration_test.cc
```diff
@@ -85,6 +85,7 @@ TEST_P(RtdsIntegrationTest, RtdsReload) {
   EXPECT_EQ("yar", getRuntimeKey("bar"));
   EXPECT_EQ("", getRuntimeKey("baz"));
 
+  const uint64_t initial_load_success = getCounterValue("runtime.load_success");
```
alyssawilk, line 88: can we have a race where this may or may not already be set? i.e. should we wait for it to get incremented here, and just always hav the below one be 1?

abeyad: Probably no race there since the PostInit would've finished executing by then, but I think hard coding 1 is clearer so I changed it below per your suggestion. Thanks!

alyssawilk (approved):

PR #2692, opened 2022-11-21
Fix a possible use-after-free with platform cert verification

### LINKED ISSUE #2691: Possible Use-After-Free in the platform cert validator.
PlatformBridgeCertValidator uses and absl::flat_hash_set to store pending certificate verifications. These are stored as PendingVerification, directly in the flat_hash_set and the validation thread holds a pointer to the PendingVerification. However, flat_hash_set does not ensure pointer stability which means it is possible for this pointer to become invalid.

RyanTheOptimist:
Fix a possible use-after-free with platform cert verification by using a unique_ptr in the flat_hash_set of pending validations. The flat_hash_set does not ensure pointer stability, but the validation thread holds a pointer to the PendingVerification, which is problematic. This PR makes PendingVerification non-moveable and non-copyable which avoids this problem.

There is also another potential use-after free in that the task posted to the dispatcher deletes the PendingValidation, but the PendingValidation touches member variables after the call to post. Reordered the call to post to avoid this.

Fixes #2691

Risk Level: low
Testing: manual
Docs Changes: N/A
Release Notes: Updated version history.

### DIFF
--- a/docs/root/intro/version_history.rst
+++ b/docs/root/intro/version_history.rst
@@ -7,7 +7,7 @@ Pending Release
 Breaking changes:
 
 - ios/android: remove ``addH2RawDomains`` method. (:issue: `#2590 <2590>`)
-- build: building on macOS now requires Xcode 14.0. (:issue:`#2544 <2544>`)
+- build: building on macOS now requires Xcode 14.1. (:issue:`#2664 <2664>`)
 - iOS: remove experimental option to force all connections to use IPv6.
 - kotlin: always use ``getaddrinfo`` DNS resolver. Remove ``addDNSFallbackNameservers``, ``enableDNSFilterUnroutableFamilies``, and ``enableDNSUseSystemResolver`` methods from the Kotlin engine builder. (:issue:`#2618 <2618>`)
 
--- a/docs/root/intro/version_history.rst
+++ b/docs/root/intro/version_history.rst
@@ -15,6 +15,7 @@ Bugfixes:
 
 - android: fix engine startup crash for when admin interface is enabled. (:issue:`#2520 <2520>`)
 - android: respect system security policy when determining whether clear text requests are allowed. (:issue:`#2528 <2528>`)
+- tls: fix a potential use-after-free with the platform cert validator. (:issue:`#2691 <2691>`)
 
 Features:
 
--- a/docs/root/intro/version_history.rst
+++ b/docs/root/intro/version_history.rst
@@ -15,7 +15,6 @@ Bugfixes:
 
 - android: fix engine startup crash for when admin interface is enabled. (:issue:`#2520 <2520>`)
 - android: respect system security policy when determining whether clear text requests are allowed. (:issue:`#2528 <2528>`)
-- tls: fix a potential use-after-free with the platform cert validator. (:issue:`#2691 <2691>`)
 
 Features:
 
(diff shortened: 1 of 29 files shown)


danzh2010 (dismissed):
Thanks for the fix!

docs/root/intro/version_history.rst
```diff
@@ -15,6 +15,7 @@ Bugfixes:
 
 - android: fix engine startup crash for when admin interface is enabled. (:issue:`#2520 <2520>`)
 - android: respect system security policy when determining whether clear text requests are allowed. (:issue:`#2528 <2528>`)
+- tls: fix a potential use-after-free with the platform cert validator. (:issue:`#2691 <2691>`)
```
jpsim, line 18: Not a big deal, but change log entries aren’t necessary for fixes to features added in that same version.

library/common/extensions/cert_validator/platform_bridge/platform_bridge_cert_validator.h
```diff
@@ -73,22 +73,15 @@ class PlatformBridgeCertValidator : public CertValidator, Logger::Loggable<Logge
           result_callback_(std::move(result_callback)),
           transport_socket_options_(std::move(transport_socket_options)) {}
 
+    // Ensure that this class is never moved or copies to guarantee pointer stability.
```
jpsim, line 76: Nit:

```suggestion
    // Ensure that this class is never moved or copied to guarantee pointer stability.
```

RyanTheOptimist: Done!

docs/root/intro/version_history.rst
RyanTheOptimist: Oh, I didn't realize that. Thanks! Reverted.

library/common/extensions/cert_validator/platform_bridge/platform_bridge_cert_validator.cc
```diff
@@ -138,7 +137,16 @@ void PlatformBridgeCertValidator::PendingValidation::verifyCertsByPlatform() {
...
+  ENVOY_LOG(trace,
+            "Finished platform cert validation for {}, post result callback to network thread",
+            host_name_);
+
+  if (parent_.platform_validator_->release_validator) {
+    parent_.platform_validator_->release_validator();
+  }
   std::weak_ptr<size_t> weak_alive_indicator(parent_.alive_indicator_);
+
+  // Once this task runs, `this` will be deleted so this must be the last statement in the file.
```
jpsim, line 149: This seems pretty brittle, is there a way to force the lifetime of `this` or otherwise only capture what's needed here? Is that something you're addressing in https://github.com/envoyproxy/envoy-mobile/pull/2696?

RyanTheOptimist: Yes, I'm planning to really fix this in #2696 (though that's not complete sorted out yet). The threading/lifetime semantics here I think need to be crisper.

jpsim:
We already run tests under asan. We could either add iOS tests to run under asan as well, or write some tests for the platform cert validator under `test/common/*` to have this validated.

jpsim (approved):
This fixes the crash I was seeing on iOS. Thanks!

PR #2699, opened 2022-11-22
Builder api for EM engine

caschoener:
Add a bunch of params to the engine builder so that rtds and sds tests can be run with `override_builder_config_ = false`.

TBD before shipping: Add these build API params to the other languages.  Wait for formatting CI to un-break.


library/common/config/config.cc
```diff
@@ -97,6 +97,184 @@ const char* socket_tag_config_insert = R"(
...
+                      - envoy_grpc:
+                          cluster_name: xds_cluster.lyft.com
+                        timeout: 300s
+                    transport_api_version: V3
+                  resource_api_version: V3
+          sni: lyft.com
+)";
+
+const char* rtds_clusters_insert = R"(
+    - name: base_h2
```
abeyad, line 241: base_h2 is not an xDS cluster, is it?  and it's already declared in the default clusters in L197 above

library/common/config/config.cc
```diff
@@ -511,22 +595,7 @@ stats_sinks: *stats_sinks
...
-            http2_delay_keepalive_timeout: *h2_delay_keepalive_timeout
-            skip_dns_lookup_for_proxied_requests: *skip_dns_lookup_for_proxied_requests
-)"
-// Needed due to warning in
-// https://github.com/envoyproxy/envoy/blob/6eb7e642d33f5a55b63c367188f09819925fca34/source/server/server.cc#L546
-R"(
-        overload:
-          global_downstream_max_connections: 0xffffffff # uint32 max
+#{custom_layers}
+#{custom_admin}
```
abeyad, line 599: don't the xDS clusters need to be added here as well through the `#{custom_clusters}` macro (or something similar)?

library/common/config/templates.h
```diff
@@ -77,6 +77,26 @@ extern const char* brotli_config_insert;
...
 
+/* Insert for default layers.
+ */
+extern const char* default_layers_insert;
+
+/* Insert that enables default stats config.
+ */
+extern const char* default_stats_config_insert;
+
+/* Insert that enables rtds clusters.
```
abeyad, line 88: I think this should be xDS clusters.... there might be one or more xDS clusters and the various xDS APIs might make use of one or more of those xDS clusters... a cluster may not be specific to just rtds or sds etc

library/cc/engine_builder.cc
```diff
@@ -155,6 +154,31 @@ EngineBuilder& EngineBuilder::enableSocketTagging(bool socket_tagging_on) {
   return *this;
 }
 
+EngineBuilder& EngineBuilder::setRtdsConfig(std::string yaml) {
+  this->custom_layers_ = yaml;
+  return *this;
+}
+
+EngineBuilder& EngineBuilder::setRtdsEndpoint(std::string hostname) {
```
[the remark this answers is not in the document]
caschoener: Good catch, this isn't being used anymore

library/cc/engine_builder.cc
```diff
@@ -302,13 +341,53 @@ std::string EngineBuilder::generateConfigStr() const {
         absl::StrReplaceAll(platform_filter_template, {{"{{ platform_filter_name }}", name}});
     insertCustomFilter(filter_config, config_template);
   }
+  if (!this->disable_stats_) {
```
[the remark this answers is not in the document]
caschoener: That's a good callout.  I wonder why this file uses `this->` so much (81 times!), as it does seem like they're all useless.  Maybe I make a separate PR to remove them all?

library/cc/engine_builder.cc
```diff
@@ -302,13 +341,53 @@ std::string EngineBuilder::generateConfigStr() const {
...
+  }
 
+  if (this->admin_yaml_ != "") {
+    absl::StrReplaceAll({{"#{custom_admin}", absl::StrCat("#{custom_admin}\n", this->admin_yaml_)}},
+                        &config_template);
+  }
+  if (this->enable_clusters_) {
+    std::string formatted_insert;
+    if (this->ports_.empty() || this->ipv_version_ == "") {
+      formatted_insert = fmt::format(rtds_clusters_insert, "127.0.0.1", "0", "127.0.0.1", "0");
```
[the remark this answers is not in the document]
caschoener: Addressed offline.  I will add a comment to explain

library/cc/engine_builder.cc
```diff
@@ -155,6 +154,31 @@ EngineBuilder& EngineBuilder::enableSocketTagging(bool socket_tagging_on) {
   return *this;
 }
 
+EngineBuilder& EngineBuilder::setRtdsConfig(std::string yaml) {
```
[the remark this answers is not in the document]
caschoener: Talked offline: easy enough to move the yaml into the config.cc bank of yaml and then have this function just take in the dynamic pieces.

library/cc/engine_builder.cc
```diff
@@ -155,6 +154,31 @@ EngineBuilder& EngineBuilder::enableSocketTagging(bool socket_tagging_on) {
...
+  return *this;
+}
+
+EngineBuilder& EngineBuilder::setRtdsEndpoint(std::string hostname) {
+  this->hostname_ = hostname;
+  return *this;
+}
+
+EngineBuilder& EngineBuilder::enableCustomClusters(bool enable_clusters) {
+  this->enable_clusters_ = enable_clusters;
```
caschoener, line 168: `enable_custom_clusters_`

library/common/config/config.cc
caschoener: Talked offline: we may want to have this yaml get passed in to the function instead.

library/common/config/config.cc
caschoener: I _think_ you just misread this one, I do add the custom clusters above (line 582)

test/common/integration/rtds_integration_test.cc
```diff
@@ -34,25 +44,22 @@ envoy::config::bootstrap::v3::LayeredRuntime layeredRuntimeConfig(const std::str
             set_node_on_first_message_only: true
     - name: some_admin_layer
       admin_layer: {{}}
-  )EOF",
+)EOF",
                                        api_type, XDS_CLUSTER);
```
caschoener, line 48: This function should just set dynamic parameters, and the yaml will be stored in config.cc

params are api_type, XDS_CLUSTER, and initial_fetch_timeout.

library/cc/engine_builder.h
```diff
@@ -61,6 +61,14 @@ class EngineBuilder {
   EngineBuilder& enableH2ExtendKeepaliveTimeout(bool h2_extend_keepalive_timeout_on);
   EngineBuilder& enforceTrustChainVerification(bool trust_chain_verification_on);
   EngineBuilder& enablePlatformCertificatesValidation(bool platform_certificates_validation_on);
+  EngineBuilder& setRtdsConfig(std::string yaml);
+  EngineBuilder& setRtdsEndpoint(std::string hostname);
+  EngineBuilder& enableCustomClusters(bool enable_clusters);
```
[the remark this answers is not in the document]
caschoener: Talked offline: these are all necessary for turning off `override_builder_config_`

library/cc/engine_builder.cc
```diff
@@ -190,8 +214,24 @@ EngineBuilder& EngineBuilder::enableH2ExtendKeepaliveTimeout(bool h2_extend_keep
   return *this;
 }
 
+EngineBuilder& EngineBuilder::setPorts(std::vector<uint32_t> ports) {
+  this->ports_ = ports;
+  return *this;
+}
+
+EngineBuilder& EngineBuilder::setIpvVersion(std::string ipv_version) {
```
alyssawilk, line 222: ditto. I'd think the only "new" think we're setting was the xDS hostname so I'm probably missing something

library/cc/engine_builder.cc
```diff
@@ -190,8 +214,24 @@ EngineBuilder& EngineBuilder::enableH2ExtendKeepaliveTimeout(bool h2_extend_keep
...
+}
+
+EngineBuilder& EngineBuilder::enableCustomTransportSocket(bool enable_transport_socket) {
+  this->enable_transport_socket_ = enable_transport_socket;
+  return *this;
+}
+
 EngineBuilder&
 EngineBuilder::enablePlatformCertificatesValidation(bool platform_certificates_validation_on) {
+
```
alyssawilk, line 234: please do go through and revert things like this and newlines that aren't necessary changes - just makes review easier.

alyssawilk (commented):
Epic change, wohoo!
here's some first pass thoughts

caschoener:
Talked with Alyssa and I'll put this on hold for a sec.  We realized that...

1) These tests are supposed to be using [DFP](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_proxy) not direct addressing.  So I will fix that in a separate PR before this one.
2) It will be wise to break this out into separate PRs, so I will start with just doing the admin piece in all languages.

Still good imo to have done it all in one place so at least we know this works.

abeyad:
> 1. These tests are supposed to be using [DFP](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_proxy) not direct addressing.  So I will fix that in a separate PR before this one.

Not sure I follow; the integration tests (e.g. rtds_integration_test.cc, sds_integration_test.cc) use clusters with a single fake upstream as an endpoint.  Where would DFP come into play?

> 2. It will be wise to break this out into separate PRs, so I will start with just doing the admin piece in all languages.

+1 on splitting out into separate PRs.

caschoener:
The clusters have a single fake upstream yes but I'm told it's better to use DFP since that's how envoy mobile actually works in practice, and this is also how the other tests work.

In other words we would point the xds cluster at `xds_cluster.lyft.com` and let DFP work out the IP.

alyssawilk:
sorry if I wasn't clear I think we want DFP for data plane.  For xDS I think for production use cases we're probably going to want logical dns, pointed at whatever the public TD hostname is (trafficdirector.google.com) and for tests it's fine to point to ip:port - I think the addXdsCluster API should take a string and if that string is an IP adress we set up the clutster we have today, and we TODO to hand the other case later?

abeyad:
> sorry if I wasn't clear I think we want DFP for data plane. For xDS I think for production use cases we're probably going to want logical dns (yet another cluster type) pointed at whatever the public TD hostname is (trafficdirector.google.com) and for tests it's fine to point to ip:port 

Ah yes, that makes perfect sense.  Thanks for clarifying!

> I think the addXdsCluster API should take a string and if that string is an IP adress we set up the clutster we have today, and we TODO to hand the other case later?

+1, makes sense to me.

jpsim:
Closing this PR as this repo is now archived. You can open a new PR in the Envoy repo if you'd like: https://github.com/envoyproxy/envoy

PR #2701, opened 2022-11-22
Rename release_validator to validation_cleanup

RyanTheOptimist:
Rename release_validator to validation_cleanup.
The release_validator function does not actually release the validator.
On iOS is does nothing and on Android it merely detaches the JVM from the current thread.
So "cleanup validation" is a better description of it's function than "release validator" and
matches the existing description "to clean up after validation completion."

### DIFF
--- a/library/common/extensions/cert_validator/platform_bridge/c_types.h
+++ b/library/common/extensions/cert_validator/platform_bridge/c_types.h
@@ -25,7 +25,7 @@ typedef envoy_cert_validation_result (*envoy_validate_cert_f)(const envoy_data* 
                                                               const char* host_name);
 
 /**
- * Function signature for calling into platform APIs to clean up after validation completion.
+ * Function signature for calling into platform APIs to clean up after validation is complete.
  */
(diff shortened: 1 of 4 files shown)


library/common/extensions/cert_validator/platform_bridge/c_types.h
```diff
@@ -27,7 +27,7 @@ typedef envoy_cert_validation_result (*envoy_validate_cert_f)(const envoy_data*
 /**
  * Function signature for calling into platform APIs to clean up after validation completion.
```
alyssawilk, line 28: long as you're in here, validation is complete?

alyssawilk (approved):
LGTM with optional nit

library/common/extensions/cert_validator/platform_bridge/c_types.h
RyanTheOptimist: Done!

alyssawilk (approved):

PR #2702, opened 2022-11-23
bridge: fix up proto registration

alyssawilk:
This fixes up issues exposed in https://github.com/envoyproxy/envoy/pull/24151 where the PlatformBridgeCertValidatorFactory was not associated with the PlatformBridgeCertValidator proto.

Risk Level: low
Testing:  https://github.com/envoyproxy/envoy/pull/24151
Docs Changes: n/a
Release Notes: n/a

### DIFF
--- a/library/common/extensions/cert_validator/platform_bridge/config.h
+++ b/library/common/extensions/cert_validator/platform_bridge/config.h
@@ -2,6 +2,7 @@
 
 #include "source/extensions/transport_sockets/tls/cert_validator/factory.h"
 
+#include "library/common/extensions/cert_validator/platform_bridge/platform_bridge.pb.h"
 #include "library/common/extensions/cert_validator/platform_bridge/platform_bridge_cert_validator.h"
 
 namespace Envoy {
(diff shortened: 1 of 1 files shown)


alyssawilk:
cc @danzh2010

library/common/extensions/cert_validator/platform_bridge/config.h
```diff
@@ -2,21 +2,28 @@
...
                                        SslStats& stats, TimeSource& time_source) override;
 
   std::string name() const override {
     return "envoy_mobile.cert_validator.platform_bridge_cert_validator";
   }
+  ProtobufTypes::MessagePtr createEmptyConfigProto() override {
+    return std::make_unique<
+        envoy_mobile::extensions::cert_validator::platform_bridge::PlatformBridgeCertValidator>();
+  }
+  std::string category() const override { return "envoy.tls.cert_validator"; }
```
RyanTheOptimist, line 26: In envoy_build_config/extensions_build_config.bzl, the extension shows up as "envoy_mobile.cert_validator.platform_bridge_cert_validator" (which matches name()). Should these be made consistent? How 'bout we change the name to ""envoy.tls.cert_validator.platform_bridge"? Or maybe this isn't required?

RyanTheOptimist (commented):
Yay! Thanks for doing this! One tiny suggestion.

library/common/extensions/cert_validator/platform_bridge/config.h
alyssawilk: name is still that string (above).  I think category is generally different than name so I copied category from CertValidatorFactory

RyanTheOptimist (approved):

PR #2709, opened 2022-11-29
android proxy: fix initial proxy settings state

Augustyniak:
Description: Upon investigation of multiple sessions with active proxy settings it seems to be highly likely that our EM proxy logic did not engage in all of the aforementioned sessions. That appears to be a direct result of the way our logic for registering to `PROXY_CHANGE` intent works. The implementation assumes that upon subscription to the intent we always receive an initial intent with the state of proxy settings. That does not seem to be guaranteed by Android as `PROXY_CHANGE` intent is not documented to be sticky (a term used by Android) and whether it's sticky or not may depend on manufacturer / type of the proxy update.
Risk Level: Low, hidden behind a feature flag and disabled by default.
Testing: Integration tests
Docs Changes: N/A
Release Notes: N/A


library/java/io/envoyproxy/envoymobile/engine/AndroidProxyMonitor.java
```diff
@@ -75,7 +78,12 @@ private ProxyInfo extractProxyInfo(final Intent intent) {
     // proxy is configured.
     //
     // See https://github.com/envoyproxy/envoy-mobile/issues/2531 for more details.
-    if (info.getPacFileUrl() != null && info.getPacFileUrl() != Uri.EMPTY) {
+    if (info.getHost() == "localhost" && info.getPort() == -1) {
```
jpsim, line 81: Is there a link to documentation or Android source code we could include in a comment here to explain why these magical values should mean there is no proxy?

test/kotlin/integration/proxying/ProxyPollPerformHTTPRequestUsingProxyTest.kt
```diff
@@ -0,0 +1,107 @@
...
+import java.util.concurrent.TimeUnit
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Mock
+import org.mockito.Mockito
+import org.robolectric.RobolectricTestRunner
+
+//                                               ┌──────────────────┐
```
jpsim, line 31: Nice diagram!

jpsim (dismissed):

library/java/io/envoyproxy/envoymobile/engine/AndroidProxyMonitor.java
```diff
@@ -35,6 +35,9 @@ private AndroidProxyMonitor(Context context, EnvoyEngine envoyEngine) {
     this.envoyEngine = envoyEngine;
     this.connectivityManager =
         (ConnectivityManager)context.getSystemService(Context.CONNECTIVITY_SERVICE);
+    // PROXY_INFO is not guaranteed to be a sticky intent so we need to trigger
+    // a manual poll of proxy settings.
+    handleProxyChange(null);
     registerReceiver(context);
```
jpsim, line 41: Since the order apparently matters here, that wasn't clear to me reading the code. Can you add a comment explaining why this order is important?

Augustyniak: done

jpsim (dismissed):

jpsim (approved):

jpsim (approved):

jpsim (approved):

PR #2696, opened 2022-11-22
Cleanup the state of PlatformBridgeCertValidator to clarify the threading semantics

RyanTheOptimist:
Cleanup the state of PlatformBridgeCertValidator to clarify the threading semantics.
Make the PendingValidation object only accessed on the worker thread.
Make the PlatformBridgeCertValidator only accessed on the main thread.

PR #2713, opened 2022-11-30
Cleanup threading/ownership semantics of PlatformBridgeCertValidator

RyanTheOptimist:
Cleanup threading/ownership semantics of PlatformBridgeCertValidator by only running static methods on the worker thread. The worker thread is owned by the Validator which join()s the thread before deleting it. So by making the worker thread only operate on data copied (or moved) into the the thread, we avoid multi-threaded access.

Risk Level: Low - Not in production
Testing: Existing unit
Docs Changes: N/A
Release Notes: N/A

### DIFF
--- a/library/common/extensions/cert_validator/platform_bridge/platform_bridge_cert_validator.cc
+++ b/library/common/extensions/cert_validator/platform_bridge/platform_bridge_cert_validator.cc
@@ -26,9 +26,9 @@ PlatformBridgeCertValidator::PlatformBridgeCertValidator(
 
 PlatformBridgeCertValidator::~PlatformBridgeCertValidator() {
   // Wait for validation threads to finish.
-  for (auto& [id, thread] : validation_threads_) {
-    if (thread.joinable()) {
-      thread.join();
+  for (auto& [id, job] : validation_jobs_) {
+    if (job.validation_thread_.joinable()) {
+      job.validation_thread_.join();
     }
   }
 }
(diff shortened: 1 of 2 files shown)


RyanTheOptimist:
@jpsim Can you take a look at this? After working on #2696 for a while, I eventually concluded that this approach was simpler/clearer. But what do you think?

library/common/extensions/cert_validator/platform_bridge/platform_bridge_cert_validator.cc
```diff
@@ -123,49 +125,69 @@ void PlatformBridgeCertValidator::verifyCertChainByPlatform(
   if (!success) {
     error_details = "PlatformBridgeCertValidator_verifySubjectAltName failed: SNI mismatch.";
```
jpsim, line 126: Is this SAN or SNI mismatch? The error message says SNI but the type is `FAIL_VERIFY_SAN`

jpsim (approved):
This does seem safer and easier to reason about. Thanks!

library/common/extensions/cert_validator/platform_bridge/platform_bridge_cert_validator.cc
RyanTheOptimist: I think it's both :) It's saying that the SNI from the ClientHello (a single hostname) is not contained in the SAN from the certificate (list of hostname). The current code uses "SNI" in the error message and updates the "fail_verify_san_" stat. I preserved this behavior in this PR, as it's not intended to change behavior.

But it would be trivial to rename FAIL_VERIFY_SAN to FAIL_VERIFY_SNI or change the string in the error message if that would be clearer? (I suspect changing the name of the stat is out of scope).

jpsim: No need to rename, I just wanted to flag in case this was a typo or they meant different things.

